Issue #25, opened 2020-11-26
Show accepting businesses on map.


elKei24:
Part of #10 

Acceptance criteria:
* [x] the app contains a map with dots for the accepting business

Out of scope:
* fetching the accepting businesses at runtime
* showing details for the accepting businesses
* setting the local area

PR #36, opened 2020-11-26
#25 show name of accepting business

elKei24:


RoAnBu:
@elKei24 why is it so slow? Is the data fetched from the server?

Clicking fast on different items also causes "null"s and randomly shows the name of a previous accepting business. Maybe that's also caused by the relatively slow animation of the snackbar?

elKei24:
> @elKei24 why is it so slow? Is the data fetched from the server?
> 
> Clicking fast on different items also causes "null"s and randomly shows the name of a previous accepting business. Maybe that's also caused by the relatively slow animation of the snackbar?

Dunno 😄 Did not care too much as this only is a proof-of-concept and a placeholder for the real details implementation. I think i you are unlucky, you click on a feature that is no accepting business and thus has no k_name property (like a standard POI or something), causing the null. We could filter for only features of a specific layer by passing layer IDs to the query (where we currently only pass an empty list).

RoAnBu (approved):
ok lets merge it, we can fix it later i guess

frontend/lib/map/full_map.dart
```diff
@@ -1,26 +1,43 @@
...
       styleString: "mapbox://styles/elkei24/ckhyn5h2l1yq519r9g3fbsogj",
+      onMapCreated: (controller) => {this._controller = controller},
+      onMapClick: this.onMapClick,
     );
   }
+
+  void onMapClick(Point<double> point, LatLng coordinates) {
+    this._controller.queryRenderedFeatures(point, [], null).then((features) => {
+          if (features?.isNotEmpty)
+            {this.onFeatureClick(json.decode(features[0]))}
```
maxammann, line 30: is features[0] here a plain geojson object?

maxammann (approved):

frontend/lib/map/full_map.dart
elKei24: I think so. Sadly, the type is specified as `dynamic` (like any, Object), but at runtime it is a string containing exactly the GeoJSON entry

PR #38, opened 2020-11-26
#24 flutter add permission management

### LINKED ISSUE #24: Setup Flutter Permissions for iOS and Android
### Problem: iOS app instantly crashes because mapbox wants to request location permission but permissions are not properly requested

Temporary fix: Add permission in .plist file

Correct fix should probably be to configure that in with Flutter

RoAnBu:
closes #24


elKei24 (approved):
Would be nice to see if the build server also can still build the app. No iOS expert, but looks good for me

maxammann (approved):

RoAnBu:
> Would be nice to see if the build server also can still build the app. No iOS expert, but looks good for me

Runs fine on iOS for me

PR #42, opened 2020-12-02
20 graphql schema generation

Schedulaar:
I added a run configuration "Generate GraphQL" which generates the graphql in sdl format to the frontend folder, where it can be used to generate the queries.

I had to upgrade artemis to make this possible (I don't know, why we used such an old version, maybe @PaulaWi do you know why?)


PaulaWi (approved):
Looks good :+1:

PaulaWi:
> I had to upgrade artemis to make this possible (I don't know, why we used such an old version, maybe @PaulaWi do you know why?)

No particular reason, the article I read just used this older version.

RoAnBu (approved):

PR #45, opened 2020-12-03
#26 set local area of map

### LINKED ISSUE #26: Set local area of map
Belongs to #10.

Accepting criteria:
* [x] local area can be set using device position
* [X] ~local area can be set using zip code~
* [x] if permission is granted, show device location on map

elKei24:
Closes #26.


PaulaWi (approved):
Works on android

RoAnBu:
@elKei24 the .ios name didnt work because it was probably already taken by the old app, so I removed the .ios suffix

PR #47, opened 2020-12-03
43 connect database

aiquita:


backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/database/Schema.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.jetbrains.exposed.dao.id.IntIdTable
+import org.jetbrains.exposed.sql.Column
+import org.jetbrains.exposed.sql.ColumnType
+import org.jetbrains.exposed.sql.Table
+import org.postgis.PGgeometry
+import org.postgis.Point
+
+object Categories : IntIdTable() {
+    val name = varchar("name", 50)
+    val iconUrl = varchar("iconUrl", 50)
```
maxammann, line 15: iconUrl is not needed as this is handled by the map

maxammann (approved):

PR #71, opened 2020-12-10
#70 use different main files for configuration

### LINKED ISSUE #70: Make app environment-aware
The app should be able to run against the deployed backend services as well as local ones, without changes to the code.

Belongs to #53

elKei24:
Closes #70.

Left to do:
* [x] maybe adapt URLs
* [x] test if it works


backend/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM gradle:jdk8 AS build
```
RoAnBu, line 1: What went wrong in this branch? This file is already in main, why should it be added again?

RoAnBu (dismissed):

backend/Dockerfile
maxammann: The reason for this is rebasing, There are two commits with the same content but different ref hash:
on main:
https://github.com/ehrenamtskarte/ehrenamtskarte/commit/eab148317e5bcee8a37f3744d3768964330695b2

in this pr: https://github.com/ehrenamtskarte/ehrenamtskarte/commit/98be6f07057364ebda70ae903628c5f381bc94d5

Somehow my commits ended up in this branch. This could be because somebody based a branch not on main but on some other branch which was rebased after that.

Conclusion: Don't rebase unless you are sure what you are doing :D

this should not be a problem here because the diffs will merge cleanly if this is merged very soon.

frontend/lib/main_local.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:ehrenamtskarte/configuration.dart';
+import 'package:flutter/material.dart';
+
+import 'app.dart';
+
+Future<void> main() async {
+  WidgetsFlutterBinding.ensureInitialized();
+  runApp(Configuration(
+      mapStyleUrl: "http://10.0.2.2:5002/style.json",
```
maxammann, line 9: I'm not sure whether this works on all environments. This maybe does not work when using a real device instead of an emulator.

better way would be to use localhost and use adb reverse.

maxammann (approved):

frontend/lib/main_local.dart
elKei24: Good point, I will change it to localhost then

backend/Dockerfile
elKei24: Yep, seams I rebased onto the wrong branch by mistake. Even if we should not have had problems merging, I dropped the duplicate commits.

maxammann (approved):

PR #74, opened 2020-12-10
#63 freinet data import

### LINKED ISSUE #63: Fetch and parse information about accepting businesses from freinet
Belongs to #11

RoAnBu:
closes #63


aiquita:
> 
> 
> closes #63

I think it is a first shot especially according to the technology. On the analysis side we have to have a deeper look into the data again. For testing purposes its suffices.

RoAnBu:
> I think it is a first shot especially according to the technology. On the analysis side we have to have a deeper look into the data again. For testing purposes its suffices.
> 
> Moreover categories are not imported yet.

Yes, not finished yet. Therefore draft PR. I just like PRs for discussions and an overview of what's being worked on.

@aiquita Do you want to continue working on the branch? Otherwise i would continue

backend/build.gradle.kts
```diff
@@ -49,9 +49,14 @@ dependencies {
     implementation("org.jetbrains.exposed", "exposed-core", exposed_version)
     implementation("org.jetbrains.exposed", "exposed-dao", exposed_version)
     implementation("org.jetbrains.exposed", "exposed-jdbc", exposed_version)
-    implementation("com.impossibl.pgjdbc-ng", "pgjdbc-ng", "0.8.6")
+    implementation("org.postgresql", "postgresql", "42.2.18")
```
Schedulaar, line 52: Why did you remove the impossibl?

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/AcceptingStoresImporter.kt
```diff
@@ -0,0 +1,76 @@
...
+            .parse<FreinetData>(freinetJson)
+
+    private fun importAcceptingStores(acceptingStores: List<FreinetAcceptingStore>) {
+        println("Inserting data into db")
+        try {
+            transaction {
+                for (acceptingStore in acceptingStores) {
+                    val address = AddressEntity.new {
+                        street = acceptingStore.street
+                        houseNumber = "10"
```
Schedulaar, line 49: Is there no house number in the data? Maybe in the street name?

backend/src/main/resources/freinet_import/categories.json
```diff
@@ -0,0 +1,38 @@
+[
```
Schedulaar, line 1: Don't we have an endpoint that we could fetch from instead of serving this statically? (Although it doesn't really matter)

Schedulaar (approved):
Minor:
You use the ```!!``` operator quite excessively... Afaik this should be avoided where possible. Maybe you can get rid of it at some places by using ```?: throw IllegalArgumentException``` or some more suitable exceptions

backend/build.gradle.kts
RoAnBu: Has no implementation for postgis columns and they don't plan to implement it. The main postgres jdbc driver worked straight away.

Does the impossibl driver have any advantages we use/wanted to use? We could maybe use different drivers for data importing and querying but idk if that is a good idea

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/AcceptingStoresImporter.kt
RoAnBu: Good question, didnt see that. @aiquita? 😄

backend/src/main/resources/freinet_import/categories.json
RoAnBu: No idea. @maxammann is there a way to see all files that can be queried from the freinet api?

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/database/Database.kt
```diff
@@ -51,12 +51,10 @@ class Database {
...
 
     val db by lazy {
-        val ds = PGDataSource()
-        ds.serverName = System.getProperty("app.postgres.host")
-        ds.databaseName = System.getProperty("app.postgres.database")
-        ds.user = System.getProperty("app.postgres.user")
-        ds.password = System.getProperty("app.postgres.password")
-
-        connect(ds)
+        val connectionUrl = "jdbc:postgresql://" +
```
maxammann, line 54: what does reWriteBatchedInserts do? Do we require it?

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/AcceptingStoresImporter.kt
```diff
@@ -0,0 +1,76 @@
...
+                        postalCode = acceptingStore.postalCode
+                        locaction = acceptingStore.location
+                        state = "de"
```
maxammann, line 52: Is state fine here or should it be "Germany"? We should define what should be in this field.

E.g. Alpha-2 codes: https://www.iban.com/country-codes

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/annotations/BooleanField.kt
```diff
@@ -0,0 +1,4 @@
+package xyz.elitese.ehrenamtskarte.importer.annotations
+
+@Target(AnnotationTarget.FIELD)
+annotation class BooleanField
```
maxammann, line 4: Minor: Kind of weird that these are required. Shouldn't Klaxon support this?

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/dataqa/FreinetDataFilter.kt
```diff
@@ -0,0 +1,10 @@
+package xyz.elitese.ehrenamtskarte.importer.dataqa
+
+import xyz.elitese.ehrenamtskarte.importer.types.FreinetAcceptingStore
+
+object FreinetDataFilter {
+    fun filterOutInvalidEntries(freinetAcceptingStores : List<FreinetAcceptingStore>) : List<FreinetAcceptingStore> {
```
maxammann, line 6: Are we sure that these should be filtered out? Can you post a list of partners which have invalid data?

Maybe we should transform the data to be valid instead of filtering the entries out.

backend/src/main/resources/freinet_import/categories.json
maxammann: Not that I know of. This seems fine!

maxammann: But thee ids seems wrong. They go from 0 to 8 in NRW.

maxammann (approved):

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/database/Database.kt
RoAnBu: It tries to batch following inserts together to make the inserts faster. However that does not seem to work because we have inserts that depend on each other. It was a try, I can remove it again I guess

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/annotations/BooleanField.kt
RoAnBu: I think the issue is that all fields in the original json are strings, so they need to be converted to bools, ints etc. @aiquita can you confirm that this was the reason?

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/dataqa/FreinetDataFilter.kt
RoAnBu: Currently i think we have less than 10 invalid entries, so not that urgent. Most of them should be fixable, however some are missing the category id which could be a problem. I think we need a concept for broken data later anyway and we also need to search for duplicates in the database etc. I think a separate user story/technical task would be good

backend/src/main/resources/freinet_import/categories.json
RoAnBu: Freinet api returns values from 1-9 (inclusive), I got the values from here: https://github.com/ehrenamtskarte/artefacts/blob/main/api_access.md

aiquita (approved):

PR #98, opened 2020-12-19
#97: Animate Store Click on Map

### LINKED ISSUE #97: Animate Navigation to Clicked Store On Map
When the user clicks on a store on the map, the map currently jumps to the target.
Make it animated!

Schedulaar:
Closes #97

### DIFF
--- a/frontend/lib/map/map/map.dart
+++ b/frontend/lib/map/map/map.dart
@@ -18,10 +18,11 @@ class Map extends StatefulWidget {
   final List<String> onFeatureClickLayerFilter;
   final bool myLocationEnabled;
 
-  const Map({this.onFeatureClick,
-    this.onNoFeatureClick,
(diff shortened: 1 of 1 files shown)


frontend/lib/map/map/map.dart
```diff
@@ -19,67 +18,68 @@ class Map extends StatefulWidget {
...
+        onMapCreated: _onMapCreated,
+        onMapClick: _onMapClick
     );
+    super.didChangeDependencies();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return _mapboxMap;
   }
```
Schedulaar, line 54: This might be a little bit hacky, but I don't see any disadvantages (yet?)

elKei24 (approved):
Well, I still do not fully understand why it works now, but it does and the code is cleaner than before, so 👍🏽

PR #100, opened 2020-12-19
#100: Enlarge selected symbol on map

Schedulaar:
Closes #100 .

### DIFF
--- a/frontend/lib/map/map/map.dart
+++ b/frontend/lib/map/map/map.dart
@@ -55,6 +55,22 @@ class _MapState extends State<Map> {
 
   _onMapCreated(controller) => this._controller = controller;
 
+  Symbol _symbol;
+  _removeSymbol() async {
+    if (_symbol != null) {
+      await _controller.removeSymbol(_symbol);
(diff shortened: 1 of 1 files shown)


frontend/lib/map/map/map.dart
```diff
@@ -65,9 +81,14 @@ class _MapState extends State<Map> {
...
         var location = new LatLng(coordinates[1], coordinates[0]);
+        await _removeSymbol();
+        await _addSymbol(location, featureInfo["properties"]["categoryId"]);
         await _bringCameraToLocation(location);
       }
-    } else if (widget.onNoFeatureClick != null) widget.onNoFeatureClick();
+    } else if (widget.onNoFeatureClick != null) {
+      await _removeSymbol();
+      widget.onNoFeatureClick();
+    }
```
Schedulaar, line 91: Actually, I only did this, because I was curious how we could actually implement this.
Unfortunately, I couldn't find any other solution than drawing a symbol on top of the actual symbol.
Do you know of any other way?

PR #101, opened 2020-12-19
99: Enlarge selected symbol

### LINKED ISSUE #99: Enlarge Selected Store on Map
When clicking on a store which is surrounded by lots of other stores, it is often unclear, which store acutally got selected.
Enlarge the selected icon.

Schedulaar:
Closes #99

### DIFF
--- a/frontend/lib/map/map/map.dart
+++ b/frontend/lib/map/map/map.dart
@@ -56,6 +56,22 @@ class _MapState extends State<Map> {
 
   _onMapCreated(controller) => this._controller = controller;
 
+  Symbol _symbol;
+  _removeSymbol() async {
+    if (_symbol != null) {
+      await _controller.removeSymbol(_symbol);
+      _symbol = null;
+    }
+  }
+
+  _addSymbol(LatLng location, int categoryId) async {
+    _symbol = await _controller.addSymbol(new SymbolOptions(
+        iconSize: 1.5,
+        geometry: location,
+        iconImage: categoryId.toString()
+    ));
+  }
+
   void _onMapClick(Point<double> point, clickCoordinates) async {
     var features = await this._controller.queryRenderedFeatures(
         point, widget.onFeatureClickLayerFilter ?? [], null);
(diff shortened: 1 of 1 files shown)


frontend/lib/map/map/map.dart
```diff
@@ -56,6 +56,20 @@ class _MapState extends State<Map> {
 
   _onMapCreated(controller) => this._controller = controller;
 
+  Symbol _symbol;
+
+  _removeSymbol() async {
```
maxammann, line 61: There is probably a ?better? way of doing this instead of adding and removing the element. We could just add a property on the element on the map called `iconSize`. Our map style could then use this if it is set.

We can use your solution though if it does not make problems. Does the symbol disappear when scrolling away?

maxammann (approved):

frontend/lib/map/map/map.dart
Schedulaar: >There is probably a ?better? way of doing this instead of adding and removing the element. We could just add a property on the element on the map called iconSize. Our map style could then use this if it is set.

Yeah, I don't like the way I draw a symbol on top of the original symbol either, but after a deep dive into the mapbox flutter library, I don't see any other possiblity (but if you do, I'd appreciate to know :) ). Afaics, there's no possibility to "change" the feature queried through the flutter bridge.

> We can use your solution though if it does not make problems. Does the symbol disappear when scrolling away?

Besides some very small spoardeous flickerings, it behaves as I would expect.

maxammann: Huch oke there is no way to write to the `"properties"` object within the queried features?

If that is the case then you are right. Really shitty lib...

maxammann: @elKei24 do you have experience with this?

elKei24: No, sadly not

maxammann: Is `featureInfo["properties"]` writeable? We would need to write into this map.

Schedulaar: well, it's writable, but it comes through a json string from the flutter bridge and hence there's no effect if you'd write anything on it. (I haven't tried it though :thinking: )

maxammann: I'm looking thourhg the API right now. Maybe updateSymbol would be nice: https://github.com/tobrun/flutter-mapbox-gl/blob/8f9838ff5db4e6d4c17e09a5088afb38430b693b/mapbox_gl_web/lib/src/mapbox_map_controller.dart#L170

maxammann: There is no way to write to the properties as far as i can see :(

PR #108, opened 2020-12-22
77 simple search

maxammann:


Schedulaar (approved):

frontend/lib/search/filter_bar.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:flutter/material.dart';
+
+class FilterBar extends StatelessWidget {
+  final Function(List<int>) onSelectedCategoriesChange;
+
+  FilterBar({this.onSelectedCategoriesChange});
```
elKei24, line 6:
```suggestion
  FilterBar({Key key, this.onSelectedCategoriesChange}) : super(key: key);
```

frontend/lib/search/filter_bar.dart
```diff
@@ -0,0 +1,21 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return SliverGrid.count(
+        crossAxisCount: 5,
+        children: Iterable.generate(5)
+            .map((i) => Container(
+                color: Theme.of(context).accentColor,
+                child: IconButton(
+                  icon: Icon(Icons.category),
+                  onPressed: () => {},
```
elKei24, line 17: Handler will be implemented in a later PR I guess?

frontend/lib/search/result_list.dart
```diff
@@ -0,0 +1,83 @@
...
+import 'package:ehrenamtskarte/widgets/error_message.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:graphql_flutter/graphql_flutter.dart';
+
+class ResultList extends StatelessWidget {
+  final String _searchText;
+  final List<int> _searchCategories;
+
+  ResultList(this._searchText, this._searchCategories);
```
elKei24, line 14:
```suggestion
  ResultList(this._searchText, this._searchCategories, {Key key}) : super(key: key);
```

frontend/lib/search/result_list.dart
```diff
@@ -0,0 +1,83 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final byIdQuery = AcceptingStoresSearchQuery(
```
elKei24, line 18: `byIdQuery` seems misleading, maybe better use `searchQuery`?

frontend/lib/search/result_list.dart
```diff
@@ -0,0 +1,83 @@
...
+            return Divider(height: 0, color: Colors.grey);
+          },
+          semanticIndexCallback: (Widget widget, int localIndex) {
+            if (localIndex.isEven) {
+              return localIndex ~/ 2;
+            }
+            return null;
+          },
+          childCount: max(0, matchingStores.length * 2 - 1),
+        ));
```
elKei24, line 71: Maybe we could extract this into a separate build function or maybe even a separate widget to reduce the complexity of this build function.

frontend/lib/search/search_page.dart
```diff
@@ -0,0 +1,66 @@
...
+              hintText: "Tippen und schreiben, um zu suchen …",
+            ),
+          ),
+          pinned: true,
+          actions: [
+            IconButton(
+              icon: const Icon(Icons.search),
+              onPressed: () => {},
+            )
+          ],
```
elKei24, line 38:
```suggestion
```
I think we can leave away the button if it does nothing anyway.

elKei24 (approved):
Looks great in general. Sadly I could not test it as the backend fails to start, but that might be only a local issue on my side.

frontend/lib/search/result_list.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:math';
+
+import 'package:ehrenamtskarte/graphql/graphql_api.graphql.dart';
+import 'package:ehrenamtskarte/map/detail/detail_view.dart';
+import 'package:ehrenamtskarte/widgets/error_message.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:graphql_flutter/graphql_flutter.dart';
```
elKei24, line 8:
```suggestion
import 'package:flutter/cupertino.dart';
import 'package:flutter/material.dart';
import 'package:graphql_flutter/graphql_flutter.dart';

import '../graphql/graphql_api.graphql.dart';
import '../map/detail/detail_view.dart';
import '../widgets/error_message.dart';
```

When possible, you should use the relative path according to the Dart style guide.

elKei24 (approved):

frontend/lib/home/home_page.dart
```diff
@@ -37,18 +39,25 @@ class _HomePageState extends State<HomePage> {
...
+          index: _currentTabIndex,
+          children: _tabs
+              .asMap()
+              .map((index, tab) => MapEntry(
+                  index,
+                  Focus(
+                      key: ValueKey(index),
+                      child: tab.widget,
+                      descendantsAreFocusable: _currentTabIndex == index)))
+              .values
```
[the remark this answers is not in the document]
Schedulaar: Alright :+1: It's part of the search-pr as the search textbox introduced a problem with showing the keyboard; hence we need some focus management...

frontend/lib/search/filter_bar.dart
Schedulaar: Yep

frontend/lib/search/result_list.dart
Schedulaar: Why is this needed?

Schedulaar: I guess this is standard in flutter... Damn that boilerplate :/

elKei24: It is probably not needed at the moment, but good style to still add it so that you do not have to touch this file later on when you need a key.
The key is necessary for example when you replace a widget in the tree by an other of the same type, to allow Flutter to detect the change by comparing the key.

frontend/lib/search/result_list.dart
Schedulaar: I extracted the list

frontend/lib/search/result_list.dart
Schedulaar: I removed cupertino.
The relative path thing is a bit tricky, as Intellij doesn't do that out of the box. Hence it's not practical to keep up such a rule i think

elKei24: Yep, you are right. I also misunderstood the style guide, one only should use the relative path if the other file is _in the same directory_.

frontend/lib/search/search_page.dart
Schedulaar: Now, it does focus the text field :)

PR #111, opened 2020-12-22
104 scan qr code

### LINKED ISSUE #104: Scan the EAK authorization token provided by the backend
Related to user story #17. The backend provides a qr code (#102) containing a json web token (jwt). The payload is unencrypted.

- [x] Provide an interface which can scan the qr code and write its content into a state

Not part of this task:
- Verifying the qr code content
- the ui displaying the digital EAK, some log message or string is sufficient

JWT might not be the best solution here, but it is super easy to create jwts in java as there is not a lot of stuff you can do wrong. Instead of a jwt we could also encode a simple json in the qrcode that contains some kind of unique string (maybe uuid, but is that safe?). That unique string must then be checked against the backend

**Additional context**
QR Code for testing.
image

The json payload is:
```json
{
  "expirationDate": "2021-12-24",
  "firstName": "Max",
  "lastName": "Mustermann"
}
```

RoAnBu:
closes #104


frontend/ios/Runner/Info.plist
```diff
@@ -43,5 +43,10 @@
 	<false/>
 	<key>NSLocationWhenInUseUsageDescription</key>
     <string>Display location on the world map</string>
+    <key>io.flutter.embedded_views_preview</key>
+    <true/>
+    <key>NSCameraUsageDescription</key>
+    <string>This app needs camera access to scan QR codes</string>
```
maxammann, line 49: This should be german

Issue #127, opened 2020-12-27
Add coordinates parameter for search


michael-markl:
As the search results should be sorted by distance to the user, the endpoint of the backend should support entering coordinates and return the results sorted by distance to the user.

User Story #118 .

PR #128, opened 2020-12-27
127: Add coordinates to search backend

Schedulaar:
Closees #127


elKei24:
I am not competent enough in the backend yet for reviewing this, but we should review and merge this quickly, so I can integrate it in #129. I did very similar changes there 😢

Schedulaar:
I looked at your chnages, but afaik, they should be easily resolvable

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/webservice/dataloader/AcceptingStoreDataLoader.kt
```diff
@@ -14,14 +14,10 @@ val acceptingStoreLoader = DataLoader<Int, AcceptingStore?> { ids ->
         runBlocking {
             transaction {
                 AcceptingStoresRepository.findByIds(ids).map {
-                    AcceptingStore(it.id.value, it.name, it.description, it.contactId.value, it.categoryId.value)
-                }.associateWithKeys<AcceptingStore, Int>({ it.id }, ids)
+                    if (it == null) null
```
maxammann, line 17: This is a little bit weird. The accepting store shouldn't be null I think

maxammann (approved):

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/webservice/dataloader/AcceptingStoreDataLoader.kt
Schedulaar: The user might always query ids for which the accepting store is null

RoAnBu (approved):

PR #129, opened 2020-12-28
#105 view on map

### LINKED ISSUE #105: View-on-map function in detail view of accepting store
When clicking on a search result its currently not possible to view the selected store on the map. 

A "view on map"-function needs to be added to the detail view.

Alternatively, a "view on map"-button could be added directly to the list item in the search result.

elKei24:
Closes #105.

I wanted to implement it using navigation, but using navigation to change state (selected tab) instead for just replacing widgets seems not possible without hacking ugly stuff.


RoAnBu:
```
======== Exception caught by gesture ===============================================================
The following NoSuchMethodError was thrown while handling a gesture:
The method 'showAcceptingStore' was called on null.
Receiver: null
Tried calling: showAcceptingStore(Instance of 'PhysicalStoreFeatureData')

When the exception was thrown, this was the stack: 
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:51:5)
#1      _HomePageState._goToMap (package:ehrenamtskarte/home/home_page.dart:86:30)
#2      DetailContent._showOnMap (package:ehrenamtskarte/map/detail/detail_content.dart:84:33)
#3      DetailContent.build.<anonymous closure> (package:ehrenamtskarte/map/detail/detail_content.dart:74:46)
#4      _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:993:19)
...
Handler: "onTap"
Recognizer: TapGestureRecognizer#7671a
  debugOwner: GestureDetector
  state: possible
  won arena
  finalPosition: Offset(210.7, 563.0)
  finalLocalPosition: Offset(87.2, 20.0)
  button: 1
  sent tap down
====================================================================================================

======== Exception caught by widgets library =======================================================
The following assertion was thrown building DetailContent(dirty, dependencies: [_LocalizationsScope-[GlobalKey#589a9], _InheritedTheme]):
Failed assertion: boolean expression must not be null

The relevant error-causing widget was: 
  DetailContent file:///Users/Roman/git/eak/ehrenamtskarte/frontend/lib/map/detail/detail_view.dart:41:17
When the exception was thrown, this was the stack: 
#0      DetailContent.build (package:ehrenamtskarte/map/detail/detail_content.dart:62:18)
#1      StatelessElement.build (package:flutter/src/widgets/framework.dart:4701:28)
#2      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4627:15)
#3      Element.rebuild (package:flutter/src/widgets/framework.dart:4343:5)
#4      ComponentElement._firstBuild (package:flutter/src/widgets/framework.dart:4606:5)
...
====================================================================================================
```

elKei24:
@RoAnBu Hm, I merged main in case it was fixed there already, and did some other small changes. Can you try again? At least I can not reproduce it now.

frontend/lib/home/home_page.dart
```diff
@@ -1,62 +1,111 @@
+import 'package:ehrenamtskarte/home/app_flows_stack.dart';
 import 'package:flutter/material.dart';
 
 import '../map/map_page.dart';
 import '../identification/identification_page.dart';
 import '../search/search_page.dart';
+import './app_flow.dart';
```
RoAnBu, line 7: Which import style do we use now? Relative or "absolute" paths? Maybe we should use the package version because the flutter plugin autoimport works that way?

RoAnBu (approved):
Looks good but tbh I did not really understand everything :D I think I would need some additional explanations

frontend/lib/home/home_page.dart
elKei24: Yep, you are right. I misunderstood the style guide at first, it says you should only use the relative version if the imported file is in exactly the same directory (like in the new line 7, but not in the lines 4-6). And, as you said, the autoimporter also does it that way, so that is a good reason for us to do it that way.

PR #149, opened 2020-12-31
#123 fix map zoom (again)

### LINKED ISSUE #123: App starts zoomed in even if no location available
When starting the app without location services, the map is zoomed in too much.
Actually, if there is no location available, it should be zoomed out such that all of Bavaria is visible.

https://user-images.githubusercontent.com/7025199/103023838-fa0f3b80-454e-11eb-9bf7-20f163f0eab6.mov

elKei24:
Previous PR did not really fix it, but this should now close #123.

There seemed to be a race condition again, so that sometimes the map zoomed in if location services were enabled, and sometimes not. This PR changes the behavior, so that the map always zooms in to the user zoom level, and zooms out to whole Bavaria if it fails to retrieve the user location. As this happens directly after initialization and the zooming happens without animation, it looks like the app was showing Bavaria from the beginning.

Additionally, this PR moves the compass to the lower left, so that it is not hidden behind the Android status bar on some devices. There also is an option to set its margin, but that would force us to find out how many pixels of our app are hidden behind the status bar (could be 0 on older devices for example).


RoAnBu:
Android zooms in to "Grösdorf" to "Markt Kipfenberg" instead of showing Bavaria when I do the following:
1. Delete the app from the phone
2. Start it again and deny location permissions
3. It directly zooms in to Grösdorf
Interestingly if I click on the "Markt Kipfenberg" store it asks me again for location permission (even though I already denied it like 10 seconds before).

Is that a problem with the Emulator? I can't test on real devices right now.

frontend/lib/map/map/map.dart
```diff
@@ -57,23 +55,27 @@ class _MapState extends State<Map> {
...
   }
 
-  _onMapCreated(controller) async {
+  Future<void> _onMapCreated(MapboxMapController controller) async {
     this._controller = controller;
+    // zoom out to initial zoom level if can not get location
     await _controller
         .requestMyLocationLatLng()
-        .then((_) => _onUserLocationUpdated());
+        .then((_) {}, // like catchError, but do not enforce return type
```
RoAnBu, line 63: Why do we do nothing here?

elKei24: Because if we get there, that only means that we successfully retrieved to location and thus can stay zoomed in, so there is nothing to do. We only have to handle the error case, because then we have to access to the current location and thus should zoom out. As the comment says, there is an alternative to `then` for such cases, but it expects a location as return type, thus i did not use it.

RoAnBu: Ah, makes sense! Thanks for the explanation!

elKei24:
> Is that a problem with the Emulator? I can't test on real devices right now.

Oh boy, you are right. Even if it only is the emulator, we should still fix it to be safe. 

Why can not Mapbox just support this usecase out of the box? Seems like a common use case to only zoom in if we know the location … 😞

elKei24:
Must be some kind of race condition again. If I run it with breakpoints, it works, if I remove them, its broken. I hate it …

frontend/lib/map/map/map.dart
Schedulaar: Is there a difference to a try catch?
https://dart.dev/codelabs/async-await#example-async-and-await-with-try-catch
I think mixing the notation of `await/async` with `then,...` is confusing

frontend/lib/map/map/map.dart
```diff
@@ -31,23 +31,21 @@ class Map extends StatefulWidget {
...
       styleString: config.mapStyleUrl,
       myLocationEnabled: widget.myLocationEnabled,
       myLocationTrackingMode: widget.myLocationEnabled
           ? MyLocationTrackingMode.Tracking
           : MyLocationTrackingMode.None,
       onMapCreated: _onMapCreated,
       onMapClick: _onMapClick,
-      onUserLocationUpdated: (location) => _onUserLocationUpdated(),
-      onCameraTrackingDismissed: _onCameraTrackingDismissed,
+      compassViewPosition: CompassViewPosition.BottomRight,
```
Schedulaar, line 48: If an acceptance store is selected, the compass is covered by the overlay then, right?

Schedulaar (approved):

frontend/lib/map/map/map.dart
elKei24: > I think mixing the notation of `await/async` with `then,...` is confusing

Agreed, I will change it

frontend/lib/map/map/map.dart
elKei24: Yes

PR #152, opened 2020-12-31
#138 replace "n.A." with null on data import

### LINKED ISSUE #138: "n. A." in accepting store contact data
Some accepting stores contain contact entries like "n. A.". We should replace them with `null` when importing, so that the frontend can decide to not display that entry.

elKei24:
Closes #138


backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/general/ImportToDatabase.kt
```diff
@@ -29,25 +28,36 @@ object ImportToDatabase {
...
     }
 
+    /**
+     * Replaces notes for unspecified entries with empty strings.
+     *
+     * Such a note consists of the letters "nA" (in that order), potentially with dots like "n.A." and/or stuffed with whitespace (e.g. " n A. ").
+     *
+     * @return empty string if `text` is a "unspecified note", `text` otherwise
+     */
+    private fun replaceUnspecifiedNote(text: String): String {
```
maxammann, line 38: The issue says that the entry should be null. In this PR an empty string is used. I think null would be way better

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/general/ImportToDatabase.kt
```diff
@@ -29,25 +28,36 @@ object ImportToDatabase {
...
+        return if (Regex("""\s*n\.?\s*A\.?\s*""").matches(text)) "" else text
```
maxammann, line 39: This regex should probably be stored as a variable in order to speedup the matching

maxammann (changes requested):

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/general/ImportToDatabase.kt
elKei24: You are right. I used the empty string because this is what undefined contact data are currently stored like. Would make more sense to change these entries to null, too 👍

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/general/ImportToDatabase.kt
elKei24: Done

backend/src/main/kotlin/xyz/elitese/ehrenamtskarte/importer/general/ImportToDatabase.kt
elKei24: Done

maxammann (approved):

frontend/schema.graphql
```diff
@@ -31,18 +31,18 @@ type AcceptingStore {
...
 type Address {
   id: Int!
-  location: String!
-  postalCode: String!
-  state: String!
-  street: String!
+  location: String
+  postalCode: String
+  state: String
+  street: String
```
Schedulaar, line 45: ### Introducing nullables to a not null-safe frontend
I dont care

(just kidding, I'm fine with it^^)

Schedulaar (approved):

frontend/schema.graphql
elKei24: 😄 Null safety is in beta since 10/2020, I hope they will finally push it to stable in the next weeks …

PR #157, opened 2021-01-02
add real linter

elKei24:
We only used the formatter before, but it does not check code smells and other non-formatting related code styling rules. 

This PR …
- sets up analysis rules (automatically detected by IntelliJ)
- uses them in the pipeline instead of the formatter
- adapts the existing code to the effective dart guidelines

This should make the pipeline less pedantic, as files do not have to be formatted _exectly_ like flutter format would format them, and instead give more valuable hints.


elKei24:
**Note:** after merging this PR, we have to adapt the branch protection rules to depend on "FE: analyze" instead of "FE: lint". We also have to force merging of this PR using admin power due to the same reason.

Schedulaar (approved):
Should be merged asap

Issue #165, opened 2021-01-07
Create form for eak generation in administration


aiquita:
Belongs to #17

PR #170, opened 2021-01-08
#165: added form for eak qr-code generation

aiquita:
- eak form tile
- page consists of multiple form tiles
- Generate QR-codes button has not functionality yet!


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -1,6 +1,17 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0">
     <option name="myName" value="Project Default" />
+    <inspection_tool class="CssUnknownProperty" enabled="true" level="WARNING" enabled_by_default="true">
+      <option name="myCustomPropertiesEnabled" value="true" />
+      <option name="myIgnoreVendorSpecificProperties" value="false" />
+      <option name="myCustomPropertiesList">
+        <value>
+          <list size="1">
+            <item index="0" class="java.lang.String" itemvalue="hegiht" />
```
maxammann, line 10: `hegiht` does not seem well spelled :D

administration/package.json
```diff
@@ -19,6 +20,8 @@
     "react": "^17.0.1",
     "react-apollo": "^3.1.5",
     "react-dom": "^17.0.1",
+    "react-router": "^5.2.0",
```
maxammann, line 23: I prefere the react-navigation router but probably it does not matter: https://www.npmjs.com/package/react-navigation

administration/src/components/generation/AddEakForm.css
```diff
@@ -0,0 +1,18 @@
+.add-eak-form-label {
```
maxammann, line 1: Should we maybe use styled-components? What do you think @Schedulaar ?

administration/src/components/generation/EakGeneration.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export class EakGeneration extends React.Component<Props, State> {
+
+    constructor(props: Props) {
+        super(props);
+        this.state = {
+            cardCreationModels: [EakGeneration.createEmptyCard()]
+        };
+
+        this.addForm = this.addForm.bind(this);
```
maxammann, line 26: The preferred way of writing components nowadays is using function components and react hooks.
That also avoids using constructors and binding funcitons to a context.

administration/src/components/generation/EakGeneration.tsx
```diff
@@ -0,0 +1,92 @@
...
+        this.reset = this.reset.bind(this);
+    }
+
+    componentDidMount(): void {
+        const firstFormColumn = document.getElementsByClassName("form-column")[0];
+        const addEakBpCard = document.getElementById("add-eak-form-card")
+        if (addEakBpCard != null) {
+            addEakBpCard.style.height = firstFormColumn.clientHeight - 50 + "px"
```
maxammann, line 34: Why is that needed? Can you maybe set the height of the `form-column`s to a fixed height? That way you would not need to recalculate the height of elements

administration/src/components/generation/EakGeneration.tsx
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    private static createEmptyCard() {
```
maxammann, line 38: I think the return type should be set here.

administration/src/models/CardCreationModel.ts
```diff
@@ -0,0 +1,8 @@
+import {CardType} from "./CardType";
+
+export interface CardCreationModel {
```
maxammann, line 3: This model is only used in EakGeneation.ts. That means you can simply inline the interface there. I think right now there is no need for a separate `models` directory.

maxammann (approved):

administration/src/components/generation/EakGeneration.tsx
aiquita: I am not happy with this too, but as far as I can see there is no option to reach my goal otherwise. 

If I set form-column to a fixed height I fear that we maybe lose responsiveness of blueprintjs. As well, I do not want to investigate pixels by hand. 

CSS-Hacking does not help in this case, because not matter what I tried a edge case remains.

maxammann (approved):

.idea/inspectionProfiles/Project_Default.xml
Schedulaar: Still not

Schedulaar (commented):
EakForm.css is empty and should be remoed

administration/src/components/generation/EakGeneration.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    const [cardCreationModels, setCardCreationModels] = useState([createEmptyCard()]);
+
+    /*componentDidMount(): void {
+        const firstFormColumn = document.getElementsByClassName("form-column")[0];
+        const addEakBpCard = document.getElementById("add-eak-form-card")
+        if (addEakBpCard != null) {
+            addEakBpCard.style.height = firstFormColumn.clientHeight - 50 + "px"
+        }
+    }*/
```
Schedulaar, line 30: Remove commented code (or uncomment it if its neeed :) )

administration/src/components/generation/AddEakForm.css
Schedulaar: Yeah, would be nice, I think

administration/src/components/generation/EakGeneration.tsx
Schedulaar: If you really **need** to measure pixels here, use a react ref, pass the height via setState and prop to the related child element and define the style property on it.

Schedulaar: What do you want to achieve by setting the height to `firstFormColumn.clientHeight - 50`?

Schedulaar:
I added styled components in #180

.idea/inspectionProfiles/Project_Default.xml
maxammann: These changes should be reverted

maxammann (approved):

Schedulaar (approved):

Issue #190, opened 2021-01-21
Add probuf schema for CardQRActivationModel


maxammann:
Add the proto file to <root>/specs

Also move schema.graphql there.

PR #192, opened 2021-01-21
#190 Add protobuf schema to specs directory

maxammann:
fixes #192


.idea/ehrenamtskarte.iml
```diff
@@ -1,2 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
RoAnBu, line 1: Is that wanted? I mean, it seems to work :D

RoAnBu (approved):

PR #195, opened 2021-01-21
191 administration protobuf

maxammann:


administration/src/util/protobuf.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const generateCardActivateModel = (fullName: string, region: number, expireIn: Duration, cardType: CardActivateModel.CardType) => {
+    const randomBytes = new Uint8Array(16); // 128 bit randomness
+    crypto.getRandomValues(randomBytes)
+
+    // https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HTOP)
+    // https://tools.ietf.org/html/rfc4226#section-4 - R6 (How long should a shared secret be? -> 160bit)
+    // https://tools.ietf.org/html/rfc4226#section-7.5 - Random Generation (How to generate a secret? -> Random)
+    const totpSecret = new Uint8Array(20);
+    crypto.getRandomValues(totpSecret)
```
Schedulaar, line 12: Have you seen this?
From https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues#usage_notes

> Usage notes
> Don't use getRandomValues() to generate encryption keys. Instead, use the generateKey() method. There are a few reasons for this; for example, getRandomValues() is not guaranteed to be running in a secure context.
> 
> There is no minimum degree of entropy mandated by the Web Cryptography specification. User agents are instead urged to provide the best entropy they can when generating random numbers, using a well-defined, efficient pseudorandom number generator built into the user agent itself, but seeded with values taken from an external source of pseudorandom numbers, such as a platform-specific random number function, the Unix /dev/urandom device, or other source of random or pseudorandom data.
>

maxammann: oh good catch, generateKey seems like its targeted only at the web crypto spec. Not sure how we can use it here.
Maybe getRandomValues() is fine as our app is hosted on .app which always has a secure context (if I understand this correctly).

The second paragraph is not that much of a problem imo. Random numbers are limited. Browsers can run out of them. Same without VMs at hetzner.

maxammann: Maybe this is better though: https://dev.to/al_khovansky/generating-2fa-one-time-passwords-in-js-using-web-crypto-api-1hfo

```
async function generateHOTP(secret, counter) {
  const key = await generateKey(secret, counter);
  const uKey = new Uint8Array(key);

  const Snum = truncate(uKey);
  // Make sure we keep leading zeroes
  const padded = ('000000' + (Snum % (10 ** 6))).slice(-6);

  return padded;
}
```

administration/src/util/protobuf.ts
```diff
@@ -0,0 +1,39 @@
+import {add, Duration, getUnixTime, setHours} from "date-fns";
+import {CardActivateModel} from "../generated/compiled";
+import {detect} from "detect-browser";
+
+const generateCardActivateModel = (fullName: string, region: number, expireIn: Duration, cardType: CardActivateModel.CardType) => {
+    const browser = detect()
+    const isIE11 = browser?.name === "ie" && browser.version === "11"
+    if (!window.isSecureContext && !isIE11) // localhost is considered secure.
+        throw Error("Environment is not considered secure nor are we using Internet Explorer.")
```
maxammann, line 9:
```suggestion
        throw Error("Environment is not considered secure nor are we using Internet Explorer 11.")
```

Schedulaar (approved):

PR #211, opened 2021-01-28
16 display qr code for verification on other devices

### LINKED ISSUE #16: Display QRCode for authorization against accepting businesses
Function to display QR-Code containing the CardQrVerifyModel in app

userstory #176

PaulaWi:


Schedulaar (commented):
I think it would be good, if you could turn this into a PR very soon. Otherwise reviewing will be very hard.

It doesn't have to  be finished, but it should go to main soon

Issue #212, opened 2021-01-28
Connect Adminstration Backend to GraphQl endpoints


Schedulaar:
The adminstration backend should be connected to the graphql endpoints for creating new ehrenamtskarten.

Relates to #83.

PR #213, opened 2021-01-28
#212: Make endpoint selectable for graphiql

Schedulaar:


backend/src/main/resources/graphiql/renderExample.js
```diff
@@ -11,139 +11,150 @@
...
-      onEditQuery: onEditQuery,
-      onEditVariables: onEditVariables,
-      onEditHeaders: onEditHeaders,
-      defaultSecondaryEditorOpen: true,
-      onEditOperationName: onEditOperationName,
-      headerEditorEnabled: true,
-  }),
-  document.getElementById('graphiql'),
-);
+ReactDOM.render(React.createElement(SelectGraphqlInstance), document.getElementById('graphiql'));
```
elKei24, line 160: Why is there another call to `React.createElement` necessary? Does not `SelectGrapphqlInstance` already create an element?

elKei24 (approved):

backend/src/main/resources/graphiql/renderExample.js
Schedulaar: `SelectGraphqlInstance` itself is only a function which cannot be passed to `ReactDOM.render`.

Schedulaar:
On Hold until discussion about GraphQL tactic is resolved...

backend/src/main/resources/graphiql/renderExample.js
elKei24: Ah, of course, thanks for the explanation 👍🏽

Schedulaar:
Closed, as we decided for going the single graphql endpoint way for now...

Issue #189, opened 2021-01-21
Backend endpoint for adding EAKs


elKei24:
Part of ???

> See https://raw.githubusercontent.com/ehrenamtskarte/artefacts/main/verification_protocol.drawio.svg?token=ABVTEL7LILXTBDIBTZJK45S7643NI for details

~ description by @Schedulaar

PR #214, opened 2021-01-29
#189: Update Verification Backend According to Verification Protocol

Schedulaar:
Bytearrays should not be transmitted using actual arrays of numbers, as this would result in a query like
```[0, 51, 5, 32, 117, 32, 42, 43]```.
Instead, we simply base64 encode it.

### CODE REVIEW

**Review by maxammann — CHANGES_REQUESTED** [review]
Wong branch name

**Review by maxammann — APPROVED** [review]

**Review by maxammann — APPROVED** [review]

File: backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardVerifier.kt [hunk 1]
```diff
@@ -1,24 +1,27 @@
...
 import com.eatthepath.otp.TimeBasedOneTimePasswordGenerator
 import org.jetbrains.exposed.sql.transactions.transaction
 import java.time.Duration
 import java.time.Instant
+import java.time.LocalDateTime
 import javax.crypto.spec.SecretKeySpec
 
 val TIME_STEP: Duration = Duration.ofSeconds(30)
-const val TOTP_LENGTH = 10
+const val TOTP_LENGTH = 6
```
maxammann, line 13: We use 6 as 10 is out of range. Actually we could use any value in range here security wise.

RoAnBu (approved):

elKei24 (approved):

PR #219, opened 2021-01-31
217: Properly hash card details

### LINKED ISSUE #217: Properly hash card details in the administration
Currently, the administration hashes the card details only via JSON.stringify. Use the binary representation as in the protocol.

Relates to #83

Schedulaar:
Closes #217

### DIFF
--- a/administration/src/util/generateHashFromCardDetails.ts
+++ b/administration/src/util/generateHashFromCardDetails.ts
@@ -10,17 +10,25 @@ export interface CardDetails {
 
 const cardDetailsToBinary = (cardDetails: CardDetails) => {
     const fullNameWithoutZero = new TextEncoder().encode(cardDetails.fullName)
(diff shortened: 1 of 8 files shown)


administration/src/util/generateHashFromCardDetails.ts
```diff
@@ -8,26 +8,43 @@ export interface CardDetails {
     cardType: CardActivateModel.CardType,
 }
 
+const cardDetailsToBinary = (cardDetails: CardDetails) => {
+    const fullNameWithoutZero = new TextEncoder().encode(cardDetails.fullName)
+    const binary = new Uint8Array(
+        8 // int64 expirationDate
```
maxammann, line 14: I think is would be good to define constants here. These can  then be reused in line 21 for example

maxammann (approved):

elKei24 (approved):

Issue #227, opened 2021-02-04
Build and provide .deb packages for backend/administration


elKei24:
Part of #53

Acceptence criteria:
- CI builds deb packages for backend/administration
- deb packages are stored somewhere (e.g. ftp folder, github releases or both)

elKei24:
Left to do:
- [ ] styles.json, sprites, …
- [ ] .deb signing

elKei24:
Nothing left to do, rest moved to #235 and #236

PR #228, opened 2021-02-04
#227 build basic backend deb package

elKei24:
Does not close issue #227 yet, but is a working intermediate step and prevents all other PRs from getting red.

Builds a basic .deb package containing the built backend and a systemd service file for starting it.

### DIFF
--- a/.idea/ehrenamtskarte.iml
+++ b/.idea/ehrenamtskarte.iml
@@ -2,11 +2,7 @@
 <module type="WEB_MODULE" version="4">
   <component name="NewModuleRootManager" inherit-compiler-output="true">
     <exclude-output />
-    <content url="file://$MODULE_DIR$">
-      <excludeFolder url="file://$MODULE_DIR$/administration" />
-      <excludeFolder url="file://$MODULE_DIR$/backend" />
-      <excludeFolder url="file://$MODULE_DIR$/frontend" />
-    </content>
+    <content url="file://$MODULE_DIR$" />
     <orderEntry type="sourceFolder" forTests="false" />
   </component>
 </module>
--- a/backend/gradle.properties
+++ b/backend/gradle.properties
@@ -2,11 +2,9 @@
 
 exposed_version=0.28.1
 
-app.postgres.host=localhost
-app.postgres.port=5432
+app.postgres.url=jdbc:postgresql://localhost:5432/ehrenamtskarte
 app.postgres.user=postgres
 app.postgres.password=postgres
-app.postgres.database=ehrenamtskarte
 
 app.import.xml=https://www.lbe.bayern.de/engagement-anerkennen/ehrenamtskarte/akzeptanzstellen/app-daten.xml
 app.import.json_1=https://bayerische_eak:
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
@@ -14,8 +14,8 @@ class WebService {
             if (!production) {
                 cfg.enableDevLogging()
                 cfg.enableCorsForAllOrigins()
-                cfg.addStaticFiles("/graphiql", "/graphiql", Location.CLASSPATH)
             }
+            cfg.addStaticFiles("/graphiql", "/graphiql", Location.CLASSPATH)
         }.start(host, port)
 
         println("Server is running at http://${host}:${port}")
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
@@ -3,18 +3,21 @@ package app.ehrenamtskarte.backend.common.webservice
 import io.javalin.Javalin
 import io.javalin.http.staticfiles.Location
 
-const val PORT = 7000
+const val DEFAULT_PORT = "7000"
 
 class WebService {
     fun start() {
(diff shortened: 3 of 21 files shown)


elKei24:
@maxammann  👌👌

maxammann (approved):

backend/gradle.properties
```diff
@@ -2,11 +2,9 @@
 
 exposed_version=0.28.1
 
-app.postgres.host=localhost
-app.postgres.port=5432
+app.postgres.url=jdbc:postgresql://localhost:5432/ehrenamtskarte
```
Schedulaar, line 5: didn't we have some problems with that?

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
```diff
@@ -3,18 +3,23 @@ package app.ehrenamtskarte.backend.common.webservice
...
         val app = Javalin.create { cfg ->
-            cfg.enableDevLogging()
-            cfg.enableCorsForAllOrigins()
-            cfg.addStaticFiles("/graphiql", "/graphiql", Location.CLASSPATH)
-        }.start(PORT)
+            if (!production) {
+                cfg.enableDevLogging()
+                cfg.enableCorsForAllOrigins()
+                cfg.addStaticFiles("/graphiql", "/graphiql", Location.CLASSPATH)
+            }
```
Schedulaar, line 18: do we disable graphiql in production? I'd probably vote for having it enabled :D

.idea/ehrenamtskarte.iml
```diff
@@ -2,11 +2,7 @@
 <module type="WEB_MODULE" version="4">
   <component name="NewModuleRootManager" inherit-compiler-output="true">
     <exclude-output />
-    <content url="file://$MODULE_DIR$">
-      <excludeFolder url="file://$MODULE_DIR$/administration" />
-      <excludeFolder url="file://$MODULE_DIR$/backend" />
-      <excludeFolder url="file://$MODULE_DIR$/frontend" />
-    </content>
+    <content url="file://$MODULE_DIR$" />
```
Schedulaar, line 5: Hmm... I think these changes should not be necessary....

docker/reverse_proxy/nginx-development.conf
```diff
@@ -1,25 +1,39 @@
...
+
+    location = /map.html {
+    }
+
+    location = /style.json {
+    	add_header 'Access-Control-Allow-Origin' '*';
+
+    	sub_filter_once off;
+    	sub_filter_types application/json;
+        sub_filter "https://tiles.staging.ehrenamtskarte.app" "http://localhost:5002";
```
Schedulaar, line 21: is that right for development?

docker/reverse_proxy/www/map.html
```diff
@@ -0,0 +1,21 @@
+<html lang="en">
+<head>
+    <script src='https://api.mapbox.com/mapbox-gl-js/v2.1.0/mapbox-gl.js'></script>
+    <link href='https://api.mapbox.com/mapbox-gl-js/v2.1.0/mapbox-gl.css' rel='stylesheet' />
+    <title>Ehrenamtskarte Preview Map</title>
+</head>
+<body>
+<div id='map' style='width: 100%; height: 100%;'></div>
+<script>
+  mapboxgl.accessToken = 'pk.eyJ1IjoibWF4YW1tYW5uIiwiYSI6ImNraHlzdGNmejBkaGsycm9hNjJjcWZzdmwifQ.3NaekT5CjDsBnBNdUr1F7g';
```
Schedulaar, line 10: whoopsie?

Schedulaar (approved):

docker/reverse_proxy/www/map.html
maxammann: Noopsie just an alreadly public key which is not used. Dummy key.

backend/gradle.properties
maxammann: This is equivalent to the previous way. I no longer concat the url in java but pass the whole one here.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
maxammann: I think this is a tool for debugging and should not be exposed in production. Also you can always use intellij for sending graphql queries.

Not sure whether this tool is intended for production and is tested against vulnerabilities.

.idea/ehrenamtskarte.iml
maxammann: They also do not harm :) i tested this. No differnce with or without these changes.

docker/reverse_proxy/nginx-development.conf
maxammann: Seems right. It gets replaced with localhost like before.

maxammann:
Can be merged from my side.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
Schedulaar: To me, its just a public access to the api, which shouldn't do any harm...

maxammann: I see this more like a standalone application which could be deployed to production. I mean it is like the administration backend. I added it again because there is no problem from security perspective as these are just static files.

maxammann:
I merged this now, for the rest we should create separate tasks

PR #230, opened 2021-02-04
222 backend verification request

### LINKED ISSUE #222: Add backend request for verification of EAK on other device
Refers to #176. Followup of #177

Necessary changes:
- Adjust to current grapqhl state in main
- ~~Adjust hashing model to fit changes like in #217~~ Moved to #226
- Use graphql provider pattern to query the backend if the given card is valid

### LINKED ISSUE #226: Hash Carddetails for verification in frontend
Refers to #176.

Adjust hashing model to fit changes like in #217. 

Current placeholder implementation: https://github.com/ehrenamtskarte/ehrenamtskarte/blob/main/frontend/lib/verification/verification_hasher.dart

RoAnBu:
closes #222 
closes #226


elKei24 (approved):
(see below)

frontend/lib/identification/identification_qr_content_parser.dart
```diff
@@ -86,24 +75,19 @@ class IdentificationQrContentParser {
           unixExpirationDate == null) {
         return QRCodeParseResult(
```
elKei24, line 76: Well, it probably won't surprise you that I still am no fan of this error handling. Especially now when we think about i18n. It smells pretty fishy that we then would have to add translations to this parser class for example, and not to the UI class.

Suggestion: we could add as example a `ParserException` class. One could then e.g. add a subclass `ExpirationDateMissingParserException` and throw an instance of that here instead of returning a `QRCodeParserResult` with `error == true` and the german error message. One could then catch `ExpirationDateMissingParserException` in the UI class separately and show the message, if we want to inform the user about what is missing in detail, or just catch all `ParserException`s at once, display a generic warning to the user and still print a nice log message with details.

frontend/graphql_queries/verification/card_verification_by_hash.graphql
```diff
@@ -0,0 +1,3 @@
+query CardVerificationByHash($card: CardVerificationModelInput!) {
+    verifyCard(card: $card)
```
Schedulaar, line 2: I think you can declare a result variable here:
```suggestion
    cardValid: verifyCard(card: $card)
```

Schedulaar (approved):

frontend/lib/identification/base_card_details.dart
```diff
@@ -2,12 +2,13 @@ class BaseCardDetails {
   final String fullName;
   final String hashSecretBase64;
   final int unixExpirationDate;
-  final String cardType;
+  final int cardType;
```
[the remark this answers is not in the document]
RoAnBu: Hm I tried, do you think it is better now? @Schedulaar Changes here: https://github.com/ehrenamtskarte/ehrenamtskarte/pull/230/commits/2e42a0c69df2903032603e15e47cabc5af1fb5c1

frontend/lib/identification/identification_qr_content_parser.dart
RoAnBu: @elKei24 Do you think this is better? https://github.com/ehrenamtskarte/ehrenamtskarte/blob/222-verification-backend-request/frontend/lib/verification/scanner/verification_qr_content_parser.dart

I haven't changed anything in the identification parser yet

elKei24: Well, the UI message is still in the same file, but I find it cleaner now 👍🏽

frontend/lib/identification/base_card_details.dart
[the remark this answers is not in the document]
Schedulaar: Yup, seems better :+1:

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardVerifier.kt
```diff
@@ -19,7 +19,7 @@ object CardVerifier {
     }
 
     private fun verifyCard(card: CardEntity, totp: Int): Boolean =
```
aiquita, line 21: Just for convenience: You can omit the return type in cases where the right side is an expression

PR #237, opened 2021-02-11
#235 package styles.json

elKei24:
Closes #235

Closes #243

PR #239, opened 2021-02-11
#236 package martin

### LINKED ISSUE #236: Package/Deploy martin

elKei24:
Contains #237. Closes #236.


maxammann (approved):

.circleci/config.yml
```diff
@@ -109,10 +113,67 @@ jobs:
...
+    working_directory: ~/project/mapbox/martin
+    steps:
+      - checkout:
+          path: ~/project
+      - run:
+          name: "Download martin"
+          command: |
+            downloadUrl=$(curl -sH "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/urbica/martin/releases/latest" | grep -o "http.*martin-Linux.*.tar.gz")
+            echo "Downloding and extracting $downloadUrl …"
+            wget -q -O- "$downloadUrl" | tar -xz
```
maxammann, line 168: It would be very cool if we would build martin from source here. Else there is not really an advantage to the current salt deployment.

maxammann (changes requested):

.circleci/config.yml
elKei24: Sure, I can do that. What is the problem with the pre-compiled Martin?

maxammann: Its just that this is not an improvement but more complexity in the deployment.
The martin from source has severy security fixes. Also it is more "trustworthy" than that binary.

maxammann: But I think this does not really have priority.

.circleci/config.yml
```diff
@@ -109,10 +113,67 @@ jobs:
...
+    steps:
+      - checkout:
+          path: ~/project
+      - run:
+          name: "Download glyphs"
+          command: |
+            wget -O /tmp/glyphs.zip https://github.com/orangemug/font-glyphs/archive/gh-pages.zip
+            unzip /tmp/glyphs.zip
+            mv font-glyphs* font-glyphs
+            rm -f /tmp/glyphs.zip
```
Schedulaar, line 140: What are the glyphs for?

Schedulaar (approved):
Can't really judge the code...

.circleci/config.yml
elKei24: They are the fonts used on the map

.circleci/config.yml
elKei24: Done

.circleci/config.yml
Schedulaar: Ah, but they have nothing to do with martin, right?

.circleci/config.yml
```diff
@@ -83,6 +90,47 @@ jobs:
...
+      - image: cimg/rust:1.50.0
+    working_directory: ~/martin
+    steps:
+      - run:
+          name: "Checkout"
+          command: |
+            mkdir -p ~/.ssh
+            touch ~/.ssh/known_hosts
+            echo "github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==" >> ~/.ssh/known_hosts
+            git clone git@github.com:urbica/martin.git ~/martin
```
maxammann, line 104: You should checkout a specific ref here.

maxammann (approved):

.circleci/config.yml
elKei24: Nope, they are just there because of this branch containing the changes of #237 already.

.circleci/config.yml
elKei24: Done

maxammann (approved):

PR #241, opened 2021-02-11
238 eak without expiration date

### LINKED ISSUE #238: Support EAKs without expiration date
Golden EAKs don't have an expiration date. 
This has to be supported by the administration frontend and the app when creating and scanning QR codes.
Refers to #176 and #17

PaulaWi:
closes #238


administration/src/components/generation/CardCreationModel.ts
```diff
@@ -6,15 +6,15 @@ export interface CardCreationModel {
     id: number,
     forename: string,
     surname: string,
-    expirationDate: Date,
+    expirationDate: Date | null,
```
maxammann, line 9: I think this should not be nullable after our last discussion.

A new flag should be introduced: `infiniteValid`

maxammann, line 9: I think this should not be nullable after our last discussion.

A new flag should be introduced: `infiniteValid`

maxammann (changes requested):

administration/src/components/generation/CardCreationModel.ts
Schedulaar: I think this should be fine. Nullables are not evil here

administration/src/components/generation/CardCreationModel.ts
```diff
@@ -6,15 +6,15 @@ export interface CardCreationModel {
...
 export const isNameValid = (value: string) => value.length > 0 && value.length < MAX_NAME_LENGTH
 
-export const isExpirationDateValid = (value: Date) => value > new Date()
+export const isExpirationDateValid = (value: Date | null) => value !== null && value > new Date()
 
 export const isValid = (cardCreationModel: CardCreationModel) =>
     isNameValid(cardCreationModel.forename)
     && isNameValid(cardCreationModel.surname)
-    && isExpirationDateValid(cardCreationModel.expirationDate)
+    && (isExpirationDateValid(cardCreationModel.expirationDate) || cardCreationModel.cardType === CardType.gold)
```
Schedulaar, line 20: I think we don't need the `cardType` check, do we? If we set CardType to gold, then we should also set expiration date to null automatically (already in the input form)

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
```diff
@@ -27,10 +27,14 @@ class CardEntity(id: EntityID<Int>) : IntEntity(id) {
...
     var cardDetailsHash by Cards.cardDetailsHash
 
-    var expirationDate: LocalDateTime
-        get() = LocalDateTime.ofEpochSecond(expirationDateEpochSeconds, 0, ZoneOffset.UTC)
+    var expirationDate: LocalDateTime?
+        get() = if (expirationDateEpochSeconds > 0) LocalDateTime.ofEpochSecond(
+            expirationDateEpochSeconds,
+            0,
+            ZoneOffset.UTC
+        ) else null
```
Schedulaar, line 35: I think we should rather make `Cards.expirationDate` nullable in the database, otherwise this seems pretty hacky.
Also I think it would be better, if we had expirationDate as a nullable already in GraphQL to make it explicit...

Schedulaar:
For protobuf nullables:
Either we declare a special number, i.e. `Long.Max` as the 
Other than that, we could follow this guy
https://itnext.io/protobuf-and-null-support-1908a15311b6
who suggest (I think) to introduce some new type, i.e. `NullableLong` with a `oneOf` selector

maxammann:
> For protobuf nullables:
> Either
> 
>     1. we declare a special number, i.e. `Long.Max` (seems a little bit better than 0) as the "null-value" of expirationdate or
> 
>     2. we could follow this guy
>        https://itnext.io/protobuf-and-null-support-1908a15311b6
>        who suggest (I think) to introduce some new type, i.e. `NullableLong` with a `oneOf` selector
> 
> 
> I'm kind of in favor of the second option here as it makes everything explicit.
> Also we should document our convention for the hashing procedure in the artefacts repo...

There is experimental Nullable support: https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md
But I'm not sure whether flutter supports this already. But we can already do this manually like michael suggested:

```proto
syntax = "proto3";

message Foo {
  // Experimental feature, not generally supported yet!
  optional int32 foo = 1;

  // Internally rewritten to:
  // oneof _foo {
  //   int32 foo = 1 [proto3_optional=true];
  // }
  //
  // We call _foo a "synthetic" oneof, since it was not created by the user.
}
```

We can just use the oneof workaround

Schedulaar (approved):

PR #247, opened 2021-02-15
231 periodical import

aiquita:


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/StoreToDatabase.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    private fun drawSuccessBar(done: Int, of: Int) {
+        if (done % 200 == 0) {
+            var bar = ""
+            repeat(done / 200) { bar += "##" }
+            var remaining = ""
+            repeat((of - done) / 200) { remaining += "__" }
+            println("[$bar$remaining]")
```
elKei24, line 61: Just a trifle: I have read one can replace lines on the console using carriage return (https://stackoverflow.com/a/11777847/3987396)

elKei24 (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/FilterRawData.kt
```diff
@@ -0,0 +1,25 @@
...
+    override fun execute(acceptingStores: List<LbeAcceptingStore>): List<LbeAcceptingStore> = acceptingStores.filter { filterLbe(it) }
+
+    private fun filterLbe(store: LbeAcceptingStore) = try {
+        val builder = FilterBuilder(store, monitor)
+        builder.filterLongitudeAndLatitude()
+                && builder.filterPostalCode()
+                && builder.filterCategory()
+
+    } catch (e: Exception) {
+        monitor.addMessage("$store was filtered out because of unknwon exception while filtering", e)
```
Schedulaar, line 19:
```suggestion
        monitor.addMessage("$store was filtered out because of unknown exception while filtering", e)
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/RawDataFilterBuilder.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    fun filterLongitudeAndLatitude()
+            = if (store.longitude.isNullOrBlank() || store.latitude.isNullOrBlank()) {
+        monitor.addMessage("$store was filtered out because longitude or latitude were invalid")
+        false
+    } else {
+        true
+    }
+
+    fun filterPostalCode() = if(store.postalCode == null || store.postalCode?.trim()?.length == 5) {
```
RoAnBu, line 15: this one might be a bit too harsh. I think there are some instances where the accepting store looks fine but the postal code included e.g. the city name. Maybe a warning would be better here instead of blocking it completely

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/ImportMonitor.kt
```diff
@@ -0,0 +1,20 @@
+package app.ehrenamtskarte.backend.stores.importer
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+class ImportMonitor(val writeToStout: Boolean) {
```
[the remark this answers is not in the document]
RoAnBu: it's used by javalin and it's really simple to use and setup in case we want to use it

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/ImportMonitor.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    val messages = ArrayList<String>()
+
+    fun addMessage(message: String) {
+        val wrappedMessage = "[" + LocalDateTime.now().format(DateTimeFormatter.ISO_LOCAL_DATE_TIME) + "]" + "\t" + message
```
RoAnBu, line 11: here slf4j would probably be helpful as that can be done automatically :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/DataImporter.kt
```diff
@@ -0,0 +1,29 @@
...
+                .addStep(Encoding(monitor)) { monitor.addMessage("== Handle encoding issues ==") }
+                .addStep(StoreToDatabase(monitor)) { monitor.addMessage("== Store remaining data to db ==") }
+        }
+
+        try {
+            pipe()
+            monitor.addMessage("== Pipeline successfully finished ==")
+        } catch (e : Exception) {
+            monitor.addMessage("== Pipeline was aborted without altering the database ==")
+        }
```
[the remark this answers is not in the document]
RoAnBu: I agree!

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Encoding.kt
```diff
@@ -0,0 +1,56 @@
...
+                it.discount?.decodeSpecialCharacters()?.sanitizeForDb(),
+                it.categoryId
+            )
+        } catch (e: Exception) {
+            monitor.addMessage("Exception while encoding $it for database import:\n$e.message")
+            null
+        }
+    }.filterNotNull()
+
+    private val unspecifiedNoteRegex = Regex("""^\s*(n\.?\s*A\.?|)\s*$""")
```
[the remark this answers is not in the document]
RoAnBu: n.a. would not be caught here right? This is where unit tests would be handy :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Download.kt
```diff
@@ -0,0 +1,43 @@
...
+
+            val xmlMapper = XmlMapper()
+            xmlMapper.registerModule(KotlinModule())
+            xmlMapper.enable(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT)
+            val lbeData = xmlMapper.readValue(response, LbeData::class.java)
+
+            return lbeData.acceptingStores
+        } catch (e: Exception) {
+            monitor.addMessage("Unknown exception while downloading data from lbe", e)
+            throw e
```
RoAnBu, line 39: Does that crash the whole backend?

RoAnBu (changes requested):
Is the category import now missing completely? How do we populate the database on the first start, for example for local testing?

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/StoreToDatabase.kt
```diff
@@ -0,0 +1,65 @@
...
+                }
+            } catch (e: Exception) {
+                monitor.addMessage("Unknown exception while storing to db", e)
+                rollback()
+                throw e
+            }
+        }
+    }
+
+    private fun drawSuccessBar(done: Int, of: Int) {
```
maxammann, line 55: I think we should remove this if we log the stdout to the system logs. Else the progressbar will end up in the logs.

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/RawDataFilterBuilder.kt
aiquita: Problem: What to with the data else?
I don not like to extend the postal code field to an text field of arbitrary length just because of some instances that do not fit into.

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Download.kt
aiquita: No it is catched above.

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Encoding.kt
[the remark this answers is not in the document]
aiquita: I do not understand this either ...
Who has build this?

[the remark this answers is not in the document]
elKei24: Eeehhm, that probably was me 😄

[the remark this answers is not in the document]
elKei24: Oh, I forgot to answer the questions. I think the brackets and the `|` are there to also detect whitespace-only entries as unspecified. Probably would be better to just make the group optional …
Not catching `n.a.` was no problem when I wrote this, as we had no `n.a.` in the data at that moment.

backend/src/main/resources/import/categories.json
```diff
@@ -34,5 +34,9 @@
   {
     "id": 8,
     "name": "Essen/Trinken/Gastronomie"
+  },
+  {
+    "id": 9,
+    "name": "Sonstiges"
   }
```
Schedulaar, line 41: I'd prefer to make category nullable, but it does not reaaly matter I guess...

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/FilterRawData.kt
```diff
@@ -0,0 +1,25 @@
...
+    override fun execute(acceptingStores: List<LbeAcceptingStore>): List<LbeAcceptingStore> = acceptingStores.filter { filterLbe(it) }
+
+    private fun filterLbe(store: LbeAcceptingStore) = try {
+        val builder = FilterBuilder(store, logger)
+        builder.filterLongitudeAndLatitude()
+                && builder.filterPostalCode()
+                && builder.filterCategory()
+
+    } catch (e: Exception) {
+        logger.info("$store was filtered out because of unknown exception while filtering", e)
```
Schedulaar, line 19: What can go wrong when filtering?  :D
```suggestion
        logger.info("$store was filtered out because of an unknown exception while filtering", e)
```

Schedulaar (approved):
I did not test it, but the code looks good to me.

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/FilterRawData.kt
aiquita: Probably nothing.

PR #252, opened 2021-02-25
Use official rust container which is based on debian

maxammann:


Schedulaar (approved):
Awesome, I love it

maxammann:
@elKei24 fixes execution on debian

elKei24:
Oh, interesting 👍

PR #261, opened 2021-03-01
#248 update some dart packages

### LINKED ISSUE #248: Upgrade dart packages
As our app evolves, so do the packages we depend on. Some of them had major version updates in the meantime. In order to make use of new features/bugfixes and avoid huge migrations in the future, we should:
- [x] upgrade all dependencies that we can upgrade
- [x] fix the code where neccessary, there are breaking changes e.g. for
    - `qr_code_scanner`
    - `graphql_flutter`
- [x] check if we can simplify our code using the new features
- [x] update mapbox library

elKei24:
Closes #248

### DIFF
--- a/frontend/android/gradle.properties
+++ b/frontend/android/gradle.properties
@@ -1,4 +1,4 @@
 org.gradle.jvmargs=-Xmx1536M
 android.useAndroidX=true
 android.enableJetifier=true
-android.enableR8=true
+# android.enableR8=true
--- a/frontend/android/gradle.properties
+++ b/frontend/android/gradle.properties
@@ -1,4 +1,4 @@
 org.gradle.jvmargs=-Xmx1536M
 android.useAndroidX=true
 android.enableJetifier=true
-# android.enableR8=true
+android.enableR8=true
--- a/frontend/lib/qr_code_scanner/qr_code_scanner_controls.dart
+++ b/frontend/lib/qr_code_scanner/qr_code_scanner_controls.dart
@@ -12,12 +12,10 @@ class QRCodeScannerControls extends StatelessWidget {
   // state and just call setState({}) to trigger redrawing of whole widget.
   final StreamController<bool> flashStreamController = StreamController();
   final StreamController<CameraFacing> cameraStreamController =
-    StreamController<CameraFacing>();
+      StreamController<CameraFacing>();
 
-  QRCodeScannerControls({
-    @required this.controller,
-    Key key
-  }) : super(key: key) {
+  QRCodeScannerControls({@required this.controller, Key key})
+      : super(key: key) {
     updateFlashStream();
     updateCameraStream();
   }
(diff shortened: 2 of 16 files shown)


elKei24:
Left to do:
- [ ] update mapbox

frontend/android/gradle.properties
```diff
@@ -1,4 +1,4 @@
 org.gradle.jvmargs=-Xmx1536M
 android.useAndroidX=true
 android.enableJetifier=true
-android.enableR8=true
+# android.enableR8=true
```
Schedulaar, line 4: Why not just remove it?

elKei24, line 4: We seem not to need this, and it is deprecated and causes warnings.

frontend/pubspec.yaml
```diff
@@ -26,12 +26,12 @@ dependencies:
   intl: ^0.16.1
   mapbox_gl: ^0.9.0
   provider: ^4.3.2+2
-  graphql_flutter: ^3.1.0
+  graphql_flutter: ^4.0.1
   url_launcher: ^5.7.10
   maps_launcher: ^1.2.2+2
-  flutter_svg: ^0.18.0
+  flutter_svg: 0.19.1
```
Schedulaar, line 32: Why not 
```suggestion
  flutter_svg: ^0.19.1
```

elKei24:
Maybe the multidex is only necessary due to disabled R8. I will check this before merging

RoAnBu (changes requested):
qr code scanner now broken on ios, i will investigate...

frontend/lib/qr_code_scanner/qr_code_scanner_controls.dart
```diff
@@ -0,0 +1,75 @@
...
+  Widget build(BuildContext context) {
+    return FutureBuilder<SystemFeatures>(
+        future: controller.getSystemFeatures(),
+        builder: (context, snapshot) =>
+      Row(
+          mainAxisAlignment: MainAxisAlignment.center,
+          crossAxisAlignment: CrossAxisAlignment.center,
+          children: <Widget>[
+            if (snapshot.data.hasFlash)
+              Container(
```
RoAnBu, line 40: On iOS in the qr_code_scanner_controls.dart, the first call on `controller.getSystemFeatures()` causes a `CameraException(404, null)`. From testing on my ipad, it takes about 60ms until the first request succeeds.

One solution could be to wrap `controller.getSystemFeatures()` with a retry mechanism like this:
```dart
  Future<SystemFeatures> tryToGetSystemFeatures() async {
    for (var i = 0; i < 10; i++) {
      await Future.delayed(Duration(milliseconds: 100));
      try {
        return await controller.getSystemFeatures();
      } on CameraException {
        continue;
      }
    }
    return null;
  }
```

A hardcoded delay on ios with like 100ms might work, too, but i dont know if it might take longer than 100ms on some older ios devices.

RoAnBu (changes requested):

frontend/lib/qr_code_scanner/qr_code_scanner_controls.dart
elKei24: Oh man 😄 I added your solution with the retry. Could you test it again?

If it still does not work, we can just throw this system feature check away and just always display both buttons as before. Found it a quick win, but did not expect more of these strange iOS bugs 😄

frontend/pubspec.yaml
elKei24: Because I tested `0.19.2` and build failed due to incompatible SDK versions. But seems `flutter pub` is clever enough to still use `0.19.1`, so we can add the `^` 👍🏽

frontend/android/gradle.properties
elKei24: Because Flutter automatically adds it again if it does not find the setting in the file (regardless of being commented out or not). 🤦🏽‍♂️
See also https://github.com/flutter/flutter/issues/52005.

frontend/lib/qr_code_scanner/qr_code_scanner_controls.dart
RoAnBu: Hm, found at least two other bugs when scanning random qr codes... I'll try to fix them

RoAnBu: Wow, seems like the base32 encoding library we use crashes the thread when given a null value which happened in my test. The interesting thing is that there is no exception that can be caught, also the debugger sees nothing. The thread just exits silently (but the app keeps running). Wtf flutter/dart

RoAnBu (approved):
Ok, everything should work now on iOS!

elKei24:
> Ok, everything should work now on iOS!

Very nice, thank you @RoAnBu!

PR #277, opened 2021-03-07
275: Add card region to frontend

Schedulaar:


frontend/pubspec.lock
```diff
@@ -1072,5 +1072,5 @@ packages:
     source: hosted
     version: "2.2.1"
 sdks:
-  dart: ">=2.10.0 <2.11.0"
-  flutter: ">=1.22.0 <2.0.0"
+  dart: ">=2.12.0-0.0 <3.0.0"
+  flutter: ">=1.22.0"
```
elKei24: Does this mean we can use the latest flutter version in Cirrus again, too?

elKei24 (approved):

RoAnBu (approved):

frontend/pubspec.lock
Schedulaar: Hmm... I guess so, I upgraded my local flutter version and I think I had to bump flutter_svg, but other than it worked...

PR #282, opened 2021-03-09
use latest flutter version and update dependencies

elKei24:


elKei24:
> Wait is that everything we need to change for Flutter 2.0?

It is even more as I updated the packages, too 😄 
But well, this does not yet make use of new features like the sound null-safety.

Schedulaar (approved):

PR #287, opened 2021-03-09
#284 login for administration backend

### LINKED ISSUE #284: Basic login to administration BE
Usage of the administration BE should only be able after login.

_Out of scope_:
- creating new users (#285)
- rights and roles (#286)

Part of #280

elKei24:
Backend part of #284. Does not yet close the ticket, as the frontend part is still missing.


elKei24:
> Please consider adding a mechanism to restore your password by email.

I consider this out of scope for now. This PR is for a basic login only, not even containing registration/rights&roles.

Schedulaar:
I agree with @elKei24 about the scope. Nevertheless, I think it would be good to have *some* sort of email mechanism built in at some point. This could also come in handy for sending "confirmation mails" for incoming applications.

aiquita:
@elKei24 Using the mail address as username simplifies things. Then you already have it at hand in order you need it later. Moreover, a chosen username has no benefit.

elKei24:
> @elKei24 Using the mail address as username simplifies things. Then you already have it at hand in order you need it later. Moreover, a chosen username has no benefit.

Well, it is easier to handle change of mail address if it only is an additional field. Moreover, we are more privacy friendly if you can opt in to receiving mails and giving us the email address if you want only.

Schedulaar:
@elKei24 As "reset your password" is a standard use case, which we definitely want to support (at some point), I guess, we always need to have an email-address anyway... Changing an email-address is rather uncommon (I think) and having a mandatory email address does not make it harder :)
I dont have any opinion on whether we should have additional user names, but to me, email address as login parameter would suffice.

elKei24:
@aiquita @Schedulaar replaced username by email

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/JwtService.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlin.NoSuchElementException
+
+object JwtService {
+    private val secret by lazy {
+        System.getenv("JWT_SECRET")
+        ?: throw NoSuchElementException("Environment variable JWT_SECRET not set")
+    }
+    private val algorithm by lazy { Algorithm.HMAC512(secret) }
+
+    fun create(administrator: Administrator): String =
```
Schedulaar, line 18: I guess:
```suggestion
    fun createToken(administrator: Administrator): String =
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/JwtService.kt
```diff
@@ -0,0 +1,30 @@
...
+        JWT.create()
+            .withClaim(JwtPayload::email.name, administrator.email)
+            .withClaim(JwtPayload::userId.name, administrator.id)
+            .withExpiresAt(Date.from(Instant.now().plus(1, ChronoUnit.HOURS)))
+            .sign(algorithm)
+
+    fun verify(token: String) = JWT.require(algorithm).build().verify(token).claims.let {
```
Schedulaar, line 25: And then consequently
```suggestion
    fun verifyToken(token: String) = JWT.require(algorithm).build().verify(token).claims.let {
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/JwtService.kt
```diff
@@ -0,0 +1,30 @@
...
+object JwtService {
+    private val secret by lazy {
+        System.getenv("JWT_SECRET")
+        ?: throw NoSuchElementException("Environment variable JWT_SECRET not set")
+    }
+    private val algorithm by lazy { Algorithm.HMAC512(secret) }
+
+    fun create(administrator: Administrator): String =
+        JWT.create()
+            .withClaim(JwtPayload::email.name, administrator.email)
```
Schedulaar, line 20: where does `JwtPayload` come from?

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/JwtService.kt
```diff
@@ -0,0 +1,30 @@
...
+            .withClaim(JwtPayload::userId.name, administrator.id)
+            .withExpiresAt(Date.from(Instant.now().plus(1, ChronoUnit.HOURS)))
```
Schedulaar, line 22: Does this mean, a token is valid for one hour until it is created (or is it extended, if the user stays active?)

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/types/Administrator.kt
```diff
@@ -0,0 +1,6 @@
+package app.ehrenamtskarte.backend.auth.webservice.schema.types
+
+class Administrator(
+    val id: Int,
```
Schedulaar, line 4: I'm not sure, whether we should leak the internal id of the user. Shouldn't matter to the outside world I guess...

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLContext.kt
```diff
@@ -0,0 +1,10 @@
+package app.ehrenamtskarte.backend.common.webservice
+
+import app.ehrenamtskarte.backend.auth.webservice.JwtPayload
+import com.expediagroup.graphql.execution.GraphQLContext
+
+data class GraphQLContext(
+    val jwtPayload: JwtPayload?
+) : GraphQLContext {
+    val isSignedIn get() = jwtPayload != null
```
Schedulaar, line 9: should we verify the payload here? Or is the payload never in the context if it is not verified?

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLHandler.kt
```diff
@@ -69,15 +73,40 @@ class GraphQLHandler(
         return result
     }
 
+    private fun getJwtTokenFromHeader(context: Context): String? {
+        val header = context.header("Authorization") ?: return null
+        val split = header.split(" ")
+        return if (split.size != 2 || split[0] != "Bearer") null else split[1]
+    }
```
Schedulaar, line 80: As this involves cross-origin headers, we have to be a little bit careful... @maxammann Do you remember what we need to take into account here s.t. browsers don't block fetches with headers from `druckerei.*` to `api.*`? I think we solved this problem in some prior project...

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -12,7 +14,8 @@ import java.time.ZoneOffset
 @Suppress("unused")
 class CardMutationService {
     @GraphQLDescription("Stores a new digital EAK")
-    fun addCard(card: CardGenerationModel): Boolean {
+    fun addCard(context: GraphQLContext, card: CardGenerationModel): Boolean {
+        if (!context.isSignedIn) throw GraphQLException("Unauthorized")
```
Schedulaar, line 18: maybe it would be nice, if we would throw sth like an actual `UnauthorizedException` and catch it in the handler and return a `401` status code... Or do we already do that magically?

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/webservice/dataloader/RegionDataLoader.kt
```diff
@@ -7,7 +7,7 @@ import org.dataloader.DataLoader
 import org.jetbrains.exposed.sql.transactions.transaction
 import java.util.concurrent.CompletableFuture
 
-const val REGION_LOADER_NAME = "CATEGORY_LOADER"
+const val REGION_LOADER_NAME = "REGION_LOADER"
```
Schedulaar, line 10: :heart_eyes:

Schedulaar (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLContext.kt
elKei24: > Or is the payload never in the context if it is not verified?

This

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/types/Administrator.kt
elKei24: I would like to keep it. Does not matter to the outside world, but to the inside. If for example we allow to change the email address at one point, the JWT still stays valid (and should), but we would fail to look up the user in our database. It makes everything easier (and probably faster) for us if we use the ID and not the email to identify users.

And I see no risk at publishing the ID. If we really would care about not giving away any metadata at all, we could use UUIDs.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLHandler.kt
elKei24: Something like that: https://www.sjoerdlangkemper.nl/2018/09/12/authorization-header-and-cors/#by-the-application ?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
elKei24: Done

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/JwtService.kt
elKei24: from line 30 😉

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/JwtService.kt
elKei24: Yes, it is valid for 1h and not extended at the moment.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLHandler.kt
Schedulaar: Probably, yes :+1: I thought, it might be harder than that though... But yeah :sweat_smile:

Schedulaar (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLHandler.kt
elKei24: @Schedulaar I would add that when implementing the login on frontend side though, so that I can test it properly.

RoAnBu (approved):

.idea/misc.xml
```diff
@@ -2,6 +2,7 @@
 <project version="4">
   <component name="ExternalStorageConfigurationManager" enabled="true" />
   <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="web" url="file://$PROJECT_DIR$/backend" />
```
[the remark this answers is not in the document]
elKei24: Does not IntelliJ want to put that there all the time on your machine?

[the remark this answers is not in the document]
RoAnBu: Yes but I think that causes intellij to only recognize the backend as a module and not recognize the flutter project. I have to rollback that line all the time, too...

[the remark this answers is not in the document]
Schedulaar: I think intellij has problems independent of this line... I checked the branch out and it works for me right now... Until intellij decides otherwise I gues...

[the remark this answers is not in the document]
elKei24: Okay, I removed the line.
Never again IntelliJ files in a repo 😅

PR #294, opened 2021-03-11
upgrade flutter in ios build

RoAnBu:


elKei24:
@RoAnBu It seems like we can not run `flutter upgrade` on the linux containers as the version is pinned there and thus allows no upgrades. We did not notice this as the cache was used for your build and thus the command was not even executed.
I moved the flutter upgrade task so that it only is executed on the iOS build. Lets see if it works now, the job is still waiting for being scheduled …

I wonder why upgrading flutter manually even is necessary. Probably CirrusCI just did not update their MacOS images yet and will do so soon. I should have noticed this on my version upgrade PR already, but the ios build was sadly cancelled there due to a newer build, so that it was not shown as failed.

Schedulaar (approved):

PR #301, opened 2021-03-11
Frontend UI improvements

### LINKED ISSUE #304: Fix status bar icons/text not readable on ios and android
On all pages:
- in darkmode
- in light mode

Also needs to be considered: Sometimes the status bar color changes on a page but doesn't change back properly

RoAnBu:
Todos for @RoAnBu:

- [x] Fix detail view header text color in light mode
- [x] Fix broken paging in search page
- [x] Make distance fancier
- [x] Refactor Widget tree
- [x] Make detail view header properly readable over icon
- [x] Use color scheme for detail view opened from the map
- [x] fix #304 

For testing you might need a local backend with database import, because the staging server might still return null errors for category 0 causing the widget to not render properly


RoAnBu:
Todos for @RoAnBu:

- [ ] Fix detail view header text color in light mode
- [ ] Fix broken paging in search page
- [ ] Make distance fancier
- [ ] Refactor Widget tree

RoAnBu:
Screenshot_1615729068
Screenshot_1615729058
Screenshot_1615729033
Screenshot_1615729018
Screenshot_1615729013
Screenshot_1615729006
Screenshot_1615728985
Screenshot_1615728965

Schedulaar:
I like the search a lot! The header of the detail view still needs some love though I think: The background images should be opaque (or so) as otherwise the title is not readable (as you can see on the Parfümerie store screenshot)

frontend/lib/category_assets.dart
```diff
@@ -28,53 +38,69 @@ const List<CategoryAsset> categoryAssets = [
...
       icon: "assets/category_icons/8.svg",
-      detailIcon: "assets/detail_headers/8_essen.svg"),
+      detailIcon: "assets/detail_headers/8_essen.svg",
+      color: Color(0xFF197489)),
+  CategoryAsset(
+      id: 9,
+      name: 'Anderes',
+      shortName: "Anderes",
+      icon: "assets/category_icons/9.svg",
+      detailIcon: "assets/detail_headers/8_essen.svg",
```
Schedulaar, line 104: Do we want essen here? Can we put `null` here?

frontend/lib/category_assets.dart
```diff
@@ -28,53 +38,69 @@ const List<CategoryAsset> categoryAssets = [
...
+      detailIcon: "assets/detail_headers/6_haus.svg",
+      color: Color(0xFF00D0C7)),
   CategoryAsset(
       id: 7,
       name: 'Freizeit/Reise/Unterkünfte',
       shortName: "Freizeit",
       icon: "assets/category_icons/7.svg",
-      detailIcon: "assets/detail_headers/7_freizeit.svg"),
+      detailIcon: "assets/detail_headers/7_freizeit.svg",
+      color: Color(0xFF007CE8)),
```
maxammann, line 91: Colors are from the svgs in map-icons/ right?

maxammann (approved):

frontend/lib/category_assets.dart
RoAnBu: yes

Schedulaar:
Maybe you can still fix the following small issues:
* Imo, the search now feels a bit too compact. You can try to
  * add some more padding to each store
  * Make the icons smaller
  * Don't use bold font for the distance label
  * Make the description with maxLines=2
* Also I think, it was quite helpful to instantly see the location of each store... I'm not really sure how to integrate it well though...
* In dark mode, on the about page the label "Mehr Informationen" is not visible (for "Herausgeber")
* In dark mode, the location button is white instead of a dark.

PR #302, opened 2021-03-11
#289 fix crash with invalid region IDs

### LINKED ISSUE #289: Region Id mismatch can prevent identification tab from rendering
https://github.com/ehrenamtskarte/ehrenamtskarte/blob/5482c6cdd9fd84a4e0fa6951dedf8a015687a5e8/frontend/lib/identification/card_detail_view.dart#L33-L38

From the documentation of `firstWhere`:

> Returns the first element that satisfies the given predicate test.
> Iterates through elements and returns the first to satisfy test.
> If no element satisfies test, the result of invoking the orElse function is returned. If orElse is omitted, it defaults to throwing a StateError.

So if a region id is not in the returned list the exception prevents the page from rendering.

A possible fix could be:
```dart
          var region = result.hasException || result.isLoading
              ? null
              : getRegions.parse(result.data).regions.firstWhere(
                  (element) => element.id == cardDetails.regionId,
                  orElse: () => null);
```

elKei24:
Closes #289

### DIFF
--- a/frontend/lib/verification/view/result_card/positive_verification_result.dart
+++ b/frontend/lib/verification/view/result_card/positive_verification_result.dart
@@ -26,8 +26,9 @@ class PositiveVerificationResult extends StatelessWidget {
         child: Column(
           children: [
             Text("Name: ${cardDetails.fullName}"),
-            Text("Ablaufdatum: $expirationDateString"),
-            Text("Landkreis: ${region.prefix} ${region.name}"),
+            Text("Gültig bis: $expirationDateString"),
+            if (region != null)
+              Text("Ausgestellt von: ${region.prefix} ${region.name}"),
         ]));
   }
 }
--- a/frontend/lib/verification/view/verification_result.dart
+++ b/frontend/lib/verification/view/verification_result.dart
@@ -52,13 +52,18 @@ class VerificationResult extends StatelessWidget {
                 "Verbindung zum Server fehlgeschlagen.", "connectionError"));
           }
           QueryResult verifyResult = snapshot.data[0];
-          QueryResult regionsResult = snapshot.data[1];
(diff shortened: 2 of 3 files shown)


frontend/lib/verification/view/verification_result.dart
```diff
@@ -52,13 +52,18 @@ class VerificationResult extends StatelessWidget {
...
-            var exception = verifyResult.hasException
-                ? verifyResult.exception.toString()
-                : regionsResult.exception.toString();
+          if (regionsResult.hasException) {
+            debugPrint(regionsResult.exception.toString());
             return NegativeVerificationResult(
-                VerificationError(exception, "verifyRequestError"));
+                VerificationError("Fehler bei der Kommunikation mit dem "
+                    "Server.", "queryRegionsError"));
           }
```
elKei24, line 67: Has nothing to do with the bug, but I think we should not just forward potentially cryptic exception messages to the user but instead log them.

frontend/lib/verification/view/result_card/positive_verification_result.dart
```diff
@@ -26,8 +26,9 @@ class PositiveVerificationResult extends StatelessWidget {
         child: Column(
           children: [
             Text("Name: ${cardDetails.fullName}"),
-            Text("Ablaufdatum: $expirationDateString"),
-            Text("Landkreis: ${region.prefix} ${region.name}"),
+            Text("Gültig bis: $expirationDateString"),
+            if (region != null)
+              Text("Ausgestellt von: ${region.prefix} ${region.name}"),
```
Schedulaar, line 31: Oh this absent null-safety :cry:

Schedulaar (approved):

RoAnBu (approved):

PR #309, opened 2021-03-17
#284 Sign in for administration frontend

### LINKED ISSUE #284: Basic login to administration BE

elKei24:
Closes #284


Schedulaar:
I added a default user:
`admin@example.com` with password `admin` which gets created if no user exists

Schedulaar (approved):

administration/__generated__/globalTypes.ts
```diff
@@ -7,6 +7,11 @@
 // START Enums and Input Objects
 //==============================================================
 
+export interface AuthDataInput {
+  email: string;
+  password: string;
```
maxammann, line 12: Should this be a passwordHash?

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
```diff
@@ -22,6 +22,10 @@ class WebService {
         println("Goto http://${host}:${port}/graphiql for a graphical editor")
 
         val graphQLHandler = GraphQLHandler()
-        app.post("/") { ctx -> graphQLHandler.handle(ctx) }
+        app.post("/") { ctx ->
+            ctx.header("Access-Control-Allow-Headers: Authorization")
+            ctx.header("Access-Control-Allow-Origin: *")
```
maxammann, line 27: This should only be done in development and not production.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Setup.kt
```diff
@@ -1,9 +1,13 @@
...
 
+import app.ehrenamtskarte.backend.auth.database.repos.AdministratorsRepository
 import org.jetbrains.exposed.sql.SchemaUtils
 
 fun setupDatabase() {
     SchemaUtils.create(
         Administrators
     )
+    if (AdministratorEntity.count() == 0L) {
+        AdministratorsRepository.insert("admin@example.com", "admin")
```
maxammann, line 11: `admin` is the PW here, right? I think we should use bcrypt here to store the passwords.

maxammann: Then we also can set a default password by just using the hashed version of it.

administration/__generated__/globalTypes.ts
maxammann: Else we are transmitting plain text passwords to the server.

maxammann (approved):
I think we should move the issues I mentioned to a separate issue as it is probably quite huge.

administration/__generated__/globalTypes.ts
Schedulaar: I created an issue. https://github.com/ehrenamtskarte/ehrenamtskarte/issues/322

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Setup.kt
Schedulaar: `admin` is the pw, yes. `AdminstratorsRepository` does use bcrypt under the hood though...

PR #311, opened 2021-03-18
Automatically configure base api url

maxammann:


Schedulaar (approved):
Can we have npm run-configs 🥺

maxammann:
This was quicker to implement and does not interfere with current ci configs

RoAnBu (approved):

PR #328, opened 2021-03-20
319: Add applications to administration

Schedulaar:


administration/src/components/applications/ApplicationsOverview.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+interface JsonField {
+    name: string,
+    translations: { de: string },
+    type: 'Array' | 'String' | 'Number' | 'Boolean' | 'Attachment',
+    value: any
+}
+
+function JsonFieldView(props: { jsonField: JsonField, baseUrl: string, token: string }) {
```
elKei24, line 20: Hm, so we build the view completely from user data? Probably too late to change this design decision, but maybe we should at least have some kind of warning on the page.
Just imagine an user building sending a malicious application building a UI here that says something like "Sorry, you have to verify again that you are allowed to see this application due to privacy reasons. Head over to myevildomain.com/login and type your credentials there". There are enough users trusting such emails; those will trust texts shown in our web application even more.

elKei24 (approved):

administration/src/components/Navigation.tsx
```diff
@@ -17,7 +17,8 @@ const Navigation = (props: Props) => {
                 <RegionSelector/>
                 <Navbar.Divider/>
                 <NavLink to={"/"}><Button minimal icon="home" text="Home"/></NavLink>
-                <NavLink to={"/eak-generation"}><Button minimal icon="people" text="Karten"/></NavLink>
+                <NavLink to={"/applications"}><Button minimal icon="form" text="Eingehende Anträge" /></NavLink>
+                <NavLink to={"/eak-generation"}><Button minimal icon="people" text="Karten erstellen"/></NavLink>
                 <NavLink to={"/accepting-stores"}><Button minimal icon="list" text="Akzeptanzstellen"/></NavLink>
```
aiquita, line 22: We should remove the Akzeptanzstellen Tab, because it will remain empty.

aiquita (approved):

maxammann (approved):

PR #335, opened 2021-03-22
Frontend verification improvements

RoAnBu:
- fix backend verification result being cached by graphql
- fix null exception when region was still loading in positive_verification_result


maxammann (approved):

maxammann:
Good catch!

Schedulaar (approved):

PR #346, opened 2021-03-25
#345: Fix jumping map

### LINKED ISSUE #345: The map jumps on startup of release builds
**Describe the bug**
On startup of the app in a release build, the map starts at the wrong position (ie. Bayern zoomed in) until it figures out where the user is and then moves the camera.
Fix this glitch by first checking where the user is and setting an appropriate initalPosition.

**To Reproduce**
1. Build release build (e.g. on Android)
2. Install, open the app and allow location permission.
3. Reopen the app

**Expected behavior**
The map should correctly initialize with the user's position if it can be determined quickly. Otherwise it should show an overview of Bavaria.

**Desktop (please complete the following information):**
 - OS: Android, Probably iOS

Schedulaar:
Closes #345

### DIFF
--- a/frontend/lib/map/map/map.dart
+++ b/frontend/lib/map/map/map.dart
@@ -16,20 +16,22 @@ typedef OnMapCreatedCallback = void Function(MapController controller);
 
 class Map extends StatefulWidget {
   static const double userLocationZoomLevel = 13;
-  static const double initialZoomLevel = 6;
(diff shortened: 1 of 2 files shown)


RoAnBu (approved):
Nice, this seems to solve all the map-jumping-magic I had on iOS. Works how I would expect it now! Very nice!

maxammann (approved):

PR #350, opened 2021-04-23
349: Delete applications

Schedulaar:


backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -29,4 +29,11 @@ class EakApplicationMutationService {
         return true
     }
 
+    @GraphQLDescription("Deletes the application with specified id")
+    fun deleteApplication(
+        context:GraphQLContext, applicationId: Int
+    ): Boolean {
+        context.enforceSignedIn()
```
maxammann, line 36: This throws an exception if not signed in right?

maxammann (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
Schedulaar: Yep, I guess so

elKei24 (approved):
Did not run it but code looks good

PR #362, opened 2021-05-23
Constrain pipeline to Flutter 2.0.6

elKei24:
This should fix the failing tests. On the long term, we should of course update to the latest version.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -18,7 +18,7 @@ fe_task_without_container_template: &FE_TASK_WITHOUT_CONTAINER_TEMPLATE
 fe_task_template: &FE_TASK_TEMPLATE
   << : *FE_TASK_WITHOUT_CONTAINER_TEMPLATE
   container:
-    image: cirrusci/flutter:latest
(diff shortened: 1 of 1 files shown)


RoAnBu:
@elKei24 same fix for ios necessary?

elKei24:
> @elKei24 same fix for ios necessary?

Yes, would be, but seems not possible unfortunately. The OSX images are only selectable by Xcode version, and are always created with the latest version of flutter (can be seen [here](https://github.com/cirruslabs/osx-images/blob/master/templates/xcode.json)).

The best would probably be to update to the latest version as fast as possible. I tried to do it (s. branch `update-dependencies`), but it does not build any more with the latest MapboxGL lib: `SDK Registry token is null. See README.md for more information.` I had a look at the Readme and tried to set the key(s), but without success.

RoAnBu:
ok maybe just ignore ios build for now?

RoAnBu (approved):

elKei24:
> ok maybe just ignore ios build for now?

I think so, yes. At least allows us to merge your artifact PR.

Issue #360, opened 2021-05-07
Attachments of applications are not deleted


michael-markl:
Currently, when we delete an application via GraphQl, only the DB entry for that application gets deleted.
The attachments of the application which are stored on disk are not yet deleted.
They should be deleted as well.

PR #366, opened 2021-07-02
#360: Use Flutter version 2.0.6

Schedulaar:
I have noticed after upgrading to the latest flutter version, that our project is not able to `flutter pub get` anymore. It seems the problem is an upgrade of dependencies inside the upgraded dart sdk.
I thought this might be a good time, to pin the flutter version used in the project with [FVM](https://fvm.app)


elKei24 (approved):
Have not tested it but looks good. Nice documentation.

maxammann (approved):

PR #369, opened 2021-08-06
fix martin config getting overwritten

elKei24:
This PR extends the package build script, so that a configuration file can be passed that will be added to the [conffiles control file](https://manpages.debian.org/testing/dpkg-dev/deb-conffiles.5.en.html).
This is used for the martin configuration file, so that changes to that file are not overwritten when updating the package. If both the package maintainers file and the local file have changed, an error is shown.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -215,7 +215,7 @@ jobs:
           name: "Build .deb"
           command: |
             cp /tmp/workspace/martin/martin .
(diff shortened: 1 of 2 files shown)


maxammann (approved):
Lets see whether this fixes it

if it never happens again then this is fixed :)

Schedulaar (approved):
hihi

PR #371, opened 2021-08-16
Preparations for initial release

maxammann:


maxammann:
New version will be 2.0.0 and build number 50

docs/development-setup.md
```diff
@@ -1,6 +1,6 @@
 # Frontend
 
-The Ehrenamtskarte app uses Flutter. You can have a look at [this guide](https://flutter.dev/docs/get-started/install) to install flutter and the tools needed to work with it.
+The Ehrenamtskarte app uses Flutter. You can have a look at [this guide](https://flutter.dev/docs/get-started/install) to install flutter and the tools needed to work with it. Alternatively, you can take a look at [README.md](../frontend/README.md) for a simpler guide which used Flutter Version Manager.
```
elKei24, line 3:
```suggestion
The Ehrenamtskarte app uses Flutter. You can have a look at [this guide](https://flutter.dev/docs/get-started/install) to install flutter and the tools needed to work with it. Alternatively, you can take a look at [README.md](../frontend/README.md) for a simpler guide which uses Flutter Version Manager.
```

frontend/ios/Flutter/Debug.xcconfig
```diff
@@ -1,2 +1,2 @@
-#include "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
+#include? "Pods/Target Support Files/Pods-Runner/Pods-Runner.debug.xcconfig"
```
elKei24, line 1: Just out of curiosity: What does this do?

frontend/ios/Podfile
```diff
@@ -0,0 +1,46 @@
+# Uncomment this line to define a global platform for your project
```
elKei24, line 1: Is this auto-generated code? If not, should I try to understand it or is it sufficient that it works for you?

frontend/ios/Runner/Info.plist
```diff
@@ -28,6 +28,8 @@
 	<string>Einlesen von QR-Codes und Abfotografieren von Dokumenten für das Bewerbungsformular.</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Standort auf der Karte und Akzeptanzstellen in der Umgebung anzeigen.</string>
+	<key>NSLocationAlwaysUsageDescription</key>
+	<string>Standort auf der Karte und Akzeptanzstellen in der Umgebung anzeigen.</string>
```
elKei24, line 32: Do we still need `NSLocationWhenInUsageDescription` then?

elKei24 (approved):
Can obviously not test iOS. Is there something for Android to test? Looks like only the IntelliJ config has changed there, right?

frontend/ios/Flutter/Debug.xcconfig
maxammann: I dont know to be honest :D this was generated by Flutter

frontend/ios/Podfile
maxammann: no need to understand

frontend/ios/Runner/Info.plist
maxammann: Yes, they are shown at different spots. The thing is because we use flutter, code for both types of usage in Swift/ObjectiveC is included in the app. That means apple thinks that both dialogs could be shown even though this is wrong.

So this is just disabling a warning by apple.

maxammann:
Ready for Review basically, but more changes to come.

Schedulaar:
Should we use the short name "Ehrenamt.BY" instead of "Ehrenamtskarte"? "Ehrenamtskarte" is split into two lines on my home screen ("Ehrenamtskart\ne")

elKei24:
> Should we use the short name "Ehrenamt.BY" instead of "Ehrenamtskarte"? "Ehrenamtskarte" is split into two lines on my home screen ("Ehrenamtskart\ne")

Good point. On mine, it is "Ehrenamts…" 😢

maxammann:
Or just Ehrenamt on the home screen and Ehrenamtskarte Bayern in the stores?

I think you usually have only one ehrenamtskarten app.

Schedulaar:
We might want to talk about this with the Ministerium, but they currently use "Ehrenamt.BY".

maxammann:
Ah right sure, then we should stick to that for now

frontend/android/app/src/main/kotlin/app/ehrenamtskarte/MainActivity.kt
```diff
@@ -1,6 +1,49 @@
...
+    private fun cleanITNRWData() {
+        val m = packageManager
+        try {
+            val info = m.getPackageInfo(packageName, 0)
+            val arcgisfile =
+                File(info.applicationInfo.dataDir + File.separatorChar + "shared_prefs" + File.separatorChar + "com.esri.arcgisruntime.settings.xml")
+            if (arcgisfile.exists()) {
+                resetApp()
+            }
+        } catch (e: PackageManager.NameNotFoundException) {
```
elKei24, line 40: Maybe better catch everything, so that the app can not crash even if something else in the cleanup fails?

frontend/android/app/src/main/kotlin/app/ehrenamtskarte/MainActivity.kt
```diff
@@ -1,6 +1,49 @@
...
+    private fun deleteRecursive(fileOrDirectory: File) {
+        if (fileOrDirectory.isDirectory) {
+            for (child in fileOrDirectory.listFiles()) {
+                deleteRecursive(child)
+            }
+        }
+        fileOrDirectory.delete()
+    }
+
+    private fun cleanITNRWData() {
```
elKei24, line 31: I would prefer cleanItNrwData, but no idea what our style guide says about it …

frontend/android/app/src/main/AndroidManifest.xml
```diff
@@ -7,7 +8,7 @@
          FlutterApplication and put your custom class here. -->
     <application
         android:name="io.flutter.app.FlutterApplication"
-        android:label="Ehrenamtskarte"
+        android:label="Ehrenamt"
```
Schedulaar, line 11: Previously, the short name was
```suggestion
        android:label="Ehrenamt.BY"
```

frontend/ios/Runner/AppDelegate.swift
```diff
@@ -7,7 +7,51 @@ import Flutter
...
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
+    clearITNRWStorage()
     GeneratedPluginRegistrant.register(with: self)
     return super.application(application, didFinishLaunchingWithOptions: launchOptions)
   }
+    
+    func clearITNRWStorage() {
+        let preferences = UserDefaults.standard
+        let key = "preferencesSQLFilterString"
```
Schedulaar, line 17: This is a string that was used by IT.NRW and that should never be used by us?
Can we document this here?

frontend/lib/entry_widget.dart
```diff
@@ -34,6 +34,7 @@ class EntryWidget extends StatelessWidget {
               darkTheme: darkTheme,
               themeMode: ThemeMode.system,
               initialRoute: initialRoute,
+              debugShowCheckedModeBanner: false,
```
Schedulaar, line 37: Is this for screenshots?

frontend/android/app/src/main/kotlin/app/ehrenamtskarte/MainActivity.kt
```diff
@@ -1,6 +1,49 @@
...
+        val m = packageManager
+        try {
+            val info = m.getPackageInfo(packageName, 0)
+            val arcgisfile =
+                File(info.applicationInfo.dataDir + File.separatorChar + "shared_prefs" + File.separatorChar + "com.esri.arcgisruntime.settings.xml")
```
Schedulaar, line 36: Please add a comment indicating that this is the "characteristic" file whose existence implies that the app was upgraded from the IT.NRW version.

frontend/android/app/src/main/kotlin/app/ehrenamtskarte/MainActivity.kt
```diff
@@ -1,6 +1,49 @@
...
 class MainActivity: FlutterActivity() {
+
+    private fun resetApp() {
+        val m = packageManager
+        try {
+            val info = m.getPackageInfo(packageName, 0)
+            deleteRecursive(File(info.applicationInfo.dataDir))
+        } catch (e: PackageManager.NameNotFoundException) {
+            Log.e("WRAPPER", "Failed to get data directory!", e)
+        }
```
Schedulaar, line 19: Do we also need to reset the shared preferences? Or are they stored within this directory?

Schedulaar (approved):

maxammann:
Will fix stuff in some other PR

frontend/android/app/src/main/AndroidManifest.xml
maxammann: I changed it silently :see_no_evil:

frontend/android/app/src/main/kotlin/app/ehrenamtskarte/MainActivity.kt
maxammann: I thin shared preferences are in the dataDir. So not need to delete them.

PR #387, opened 2021-10-07
372: Upgrade flutter dependencies

Schedulaar:


maxammann (approved):
Works on iOS

Better than before!

PR #412, opened 2021-11-15
Fix permissions on iOS

maxammann:
This PR contains fixes and improvement to the location permission detection. (especially for iOS)


maxammann:
Tested on iOS

PR #416, opened 2021-11-17
Fix Map Rotation

maxammann:
Sadly I dont know how to fix this else. I don't even know where to start...

### DIFF
--- a/frontend/android/app/src/main/AndroidManifest.xml
+++ b/frontend/android/app/src/main/AndroidManifest.xml
@@ -12,6 +12,7 @@
         <activity
             android:name=".MainActivity"
             android:launchMode="singleTop"
+            android:screenOrientation="portrait"
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
             android:hardwareAccelerated="true"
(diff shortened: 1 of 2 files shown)


Schedulaar:
I assume, the bug has to do with the changes of the FAB. The bug only occurs for me on the right hand side of the screen when rotated to landscape... Maybe you can have another look before we fix the orientation to portrait.
Afaik, rotating the device was no problem prior to the refactorings.

maxammann:
> I assume, the bug has to do with the changes of the FAB. The bug only occurs for me on the right hand side of the screen when rotated to landscape... Maybe you can have another look before we fix the orientation to portrait.
> Afaik, rotating the device was no problem prior to the refactorings.

This bug is already old. I noticed it already a few months ago, but could not reproduce it.
Now I had the idea of rotating my device first, and then atarting the app.

Has nothing to do with the fab.

Schedulaar:
Maybe this workaround is still not worth it. For tablets for example, it is very unusual to use portrait mode for example...

We should work an an upstream fix.

btw: does something like that also occus on ios?

maxammann:
I'd say for a short-time fix this is the way to go. Avoiding this bug is very important imo as it makes the app unusable if starting in landscape mode.
I can create a new issue which tries to fix this properly.

maxammann:
Happens only on Android so far.

PR #422, opened 2021-11-25
Move map style

### LINKED ISSUE #421: Move map style to a separate repository
The new home of the map style will be https://github.com/digitalfabrik/ehrenamtskarte-maplibre-style

maxammann:
Styles for the ehrenamtskarte live now here: https://maps.tuerantuer.org/styles/ehrenamtskarte/map.html
The integreat map style is also on this host. The old style.json is still accessible and will always be accessible in the state it currently is.

At some point in the future we can remove the style.json from tiles.ehrenamtskarte.app.

fixes #421


.circleci/config.yml
```diff
@@ -181,22 +181,13 @@ jobs:
...
-          name: "Download glyphs"
-          command: |
-            wget -O /tmp/glyphs.zip https://github.com/orangemug/font-glyphs/archive/gh-pages.zip
-            unzip /tmp/glyphs.zip
-            mv font-glyphs* font-glyphs
-            rm -f /tmp/glyphs.zip
-      - run:
-          name: "Build .deb"
-          command: ~/project/scripts/pack_deb.sh -v "0.$CIRCLE_BUILD_NUM" -m . -d "Mapbox styles.json with dependencies for the Ehrenamtskarte app" -n "eak-map-styles"
+          name: "Download legacy styles deb"
```
klinzo, line 184: Why do we download the old legacy styles here?

klinzo (approved):

.circleci/config.yml
maxammann: If we just remove the legacy styles, then current installations will fail.

PR #429, opened 2021-11-28
Prevent line breaks in category names

### LINKED ISSUE #391: Category titles are too long
**Describe the bug**
Sometimes the titles are breaking if they are too long. It also depends on screen size.

**To Reproduce**
1. Go to 'Suche'

**Expected behavior**
It should fit in one line.

**Screenshots**
image

**Smartphone (please complete the following information):**
 - Device: iPhone X

klinzo:
Fixes #391 

To increase the available space of the filter buttons with long names and therefore prevent line breaks sort by name length and put all categories with longer names in second row (less items -> more space).

Should also work fine on larger devices.


frontend/lib/search/filter_bar_button.dart
```diff
@@ -59,10 +60,24 @@ class _FilterBarButtonState extends State<FilterBarButton>
...
+      65.0,
+      smallWidth);
+
+    var numLargeElementsPerRow = totalWidth ~/ largeWidth;
+    var isSecondRow = widget.index >= max(minNumberElements, numLargeElementsPerRow);
+
+    // In the second row we can use larger width as there are only 4 categories
+    var width = isSecondRow
+      ? largeWidth
+      : smallWidth;
```
maxammann, line 80: I'm not sure whether this added complexity is worth it. How about refactorign the bar such that we just have two horizontal category lists?

Then we just manually put the elements there and set a custom width for each list. Even with this implementation we will have the magic values 80 and 65.

But I have to say that I do not really get how the items are behaving according to the definition above.

frontend/lib/search/filter_bar_button.dart
```diff
@@ -59,10 +60,24 @@ class _FilterBarButtonState extends State<FilterBarButton>
...
+      80.0,
+      (totalWidth - minNumberElements * paddingPerElement) / minNumberElements);
+
+    // At least 65 px are needed to fit all category names
+    var largeWidth = max(
+      65.0,
+      smallWidth);
+
+    var numLargeElementsPerRow = totalWidth ~/ largeWidth;
+    var isSecondRow = widget.index >= max(minNumberElements, numLargeElementsPerRow);
```
maxammann, line 75: On a tablet there is no second row, so would this then be always false?

frontend/lib/search/filter_bar_button.dart
maxammann: So my idea would be to simplify this a little bit by making it even more explicit. We know we have 1-2 rows and we know exactly how wide each icon is.

frontend/lib/search/filter_bar_button.dart
```diff
@@ -59,10 +60,24 @@ class _FilterBarButtonState extends State<FilterBarButton>
...
-        80.0,
-        (totalWidth - minNumberElements * paddingPerElement) /
-            minNumberElements);
+
+    // Width allowing at least minNumberElements in one row
+    var smallWidth = min(
+      80.0,
+      (totalWidth - minNumberElements * paddingPerElement) / minNumberElements);
+
+    // At least 65 px are needed to fit all category names
```
maxammann, line 69: I'm not sure whether this is specified in px or dp. One at least depends on the DPI. So I'm not sure whether this is a constant.

maxammann (approved):

maxammann:
We can still merge and try and experiment :D

frontend/lib/search/filter_bar_button.dart
klinzo: Yes

frontend/lib/search/filter_bar_button.dart
klinzo: So we have always to rows, one with 5 and one with 4 categories? I think that does not make sense on tablets

maxammann: I'm not sure about which approach is better. On tablets we would ofc add all icons to a single list.

I think the difference is **where** we handle this logic. In this PR it is handled in the button widget. The other approach would handle it at the level of the filter bar widget.

klinzo: Moved magic numbers to constants. Is there a reason why it needs to be handled in the filter bar widget? Otherwise I'd keep it as it works just fine

frontend/lib/search/filter_bar_button.dart
```diff
@@ -59,10 +60,24 @@ class _FilterBarButtonState extends State<FilterBarButton>
     const paddingPerElement = 8;
     const minNumberElements = 5;
     var totalWidth = MediaQuery.of(context).size.width;
-    var width = min(
-        80.0,
-        (totalWidth - minNumberElements * paddingPerElement) /
-            minNumberElements);
+
+    // Width allowing at least minNumberElements in one row
```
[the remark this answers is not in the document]
maxammann: I guess the answer is no?

frontend/lib/search/filter_bar_button.dart
maxammann: So is it a constant?

frontend/lib/search/filter_bar_button.dart
maxammann: No we can keep it where it is. I just wanted to raise the question whether it can be done in a simpler way, because these min/max formulas are quite complex.

frontend/lib/search/filter_bar_button.dart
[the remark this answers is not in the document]
klinzo: Yes, you were right, moved the comment accordingly

frontend/lib/search/filter_bar_button.dart
klinzo: Okay, maybe lets wait for a review from @Schedulaar

frontend/lib/search/filter_bar_button.dart
```diff
@@ -56,14 +60,30 @@ class _FilterBarButtonState extends State<FilterBarButton>
...
+        maxButtonWidth,
+        // Width allowing at least minNumberElements in one row
         (totalWidth - minNumberElements * paddingPerElement) /
             minNumberElements);
 
+    var largeWidth = max(requiredCategoryLongTitleWidth, smallWidth);
+
+    var numLargeElementsPerRow = totalWidth ~/ largeWidth;
+    var isSecondRow =
+        widget.index >= max(minNumberElements, numLargeElementsPerRow);
```
Schedulaar, line 82: If I understand correctly, `max(minNumberElements, numLargeElementsPerRow)` is the number of elements in the first row. But why do we use `numLargeElementsPerRow` here as in the first row only smallWidth is used?

Shouldn't it rather be `max(minNumberElements, numSmallElementsPerRow)` with `numSmallElementsPerRow = totalWidth ~/ smallWidth`?

frontend/lib/search/filter_bar_button.dart
Schedulaar: Yeah, the calculations seem quite error prone to me as well, although I don't really have a better suggestion at the top of my head :/

frontend/lib/search/filter_bar_button.dart
```diff
@@ -56,14 +60,30 @@ class _FilterBarButtonState extends State<FilterBarButton>
 
   @override
   Widget build(BuildContext context) {
+    const maxButtonWidth = 80.0;
     const paddingPerElement = 8;
     const minNumberElements = 5;
+
+    // At least a width of 65 is needed to fit all category names
+    const requiredCategoryLongTitleWidth = 65.0;
```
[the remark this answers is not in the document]
Schedulaar: Ah it's not really a minimum though... Nevermind I guess^^

Schedulaar (approved):

frontend/lib/search/filter_bar_button.dart
klinzo: Should be dp/dip. Not sure if that is a constant.

frontend/lib/search/filter_bar_button.dart
klinzo: Made things more explicit now. I'd merge this as I tested this on multiple devices with reasonable screen size now and did not experience any issues.

frontend/lib/search/filter_bar_button.dart
klinzo: But I tested it on multiple devices now and seems to work fine.

klinzo:
I tested this on multiple devices and it works just fine everywhere:
800x1280px:
image
1080x2160px:
image
480x800px:
image
... and even on a 2"7 with 240x320px:
image

PR #431, opened 2021-12-02
Linting

### LINKED ISSUE #413: Introduce proper linting rules
Right now we do not have linting rules for example for unused imports. I suppose there are many more cool and beneficial linting rules.

Lightweight refactoring can also be done in this task.

klinzo:
Fixes #413 

* Fixes `dart format`  CI check
* Adds `flutter analyze` CI check
* Adds pre-commit hook to automatically run `dart format` before commit
* [Adjust line length to 120 chars (previously 80)](https://github.com/digitalfabrik/ehrenamtskarte/pull/431/commits/d7f8b0d09f40e124c9d709110d8599aec9ccc116)
* Adds linting rules from [lint](https://pub.dev/packages/lint)
* Changes some analyze rules to error (e.g. `unused_error`)
* Disable [implicit_casts](https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks)


frontend/lib/location/determine_position.dart
```diff
@@ -98,14 +98,15 @@ Future<RequestedPosition> determinePosition(
...
+  String rationale = "Erlauben Sie der App Ihren Standort zu benutzen, um Akzeptanzstellen in Ihrer Umgebung anzuzeigen.",
   Future<void> Function()? onDisableFeature,
   Future<void> Function()? onEnableFeature,
 }) async {
   final serviceEnabled = await Geolocator.isLocationServiceEnabled();
   if (!serviceEnabled) {
     if (requestIfNotGranted) {
-      final result = await showDialog(context: context, builder: (context) => const LocationServiceDialog());
+      final bool? result = await showDialog<bool>(context: context, builder: (context) => const LocationServiceDialog());
+      // TODO: Is this check correct? Both buttons in LocationServiceDialog return true
```
klinzo, line 110: Is this correct @maxammann? Same check is used somewhere below

frontend/lib/map/map/map.dart
```diff
@@ -162,7 +162,9 @@ class _MapState extends State<Map> implements MapController {
       int distSort(a, b) {
         final cA = a["geometry"]["coordinates"];
         final cB = b["geometry"]["coordinates"];
+        // ignore: argument_type_not_assignable
```
klinzo, line 166: Other solutions?

klinzo:
While there are lots of changes, the important things to review are relatively small:
https://github.com/digitalfabrik/ehrenamtskarte/pull/431/files/0faf9323d8a7bb50519a9a4f917e13ad6d807ead..4c15ecd0d1ae407534515556ca92eff86bc8875d

All other linting rules were done in a separate commit and are easily reviewable if wanted. Let me know what you think.

I'll create an issue to enable the rule `use_build_context_synchronously` as it seems relatively important but would have changed behavior too much to include in this PR.

frontend/lib/location/determine_position.dart
```diff
@@ -95,19 +94,21 @@ Future<RequestedPosition> determinePosition(BuildContext context,
...
   if (!serviceEnabled) {
     if (requestIfNotGranted) {
-      final result = await showDialog(
-          context: context,
-          builder: (context) => const LocationServiceDialog());
-      if (result) {
+      final bool? result =
+          await showDialog<bool>(context: context, builder: (context) => const LocationServiceDialog());
+      // TODO: Is this check correct? Both buttons in LocationServiceDialog return true
+      if (result == true) {
```
klinzo, line 111: Is this correct @maxammann? Same check is used somewhere below

frontend/lib/map/map/map.dart
Schedulaar: What is the problem? Maybe you need to add type declarations to `cA` and `cB`?

frontend/lib/qr_code_scanner/qr_code_scanner_page.dart
```diff
@@ -13,7 +13,7 @@ class QrCodeScannerPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return QrCodeScanner(
-      onCodeScanned: (code) async => await _onCodeScanned(context, code),
+      onCodeScanned: (code) async => _onCodeScanned(context, code),
```
Schedulaar, line 16: Crazy, this seems wrong. Why don't we have to await `_onCodeScanned`? Is it necessary that `onCodeScanned` is async?

frontend/lib/search/results_loader.dart
```diff
@@ -97,7 +96,7 @@ class ResultsLoaderState extends State<ResultsLoader> {
         // Params are outdated.
         // If we're still at the first key, we must manually retrigger fetching.
         if (pageKey == _pagingController.firstPageKey) {
-          return await _fetchPage(pageKey);
+          return _fetchPage(pageKey);
```
Schedulaar, line 99: Do we remove the await because we are returning a Future here?
I like the await syntax more, as it is clearer what is async and what's not 🤔

Schedulaar (approved):

frontend/lib/qr_code_scanner/qr_code_scanner_page.dart
klinzo: This is probably just personal preference, I think we had the same discussion in integreat a while back with js. The rule used is this one: https://dart-lang.github.io/linter/lints/unnecessary_await_in_return.html
Basically it is just not necessary to await an async functio if it is returned since the return value is a Future already.
So if you want we can revert this.

frontend/lib/search/results_loader.dart
klinzo: See answer above

frontend/lib/map/map/map.dart
klinzo: Sorry, wasn't clear here. I rather wanted to ask if there are any types that can be used because I did not find any. It would be possible to define some by ourselves if there are none.

frontend/analysis_options.yaml
```diff
@@ -1,13 +1,164 @@
...
+    unrelated_type_equality_checks: error
+    unsafe_html: error
+    use_full_hex_values_for_flutter_colors: error
+    use_function_type_syntax_for_parameters: error
+    use_named_constants: error
+    use_setters_to_change_properties: error
+    use_string_buffers: error
+    use_test_throws_matchers: error
+    valid_regexps: error
+    void_checks: error
```
maxammann, line 153: These are a lot of custom rules. Why are they not part of https://github.com/passsy/dart-lint

How much difference would you expect if we stick with https://pub.dev/packages/flutter_lints and add again more pedantic rules? I would like to avoid third-party dependencies as much as possible.

frontend/lib/graphql/graphql_api.graphql.dart
```diff
@@ -32,8 +32,7 @@ class GetRegionsById$Query$Region extends JsonSerializable with EquatableMixin {
 class GetRegionsById$Query extends JsonSerializable with EquatableMixin {
   GetRegionsById$Query();
 
-  factory GetRegionsById$Query.fromJson(Map<String, dynamic> json) =>
-      _$GetRegionsById$QueryFromJson(json);
+  factory GetRegionsById$Query.fromJson(Map<String, dynamic> json) => _$GetRegionsById$QueryFromJson(json);
```
maxammann, line 35: These files should be excluded from analyzer and linter.

frontend/lib/identification/card_detail_view/verification_qr_code_view.dart
```diff
@@ -33,7 +33,7 @@ class _VerificationQrCodeViewState extends State<VerificationQrCodeView> {
     _otpCode = widget._otpGenerator.generateOTP(_resetQrCode);
   }
 
-  _resetQrCode() {
+  void _resetQrCode() {
```
maxammann, line 36: I think this void is unnecessary. I would prefer not having to write void if this results in an error.

frontend/lib/location/determine_position.dart
maxammann: The code here is correct. The only question is why LocationServiceDialog return true for both actions.

frontend/lib/map/map/map_with_futures.dart
```diff
@@ -9,26 +9,25 @@ class MapWithFutures extends StatelessWidget {
...
-      required this.onFeatureClick,
-      required this.onFeatureClickLayerFilter,
-      required this.onMapCreated})
-      : super(key: key);
+  const MapWithFutures({
+    Key? key,
+    required this.onNoFeatureClick,
+    required this.onFeatureClick,
+    required this.onFeatureClickLayerFilter,
+    required this.onMapCreated,
```
maxammann, line 17: :+1: for enforced trailing commas

frontend/lib/qr_code_scanner/qr_code_scanner_page.dart
maxammann: Both is fine for me, but I generally prefer lints which yield shorter and less-boilerplate code. So I'm fine with disallowing await here.

maxammann: Do we have a lint for "unhandled promises". So for example what happens if you remove the "async" here? Does it show an error like when you add the "await"?

maxammann (changes requested):

maxammann:
And please squash and merge the changes before merging such that we only have a single commit.

frontend/analysis_options.yaml
klinzo: That are not custom rules but I rather just set them to error. All those rules are enabled in lint as well, but perhaps with just warning or info severity. Actually I copied those from lint.

klinzo: I'd rather use a package with recommended rules and wide adoption across other projects then setting up lots of custom rules. `flutter_lints` was completely enabled before already, so those rules are very (too) loose imo.

frontend/lib/identification/identification_page.dart
```diff
@@ -19,45 +19,49 @@ class IdentificationPage extends StatelessWidget {
...
-          ));
-        },
-      );
-    });
+      },
+    );
   }
 
-  void _showVerificationDialog(context, SettingsModel settings) async {
+  Future<void> _showVerificationDialog(BuildContext context, SettingsModel settings) async {
```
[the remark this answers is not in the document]
klinzo: Why is that unnecessary? The return type is clearly `Future<void>` and not `void`

frontend/pubspec.yaml
```diff
@@ -60,6 +60,11 @@ dependencies:
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^1.0.0
+  collection: ^1.15.0
+  meta: ^1.7.0
+  http: ^0.13.4
+  equatable: ^2.0.3
+  fixnum: ^1.0.0
```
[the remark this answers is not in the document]
klinzo: These were used in the code before but not explicitly stated as dependencies

frontend/lib/location/determine_position.dart
klinzo: Do you want to have a look or should I open an issue?

frontend/lib/qr_code_scanner/qr_code_scanner_page.dart
klinzo: Yes, there is an error if async is removed. I'd also favor the current with no `await` if returned. So will keep as it is.

.github/hooks/pre-commit
```diff
@@ -0,0 +1,17 @@
...
+#
+# Install this using:
+# ln -s ../../.github/hooks/pre-commit .git/hooks/pre-commit
+#
+
+FILES=$(git diff --cached --name-only --diff-filter=ACMR "*.dart" | sed 's| |\\ |g')
+[ -z "$FILES" ] && exit 0
+
+# Prettify all selected files
+echo "$FILES" | xargs dart format -l 120
```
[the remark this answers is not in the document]
klinzo: What should happen if it is not in $PATH? Also how do you do that? Could you just change this to your liking?

frontend/lib/graphql/graphql_api.graphql.dart
klinzo: They are, but not from format (they weren't before as well). These changes are just because of changing max line length.

frontend/lib/identification/protobuf/card_activate_model.pb.dart
```diff
@@ -16,52 +16,25 @@ export 'card_activate_model.pbenum.dart';
 
 class CardActivateModel extends $pb.GeneratedMessage {
   static final $pb.BuilderInfo _i = $pb.BuilderInfo(
-      const $core.bool.fromEnvironment('protobuf.omit_message_names')
-          ? ''
-          : 'CardActivateModel',
+      const $core.bool.fromEnvironment('protobuf.omit_message_names') ? '' : 'CardActivateModel',
```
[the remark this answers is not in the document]
klinzo: See answer above, this is just formatting and not linting/analyzing.

frontend/lib/identification/card_detail_view/verification_qr_code_view.dart
klinzo: We use this rule https://dart-lang.github.io/linter/lints/always_declare_return_types.html which enforces return types for all functions. Quite reasonable imo and it is probably not possible to exclude void. I can remember having the same discussion in Integreat :D But if you want we can disable the rule completely. @Schedulaar what do you think?

frontend/lib/search/search_page.dart
```diff
@@ -20,42 +20,47 @@ class _SearchPageState extends State<SearchPage> {
...
                       "Suchresultate".toUpperCase(),
                       style: const TextStyle(color: Colors.grey),
                     ),
-                    const Expanded(
-                        child: Padding(
-                            padding: EdgeInsets.only(left: 8),
-                            child: Divider()))
-                  ]))),
-          ResultsLoader(
+                    const Expanded(child: Padding(padding: EdgeInsets.only(left: 8), child: Divider()))
```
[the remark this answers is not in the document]
klinzo: Not sure if I understand the question. We did not have a linting/format rule to force each child to have its own line. This was rather just the case because of the shorter previous line length. AFAIK there are no formatting options except from line length, so probably not possible to force children to have its one line.

frontend/lib/identification/card/eak_card.dart
```diff
@@ -25,94 +24,89 @@ class EakCard extends StatelessWidget {
...
-                            title: headerTitle, scaleFactor: scaleFactor),
-                        EakCardHeaderLogo(
-                          title: "Freistaat Bayern",
-                          scaleFactor: scaleFactor,
-                          logo: const Image(
-                              image: AssetImage("assets/wappen-bavaria.png"),
-                              fit: BoxFit.contain),
-                        )
-                      ])),
+                aspectRatio: 6 / 1,
```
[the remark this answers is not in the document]
klinzo: No changes from me here, just applied correct formatting. AFAIK there is no option to change this.

frontend/lib/identification/identification_qr_scanner_page.dart
```diff
@@ -14,46 +14,59 @@ class IdentificationQrScannerPage extends StatelessWidget {
...
-      Expanded(
-          child: QrCodeScannerPage(
-        onCodeScanned: (code) async => _onCodeScanned(context, code),
-      ))
-    ]));
+    );
   }
 
-  void _onCodeScanned(BuildContext context, String code) async {
+  Future<void> _onCodeScanned(BuildContext context, String code) async {
```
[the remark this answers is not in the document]
klinzo: Hmm kind of, it is incorrect the same way as `void` is incorrect as return type of an async function in typescript. It returns a `Future`/`Promise` and not just `void`. However, you can treat the function as `void` if you don't await it.

frontend/lib/map/map/map.dart
klinzo: Done

frontend/lib/location/determine_position.dart
klinzo: Fixed this like suggested below

.github/hooks/pre-commit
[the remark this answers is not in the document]
maxammann: The hook should abort with an error explaining the issue in my opinion. Here is how to do it: https://stackoverflow.com/a/677212/1763110

We also recommend to use `fvm`. In this case the dart command does not need to be in the path. Instead you use `fvm dart`.

So actually this script should use fvm instead of `dart` directly.

frontend/analysis_options.yaml
maxammann: 200 starts does not mean it is adopted across projects. I think this is just a dependency which could cause problems in the future. While just copy pasting the rules into one file should work correctly.

frontend/lib/graphql/graphql_api.graphql.dart
maxammann: They still should be excluded. We could create an extra task for this, but this is the first task which introduces `dart fmt` properly. Changes to these files should only be done by the graphql/protobuf tools.

frontend/lib/identification/card/eak_card.dart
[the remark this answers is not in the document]
maxammann: The previous linting setup had 4 spaces for this use case. I did run the formatter earlier already. So all files already were formatted before this branch was forked.

frontend/lib/identification/card_detail_view/verification_qr_code_view.dart
maxammann: I think this lint can be disabled. In my opinion it is very clear what this is doing:

```
_resetQrCode() {
    // ...
}
```

Writing the following instead does not add to the readability:

```
void _resetQrCode() {
    // ...
}
```

This is also the default in all modern languages like Rust, Kotlin, Dart etc.

frontend/lib/identification/identification_page.dart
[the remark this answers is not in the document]
maxammann: I think my comment was to the wrong line here. See discussion below.

frontend/lib/identification/protobuf/card_activate_model.pb.dart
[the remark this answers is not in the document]
maxammann: Then it should be excluded from formatting. Should be done in this task or a future one.

frontend/lib/search/search_page.dart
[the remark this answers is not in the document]
maxammann: I see yes, seems like line length is the only which causes children to go onto a new line most of the cases. I did not find such a linting rule.

frontend/lib/identification/identification_qr_scanner_page.dart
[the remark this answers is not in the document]
maxammann: So there is an actual difference: https://dartpad.dev/?id=a182bddb058e612317ce7634ed149944


`void _onCodeScanned` would be fine here as it states that it is disallowed to await the result. This is beneficial if you need to pass this function to some library which is not able to await the result.

So both is fine and has its usecaes.

frontend/pubspec.yaml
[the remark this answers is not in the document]
klinzo: Yes, I am:
* [collection](https://github.com/digitalfabrik/ehrenamtskarte/blob/49d668d7a68a2f8b21ba2dd8551fdd5a56763165/frontend/lib/about/license_page.dart#L7)
* [meta](https://github.com/digitalfabrik/ehrenamtskarte/blob/49d668d7a68a2f8b21ba2dd8551fdd5a56763165/frontend/lib/category_assets.dart#L3)
* [http](https://github.com/digitalfabrik/ehrenamtskarte/blob/49d668d7a68a2f8b21ba2dd8551fdd5a56763165/frontend/lib/graphql/upload_parser.dart#L1)
* [equatable](https://github.com/digitalfabrik/ehrenamtskarte/blob/49d668d7a68a2f8b21ba2dd8551fdd5a56763165/frontend/lib/identification/card/eak_card.dart#L1)
* [fixnum](https://github.com/digitalfabrik/ehrenamtskarte/blob/49d668d7a68a2f8b21ba2dd8551fdd5a56763165/frontend/lib/verification/scanner/verification_encoder.dart#L3)

These are just examples, there are more occurrences.

frontend/lib/identification/identification_qr_scanner_page.dart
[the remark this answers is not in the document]
klinzo: Then it is just possible to do this:
```dart
void test() {
 test1();
}

Future<void> test1() async {
  
}
```
No need to await a function you don't want to await. I mean I don't really care as I am new to the project, I don't want to force you to use some rules you don't want to. Just thought (and still think) that these rules really make sense.

frontend/lib/identification/card_detail_view/verification_qr_code_view.dart
klinzo: I'm fine with disabling it, just keep in mind that we don't have other return types on functions as well then.

frontend/lib/graphql/graphql_api.graphql.dart
klinzo: Created an issue: https://github.com/digitalfabrik/ehrenamtskarte/issues/437

frontend/lib/identification/protobuf/card_activate_model.pb.dart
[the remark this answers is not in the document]
klinzo: Created an issue: https://github.com/digitalfabrik/ehrenamtskarte/issues/437

frontend/analysis_options.yaml
klinzo: What kind of problems do you think there could be?

.github/hooks/pre-commit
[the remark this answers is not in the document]
klinzo: Done. Sadly this does not work for me when committing from IntelliJ as the path env is different and does not include fvm for me. Dart and flutter are found though :/ Seems to be this issue: https://youtrack.jetbrains.com/issue/IDEA-220910
Will probably just keep a modified version of the hook on my machine.

frontend/lib/identification/card/eak_card.dart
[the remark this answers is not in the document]
klinzo: Okay so I investigates why this happens and it is because of the trailing commas/where the closing brackets are placed: https://github.com/digitalfabrik/ehrenamtskarte/pull/431/commits/27f144e70fef8f611b62c5a1cd901ee84eb992b0#diff-b2fe69abdd5b15dfac1487fed9af498543f5a23894799ff3bfab54d1db899f9aR54
Without a trailing comma and brackets in the same line (`]))`) it uses 4 spaces. So this is definitely a format issue. Not sure what one should do about that here.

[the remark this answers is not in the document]
klinzo: E.g. in line 68 it still uses 4 spaces, so it is mixed. Same as before (also sometimes 2 and sometimes 4 spaces).

frontend/pubspec.yaml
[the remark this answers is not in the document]
maxammann: wild.

frontend/analysis_options.yaml
maxammann: Any non-offical dependency can become outdated and unmaintained. We have several of such packages in the integreat project. If it is not really required to have a dependency then it is better to avoid it.

So just copy pasting the rules over could be more beneficial over the time.

.github/hooks/pre-commit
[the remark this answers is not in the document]
maxammann: Yes fvm is probably not in the path of intellij with a default installation. This can be changed by adding fvm to the path in ~/.xprofile file usually.

maxammann (approved):

frontend/analysis_options.yaml
klinzo: Lets just paste it over once we run into problems.

PR #461, opened 2022-01-18
Do not filter locations which do not have coordinates

### LINKED ISSUE #443: Do not filter locations which do not have coordinates
https://github.com/digitalfabrik/ehrenamtskarte/pull/441#discussion_r782001077)](https://github.com/digitalfabrik/ehrenamtskarte/pull/441#discussion_r782001077

klinzo:
Fixes #443 

This currently has no effect since there is only one store with no coordinates (and all other required attributes set):
> 'Kunstverein Würzburg e.V. (\"Arte Noah\"), Würzburg' was filtered out because longitude or latitude are null

It is not possible to get correct coordinates here since the street and plz are those of a Postfach:
``` xml
<name><![CDATA[Kunstverein Würzburg e.V. (\"Arte Noah\")]]></name>
<strasse><![CDATA[Postfach]]></strasse>
<hausnummer><![CDATA[110937]]></hausnummer>
<plz><![CDATA[97035]]></plz>
```

So the store is only filtered out in a different place. I think I'd merge this anyway as this may change over time and it does not hurt. If no coordinates are found we just filter it out later on.


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Sanitize.kt
```diff
@@ -23,6 +23,10 @@ class Sanitize(private val logger: Logger, httpClient: HttpClient) : PipelineSte
         val stateBbox = featureFetcher.queryFeatures(listOf(Pair("state", STATE))).first().bbox
 
         input.map { it.sanitize() }.filter {
+            if (it.longitude == null || it.latitude == null) {
+                logger.info("'${it.name}, ${it.location}' was filtered out because longitude or latitude are null")
+                return@filter false
+            }
```
Schedulaar, line 29: So does this mean, we filter them out at a later point if they make it to the "Sanitize" step?

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Store.kt
```diff
@@ -44,7 +44,7 @@ class Store(private val logger: Logger, private val manualImport: Boolean) : Pip
                     PhysicalStoreEntity.new {
                         storeId = storeEntity.id
                         addressId = address.id
-                        coordinates = Point(acceptingStore.longitude, acceptingStore.latitude)
+                        coordinates = Point(acceptingStore.longitude!!, acceptingStore.latitude!!)
```
Schedulaar, line 47: Why are these not null here?

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Sanitize.kt
klinzo: We filter locations without coordinates here instead of in the `Filter` step

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Store.kt
klinzo: Because they are filtered out in the `Sanitize` step where your other comment was. If we forget to do that, an exception is thrown here and the current store is skipped/not inserted in the database

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Sanitize.kt
Schedulaar: Then the title of the PR is wrong? We simply delay filtering accepting stores without coordinates, right?

Schedulaar (approved):
Maybe it makes sense, to have a `PreSanitizeFilter` and a `PostSanitizeFilter` for clarification 🤔

maxammann (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Sanitize.kt
klinzo: Only kind of. Stores without coordinates are first sanitized and only if they still have no coordinates afterwards filtered.

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/FilterBuilder.kt
```diff
@@ -1,59 +0,0 @@
-package app.ehrenamtskarte.backend.stores.importer
```
klinzo, line 1: Inlined to `PreSanitizeFilter`

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Filter.kt
```diff
@@ -1,21 +0,0 @@
-package app.ehrenamtskarte.backend.stores.importer.steps
```
klinzo, line 1: Renamed to `PreSanitizeFilter`

klinzo:
> Maybe it makes sense, to have a `PreSanitizeFilter` and a `PostSanitizeFilter` Pipeline steps for clarification thinking

Done

Schedulaar:
I had to restart the CI from a failed state...

PR #470, opened 2022-01-23
Reenable rotation to and from landscape mode on Android

### LINKED ISSUE #419: Reenable rotation to and from landscape mode on Android
**Describe the bug**
With https://github.com/ehrenamtskarte/ehrenamtskarte/issues/415 we set the orientation to portrait. While this is acceptable for phones, on tablets the preferred orientation is actually landscape.

**To Reproduce**
See https://github.com/ehrenamtskarte/ehrenamtskarte/issues/415

**Desktop (please complete the following information):**
 - OS: Android Emulator and Real Device

maxammann:
fixes #419


Schedulaar:
I just installed the app-release.apk from https://app.circleci.com/pipelines/github/digitalfabrik/ehrenamtskarte/912/workflows/8920355f-b301-4e08-bc96-1f93d2031941/jobs/6414/artifacts, but I found the issue sitll persists. I can rotate the app again, but there is still an area (at the right side in my case) after rotating to landscape which is unresponsive. :/

maxammann:
hm I suspect that I only fixed it for debug mode :/ I'll test it again.

maxammann:
@Schedulaar feel free to test again :)

frontend/lib/map/map/screen_parent_resizer.dart
```diff
@@ -31,12 +31,14 @@ class _ScreenParentResizer extends State<ScreenParentResizer> {
 
     final currentOrientation = MediaQuery.of(context).orientation;
     final isOverDrawingSquare = currentOrientation != _lastBuildOrientation;
-
     if (isOverDrawingSquare) {
+      // Delay setState by 2 frames. If we delay only by one frame, then Flutter is able to optimize this away.
```
Schedulaar, line 35: 😱

Schedulaar:
@maxammann I am sorry that I have to disappoint you, but it's still not working on my device :/

maxammann:
Oke, weird the ci build is not working :thinking:

maxammann:
@Schedulaar it seems only not to work if the intro slides are shown before that. Can you try to restart the app (latest build) and see whether the issue is still there?

Schedulaar:
Yep, it does indeed work after an app restart. However, I have noticed that sometimes (like 1 out of 50 times) there's a black screen where the map should be...

frontend/lib/entry_widget.dart
```diff
@@ -27,23 +27,26 @@ class EntryWidget extends StatelessWidget {
...
+          });
 
           if (settings.firstStart) {
-            routes.addAll(<String, WidgetBuilder>{
-              '/intro': (context) => IntroScreen(
-                    onFinishedCallback: () => settings.setFirstStart(enabled: false),
-                  ),
-            });
-            initialRoute = '/intro';
+            initialRoute = 'intro';
```
Schedulaar, line 42: First, do `routes.addAll( [...] )`, then declare
```dart
  String initialRoute = settings.firstStart ? 'intro' : '/';
```

frontend/lib/map/map/screen_parent_resizer.dart
```diff
@@ -0,0 +1,58 @@
+import 'dart:math';
+import 'package:flutter/material.dart';
+
+/// A widget which resizes the child widget twice when the orientation changes.
+///
+/// * Firstly, the child is resized to
+///   `width = height = max(screenWidth, screenHeight)`. This creates a square.
+/// * Secondly, the child is resized to
+///   `width = parentWidth, height = parentHeight. This sizes the child according to the parent constraints.
```
Schedulaar, line 9: Are you sure, these two steps are definitely necessary? Is it possible to omit the first step?

frontend/lib/map/map/screen_parent_resizer.dart
```diff
@@ -0,0 +1,58 @@
...
+        (_) => WidgetsBinding.instance!.addPostFrameCallback(
+          (_) => setState(() {
+            _lastBuildOrientation = currentOrientation;
+          }),
+        ),
+      );
+    }
+
+    final width = isOverDrawingSquare ? maximum : MediaQuery.of(context).size.width;
+    final height = isOverDrawingSquare ? maximum : MediaQuery.of(context).size.height;
```
Schedulaar, line 47: I assume, we have to choose other values than `MediaQuery.of(context).size.height`: Currently, the info button for tiles attribution dialog is not shown (I assume, that the height is slightly too large with the media query height).

Maybe we can use the OverflowBox only in the case of `isOverDrawingSquare` and otherwise simply return `child` as is?

Schedulaar (commented):
This seems (almost) perfect! Having the intro slides not above the main route is way better in my opinion. That way, if the location permissions are given in the intro slides, the app will start initialized properly after finishing the intro slides.

frontend/lib/map/map/screen_parent_resizer.dart
maxammann: Sadly it is required

klinzo:
@maxammann I'd review this tomorrow if you want :)

Schedulaar (commented):
Again closer to perfect, but not yet exactly there: When I rotate to landscape (given I start the app in portrait mode), the info icon is again not shown (because the map's actual size is the huge square).

Accordingly, if the app is started in landscape mode and then rotated to portrait mode, the same thing happens.

maxammann:
@Schedulaar thanks for noticing! I get now the correct constrains within the map should be placed.

frontend/lib/map/map/screen_parent_resizer.dart
```diff
@@ -0,0 +1,98 @@
+import 'dart:math';
+import 'package:flutter/material.dart';
+
+/// A widget which resizes the child widget twice when the orientation changes.
+///
+/// * Firstly, the child put in a container with the following constraints:
+///   `width = height = max(screenWidth, screenHeight)`. This creates a square.
+/// * Secondly, the child is resized to
+///   `width = parentWidth, height = parentHeight. This sizes the child according to the parent constraints. This
+///   resize only happens if the child is already initialized. If it is not yet initialized then the resize is delayed.
```
Schedulaar, line 10: I think this is outdated now? At least the second line...

Schedulaar (approved):
Buttery smooth now 😍

frontend/lib/map/map/screen_parent_resizer.dart
maxammann: Yes, thanks!

maxammann:
@klinzo feel free to test tomorrow :) then I'll merge this!

klinzo (approved):
Tested on android, did not experience any unresponsiveness.

PR #474, opened 2022-01-26
Add font to PDF

### LINKED ISSUE #354: jsPDF does not support utf8 out of the box
**Describe the bug**
jsPDF does not support utf8 out of the box.
Nevertheless, we should be able to render names from different countries from all over the world.
Also we'd like to have the German quotation marks for the sentence in the manual
```
1. Laden Sie sich die App "Ehrenamtskarte" herunter.
```

**To Reproduce**
1. Go to Administration, Karten drucken, Name etwa عائشة
2. Click on 'Karten drucken'
3. See errorneous PDF:

**Expected behavior**
The PDF should render the arabic name fine.

**Screenshots**
image

maxammann:
fixes #354


administration/src/components/generation/PdfFactory.ts
```diff
@@ -65,6 +65,12 @@ export function generatePdf(models: CardActivateModel[], region: Region) {
         unit: 'mm',
         format: 'a4'
     });
+    
+    let response = uint8ArrayToBase64(new Uint8Array(await (await fetch("/pdf-fonts/NotoSans-Regular.ttf")).arrayBuffer()));
```
maxammann, line 69: Error handling is missing here

docker/reverse_proxy/nginx-staging.conf
```diff
@@ -31,7 +31,7 @@ server {
     server_name api.staging.ehrenamtskarte.app;
 
     location / {
-        set $allowed_origin 'https://druckerei.staging.ehrenamtskarte.app';
+        set $allowed_origin '*';
```
klinzo, line 34: Why do we want to allow all origins? Can't we just allow the druckerei and the ttf download?

administration/src/components/generation/PdfFactory.ts
klinzo:
```suggestion
    const response = uint8ArrayToBase64(new Uint8Array(await (await fetch("/pdf-fonts/NotoSans-Regular.ttf")).arrayBuffer()));
```

klinzo (commented):
Not tested as I have never worked with the administration yet

docker/reverse_proxy/nginx-staging.conf
maxammann: So we would need to change this to 'https://druckerei.staging.ehrenamtskarte.app http://localhost:7000'`. Otherwise the staging server is not usable from a dev environment. Because this is not security critical I would just allow any host+port to tuse the API.

Schedulaar: I thought browsers would always allow localhost as an origin?

Schedulaar: I think security-wise, it may be useful to prevent other websites to use our API (to prevent users from using other third-party websites as a client, we also prevent phishing passwords through these third parties (?) )

Schedulaar (approved):
Also not tested

docker/reverse_proxy/nginx-staging.conf
maxammann: Yes absolutely but this is only the staging server. I would say usability goes before security here.

maxammann: > I thought browsers would always allow localhost as an origin? 

CORS is also active on localhosz

Schedulaar: > Yes absolutely but this is only the staging server. I would say usability goes before security here.

Ah, yeah, true!

PR #475, opened 2022-01-26
Check if glyphs are available

maxammann:
This PR adds two things
* Suggestion: A new result type which is in my opinion way better than throwing exceptions 
* Code for checking glyphs

I would like to refactor the web frontend soon with the Result type

### DIFF
--- a/administration/src/components/generation/GenerationController.tsx
+++ b/administration/src/components/generation/GenerationController.tsx
@@ -9,6 +9,7 @@ import downloadDataUri from "../../util/downloadDataUri";
 import generateCards from "./generateCards";
 import RegionSelector from "../RegionSelector";
 import {RegionContext} from "../../RegionProvider";
+import {Exception} from "../../exception";
 
 enum Mode {
     input,
@@ -37,7 +38,19 @@ const GenerationController = () => {
             setMode(Mode.finished)
         } catch (e) {
             console.error(e)
-            AppToaster.show({message: "Etwas ist schiefgegangen.", intent: "danger"})
+            if (e instanceof Exception) {
+                switch (e.data.type) {
+                    case "pdf-generation":
+                        AppToaster.show({message: "Etwas ist schiefgegangen beim erstellen der PDF.", intent: "danger"})
+                        break;
+                    case "unicode":
+                        AppToaster.show({message: `Ein Zeichen konnte nicht in der PDF eingebunden werden: ${e.data.unsupportedChar}`, intent: "danger"})
+                        break;
+                }
+            } else {
+
+                AppToaster.show({message: "Etwas ist schiefgegangen.", intent: "danger"})
+            }
             setMode(Mode.input)
         }
     }
--- a/administration/src/components/generation/GenerationController.tsx
+++ b/administration/src/components/generation/GenerationController.tsx
@@ -48,7 +48,6 @@ const GenerationController = () => {
                         break;
                 }
             } else {
-
                 AppToaster.show({message: "Etwas ist schiefgegangen.", intent: "danger"})
             }
             setMode(Mode.input)
--- a/administration/src/components/generation/PdfFactory.ts
+++ b/administration/src/components/generation/PdfFactory.ts
@@ -5,6 +5,7 @@ import {CardActivateModel} from "../../generated/compiled";
 import uint8ArrayToBase64 from "../../util/uint8ArrayToBase64";
 import {format, fromUnixTime} from "date-fns";
 import {getRegions_regions as Region} from "../../graphql/regions/__generated__/getRegions";
+import {Exception} from "../../exception";
 
 type TTFFont = {
     /**
@@ -82,6 +83,18 @@ Aussteller: ${region.prefix} ${region.name}`,
     });
 }
 
+function checkForeignText(doc: jsPDF, text: string): string | null {
+    let font = doc.getFont();
+    
+    for (let i = 0; i < text.length; i++) {
+        if (font.metadata.characterToGlyph(text.charCodeAt(i)) == 0) {
+            return text.charAt(i)
+        }
+    }
+
+    return null
+}
+
 export function generatePdf(font: TTFFont, models: CardActivateModel[], region: Region) {
     const doc = new jsPDF({
         orientation: 'portrait',
@@ -95,6 +108,16 @@ export function generatePdf(font: TTFFont, models: CardActivateModel[], region: 
     doc.setFont(font.name)
 
     for (let k = 0; k < models.length; k++) {
+        let model = models[k];
+        let unsupportedChar = checkForeignText(doc, model.fullName);
+        
+        if (unsupportedChar) {
+            throw new Exception({
+                type: "unicode",
+                unsupportedChar
+            })
+        }
+        
         addLetter(doc, models[k], region)
         if (k !== models.length - 1)
             doc.addPage()
@@ -107,5 +130,12 @@ export function generatePdf(font: TTFFont, models: CardActivateModel[], region: 
         creator: "Bayern"
     });
 
-    return new Blob([doc.output('blob')], {type: "application/pdf"})
+    try {
+        let output = doc.output('blob');
+        return new Blob([output], {type: "application/pdf"})
+    } catch {
+        throw new Exception({
+            type: "pdf-generation",
+        })
+    }
 }
--- a/administration/src/components/generation/PdfFactory.ts
+++ b/administration/src/components/generation/PdfFactory.ts
@@ -87,7 +87,7 @@ function checkForeignText(doc: jsPDF, text: string): string | null {
     let font = doc.getFont();
     
     for (let i = 0; i < text.length; i++) {
-        if (font.metadata.characterToGlyph(text.charCodeAt(i)) == 0) {
+        if (font.metadata.characterToGlyph(text.charCodeAt(i)) === 0) {
             return text.charAt(i)
         }
     }
--- a/administration/src/exception.ts
+++ b/administration/src/exception.ts
@@ -1,0 +1,28 @@
+
+type PDFGenerationError = {
+    type: "pdf-generation"
+}
+
+type PDFUnexpectedUnicodeError = {
+    type: "unicode"
+    unsupportedChar: string
+}
+
+type ErrorData = {
+    message?: string
+} & (PDFGenerationError | PDFUnexpectedUnicodeError)
+
+
+export class Exception extends Error {
+    public data: ErrorData;
+    
+    constructor(error: ErrorData) {
+        super()
+        if (Error.captureStackTrace) {
+            Error.captureStackTrace(this, Exception)
+        }
+
+        this.name = 'Exception'
+        this.data = error
+    }
+}


administration/src/util/result.ts
```diff
@@ -0,0 +1,96 @@
+type ResultError = { type: string }
```
klinzo, line 1: I think it would be nice to have the possible types as string literal types here

administration/src/components/generation/PdfFactory.ts
```diff
@@ -5,6 +5,34 @@ import {CardActivateModel} from "../../generated/compiled";
...
+    type: "pdf-generation"
+}
+
+type UnexpectedUnicodeError = {
+    type: "unicode"
+    unsupportedChar: string
+}
+
+function checkForeignText(doc: jsPDF, text: string): string | null {
+    let font = doc.getFont();
```
klinzo, line 22:
```suggestion
    const font = doc.getFont();
```

administration/src/components/generation/PdfFactory.ts
```diff
@@ -85,5 +122,6 @@ export function generatePdf(models: CardActivateModel[], region: Region) {
         creator: "Bayern"
     });
 
-    return new Blob([doc.output('blob')], {type: "application/pdf"})
+    let blob = doc.output('blob');
```
klinzo, line 125:
```suggestion
    const blob = doc.output('blob');
```

administration/src/components/generation/PdfFactory.ts
```diff
@@ -5,6 +5,34 @@ import {CardActivateModel} from "../../generated/compiled";
 import uint8ArrayToBase64 from "../../util/uint8ArrayToBase64";
 import {format, fromUnixTime} from "date-fns";
 import {getRegions_regions as Region} from "../../graphql/regions/__generated__/getRegions";
+import {Err, Ok, Result} from "../../util/result";
+
+export type PDFError = UnexpectedUnicodeError | PDFGenerationError
+
+type PDFGenerationError = {
```
klinzo, line 12: Is this used anywhere? Perhaps you should wrap the whole thing in a try catch and return this err on catch?

administration/src/components/generation/PdfFactory.ts
```diff
@@ -73,7 +101,16 @@ export function generatePdf(models: CardActivateModel[], region: Region) {
     doc.setFont("NotoSans-Regular")
 
     for (let k = 0; k < models.length; k++) {
-        addLetter(doc, models[k], region)
+        let model = models[k];
+        let unsupportedChar = checkCardActivateModel(doc, model);
```
klinzo, line 105:
```suggestion
        const model = models[k];
        const unsupportedChar = checkCardActivateModel(doc, model);
```

klinzo (commented):
Like it :+1: Not tested.

administration/src/util/result.ts
maxammann: This is the very very very generic case. This is just the type-shape of Errors in general. In another file an error type specific for the administration frontend could be defined.

administration/src/components/generation/PdfFactory.ts
maxammann: I think this is worth discussing. There can always be unexpected crashes, but I dont think we should just add try catches for errors we do not expect.

klinzo: But then this is currently unused and could/should be removed, right?

maxammann: oh sorry, yes, I forgot to use this error type. I planned to use it around the .output() function :) thanks!

administration/src/components/generation/PdfFactory.ts
Schedulaar, line 105:
```
        let unsupportedChar = checkCardActivateModel(doc, model);
```
The variable name is weird, as it is not clear from the function called what it means.
I think you should either rename the variable or the function.

Schedulaar (approved):
I do like the Rust way of handling errors as well, but I do not think, that it should be applied to JS too much. We will never get rid of `try .. catch` blocks anyways as we use third party libraries. The problem here is, that JS has `throw` baked in and it is widely used. Therefore, if we use the Rust way here, we will get some strange mix of error handling throughout the code which is never appreciable.

Also I do think that we should (at some points) catch any error in a predefined process, so the rest of the app does not crash and go blank and instead it shows some nice error message, that something went wrong. (However this can also be handled at some global place, I guess).

If you want to catch predefined errors you can simply create a new Error Type and throw and catch a specific error type. (Of course you do not have the nice, exhaustive handling that Rust has that way). We could however add jsdoc's @throws documentation to these functions (or is that already outdated?)

We can leave it for the PDF process. However note that if e.g. jsPDF has some internal error, we will (probably) leave the user with an unusable, broken app. (or there might be no feedback to the user pushing the "Print PDF" button).

maxammann:
> I do like the Rust way of handling errors as well, but I do not think, that it should be applied to JS too much. We will never get rid of `try .. catch` blocks anyways as we use third party libraries. The problem here is, that JS has `throw` baked in and it is widely used. Therefore, if we use the Rust way here, we will get some strange mix of error handling throughout the code which is never appreciable.
> 
> Also I do think that we should (at some points) catch any error in a predefined process, so the rest of the app does not crash and go blank and instead it shows some nice error message, that something went wrong. (However this can also be handled at some global place, I guess).
> 
> If you want to catch predefined errors you can simply create a new Error Type and throw and catch a specific error type. (Of course you do not have the nice, exhaustive handling that Rust has that way). We could however add jsdoc's @throws documentation to these functions (or is that already outdated?)

I agree with all of your points, but my conclusion is different, because the alternatives are very bad. Adding a new error type results in the following code:

```ts
try {
  myroutine(); // There's a couple of errors thrown here
} catch (e) {
  if (e instanceof TypeError) {
    // A TypeError
  } else if (e instanceof RangeError) {
    // Handle the RangeError
  } else if (e instanceof EvalError) {
    // you guessed it: EvalError
  } else if (typeof e === "string") {
    // The error is a string
  } else if (axios.isAxiosError(e)) {
    // axios does an error check for us!
  } else {
    // everything else  
    logMyErrors(e);
  }
}
```
(https://fettblog.eu/typescript-typing-catch-clauses/)

The way I see it is that after 1-2 years of development nobody knows which error can happen when calling any function. 
I even forgot which error classes we have in the Integreat api-client or in the Integreat app :D
By introducing a `Result` type early on this is clearly documented in the signature of a function. I would suggest that we never `throw` in our application and always use Result. If we call external libraries then we should check the libraries documentation whether it can fail. If it can fail, then handle the errors and wrap them in a `Result`. This should avoid the "mix of error handling throughout the code".

While this is a non-default JS way of handling errors, this should work fine in an application. Library developers should never do something like that. They should stick to the language default.

I agree that we should avoid the app getting blank or anything like that. This can be achieved by Error Boundaries: https://reactjs.org/docs/error-boundaries.html
Uncought errors in onClick etc. will go silent nonetheless.

administration/src/util/result.ts
```diff
@@ -0,0 +1,96 @@
...
+export function Err<E extends ResultError, T = never>(error: E): Result<T, E> {
+    return new _Err<T, E>(error);
+}
+
+/**
+ * This type was adopted from https://github.com/badrap/result
+ */
+export type Result<T, E extends ResultError = ResultError> =
+    | _Ok<T, E>
+    | _Err<T, E>;
```
Schedulaar, line 96: Is there a reason to copy&paste the code instead of using the npm version?
Maybe you want to document the changes you made to the original code here.

administration/src/components/generation/GenerationController.tsx
```diff
@@ -30,17 +31,30 @@ const GenerationController = () => {
...
+            },
+            (error: PDFError) => {
+                console.error(error)
+                switch (error.type) {
+                    case "pdf-generation":
+                        AppToaster.show({message: "Etwas ist schiefgegangen.", intent: "danger"})
+                        break;
+                    case "unicode":
+                        AppToaster.show({message: `Ein Zeichen konnte nicht in der PDF eingebunden werden: ${error.unsupportedChar}`, intent: "danger"})
+                        break;
```
Schedulaar, line 48: Is the `switch` statement exhaustive in typescript?

Schedulaar (approved):
I dont have a strong preference between both methods.
However, in this particular case, I'd like to see some fallback catching internal errors of jsPDF, as in the complexity of PDF creation there might always occur unexpected errors.

I think we can hook ourselves into the global onError callback and trigger the AppToaster?

administration/src/components/generation/GenerationController.tsx
maxammann: kind of, I think there is at least a warning

maxammann:
> I dont have a strong preference between both methods. However, in this particular case, I'd like to see some fallback catching internal errors of jsPDF, as in the complexity of PDF creation there might always occur unexpected errors.
> 
> I think we can hook ourselves into the global onError callback and trigger the AppToaster?

I think that will end badly :see_no_evil: 
I would only do this if it is documented and allowed by the spec to overwrite for example `console.error`
Yes when calling third party libraries I would always wrap it in a try-catch if I would suspect that it is "failable".

administration/src/util/result.ts
maxammann: Yes, it would add a third party dependency. The code we are using now is very simple in my opinion. I would even say it is pretty much bug-less and feature-complete.

I think there is no need to document changes. Let's just pretend I wrote all myself.

maxammann:
> I think we can hook ourselves into the global onError callback and trigger the AppToaster?

There is an event handler for this: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror#window.addeventlistenererror

I think this needs maybe some trial and error to gain some experience under which circumstances this is called. The documentation is not that clear about it imo.

Schedulaar:
I think it's pretty well documented [here]( https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror). I it (for example) called when a java runtime error is thrown (and not caught anywhere).
I think you only have to be careful not to create/throw an error inside the onError handler to avoid an endless loop of death^^

maxammann:
Lets decide this on Thursday: Legacy JS way vs. Results

Schedulaar:
@maxammann "Legacy JS way" is not really legacy 😅

maxammann:
> @maxammann "Legacy JS way" is not really legacy 😅 

Maybe its the wrong word, lets use traditional/usual

maxammann:
@klinzo @Schedulaar what do you think about this? It uses the traditional error handling method, combined with a "sum type" using a TS switch-case

Schedulaar:
A TS switch-case is exhaustive? If that is the case, then I disagree with the current solution, as long as you don't rename Exception to PDFException or so. Otherwise, we'd have to include a path for every PDF-Generation-Kind-Of-Error at any part in the app where we would catch errors, right?

maxammann:
> A TS switch-case is exhaustive? If that is the case, then I disagree with the current solution, as long as you don't rename Exception to PDFException or so. Otherwise, we'd have to include a path for every PDF-Generation-Kind-Of-Error at any part in the app where we would catch errors, right?

It does not force you to handle all cases, but suggests it it autocomplete.

Schedulaar (approved):
Ah, alright 👍🏽

maxammann:
I would just suggest to use this error class (the Exception class) whenever possible :) probably will try my best during the refactoring

PR #468, opened 2022-01-21
Sanitize street and house numbers

klinzo:
Fixes #467

PR #480, opened 2022-01-28
Split and rename pipeline steps

klinzo:
Follow up on #468. Renames pipeline steps to include `Lbe` and splits the `Map` step into `MapFromLbe` and `SanitizeAddress`.


maxammann (approved):
Looks great!

PR #486, opened 2022-02-14
Change icon of Freizeit/Reisen category

### LINKED ISSUE #417: Change icon of Freizeit/Reisen category
**Describe the bug**
The current icon looks like a trash bin. Try to find a better one.

klinzo:
Fixes #417

Should I also adapt all other icons to use maki? We could also use this suitcase as icon:
image


maxammann:
I would say that we change all icons if they do not divert too much.

maxammann (approved):

maxammann (changes requested):
I just remembered we also need to change them on the map. This needs to be done here: https://github.com/digitalfabrik/ehrenamtskarte-maplibre-style

Schedulaar:
I suggest, we either replace all icons with the new icon set or none :D But I do like the palm+island a lot better than the suit case :)

klinzo:
Some icons are fairly similar, some are quite a bit different/no equivalent available.

klinzo:
@maxammann @Schedulaar please have a look. I am not really a fan of the icons tbh.

Schedulaar:
To quickly toggle to rich diff for all files in the PR:
```js
[...document.getElementsByTagName("button")].filter(btn => btn.getAttribute("aria-label") == "Display the rich diff").forEach(btn => btn.click())
```
You can also add a bookmark with the following contents to trigger it:
```
javascript:[...document.getElementsByTagName("button")].filter(btn => btn.getAttribute("aria-label") == "Display the rich diff").forEach(btn => btn.click())
```

maxammann:
> Except for the bank-icon I think all replacements are okay.

I think icon "2" is weird. The category is sport, right?

maxammann:
How about this for sports?
soccer

klinzo:
> How about this for sports? soccer (cultural backgorund heh :D)

Fine for me, that would have been my second choice. If all other icons work, for you, I'll open a PR for the mapstyle as well.

klinzo:
Where should I document the icon source?

maxammann:
For me it is obvious that icons come from maki :)

Schedulaar:
We could add a README.md in `frontend/assets/category-icons` with a content similar to
```markdown
# Category Icons
This folder contains the icons of the different categories of accepting stores.
They should always be in-sync with the icons used in [https://github.com/digitalfabrik/ehrenamtskarte-maplibre-style](https://github.com/digitalfabrik/ehrenamtskarte-maplibre-style).
```

maxammann:
I merged the icon task for the map

maxammann:
New spritesheet is deployed here: https://maps.tuerantuer.org/styles/ehrenamtskarte/sprites.png

Note, that app installations which still use the old legacy map style wont get the updated icons.

klinzo:
@maxammann anything missing here? You have still requested changes.

maxammann (approved):

PR #496, opened 2022-02-21
Fix import

### LINKED ISSUE #490: Fix import
**Describe the bug**
With the new data format in bavaria, we are required to adjust the parsing of the XML. New fields are `p_eak_agentur_id` and `anbieter_name`. If any of the values is an empty string, it should be interpretet as None/null. The first is an empty string or a number, the latter a string or an empty string.

klinzo:
Fixes #490

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/FilterDuplicates.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/FilterDuplicates.kt
@@ -23,6 +23,8 @@ class FilterDuplicates(private val logger: Logger) : PipelineStep<List<Accepting
         val store = last()
 
         logger.logRemoveDuplicates(store, size - 1)
+        val issuers = map { "'${it.districtName ?: it.freinetId.toString()}'" }.toSet().joinToString()
+        logger.info("Duplicates issued by $issuers")
 
         val location = lastValue("locations") { it.location }
         val categoryId = lastValue("categoryIds") { it.categoryId }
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/types/AcceptingStore.kt
```diff
@@ -15,6 +15,8 @@ data class AcceptingStore(
     val telephone: String?,
     val website: String?,
     val discount: String?,
+    val freinetId: Int?,
+    val districtName: String?,
```
klinzo, line 19: Not sure about naming here. We want two separate properties, right? And not just one `issuerId` containing either the freinet id or the district name?

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/FilterDuplicates.kt
```diff
@@ -23,6 +23,8 @@ class FilterDuplicates(private val logger: Logger) : PipelineStep<List<Accepting
         val store = last()
 
         logger.logRemoveDuplicates(store, size - 1)
+        val issuers = map { "'${it.districtName ?: it.freinetId.toString()}'" }.toSet().joinToString()
```
Schedulaar, line 26: Is this the same as 
```kotlin
        val issuers = it.districtName ?: it.freinetId.toString()
```
?

Schedulaar (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/types/AcceptingStore.kt
maxammann: Merging those will be part of an other task

maxammann (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/FilterDuplicates.kt
klinzo: No, this lists the freinetIds/districtNames of all duplicates. So it is something like `'508', '963'`, `'Landkreis Fürth', '766'` or `'Landkreis Bad Tölz-Wolfratshausen'`, i.e. multiple freinetIds and districtNames for logging purposes.

Schedulaar: Ah, `map` calls "this.map"... I was confused because of the syntax :) Thanks for the clarification.

PR #497, opened 2022-02-24
Add documentation

klinzo:
This was a request by the supervisor of my Praxismodul :grinning:


maxammann (approved):
I would prefer to have the descriptions at the class-level instead of the function level, but I'm fine how it is right now :)

PR #531, opened 2022-04-13
Backend configuration

klinzo:


klinzo:
@maxammann not sure how do I get the config.yml file to `/etc/entitlementcard`? I copied it manually on my machine, seems like everything is working fine already.

Can I just remove this line now? https://github.com/digitalfabrik/entitlementcard/blob/b78f97f2ae852340671581e02f2e9f3562d7120c/docker-compose.staging.yml#L6

backend/build.gradle.kts
```diff
@@ -75,3 +76,7 @@ application {
 tasks.withType<JavaExec>().configureEach {
     systemProperties = properties
 }
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
```
klinzo, line 80: Without this I got the following error:
> Cannot inline bytecode built with JVM target 1.8 into bytecode that is being built with JVM target 1.6

Got this from here: https://stackoverflow.com/a/50991772

Is this fine?

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
```diff
@@ -3,28 +3,31 @@ package app.ehrenamtskarte.backend
...
 import java.io.File
 
 class Entry : CliktCommand() {
-    companion object {
-        fun isProductionEnvironment(): Boolean {
-            return System.getProperty("app.production", "").isNotEmpty()
-        }
-    }
+    private val config by option().file(canBeDir = false, mustBeReadable = true).required()
+    private val projectId by option().required()
```
klinzo, line 21: Sadly there is not really a built in functionality of clikt to automatically read properties of data classes, therefore we'd have to manually add options and mapping for all properties to be able to overwrite values of the config. Should I do that? And should I do it for all properties?
I am referring to the following from the issue #531 :
> It might be useful to override config options through the CLI. Maybe our CLI framework offers a feature for this. If not done in this task, please create a new issue.

klinzo:
This line in the docs needs probably adjustments now: https://github.com/digitalfabrik/entitlementcard/blob/b78f97f2ae852340671581e02f2e9f3562d7120c/docs/staging-setup.md?plain=1#L126

backend/build.gradle.kts
maxammann: Thats fine. Its becaude of the jackson yaml module, right?

backend/gradle.properties
```diff
@@ -1,15 +1,3 @@
...
 
 exposed_version=0.28.1
-
-app.postgres.url=jdbc:postgresql://localhost:5432/ehrenamtskarte
-app.postgres.user=postgres
-app.postgres.password=postgres
-
-app.import.xml=https://www.lbe.bayern.de/engagement-anerkennen/ehrenamtskarte/akzeptanzstellen/app-daten.xml
-app.import.json_1=https://bayerische_eak:
-app.import.json_2=@freinet-online.de/files/api_bayerische_eak/download.php?get_file=freinet_online_akzeptanzpartner.json
```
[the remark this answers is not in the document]
maxammann: Probably true. The freinet importer is gone probably.

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
maxammann: I think it has no high priority. So a ticket with your findings is enough. You can assign low priority.

backend/build.gradle.kts
klinzo: No, because of the `requireObject` method of clickt

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
```diff
@@ -3,28 +3,31 @@ package app.ehrenamtskarte.backend
...
+import com.github.ajalt.clikt.parameters.types.file
 import java.io.File
 
 class Entry : CliktCommand() {
-    companion object {
-        fun isProductionEnvironment(): Boolean {
-            return System.getProperty("app.production", "").isNotEmpty()
-        }
-    }
+    private val config by option().file(canBeDir = false, mustBeReadable = true).required()
```
maxammann, line 20: This should be optional. Some search for a config should be performed here. First look in the current working directory. If there is no `config.yml` then look in `/etc/entitlementcard.yml/` and if no config if found we could warn and choose a hardcoded default.

.idea/runConfigurations/Import.xml
```diff
@@ -16,8 +16,7 @@
       <option name="taskNames">
         <list>
           <option value="run" />
-          <option value="run" />
-          <option value="--args=&quot;import&quot;" />
+          <option value="--args=&quot;--config /etc/entitlementcard/config.yml --project-id bayern.ehrenamtskarte.app import&quot;" />
```
maxammann, line 19: These run configs are not platform independent. They should run without changing `/etc`.

.idea/runConfigurations/GraphQL_Schema_Export.xml
```diff
@@ -16,8 +16,7 @@
       <option name="taskNames">
         <list>
           <option value="run" />
-          <option value="run" />
-          <option value="--args=&quot;graphql-export ../specs/backend-api.graphql&quot;" />
+          <option value="--args=&quot;--config /etc/entitlementcard/config.yml --project-id bayern.ehrenamtskarte.app graphql-export ../specs/backend-api.graphql&quot;" />
```
maxammann, line 19: I think project-id should not be required for this.

maxammann (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
klinzo: Done in this ticket only for properties that I thought may be configured separately.

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
klinzo: Done (without the warning part)

maxammann (approved):

maxammann:
Merged with force because CDN is down.

PR #537, opened 2022-05-10
Add projects to database

### LINKED ISSUE #509: Add Project ID to Database
Every object stored in the database requires a reference to a project ID.

 The following objects for example need a reference to a project:
* Store
* Region
* Card

A new table must be introduced which lists the projects. Also create the configured projects (#506) in the database if they do not yet exist. The primary key for the table will be the project ID like `bayern.ehrenamtskarte.app`.

None

**Additional context**
None

steffenkleinle:
Fixes #509

This adds the projects of the backend configuration to the database. Regions now always reference/belong to a project. Stores now always reference/belong to a region (a mock region is used for the eak stores for now, should be changed in the future by automatically creating the regions using the app-daten.xml).

Regions and projects are now automatically updated and deleted as well instead of always deleting and creating everything again. This should prevent errors when trying to delete regions referenced in cards or stores.


backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Schema.kt
```diff
@@ -1,12 +1,15 @@
...
+import app.ehrenamtskarte.backend.projects.database.Projects
 import org.jetbrains.exposed.dao.IntEntity
 import org.jetbrains.exposed.dao.IntEntityClass
 import org.jetbrains.exposed.dao.id.EntityID
 import org.jetbrains.exposed.dao.id.IntIdTable
+import org.jetbrains.exposed.sql.ReferenceOption
 
 object Regions : IntIdTable() {
-    val regionIdentifier = varchar("regionIdentifier", 5).uniqueIndex() // 5-stelliger Kreisschlüssel
+    val projectId = reference("projectId", Projects, onDelete = ReferenceOption.CASCADE)
```
klinzo, line 11: Should I set `onDelete = ReferenceOption.CASCADE` for cards as well?

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -1,11 +1,418 @@
...
+                dbRegion.name = eakRegion[1]
+                dbRegion.prefix = eakRegion[0]
+                dbRegion.website = eakRegion[3]
+            }
+        }
+    }
+}
+
+private fun ProjectEntity.regionIdentifier(): String {
+    return project.hashCode().toString().slice(0..4)
```
klinzo, line 417: any better ideas on how to create a unique 5 digit id?

backend/build.gradle.kts
```diff
@@ -80,3 +80,7 @@ tasks.withType<JavaExec>().configureEach {
 tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
     kotlinOptions.jvmTarget = "1.8"
 }
+
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
+    kotlinOptions.jvmTarget = "1.8"
+}
```
klinzo, line 86:
```suggestion
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/projects/database/Setup.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    transaction {
+        val dbProjects = ProjectEntity.all()
+
+        // Delete projects not existing anymore in database
+        dbProjects.forEach { dbProject ->
+            if (config.projects.none { it.id == dbProject.project }) {
+                dbProject.delete()
+            }
+        }
```
maxammann, line 20: I would say that we should not delete them, but let them be ophans in the DB :D

Else its possible that data gets deleted if I change/malformat the config.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -1,11 +1,420 @@
...
+        val dbRegions = RegionEntity.all()
+
+        // Delete old regions in database
+        dbRegions.forEach { dbRegion ->
+            val isEakRegion = eakRegions.any { it[2] == dbRegion.regionIdentifier }
+            val isProjectRegion = projects.any { it.regionIdentifier() == dbRegion.regionIdentifier }
+            if (!isEakRegion && !isProjectRegion) {
+                dbRegion.delete()
+            }
+        }
```
maxammann, line 379: Should not happen because its dangerous.

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/database/Schema.kt
```diff
@@ -46,6 +48,7 @@ class AcceptingStoreEntity(id: EntityID<Int>) : IntEntity(id) {
 }
 
 object PhysicalStores : IntIdTable() {
+    val regionId = reference("regionId", Regions, onDelete = ReferenceOption.CASCADE)
```
maxammann, line 51:
```suggestion
    val regionId = reference("regionId", Regions)
```

maxammann (changes requested):

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
steffenkleinle: It has to be unique (so null, undefined, mock) is not possible for both. The project id is not 5 chars long.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Schema.kt
steffenkleinle: Since we decided not to delete projects and regions anymore this can be removed again.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -1,11 +1,418 @@
 package app.ehrenamtskarte.backend.regions.database
 
+import app.ehrenamtskarte.backend.projects.database.ProjectEntity
 import org.jetbrains.exposed.sql.SchemaUtils
+import org.jetbrains.exposed.sql.transactions.transaction
 
-fun setupDatabase(executeScript: (path: String) -> Unit) {
+private val eakRegions = listOf(
```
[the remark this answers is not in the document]
Schedulaar: I guess it would be nice, if you could still move this data stuff to a separate file, so we don't mix code and data.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/constants.kt
```diff
@@ -0,0 +1,4 @@
+package app.ehrenamtskarte.backend.common.webservice
+
+const val DEFAULT_PROJECT = "bayern.ehrenamtskarte.app"
+const val EAK_PROJECT = DEFAULT_PROJECT
```
Schedulaar, line 4:
```suggestion
const val EAK_BAYERN_PROJECT = "bayern.ehrenamtskarte.app"
const val DEFAULT_PROJECT = EAK_BAYERN_PROJECT
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -1,11 +1,413 @@
...
+    transaction {
+        val projects = ProjectEntity.all()
+        val dbRegions = RegionEntity.all()
+
+        // Create a mock region for each project to save the stores to
+        projects.forEach { project ->
+            // Project ids and therefore the mock regions should never change
+            // Therefore we only need to create the mock region if not already existing
+            if (dbRegions.none { it.regionIdentifier == project.regionIdentifier() }) {
+                // TODO #538: Perhaps no mock region will be needed anymore
```
Schedulaar, line 378:
```suggestion
        // Create a dummy region for each project to save the stores to
        projects.forEach { project ->
            // Project ids and therefore the dummy regions should never change
            // Therefore, we only need to create the dummy region if it does not already exist
            if (dbRegions.none { it.regionIdentifier == project.regionIdentifier() }) {
                // TODO #538: Perhaps no dummy region will be needed anymore
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -1,11 +1,413 @@
...
+                RegionEntity.new {
+                    projectId = project.id
+                    name = project.project
```
Schedulaar, line 381: optional:
```suggestion
                    name = "DUMMY FOR " + project.project
```

backend/src/main/resources/sql/create_regions.sql
```diff
@@ -1,92 +0,0 @@
-INSERT INTO Regions ("prefix", "name", "regionIdentifier", "website") VALUES ('Landkreis', 'Aichach-Friedberg', '09771', 'https://lra-aic-fdb.de/hier-leben/ehrenamt/bayerische-ehrenamtskarte') ON CONFLICT ("regionIdentifier") DO UPDATE SET "prefix"='Landkreis', "name"='Aichach-Friedberg', "website"='https://lra-aic-fdb.de/hier-leben/ehrenamt/bayerische-ehrenamtskarte';
```
Schedulaar, line 1: There was documentation on how to create this sql file. We should probably adjust that documentation.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
Schedulaar: I think we should probably make the regionIdentifier nullable and use null. A dummy region does not have a kreisschlüssel nevertheless :) (also for other projects this kreisschlüssel semantic might be too specific anyways...)

(
In a unique, but nullable column, null values can appear multiple times, they are not subject to the uniqueness constraint: From https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS

> In general, a unique constraint is violated if there is more than one row in the table where the values of all of the columns included in the constraint are equal. However, two null values are never considered equal in this comparison. That means even in the presence of a unique constraint it is possible to store duplicate rows that contain a null value in at least one of the constrained columns. This behavior conforms to the SQL standard, but we have heard that other SQL databases might not follow this rule. So be careful when developing applications that are intended to be portable.

)

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/steps/Store.kt
```diff
@@ -19,6 +21,8 @@ class Store(config: ImportConfig, private val logger: Logger) : PipelineStep<Lis
 
     override fun execute(input: List<AcceptingStore>) {
         transaction {
+            // TODO #538: The right region should be used instead of the mock region
```
Schedulaar, line 24:
```suggestion
            // TODO #538: The right region should be used instead of the dummy region
```

Schedulaar (approved):

maxammann:
"A wild Schedular appeared."

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
steffenkleinle: As I have to use the name to query the region I'd keep it as just the project id without the dummy.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
steffenkleinle: Done

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/webservice/schema/types/Region.kt
```diff
@@ -5,5 +5,5 @@ data class Region(
     val id: Int,
     val prefix: String,
     val name: String,
-    val regionIdentifier: String
+    val regionIdentifier: String?
```
steffenkleinle, line 8: @maxammann @Schedulaar does this have any implications for the frontend/administration? Anywhere this needs to be handled?

maxammann: It could be used in the `administration` frontend. But that would be no problem, because that does not need to keep compatibility.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -1,11 +1,53 @@
...
+                    projectId = eakProject.id
+                    name = eakRegion[1]
+                    prefix = eakRegion[0]
+                    regionIdentifier = eakRegion[2]
+                    website = eakRegion[3]
+                }
+            } else {
+                dbRegion.name = eakRegion[1]
+                dbRegion.prefix = eakRegion[0]
+                dbRegion.website = eakRegion[3]
```
maxammann, line 49: Does this actually perform an update in the DB?

maxammann (approved):
Looks good!

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
steffenkleinle: Yes, see https://github.com/JetBrains/Exposed/wiki/DAO#update
> Update a value of a property similar to any property in a Kotlin class:
movie.name = "Episode VIII – The Last Jedi"

backend/src/main/resources/sql/create_regions.sql
steffenkleinle: Will do in #540

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
maxammann: magic

maxammann: @steffenkleinle but should we flush the updates to be sure that they are done in time?

steffenkleinle: I think it should be handled already, it definitely happens
> before next select * from the database.

which should be enough for us, right?

maxammann: yes, but if the server crashes before that the changes are not executed. I'm not sure whether that could cause problems.

steffenkleinle: yeah, but then no changes are executed at all since we rollback the changes, right?

maxammann: Not sure, this all depends on how this code is called

PR #540, opened 2022-05-13
Whitelabel backend api

### LINKED ISSUE #508: Add Project ID to API
Each GraphQL query/mutation should require a project id. For compatibility reasons this field should still be optional and the fallback `bayern.ehrenamtskarte.app` should be used.

TODO

None

**Additional context**
None

steffenkleinle:
Fixes #508


steffenkleinle:
@Schedulaar would be nice if you could review this. The data loaders are now not used directly anymore for the `findByIds` graphql queries. If I understood you correct they are still used under the hood though since they are registered in the registry, right?
I think there is no way to supply a project to the data loaders to do the selecting there, right? The only other way I could think of was to first select the ids of all stores/regions matching the project and then do another query using the data loaders to fetch the actual other columns.

Schedulaar (commented):
Sooo I think the main issue is, that there is no clear location, where we would fail if an id that is requested in a graphql query does not exist in the DB. Error handling was not really handled explicitly in the backend for this case (yet :) ).

I think in such a case, your changes break the dataloaders in a very weird way as it expects outputs[i] corresponds to inputs[i].
I think as a general rule, we might employ:
**"DataLoaders and Repository Functions of type findByIds return a List of nullables (where output{i}=null means there was no element with id input[i]).
 GraphQL Services that use DataLoaders or Repository functions of type findByIds check if there is a null element in the list (and give an appropriate error to the API user)"**

We might also fail early (say in Dataloaders and repository functions) however, I am unsure what other side effects this might have {and the graphql service should still handle the failure explicitly to give a good error message to the user).

Hence:
I would suggest reverting all changes to the dataloaders, the sortByKey and the repository.findByIds functions.
For the adjusted graphql requests (with the argument `project`), I would suggest creating new functions in the repositories (.findAllByProject(project) and .findByIdsInProject(ids, project)) and handle null values as described above.

steffenkleinle:
> Sooo I think the main issue is, that there is no clear location, where we would fail if an id that is requested in a graphql query does not exist in the DB. Error handling was not really handled explicitly in the backend for this case (yet :) ).
> 
> I think in such a case, your changes break the dataloaders in a very weird way as it expects outputs[i] corresponds to inputs[i]. I think as a general rule, we might employ: **"DataLoaders and Repository Functions of type findByIds return a List of nullables (where output[i]=null means there was no element with id input[i]). GraphQL Services that use DataLoaders or Repository functions of type findByIds check if there is a null element in the list (and give an appropriate error to the API user)"**

Hmm I see, didn't think of that. Thanks for your opinion :)

> Hence: I would suggest reverting all changes to the dataloaders, the sortByKey and the repository.findByIds functions. For the adjusted graphql requests (with the argument `project`), I would suggest creating new functions in the repositories (.findAllByProject(project) and .findByIdsInProject(ids, project)) and handle null values as described above.

I agree with reverting all dataloaders, sortByKey and findByIds. However, I don't think we should have endpoints available not taking a project as param. There is no usecase where we want to query for regions or stores of multiple projects and it will perhaps be missed that there are multiple projects in the future. Also, this leads to old eak bayern app installations (which do not use the new endpoints) receiving regions and stores of other projects once we add regions/stores for other projects.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
```diff
@@ -1,20 +1,29 @@
...
 
     fun findByIds(ids: List<Int>) =
         RegionEntity.find { Regions.id inList ids }.sortByKeys({ it.id.value }, ids)
 
-    fun findById(id: Int) =
-        RegionEntity.find { Regions.id eq id }.singleOrNull()
-
-    fun findByRegionIdentifiers(ids: List<String>) =
-        RegionEntity.find { Regions.regionIdentifier inList ids }.sortByKeys({ it.regionIdentifier }, ids)
-
```
maxammann, line 19: I think this endpoint should stay. Or was it unused?

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/webservice/schema/RegionsQueryService.kt
```diff
@@ -1,28 +1,34 @@
...
-        environment.getDataLoader<Int, Region>(REGION_LOADER_NAME).loadMany(params.ids)
+    @GraphQLDescription("Returns regions queried by ids in the given project.")
+    fun regionsByIdInProject(project: String, ids: List<Int>): List<Region?> = transaction {
+        RegionsRepository.findByIdsInProject(project, ids).map {
+            if (it == null) null
+            else Region(it.id.value, it.prefix, it.name, it.regionIdentifier)
+        }
+    }
+
+    @Deprecated("Deprecated in favor of project specific query", ReplaceWith("regionsInProject"))
```
maxammann, line 27: Looks good!

maxammann (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
steffenkleinle: Both functions were unused (and they are no endpoints), so I think it is fine to remove them.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/webservice/schema/RegionsQueryService.kt
```diff
@@ -1,28 +1,34 @@
...
     }
 
-    @GraphQLDescription("Returns regions queried by ids.")
-    fun regionsById(
-        params: IdsParams,
-        environment: DataFetchingEnvironment
-    ): CompletableFuture<List<Region>> =
-        environment.getDataLoader<Int, Region>(REGION_LOADER_NAME).loadMany(params.ids)
+    @GraphQLDescription("Returns regions queried by ids in the given project.")
+    fun regionsByIdInProject(project: String, ids: List<Int>): List<Region?> = transaction {
```
steffenkleinle, line 20: Changing the return type from `CompletableFuture<List<Region>>` to `List<Region?>` is not a breaking change, is it?

maxammann: You can check the generated schema I think.

maxammann:
> One other question: Should I add a flag to the dummy regions to prevent them from being delivered to the apps in the region endpoints? AFAICS the regions endpoints are only used for card identification/registration though which is not enabled in the apps yet, right?

I think that is not important right now. We can just do it later if its a problem for us.

steffenkleinle:
Sorry for rebasing

maxammann:
Please avoid force pushing after a review has been done :see_no_evil:

steffenkleinle:
> Please avoid force pushing after a review has been done see_no_evil

yes, thats why I apologized :)

maxammann:
> > Please avoid force pushing after a review has been done see_no_evil
> 
> yes, thats why I apologized :)

sorry did not see it :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardQueryService.kt
```diff
@@ -8,7 +9,7 @@ import org.postgresql.util.Base64
 @Suppress("unused")
 class CardQueryService {
     @GraphQLDescription("Returns whether there is a card with that hash registered for that this TOTP is currently valid")
-    fun verifyCard(card: CardVerificationModel): Boolean {
-        return CardVerifier.verifyCardHash(Base64.decode(card.cardDetailsHashBase64), card.totp)
+    fun verifyCard(project: String = DEFAULT_PROJECT, card: CardVerificationModel): Boolean {
```
Schedulaar, line 12:
```suggestion
    fun verifyCard(project: String, card: CardVerificationModel): Boolean {
```

I think we can also break this endpoint.

Schedulaar (approved):
Although it is Okay to break the API for production, I would think it makes sense to have the create+verify endpoints adjusted in the backend and frontend asap. Otherwise, there's a lot of broken code in the repo.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/webservice/schema/RegionsQueryService.kt
steffenkleinle: it does not :+1:

steffenkleinle:
> Although it is Okay to break the API for production, I would think it makes sense to have the create+verify cards endpoints adjusted in the backend and frontend asap. Otherwise, there's a lot of broken code in the repo.

I think we can only do this once we have started with whitelabeling the frontend and administration and have access to the project id.

PR #558, opened 2022-09-19
[Administration] Upgrade npm packages and use new GraphQL code generator

michael-markl:


administration/package.json
```diff
@@ -3,56 +3,56 @@
...
+    "generate-graphql": "graphql-codegen --config graphql-codegen.yml",
     "generate-protobuf": "run-script-os",
-    "generate-protobuf:linux": "mkdir -p src/generated && pbjs -t static-module -w es6 -o src/generated/compiled.js ../specs/card_activate_model.proto --force-long && pbts -o src/generated/compiled.d.ts src/generated/compiled.js",
-    "generate-protobuf:win32": "mkdir src\\generated | pbjs -t static-module -w es6 -o src\\generated\\compiled.js ..\\specs\\card_activate_model.proto --force-long | pbts -o src\\generated\\compiled.d.ts src\\generated\\compiled.js",
-    "postinstall": "npm run generate-protobuf"
+    "generate-protobuf:linux": "mkdir -p src/generated && pbjs -t static-module -w es6 -o src/generated/protobuf.js ../specs/card_activate_model.proto --force-long && pbts -o src/generated/protobuf.d.ts src/generated/protobuf.js",
+    "generate-protobuf:win32": "mkdir src\\generated | pbjs -t static-module -w es6 -o src\\generated\\protobuf.js ..\\specs\\card_activate_model.proto --force-long | pbts -o src\\generated\\protobuf.d.ts src\\generated\\protobuf.js",
```
maxammann, line 54: Should we remove that? Only difference is backslashes. WSL should handle that.

administration/src/components/generation/GenerationForm.tsx
```diff
@@ -84,6 +84,7 @@ const GenerationForm = (props: Props) => {
                     {cardCreationModels.length === 0 && "Legen Sie zunächst eine Karte an."}
                 </Tooltip>
             </ButtonBar>
+            { /* @ts-ignore */}
```
maxammann, line 87: Ohoh.

maxammann (approved):
Not yet tested.

maxammann:
Merge if you feel comfortable.

administration/package.json
michael-markl: I'd just keep it for now until it breaks something for us.

administration/src/components/generation/GenerationForm.tsx
michael-markl: Yeah, the npm package react-flip-move is not maintained anymore and has wrong ts types... We should probably switch to a different library (eg. [react-flip-toolkit](https://github.com/aholachek/react-flip-toolkit).)

michael-markl:
The only thing I am uncomfortable with is the annoying issue https://github.com/zxing-js/library/issues/499 which makes the zxing library shout a huge number of warnings (involving source maps) when building.

maxammann:
> The only thing I am uncomfortable with is the annoying issue [zxing-js/library#499](https://github.com/zxing-js/library/issues/499) which makes the zxing library shout a huge number of warnings (involving source maps) when building.

I left a comment in that issue to give it a bump. I would say we just merge with warnings and ignore them. Having warnings is better than not having source maps :D

michael-markl:
Seems good to me!

PR #565, opened 2022-09-26
516: Add Project Config to administration

michael-markl:


administration/src/graphql/regions/getRegions.graphql
```diff
@@ -1,5 +1,5 @@
-query getRegions {
-    regions {
+query getRegions($project: String!) {
```
maxammann, line 1: Can we already remove the getRegions endpoint from the backend? We can break the API as we want on the whitelabel branch.

maxammann (approved):

administration/src/graphql/regions/getRegions.graphql
michael-markl: Yep, we can remove all deprecated endpoints after this.

PR #571, opened 2022-10-02
Upgrade to React 18

michael-markl:
To complete the upgrade to react 18, we need to use createRoot from react-dom/client.
Moreover, I had to adjust the creation of the AppToaster, which uses the deprecated old API.

### DIFF
--- a/administration/src/components/AppToaster.tsx
+++ b/administration/src/components/AppToaster.tsx
@@ -1,0 +1,13 @@
+import {Position, Toaster} from "@blueprintjs/core";
+import {createRoot} from "react-dom/client";
+
+/** Singleton toaster instance. Create separate instances for different options. */
+let toasterRef: Toaster | null
+createRoot(document.getElementById("toaster")!).render(
+    <Toaster ref={ref => {
+        toasterRef = ref
+    }} position={Position.TOP}/>
+)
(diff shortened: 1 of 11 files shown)


administration/src/components/AppToaster.tsx
```diff
@@ -0,0 +1,13 @@
...
+/** Singleton toaster instance. Create separate instances for different options. */
+let toasterRef: Toaster | null
+createRoot(document.getElementById("toaster")!).render(
+    <Toaster ref={ref => {
+        toasterRef = ref
+    }} position={Position.TOP}/>
+)
+export const getAppToaster = () => {
+    return toasterRef!
+}
```
maxammann, line 13: Oke this is weird. so the toaster has an own custom virtual dom? weird.

maxammann (approved):

administration/src/components/AppToaster.tsx
michael-markl: I agree! I adjusted it to use the same DOM now. Maybe you can take another look.

maxammann (approved):
I think this is better, if no null ptr exceptions can happen 👍🏼

michael-markl:
Yeah, there shouldn't be any (TM).

PR #573, opened 2022-10-02
Remove deprecated APIs

### LINKED ISSUE #570: Remove all deprecated GraphQL APIs
There is a set of APIs that are deprecated (as they do not contain the projectId).
Currently, we can break the former API without breaking the user experience. Hence, it makes sense to just remove all deprecated APIs.

michael-markl:
Closes #570


maxammann (approved):
Looking good!

PR #574, opened 2022-10-03
208: Add role concept

### LINKED ISSUE #280: Login for administration backend
As a county official, I want to be sure that no unauthorized person can create digital EAKs in my name.

michael-markl:
This PR adds the basic functionalities of a role concept for the administration backend.

Each user in the backend now has an associated (but nullable) regionId and a role.
There are four roles:
* Project Admin:
  * Can create regions
  * Can (de)nominate Region Admins for all regions
* Region Admin: 
  * Can (de)nominate Region Manager for own region,
  * Can create cards for own region
  * Can manage applications for own region
* Region Manager: 
  * Can create cards for own region
  * Can manage applications for own region

Suggestion for review: Review the changes of `backend` before the changes of `administration`.

Closes #280.


administration/src/RegionProvider.tsx
```diff
@@ -1,12 +1,29 @@
...
-export const RegionContext = createContext<[Region | null, (region: Region | null) => void]>([null, () => {
-}])
+export const RegionContext = createContext<Region | null>(null)
 
 const RegionProvider = ({children}: { children: ReactNode }) => {
-    const [region, setRegionId] = useState<Region | null>(null)
-    return <RegionContext.Provider value={[region, setRegionId]}>
+    const {projectId} = useContext(ProjectConfigContext)
+    const [authContextData,] = useContext(AuthContext)
+    const userRegionId = authContextData?.administrator.regionId
```
maxammann, line 13: This could be refactored into a hook: `useRegion` o `useAuthContext`

administration/src/RegionProvider.tsx
```diff
@@ -1,12 +1,29 @@
...
+    })
+    if (loading) return <Spinner/>
+    if (error || !data) return <Button icon="repeat" onClick={() => refetch()}/>
+    const region = data.regions.find(region => region.id === userRegionId)
+    if (!region && typeof userRegionId === "number") {
+        return <>
+            <p>Your region was not found.</p>
+            <Button icon="repeat" onClick={() => refetch()}/>
+        </>
+    }
```
maxammann, line 25: I think a Provider should not be able to block the rendering of child components.

All of this could just be refactored into a hook: `useRegion` which calls `useQuery`. I would suspect `useQuery` to cache results to queries do not get re-executed.

administration/src/components/Navigation.tsx
```diff
@@ -1,24 +1,40 @@
...
+import {Role} from "../generated/graphql";
 
 interface Props {
     onSignOut: () => void
 }
 
 const Navigation = (props: Props) => {
+    const region = useContext(RegionContext)
+    const [authContextData,] = useContext(AuthContext)
+    const role = authContextData?.administrator.role
```
maxammann, line 16: Could be refactored using a hook.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -1,40 +1,54 @@
...
     fun deleteApplication(
-        context: GraphQLContext,
-        applicationId: Int
+        context: GraphQLContext, applicationId: Int
     ): Boolean {
-        context.enforceSignedIn()
+        val jwtPayload = context.enforceSignedIn()
+
+        // We throw an UnauthorizedException here, as we do not know whether there was an application with id
+        // `applicationId` and whether this application was contained in the user's project & region.
```
maxammann, line 43: Comment should be moved down.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
```diff
@@ -9,6 +10,8 @@ import org.jetbrains.exposed.dao.id.IntIdTable
 object Administrators : IntIdTable() {
     val email = varchar("email", 100).uniqueIndex()
     val projectId = reference("projectId", Projects)
+    val regionId = optReference("regionId", Regions)
+    val role = varchar("role", 50)
```
maxammann, line 14: 16 should be enough I think, 50 is fine too

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/repos/AdministratorsRepository.kt
```diff
@@ -32,14 +35,22 @@ object AdministratorsRepository {
...
-    fun insert(project: String, email: String, password: String) {
+    fun insert(project: String, email: String, password: String, role: Role, regionId: Int? = null) {
         val projectId = ProjectEntity.find { Projects.project eq project }.firstOrNull()?.id
             ?: throw IllegalArgumentException("Project does not exist.")
 
+        if (role in setOf(Role.REGION_ADMIN, Role.REGION_MANAGER) && regionId == null) {
+            throw IllegalArgumentException("Role ${role.db_value} needs to have a region assigned.")
+        } else if (role in setOf(Role.PROJECT_ADMIN) && regionId != null) {
+            throw java.lang.IllegalArgumentException("Role ${role.db_value} cannot have a region assigned.")
+        }
```
maxammann, line 46: I think we are missing some cases here, what if the user has the role NO_RIGHTS?
I think this would pass this ifelseif.

maxammann (approved):
I did not find where UnauthroizedExceptions are handled. Do they properly set the correct HTTP codes?

Looks good!

administration/src/RegionProvider.tsx
maxammann: We might want to use some pattern for storing "global mutable/immutable state". There is `useReducer` now in react, which when paired with `useContext` makes redux unneccassary: https://reactjs.org/docs/hooks-reference.html#usereducer

In fact, `useState` is just a simpler interface to `useReducer`

maxammann: I could show you next week how I used both in some other project. We can also refactor this later though.
I'm not sure what is the "right" way of handling application state nowadays. btw, redux sucks.

michael-markl: Hmm... blocking child components, because data they depend on hasn't finished loading yet, sounds reasonable to me. Not sure if other solutions would be a lot better. Is it just the name "Provider" you have problems with? Should we rename it to RegionsLoader or similar?
I don't think managing the state in a more complex way would help here :D

administration/src/components/Navigation.tsx
michael-markl: Not sure whether that increases or decreases readability here for those two lines

michael-markl: I've refactored the AuthContext a bit so that it's more readable.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
michael-markl: Yeah. let's do 32 :D With 16, I am too afraid we will come up with long role names :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/repos/AdministratorsRepository.kt
michael-markl: This is not an authorization check. I am just checking here whether the (role, region) combination makes sense. We could also enforce this using a DB constraint. what do you think?

administration/src/RegionProvider.tsx
maxammann: Alright lets just merge this. In the end the code quality of the frontend does not matter right now much.

Basically the alternative is the following: Instead of blocking the rendering of the whole app at the Provider level, provide a null value and delay the rendering of each component which depends on the role.

Depending on the usecase either one or the other is better.

The name `Provider` is just fine here, as it is a React Provider.

michael-markl: I'll just merge this for now. We can discuss some other approach later maybe. I think it is the easiest solution for now...

PR #590, opened 2022-10-17
Fix submodule fetching for backend builds

maxammann:


maxammann:
> Works on my machine

PR #592, opened 2022-10-24
Postponed: Add Web target

michael-markl:


frontend/web/manifest.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "frontend",
+    "short_name": "frontend",
```
maxammann, line 3: Probably needs whitelabeling aswell. We could also just build this only for the Ehrenamtskarte Bayern app.

In both cases -> New issue about whitelabeling and do that in a separate issue.

maxammann:
```suggestion
    "name": "Ehrenamtskarte Bayern",
    "short_name": "Ehrenamt",
```

maxammann (commented):
Looks good already! Thinks to do in this PR:

* Add a CI job so we know when things break
* Create a task for whitelabeling this
* Somehow integreate this into our build concept. E.g. we want to define a configuration for the web target for each build config. So that we can exclude the tabs for verification in the web version (of maybe not, if we want to allow verification :D).

ztefanie:
@michael-markl What is your plan with this PR? Should we close it or should we celebrate it's first birthday soon :D :partying_face:

maxammann:
I'm all for celebrating. Then merge on its birthday :P

f1sh1918:
Hm shouldn't be that big issue to finish the task. Not sure how good the `SecureStorage` works for web but then we could also include verification.
@dkehne do you think this could be useful soonish?
Maybe i can make it till birthday :)

michael-markl:
I think we should not ship verification in the web app.

Also, we should check again, if the Flutter web target is accessible at all.
If not, we might not be able (or we might consider not) to sell the web target.

maxammann:
Hm actually that is a valid point. There is no secure storage in web browsers. So maybe it is actually risky to expose such a thing.


Let's just close?

michael-markl:
I think the original idea was never to bring "Ausweisen" to the web target, but only the map and the search of accepting stores. That's still a valid use case for a web app.

f1sh1918:
I think the biggest web target issue is lacking seo Support.
Even I think it's nice to just check stores on web but may not justify the effort

maxammann:
Yeah I think we should invest the time to disable the Ausweisen tab. We probably don't want to use experimental stuff.

The map functionality would be fun though.

ztefanie:
Tomorrow we can selebrate the second birthday of this PR. I will now close this, feel free to reopen it if you have a different opinion on it. We can still reopen this anytime we want to invest time in this feature.

PR #596, opened 2022-10-25
286 Manage Users

michael-markl:


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ViewAdministratorsQueryService.kt
```diff
@@ -0,0 +1,73 @@
...
+    @GraphQLDescription("Returns all administrators in a project. This function requires the role PROJECT_ADMIN.")
+    fun getUsersInProject(
+        project: String,
+        dfe: DataFetchingEnvironment
+    ): List<Administrator> {
+        val context = dfe.getContext<GraphQLContext>()
+        val jwtPayload = context.enforceSignedIn()
+
+        return transaction {
+            val user = AdministratorEntity.findById(jwtPayload.userId)
```
maxammann, line 29:
```suggestion
            val admin = AdministratorEntity.findById(jwtPayload.userId)
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ViewAdministratorsQueryService.kt
```diff
@@ -0,0 +1,73 @@
...
+                throw UnauthorizedException()
+            }
+            val administrators = AdministratorEntity.find { Administrators.projectId eq projectId }
+
+            administrators.map {
+                Administrator(
+                    it.id.value,
+                    it.email,
+                    it.regionId?.value,
+                    Role.fromDbValue(it.role)!!
```
maxammann, line 41: Minor: Throw some different exception instead of NPE when the role is unknown.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -25,4 +26,13 @@ object Authorizer {
...
         )
     }
+
+    fun mayViewUsersInProject(user: AdministratorEntity?, projectId: Int): Boolean {
+        return user?.projectId?.value == projectId && user.role == Role.PROJECT_ADMIN.db_value
+    }
+
+    fun mayViewUsersInRegion(user: AdministratorEntity?, region: RegionEntity): Boolean {
+        return (user?.role == Role.REGION_ADMIN.db_value && user.regionId == region.id) ||
+            (user?.role == Role.PROJECT_ADMIN.db_value && user.projectId == region.projectId)
```
maxammann, line 36:
```suggestion
            mayViewUsersInProject(user, region.projectId)
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -25,4 +26,13 @@ object Authorizer {
             Role.REGION_ADMIN.db_value
         )
     }
+
+    fun mayViewUsersInProject(user: AdministratorEntity?, projectId: Int): Boolean {
+        return user?.projectId?.value == projectId && user.role == Role.PROJECT_ADMIN.db_value
```
maxammann, line 31: Why are we checking here `projectId.value` whereas below we are checking the `projectId` directly?

administration/src/components/users/UsersTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+    </>
+  )
+}
+
+const roleToText = (role: Role): String => {
+  switch (role) {
+    case Role.NoRights:
+      return 'Keine'
+    case Role.ProjectAdmin:
+      return 'Administrator'
```
maxammann, line 77: Maybe:
```suggestion
      return 'Administrator Berechtigungen'
```

maxammann (approved):
Looks good!

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
michael-markl: Below, we are handling with `EntityID<Int>` on both sides; here we are handling with an `Int` and an `EntityID<Int>`.

administration/src/components/users/UsersTable.tsx
michael-markl: Hmm does not feel better to me. (But I am also not too happy with my naming). Will leave it like this for now.
Will add an explanation for the different roles, though.

PR #637, opened 2022-11-20
Reset Application Form on Error or on Version-Change

michael-markl:
This resets the application form if
1) the version of the application form changes. The version is determined by the commit id that touched the administration/src/application folder the last time.
2) an error occurs while rendering (this shouldn't happen in production, but acts as a safety net. For dev, it might happen when fiddling around with the application form)

(Fun fact: I tried hard to make GitHub recognize, I just renamed the useLocallyStoredState file and appended some content, but it just didn't want to understand that in its diff... 🙀 )


administration/application_commit.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# This script determines the last commit of the folder administration/src/application.
+# This is used as a version for the application form. The form should reset whenever there are changes to the source
+# code of the application form.
+
+basedir=$(dirname "$0")
```
maxammann, line 7: Add here some script to check if the dependencices are installed.

```
if ! command -v git &> /dev/null
then
    echo "Git is not installed."
    exit
fi

```

administration/package.json
```diff
@@ -54,8 +54,8 @@
     "typescript": "^4.8.3"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
+    "start": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh); react-scripts start",
```
maxammann, line 57: Move `set -ae` to the script.

administration/src/application/ApplicationErrorBoundary.tsx
```diff
@@ -0,0 +1,35 @@
...
+      await localforage.removeItem(globalArrayBuffersKey)
+    } catch (e) {
+      console.error('Another error occurred while resetting the application state.', e)
+    } finally {
+      this.setState({ resetting: false })
+    }
+  }
+
+  render() {
+    if (this.state.resetting) return null
```
maxammann, line 22: Add braces. Shouldn't such things be cought by our linter?

maxammann (approved):
Looks good!

administration/src/application/components/ApplyController.tsx
```diff
@@ -5,30 +5,35 @@ import '@fontsource/roboto/700.css'
...
+import useVersionedLocallyStoredState from '../useVersionedLocallyStoredState'
 import DiscardAllInputsButton from './DiscardAllInputsButton'
 import { useGarbageCollectArrayBuffers, useInitializeGlobalArrayBuffersManager } from '../globalArrayBuffersManager'
 import ApplicationForm from './forms/ApplicationForm'
 import { useCallback, useMemo } from 'react'
 import { SnackbarProvider, useSnackbar } from 'notistack'
+import ApplicationErrorBoundary from '../ApplicationErrorBoundary'
 
-const applicationStorageKey = 'applicationState'
+const lastCommitForApplicationForm = process.env.REACT_APP_APPLICATION_COMMIT as string
```
maxammann, line 16: Maybe add a comment here where the variable is coming from, and that a preprocessor is string-replacing the process.env. stuff.

administration/package.json
michael-markl: The `set -ae` is necessary here, to make the script exit if it fails. Putting it into the script has no effect.

michael-markl: We could also do something like:

```suggestion
    "start": "c=$(./application_commit.sh) && REACT_APP_APPLICATION_COMMIT=$c react-scripts start",
```

That seems to work for me as well. Is that better?

administration/src/application/ApplicationErrorBoundary.tsx
michael-markl: probably our linter allows that currently.

administration/package.json
michael-markl: @maxammann lmk what you think. You can also merge this otherwise.

maxammann: Hm I wonder why that works but the initial version without the `set`

Just tested on macOS with zsh, the set version works so lets merge it.

PR #648, opened 2022-11-30
625: Application end screen

f1sh1918:
- [x] implement application end screen dialog
- [x] clear form data  

If there is a proper text or text template to use, just let me know

Added issue for deep linking that has to be discussed https://github.com/digitalfabrik/entitlementcard/issues/661


administration/src/application/components/ApplyController.tsx
```diff
@@ -10,14 +10,16 @@ import useLocallyStoredState from '../useLocallyStoredState'
 import DiscardAllInputsButton from './DiscardAllInputsButton'
 import { useGarbageCollectArrayBuffers, useInitializeGlobalArrayBuffersManager } from '../globalArrayBuffersManager'
 import ApplicationForm from './forms/ApplicationForm'
-import { useMemo } from 'react'
+import { useMemo, useState } from 'react'
 import { SnackbarProvider, useSnackbar } from 'notistack'
+import BasicDialog from '../../components/BasicDialog'
```
michael-markl, line 15: Please move this file somewhere under src/application. We don't want to use "share" components between the application form and the digitale druckerei

administration/src/components/BasicDialog.tsx
```diff
@@ -0,0 +1,45 @@
+import { Icon, IconSize } from '@blueprintjs/core'
```
michael-markl, line 1: Only use @mui for the application form

michael-markl (changes requested):

maxammann (approved):
Looks good!

maxammann:
(Not tested yet)

administration/src/components/BasicDialog.tsx
f1sh1918: gonna be reverted since we use end screen instead of dialog

michael-markl (approved):

administration/src/application/components/ApplyController.tsx
```diff
@@ -57,14 +66,25 @@ const ApplyController = () => {
         application: applicationInput,
       },
     })
+
+    setFormSubmitted(true)
+    setState(() => ApplicationForm.initialState)
```
michael-markl, line 71: This should only be called, if `addBlueEakApplication` is successful. we need to await the promise, check if the status code is 200 and only then do setFormSubmitted(true), ....
Also: Please set the loading prop of ApplicationForm.Component according to the loading state of the request.

michael-markl (changes requested):

administration/src/application/components/ApplyController.tsx
michael-markl: @f1sh1918 Sorry,  I think I didn't explain it well: Please await the promise returned by `addBlueEakApplication()`, and add some try/catch block. In the success case, `setFormSubmitted(true)` and `setState(()=> ApplicationForm.initialState)` should be called. In the error case, a toast should be displayed.

f1sh1918: Ah okay that makes sense. Maybe I also didn't Listen carefully. Gonna adjust it

administration/src/application/components/ApplyController.tsx
```diff
@@ -57,14 +66,25 @@ const ApplyController = () => {
         application: applicationInput,
       },
     })
+
+    setState(() => ApplicationForm.initialState)
+    setFormSubmitted(true)
   }
+  const successText = `Ihr Antrag für die Ehrenamtskarte wurde erfolgreich übermittelt.
+            Sie können das Fenster schließen.`
```
maxammann, line 74: Maybe add a note, that users can return now to the app.

maxammann (approved):

administration/src/application/components/ApplyController.tsx
michael-markl: I probably wouldnt do that, as the user might not come from the app (but from some other website).

administration/src/application/components/ApplyController.tsx
michael-markl: Yeah I think so!

michael-markl (approved):
Looks good to me !

administration/src/application/components/ApplyController.tsx
```diff
@@ -9,18 +9,42 @@ import useVersionedLocallyStoredState from '../useVersionedLocallyStoredState'
...
+    onError: error => {
+      console.error(error)
+      appToaster?.show({ intent: 'danger', message: 'Beim Absenden des Antrags is ein Fehler aufgetreten.' })
+    },
+    onCompleted: result => {
+      if (result) {
+        setState(() => ApplicationForm.initialState)
+        setFormSubmitted(true)
+      } else {
+        appToaster?.show({ intent: 'danger', message: 'Beim Absenden des Antrags is ein Fehler aufgetreten.' })
```
michael-markl, line 44: Sorry I have another nit:
Could you please use the `enqueueSnackbar` that we get from `notistack`'s `useSnackbar` (which is already called in line 58) instead of `appToaster?.show`.
The reason is that `useAppToaster` uses the blueprint framework, whereas here for the application we use the @mui's alternative (and shouldn't use any blueprint stuff)

f1sh1918: adjusted

PR #649, opened 2022-12-01
608: link to application

sarahsporck:
I decided to link to the browser page. 

➕ the user can use stored form data from the browser which might ease filling the form on a mobile device.
➕ we don't need to implement a view for no internet connection

➖ the url is visibly exposed (though i am not sure if that is a problem), but might increase the chance of spam sent through the application interfaces.
➖ the view is not integrated into the app ➡️ navigating back and force is more than one tap (but also not really necessary)


maxammann:
> ➖ the url is visibly exposed (though i am not sure if that is a problem), but might increase the chance of spam sent through the application interfaces.

I dont see a problem, because cities will link to the form from their website.

➖ the view is not integrated into the app ➡️ navigating back and force is more than one tap (but also not really necessary)

Acceptable imo :)

frontend/lib/intro_slides/intro_screen.dart
```diff
@@ -32,6 +32,19 @@ class IntroScreenState extends State<IntroScreen> {
         styleDescription: theme.textTheme.bodyText1?.apply(fontSizeFactor: 1.2),
       ),
     );
+    slides.add(
+      Slide(
+        title: "Wie kann ich eine Ehrenamtskarte beantragen?",
+        description: "In dem Formular geben Sie Informationen über Sie und Ihre "
```
maxammann, line 38: We can fine-tune those texts at the very end, just before publication of the updated app.

maxammann (approved):

PR #669, opened 2022-12-13
615: add creator to card information

sarahsporck:


backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
```diff
@@ -19,6 +20,7 @@ object Cards : IntIdTable() {
     val issueDate = timestamp("issueDate").defaultExpression(CurrentTimestamp())
     val revoked = bool("revoked")
     val regionId = reference("regionId", Regions)
+    val creatorId = reference("creatorId", Administrators)
```
maxammann, line 23: Alternative:
```suggestion
    val issuerId = reference("issuerId", Administrators)
```

We could also change regionId to `issuingRegionId`.

maxammann (approved):
Looks good! The naming suggestion is optional.

michael-markl (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
sarahsporck: i agree to changing creator to issuer as this resembles jwt standards, but for the region i'd keep the current naming, as it is already pretty clear what is meant.

PR #679, opened 2022-12-19
655: replace qr_code_scanner with mobile_scanner

sarahsporck:


sarahsporck:
@maxammann as you wished :)

frontend/lib/qr_code_scanner/qr_overlay_shape.dart
```diff
@@ -0,0 +1,178 @@
+import 'dart:math';
+
+import 'package:flutter/material.dart';
+
+// temporary solution
+// checkout https://github.com/juliansteenbakker/mobile_scanner/issues/250#issuecomment-1209723708
```
maxammann, line 6: This issue mentions that a solution is being worked on which is already on their main: https://github.com/juliansteenbakker/mobile_scanner/blob/master/example/lib/barcode_scanner_window.dart

maxammann: We could temporarlily depend on their Github version instead of the published release if that simplifies this file.

maxammann (commented):
Testing now on iOS and Android

maxammann:
And I suspect this does not compile on iOS right now, because it is stuck in the check job

maxammann:
Tested on iOS (release and debug)

maxammann:
I also fixed linting

frontend/lib/qr_code_scanner/qr_overlay_shape.dart
michael-markl: If I understand correctly, this is just a visual overlay to allow scanning qr codes inside of the overlay? We could just ignore that for now, until the feature has officially landed in mobile_scanner. (So we could just remove this overlay shape and the `_calculateScanArea` function in `qr_code_scanner_controls.dart`)

However, if it works without problems like this, we can also leave it like this I think :) (And maybe refactor once the feature has landed)

michael-markl (approved):
(Only code review)

sarahsporck:
> If I understand correctly, this is just a visual overlay to allow scanning qr codes inside of the overlay? We could just ignore that for now, until the feature has officially landed in mobile_scanner. (So we could just remove this overlay shape and the _calculateScanArea function in qr_code_scanner_controls.dart)
> 
> However, if it works without problems like this, we can also leave it like this I think :) (And maybe refactor once the feature has landed)

for me it works without problems. And it seems it will take a little bit until integration, therefore I'd keep the solution as it is.

Tested on Android again. Works fine.
@maxammann do you want to test it for ios once more please?

maxammann:
> Tested on Android again. Works fine.
> @maxammann do you want to test it for ios once more please?

No need for that if it works on Android without that crash merge :)

PR #683, opened 2023-01-04
Reduce code verbosity of forms

michael-markl:
This PR reduces code verbosity for forms by introducing utility functions for compound forms.

(
This was a bit of a fun side project to get to know typescript better^^. My typescript wishlist:
https://github.com/microsoft/TypeScript/issues/12936
and 
https://github.com/microsoft/TypeScript/issues/14466
)


maxammann (approved):
Looks good, I did a quick review of the code and a smoke test.

Maybe some minor thing we can do in this PR: I would propose to make the "Ich beantrage neben der physischen auch die digitale Ehrenamtskarte." opt-out so it is by default selected.

I think most people would like one but maybe dont set it because they dont know what it is.

PR #695, opened 2023-01-10
Prepare Administration for Nürnberg

michael-markl:
This PR:
* Adds a single region for the nürnberg pass project (region "Stadt Nürnberg")
* Hides application based features based on a project-config setting.
* Slightly adjust the navigation header: Remove the explicit "Home" button, make it more compact


administration/src/Router.tsx
```diff
@@ -47,8 +45,13 @@ const Router = () => {
                   <Navigation onSignOut={signOut} />
                   <Main>
                     <Routes>
-                      <Route path={'/applications'} element={<ApplicationsController token={authData.token} />} />
-                      <Route path={'/region'} element={<RegionsController />} />
+                      {projectConfig.applicationFeatureEnabled ? (
```
maxammann, line 48: Should we also add a check for /apply-for-eak?

maxammann (approved):
Looks good!

PR #700, opened 2023-01-10
Do not store user info in administration

### LINKED ISSUE #619: Add a logout button to the RegionProvider/Add an error component to reset complete state

There is an error which can happen when you are logged into a user of another project. The following case is hit.

https://github.com/digitalfabrik/entitlementcard/blob/816c30a7ad426c80db3fcf81f31118bc863f63ee/administration/src/RegionProvider.tsx#L19-L24

We should add a button to reset the apps state as else its not possible to recover.

Alternatively we could reduce the state which is included in the localstorage next to the token. Right now the region id is stored there. Also the internal user id. Both can change over time (e.g. when changing the database).

**Additional context**

michael-markl:
As discussed, we do not want to store anything else than the adminId in the token.
Moreover, we do not want to store the user info (like email, role, etc.) in the local storage.
Hence, I untied the "getting user info" and "login" actions: 
I removed the user info from the result of the SignIn mutation and added a new `whoAmI` query.   

(Also closes #619.)


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/repos/AdministratorsRepository.kt
```diff
@@ -34,7 +34,7 @@ object AdministratorsRepository {
         return resultRow?.let {
             val user = AdministratorEntity.wrapRow(it)
             val passwordHash = user.passwordHash
-            if (passwordHash !== null && PasswordCrypto.verifyPassword(password, passwordHash)) {
+            if (passwordHash != null && PasswordCrypto.verifyPassword(password, passwordHash)) {
```
maxammann, line 37: Why those changes?

maxammann (approved):
Awesome! thanks for changing this that quickly!

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/repos/AdministratorsRepository.kt
michael-markl: In Kotlin, structural equality (what we usually want) is the `==` operator (whereas `===` is the ref equality). (https://kotlinlang.org/docs/equality.html)
That's quite annoying because for JS we usually want to use the `===` operator. (I guess JS is to blame here)

PR #704, opened 2023-01-11
Setup SMTP

### LINKED ISSUE #583: Setup SMTP in backend
To send password reset emails, we need to have some SMTP client setup in the backend.

We need separate SMTP configs for each project (to have project-specific email addresses).
We ordered an SMTP server from hetzner for Nürnberg and Bayern. Logins are in our PW manager.

- [ ] Add a config entry for one SMTP server per project.
- [ ] Setup keine-antwort@sozialpass.app in backend
- [ ] Setup keine-antwort@ehrenamtskarte.app in backend
- [ ] Send @maxammann the config you used for testing the email

SimpleJavaMail is a promising first try: https://www.simplejavamail.org/configuration.html#section-transport-strategy-tls

If authentication does not work on the second try using SMTP + TLS then contact @maxammann (we will create a SMTP relay server in that case).

michael-markl:
Closes #583

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
@@ -32,8 +32,11 @@ class Entry : CliktCommand() {
     private val postgresPassword by option()
     private val geocoding by option().choice("true", "false").convert { it.toBoolean() }
     private val geocodingHost by option()
+    private val smtpUsernameEakBayern by option()
     private val smtpPasswordEakBayern by option()
+    private val smtpUsernameNuernbergPass by option()
     private val smtpPasswordNuernbergPass by option()
+    private val smtpUsernameShowcase by option()
     private val smtpPasswordShowcase by option()
 
     override fun run() {
@@ -43,6 +46,12 @@ class Entry : CliktCommand() {
             projects = backendConfiguration.projects.map { project ->
                 project.copy(
                     smtp = project.smtp.copy(
+                        username = when (project.id) {
+                            EAK_BAYERN_PROJECT -> smtpUsernameEakBayern
+                            NUERNBERG_PASS_PROJECT -> smtpUsernameNuernbergPass
+                            SHOWCASE_PROJECT -> smtpUsernameShowcase
+                            else -> null
+                        } ?: project.smtp.username,
                         password = when (project.id) {
                             EAK_BAYERN_PROJECT -> smtpPasswordEakBayern
                             NUERNBERG_PASS_PROJECT -> smtpPasswordNuernbergPass
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
@@ -1,10 +1,7 @@
 package app.ehrenamtskarte.backend
 
 import app.ehrenamtskarte.backend.common.database.Database
-import app.ehrenamtskarte.backend.common.webservice.EAK_BAYERN_PROJECT
 import app.ehrenamtskarte.backend.common.webservice.GraphQLHandler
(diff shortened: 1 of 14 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
```diff
@@ -29,11 +32,35 @@ class Entry : CliktCommand() {
...
+                        } ?: project.smtp.username,
+                        password = when (project.id) {
+                            EAK_BAYERN_PROJECT -> smtpPasswordEakBayern
+                            NUERNBERG_PASS_PROJECT -> smtpPasswordNuernbergPass
+                            SHOWCASE_PROJECT -> smtpPasswordShowcase
+                            else -> null
+                        } ?: project.smtp.password
+                    )
+                )
+            },
```
maxammann, line 63: While the CLI config is definitely nice, we won't need it in production use.
I'm not sure if we really need this feature.

backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
```diff
@@ -49,21 +51,26 @@ data class BackendConfiguration(
             .registerModule(
                 KotlinModule.Builder().build()
             ).registerModule(JavaTimeModule())
+        private val logger = LoggerFactory.getLogger(BackendConfiguration::class.java)
 
         fun load(configFile: File?): BackendConfiguration {
-            val fallbackResource = ClassLoader.getSystemResource("config/config.yml")
-                ?: throw Error("Fallback backend configuration resource 'config/config.yml' missing!'")
+            val fallbacks = listOfNotNull(
+                ClassLoader.getSystemResource("config/config.local.yml"),
```
maxammann, line 58: 👍🏼

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/SendMail.kt
```diff
@@ -1,5 +1,19 @@
 package app.ehrenamtskarte.backend.mail
 
-fun sendMail(to: String, subject: String, message: String) {
-    println("SENDING PSEUDO MAIL TO $to\nSUBJECT: $subject\n$message")
+import app.ehrenamtskarte.backend.config.SmtpConfig
+import org.simplejavamail.email.EmailBuilder
+import org.simplejavamail.mailer.MailerBuilder
+
+fun sendMail(smtpConfig: SmtpConfig, fromName: String, to: String, subject: String, message: String) {
+    MailerBuilder.withSMTPServer(smtpConfig.host, smtpConfig.port, smtpConfig.username, smtpConfig.password)
```
maxammann, line 8: ~~I think we need here: `withTransportStrategy "SMTP_TLS"`~~

~~Else it uses port 25 which is unencrypted~~

maxammann: Oke probably not if we specify port 587

maxammann (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
michael-markl: I can remove it again, if we don't need it.

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/SendMail.kt
michael-markl: It worked, so I guess, the config is correct :D

michael-markl (commented):
> Did you already test sending emails through the 2 referenced accounts? 

I tested sending emails through sozialpass only (which worked). I couldn't find the ehrenamtskarte credentials in passbolt.

> Can you send me the config view PM (without pws)?

Not sure what you mean by that. You can just copy the config.yml to config.local.yml and replace the password & username of each smtp config

Issue #504, opened 2022-03-01
Frontend Whitelabeling


maxammann:
As a user I want to be able to use apps which are specific to a project, e.g. Bayern, Nürnberg,.. There should be one app per project.

## Dimensions

There are multiple dimensions for creating builds:
* Dimension 1: Release, Debug, Profile build
* Dimension 2: Production, Staging, Local Server
* Dimension 3: Nürnberg, Bavaria, BaWü etc.
* Dimension 4: FLOSS or GAPPS

Any combination of the 4 dimension makes sense:
* Release, Production, Nürnberg, FLOSS
* Debug, Staging, Nürnberg, GAPPS

Dimension 1, 2 and 4 are solved through command line flags. Dimension 3 is solved through a Build Config.

## Acceptance Criteria:
 - [ ] There are multiple build configs for Nürnberg and Bayern
 - Each app has a different:
   - [ ] Name
   - [ ] Logo
   - [ ] Color Theme
   - [ ] Impressum, Datenschutzerklärung
   - [x] Project ID 

## Technical Tasks:
 - [x] #517
 - [x] #513
 - [x] #514
 - [x] #515
 - [x] #512
 - [x] #521 
 - [x] #522 
 - [x] #523
 - [x] #524
 - [x] #542
 - [x] #564

maxammann:
All tasks are done, closing

PR #716, opened 2023-01-13
501: Some more frontend whitelabeling

michael-markl:
This PR includes:
* Change icon in about tab (and tidy up some other icon stuff)
* Extract disclaimerText and  publisherText into the build config (Nürnberg's versions are still just a draft, we're waiting for Nürnberg here)
* Adjust the initial map position in the case of missing user location depending on the build config (ie. an overview of bavaria vs nürnberg)

Related to #504


frontend/assets/icon/README.md
```diff
@@ -0,0 +1 @@
+The icons in this folder are currently not used.
```
maxammann, line 1: If they are not used, should we maybe move them out of the assets directory?

maxammann (approved):

PR #726, opened 2023-01-16
Redesign Ausweisen Without Activated Card

michael-markl:


frontend/lib/identification/no_card_view.dart
```diff
@@ -14,96 +14,117 @@ class NoCardView extends StatelessWidget {
...
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+}
 
-    Widget wrapIntrinsic(Widget widget) => isLandscape ? IntrinsicHeight(child: widget) : IntrinsicWidth(child: widget);
+class _TapableCardWithArea extends StatelessWidget {
```
maxammann, line 56: We might want to move this to a separate file.

maxammann (approved):

frontend/lib/identification/no_card_view.dart
michael-markl: I would only do it, once we have usages other than NoCardView

Issue #690, opened 2023-01-09
Limit length of dynamic data in QR code


maxammann:
**Is your feature request related to a problem? Please describe.**
We need to limit the length of dynamic data in the QR code. Right now dynamic data is:

* Name (On average 15chars (US), for Germany we estimate 30 on average)
* ~VAGNumber (probably around 10chars)~ (VAGNumber consists of 8 digits and can be encoded as uint32)

**Describe the solution you'd like**
All other data is of static length. Calculate the worst case data usage for QR code data and then compare it to the current maximum. We use a version 7 QR code with error correction L right now: https://www.qrcode.com/en/about/version.html That means we can store 224 base64 chars right now.

If we do not meet the requirements in terms of space we should truncate the name or abort with an error message: "Name too long, please abbreviate"

**Describe alternatives you've considered**
None.

**Additional context**
None

sarahsporck:
**Please confirm.**

---

A protobuf message is encoded as follows:
- (in our case) 1 byte for the tag number + x bytes for the data
- if the data has a variable length (submessages, strings or bytes) then there are additionally y bytes for the length
- if data is left out (as in one of) -> no bytes are used for missing data

so i'll write `<bytes_tag> + <bytes_len> + <bytes_data>` if there is a length
messages look like this `<bytes_tag> + <bytes_len>`
otherwise (int32 and uint32) i'll write `<bytes_tag> + <bytes_data>`

**Reference**: 
- https://developers.google.com/protocol-buffers/docs/encoding 
- https://stackoverflow.com/questions/30915704/maximum-serialized-protobuf-message-size

## Calculation

I considered the nuernberg data as it can get larger than tha bavarian data. Also the activation code takes more bytes.
So the protobuf encoded qrcode data consists of:
- qr_code = 1 + 1
    - dynamic_activation_code = 1 + 1
        - pepper = 1 + 1 + 16
        - totpsecret = 1 + 1 + 20
        - ec_signature = 1 + 1 + 32 (?)
        - card_info = 1 + 1
            - name = 1 + 1 + x
            - expiration_date = 1 + 5 
            - extensions = 1 + 1
                - birthday_ext = 1 + 1
                    - birthday = 1 + 10
                - nuernberg_pass_number = 1 + 1
                    - pass_number = 1 + 5

## Results

[...]

PR #737, opened 2023-01-20
690: Validate size of card blueprints

sarahsporck:
We decided to go for 50 bytes for the name.

### DIFF
--- a/administration/src/cards/CardBlueprint.ts
+++ b/administration/src/cards/CardBlueprint.ts
@@ -1,8 +1,8 @@
-import { CardExtensions, CardInfo, DynamicActivationCode } from '../generated/card_pb'
+import { CardExtensions, CardInfo, DynamicActivationCode, QrCode } from '../generated/card_pb'
 import { dateToDaysSinceEpoch } from './validityPeriod'
 import { ExtensionHolder } from './extensions'
 
-const MAX_NAME_LENGTH = 60 // TODO: Select proper max value
+const MAX_NAME_LENGTH = 50
 const TOTP_SECRET_LENGTH = 20
 const PEPPER_LENGTH = 16
 
@@ -28,7 +28,8 @@ export class CardBlueprint {
   }
 
   isFullNameValid(): boolean {
-    return this.fullName.length > 0 && this.fullName.length < MAX_NAME_LENGTH
+    const encodedName = new TextEncoder().encode(this.fullName)
+    return this.fullName.length > 0 && encodedName.length < MAX_NAME_LENGTH
   }
 
   isExpirationDateValid(): boolean {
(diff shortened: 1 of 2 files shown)


administration/src/components/cards/CreateCardForm.tsx
```diff
@@ -42,7 +42,7 @@ const CreateCardForm = (props: Props) => {
           <InputGroup
             placeholder='Erika Mustermann'
             autoFocus
-            intent={cardBlueprint.isFullNameValid() ? undefined : Intent.DANGER}
+            intent={cardBlueprint.isFullNameValid() && cardBlueprint.hasValidSize() ? undefined : Intent.DANGER}
```
sarahsporck, line 45: Imo it's enough to highlight where the creator of the card has to make adjustments. Especially as this seems to be an infrequent (but still possible) situation.

administration/src/cards/CardBlueprint.ts
```diff
@@ -42,10 +43,25 @@ export class CardBlueprint {
...
   }
 
+  hasValidSize(): boolean {
+    // every four characters in a base64 encoded string correspond to three bytes
+    // the qr code can hold 224 characters
+    // therefore 224 / 4 * 3 = 168
+    return this.bytesOfProtobuf() <= 168
+  }
+
+  bytesOfProtobuf(): number {
```
michael-markl, line 59:
```suggestion
  sizeOfProtobuf(): number {
```

administration/src/cards/CardBlueprint.ts
```diff
@@ -42,10 +43,25 @@ export class CardBlueprint {
...
+      (this.isExpirationDateValid() || this.hasInfiniteLifetime()) &&
+      // Number of bytes is valid
+      this.hasValidSize()
     )
   }
 
+  hasValidSize(): boolean {
+    // every four characters in a base64 encoded string correspond to three bytes
+    // the qr code can hold 224 characters
+    // therefore 224 / 4 * 3 = 168
```
michael-markl, line 55:
```suggestion
    // every four characters in a base64 encoded string correspond to three bytes
    // the qr code can hold 224 characters
    // therefore 224 / 4 * 3 = 168
    // See https://github.com/digitalfabrik/entitlementcard/issues/690  for more context.
``` 

It would be nice, if you could additionally link the issue https://github.com/digitalfabrik/entitlementcard/issues/690 in this comment to give even more context :)

michael-markl (approved):

administration/src/components/cards/CreateCardForm.tsx
michael-markl: Agreed. Maybe you can add a comment above, something like:
"If the size of the card is too large, show a warning at the name field as it is the only dynamically sized field".

PR #739, opened 2023-01-22
Use sint32 for birthday

michael-markl:
Related to the discussion in #690

### DIFF
--- a/administration/src/cards/extensions.tsx
+++ b/administration/src/cards/extensions.tsx
@@ -46,6 +46,7 @@ export const region_extension: Extension<RegionState> = {
 export interface BirthdayState {
   birthday: number
 }
+
 const initialBirthdayDate = dateToDaysSinceEpoch(new Date('1980-01-01T00:00+00:00'))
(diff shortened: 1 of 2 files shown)


administration/src/cards/extensions.tsx
```diff
@@ -60,16 +61,12 @@ export const birthday_extension: Extension<BirthdayState> = {
...
-            }
-            setState({
-              birthday: dateToDaysSinceEpoch(value),
-            })
+          value={daysSinceEpochToDate(state?.birthday ?? initialBirthdayDate)}
+          parseDate={value => {
+            const millis = Date.parse(value)
+            return isNaN(millis) ? false : new Date(millis)
           }}
+          onChange={value => setState(value !== null ? { birthday: dateToDaysSinceEpoch(value) } : null)}
```
michael-markl, line 69: These changes also fix, that the user was not able to set 1/1/1970 as a birthday.

maxammann (approved):

PR #743, opened 2023-01-23
675: Support static qr codes in backend and administration

sarahsporck:
I did not create a separate mutation/query for the static cards, so far as the only difference is the presence of the totp secret. 

Would it make sense to verify the project in the backend, so that you cannot create a static qrcode for bayern? I personally think that static qrcodes should be treated as a feature, that can be turned on for every project. But am not sure how to verify this correctly in the backend.

### DIFF
--- a/administration/src/cards/PdfFactory.ts
+++ b/administration/src/cards/PdfFactory.ts
@@ -5,7 +5,7 @@ import uint8ArrayToBase64 from '../util/uint8ArrayToBase64'
 import { format } from 'date-fns'
 import { Exception } from '../exception'
 import { Region } from '../generated/graphql'
-import { DynamicActivationCode, QrCode } from '../generated/card_pb'
+import { DynamicActivationCode, QrCode, StaticVerifyCode } from '../generated/card_pb'
 import { daysSinceEpochToDate } from './validityPeriod'
 
 type TTFFont = {
@@ -31,7 +31,12 @@ export async function loadTTFFont(name: string, fontStyle: string, path: string)
   }
 }
 
-function addLetter(doc: jsPDF, activationCode: DynamicActivationCode, region: Region) {
+function addLetter(
+  doc: jsPDF,
+  region: Region,
+  activationCode: DynamicActivationCode,
+  staticVerifyCode: StaticVerifyCode | null
+) {
   const info = activationCode.info!
 
   const pageSize = doc.internal.pageSize
@@ -108,6 +113,19 @@ Aussteller: ${region.prefix} ${region.name}`,
       align: 'center',
     }
   )
+
+  if (staticVerifyCode) {
+    doc.addPage()
+    const qrCodeText2 = uint8ArrayToBase64(
+      new QrCode({
+        qrCode: {
+          value: staticVerifyCode,
+          case: 'staticVerifyCode',
+        },
+      }).toBinary()
+    )
+    drawjsPDF(qrCodeText2, qrCodeX, qrCodeY, qrCodeSize, doc)
+  }
 }
 
 function checkForeignText(doc: jsPDF, text: string): string | null {
@@ -122,7 +140,12 @@ function checkForeignText(doc: jsPDF, text: string): string | null {
   return null
 }
 
-export function generatePdf(font: TTFFont, activationCodes: DynamicActivationCode[], region: Region) {
+export function generatePdf(
+  font: TTFFont,
+  region: Region,
+  activationCodes: DynamicActivationCode[],
+  staticCodes: StaticVerifyCode[] | null
+) {
   const doc = new jsPDF({
     orientation: 'portrait',
     unit: 'mm',
@@ -145,7 +168,7 @@ export function generatePdf(font: TTFFont, activationCodes: DynamicActivationCod
       })
     }
 
-    addLetter(doc, activationCodes[k], region)
+    addLetter(doc, region, activationCodes[k], staticCodes![k])
     if (k !== activationCodes.length - 1) doc.addPage()
   }
 
--- a/administration/src/cards/activation.ts
+++ b/administration/src/cards/activation.ts
@@ -4,16 +4,18 @@ import {
   AddCardMutation,
   AddCardMutationVariables,
   CardGenerationModelInput,
+  CodeType,
   Region,
 } from '../generated/graphql'
 import hashCardInfo from './hash'
 import uint8ArrayToBase64 from '../util/uint8ArrayToBase64'
 import { ApolloClient } from '@apollo/client'
 
-export async function activateCard(
+export async function activateCard<T extends DynamicActivationCode | StaticVerifyCode>(
   client: ApolloClient<object>,
-  activationCode: DynamicActivationCode | StaticVerifyCode,
-  region: Region
+  activationCode: T,
+  region: Region,
+  codeType: T extends DynamicActivationCode ? CodeType.Dynamic : CodeType.Static
 ) {
   const cardInfoHash = await hashCardInfo(activationCode.pepper, activationCode.info!)
   const expirationDay = activationCode.info!.expirationDay
@@ -24,6 +26,7 @@ export async function activateCard(
     cardInfoHashBase64: uint8ArrayToBase64(cardInfoHash),
     totpSecretBase64: totpSecret,
     regionId: region.id,
+    codeType,
   }
 
   return await client.mutate<AddCardMutation, AddCardMutationVariables>({
@@ -32,13 +35,14 @@ export async function activateCard(
   })
 }
 
-export async function activateCards(
+export async function activateCards<T extends DynamicActivationCode | StaticVerifyCode>(
   client: ApolloClient<object>,
-  activationCodes: DynamicActivationCode[] | StaticVerifyCode[],
-  region: Region
+  activationCodes: T[],
+  region: Region,
+  codeType: T extends DynamicActivationCode ? CodeType.Dynamic : CodeType.Static
 ) {
   const results = await Promise.all(
-    activationCodes.map(async activationCode => activateCard(client, activationCode, region))
+    activationCodes.map(async activationCode => activateCard(client, activationCode, region, codeType))
   )
 
   const firstFailure = results.find(result => !result.data?.success)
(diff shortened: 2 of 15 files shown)


administration/src/components/cards/CreateCardsController.tsx
```diff
@@ -35,14 +37,23 @@ const CreateCardsController = () => {
...
       const activationCodes = cardBlueprints.map(cardBlueprint => {
         return cardBlueprint.generateActivationCode()
-      })
+      })      
+      const staticCodes = projectConfig.staticQrCodesEnabled ? cardBlueprints.map(cardBlueprints => {
+        return cardBlueprints.generateStaticVerifyCode()
+      }) : null
+            
+      const font = await loadTTFFont('NotoSans', 'normal', '/pdf-fonts/NotoSans-Regular.ttf')
+      const pdfDataUri = generatePdf(font, region, activationCodes, staticCodes)
```
sarahsporck, line 49: This rearrangement should fix that cards are created, even though PDF generation fails.

administration/src/cards/PdfFactory.ts
```diff
@@ -5,7 +5,7 @@ import uint8ArrayToBase64 from '../util/uint8ArrayToBase64'
 import { format } from 'date-fns'
 import { Exception } from '../exception'
 import { Region } from '../generated/graphql'
-import { DynamicActivationCode, QrCode } from '../generated/card_pb'
+import { DynamicActivationCode, QrCode, StaticVerifyCode } from '../generated/card_pb'
 import { daysSinceEpochToDate } from './validityPeriod'
```
maxammann, line 10: I'll refactor this file a lot in my task, so we can just merge this as is.

administration/src/cards/activation.ts
```diff
@@ -12,15 +12,17 @@ import { ApolloClient } from '@apollo/client'
...
 ) {
   const cardInfoHash = await hashCardInfo(activationCode.pepper, activationCode.info!)
   const expirationDay = activationCode.info!.expirationDay
+  const totpSecret =
+    activationCode instanceof DynamicActivationCode ? uint8ArrayToBase64(activationCode.totpSecret) : null
   const card: CardGenerationModelInput = {
     cardExpirationDay: expirationDay ?? null, // JS number can represent integers up to 2^53, so it can represent all values of an uint32 (protobuf)
     cardInfoHashBase64: uint8ArrayToBase64(cardInfoHash),
-    totpSecretBase64: uint8ArrayToBase64(activationCode.totpSecret),
+    totpSecretBase64: totpSecret,
```
maxammann, line 25: I think we should discuss here if we want to determine here whether: totpSecret == null means that the card is static.

There is some implicit knowledge involved here.

administration/src/components/cards/CreateCardsController.tsx
```diff
@@ -35,14 +37,22 @@ const CreateCardsController = () => {
...
+        : null
+
+      const font = await loadTTFFont('NotoSans', 'normal', '/pdf-fonts/NotoSans-Regular.ttf')
+      const pdfDataUri = generatePdf(font, region, activationCodes, staticCodes)
 
       await activateCards(client, activationCodes, region)
 
-      const font = await loadTTFFont('NotoSans', 'normal', '/pdf-fonts/NotoSans-Regular.ttf')
-      const pdfDataUri = generatePdf(font, activationCodes, region)
+      if (staticCodes) await activateCards(client, staticCodes, region)
```
maxammann, line 55: I htink you already fixed there, that if the PDF generation fails, then the cards dop not get activated.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
```diff
@@ -13,7 +13,7 @@ const val CARD_INFO_HASH_LENGTH = 32 // Using SHA256-HMAC
 const val TOTP_SECRET_LENGTH = 20
 
 object Cards : IntIdTable() {
-    val totpSecret = binary("totpSecret", TOTP_SECRET_LENGTH)
+    val totpSecret = binary("totpSecret", TOTP_SECRET_LENGTH).nullable()
```
maxammann, line 16: See discussion in activation.ts

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardVerifier.kt
```diff
@@ -15,19 +15,28 @@ val TIME_STEP: Duration = Duration.ofSeconds(30)
...
+            !card.revoked
     }
 
     private fun verifyCard(card: CardEntity, totp: Int, timezone: ZoneId): Boolean {
-        val expirationDay = card.expirationDay
-        return (expirationDay == null || isOnOrBeforeToday(daysSinceEpochToDate(expirationDay), timezone)) &&
+        return !isExpired(card.expirationDay, timezone) &&
             !card.revoked &&
             isTotpValid(totp, card.totpSecret)
     }
```
maxammann, line 32: I feel like we have here a tention between implicitly having static and dynamic cards in a single DB table and still having different means of verifying them.

maxammann (changes requested):
The PR looks good overall! Though I think we should discuss whether:
* new DB schema is fine
* new GraphQL schema is fine

administration/src/cards/activation.ts
michael-markl: I am not entirely sure, how we should handle it, but to me, the current approach is a bit too implicit.
I would be okay with having a single mutation, but adding an extra field like `codeType: STATIC | DYNAMIC` and additional backend validation checking `codeType == STATIC  <=>  totpSecret == NULL`.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
michael-markl: We could also have another column `codeType` with values `STATIC | DYNAMIC` and a DB constraint that checks `codeType == STATIC <=> TOTPSECRET == NULL`
This would make it more explicit when we are dealining with a static vs a dynamic card.

michael-markl (approved):

administration/src/cards/activation.ts
sarahsporck: agreed.

administration/src/components/cards/CreateCardsController.tsx
sarahsporck: yes

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardVerifier.kt
sarahsporck: I think having two functions enforces the feeling of tension, but basicly the only difference is that a static card does not have a totp secret 🤷🏼‍♀️

maxammann (approved):
Looks good to me!

maxammann:
ping me after merge so I can reset the production DB

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardVerifier.kt
```diff
@@ -15,19 +15,28 @@ val TIME_STEP: Duration = Duration.ofSeconds(30)
...
-        return verifyCard(card, totp, timezone)
+        return if (totp != null) verifyCard(card, totp, timezone) else verifyStaticCard(card, timezone)
+    }
+
+    private fun verifyStaticCard(card: CardEntity, timezone: ZoneId): Boolean {
+        return !isExpired(card.expirationDay, timezone) &&
+            !card.revoked
     }
 
     private fun verifyCard(card: CardEntity, totp: Int, timezone: ZoneId): Boolean {
```
michael-markl, line 28:
```suggestion
    private fun verifyDynamicCard(card: CardEntity, totp: Int, timezone: ZoneId): Boolean {
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/types/CardVerificationModel.kt
```diff
@@ -2,5 +2,5 @@ package app.ehrenamtskarte.backend.verification.webservice.schema.types
 
 data class CardVerificationModel(
     val cardInfoHashBase64: String,
-    val totp: Int
+    val totp: Int?
```
michael-markl, line 5: Can you add in CodeType here as well (and the validation check in the query endpoint) to make the query explicit as well?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardVerifier.kt
```diff
@@ -15,19 +15,28 @@ val TIME_STEP: Duration = Duration.ofSeconds(30)
 const val TOTP_LENGTH = 6
 
 object CardVerifier {
-    fun verifyCardHash(project: String, cardHash: ByteArray, totp: Int, timezone: ZoneId): Boolean {
+    fun verifyCardHash(project: String, cardHash: ByteArray, totp: Int?, timezone: ZoneId): Boolean {
         val card = transaction { CardRepository.findByHashModel(project, cardHash) } ?: return false
-        return verifyCard(card, totp, timezone)
+        return if (totp != null) verifyCard(card, totp, timezone) else verifyStaticCard(card, timezone)
+    }
```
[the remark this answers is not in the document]
sarahsporck: after finishing the frontend task I agree with you, as the verification is separated either way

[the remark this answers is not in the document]
sarahsporck: will do this in the frontend task.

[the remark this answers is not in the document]
michael-markl: Perfect! :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/types/CardVerificationModel.kt
michael-markl: I guess this will also be resolved in the frontend pr?

PR #747, opened 2023-01-24
675: Add support for static qr codes in frontend

sarahsporck:


frontend/lib/identification/verification_workflow/verification_qr_code_processor.dart
```diff
@@ -42,6 +70,12 @@ void _assertConsistentDynamicVerifyCode(DynamicVerifyCode verifyCode) {
   }
 }
 
+void _assertConsistentStaticVerifyCode(StaticVerifyCode verifyCode) {
+  if (!verifyCode.hasPepper()) {
+    throw QrCodeFieldMissingException("hashSecretBase64");
```
michael-markl, line 75:
```suggestion
    throw QrCodeFieldMissingException("pepper");
```

michael-markl (approved):
This is probably missing in the issue description, but I think we agreed on the following:
In the case, a static card is being verified, the user (the verifier) should be asked "more thoroughly" to check the personalausweis (for example, once the server has verified the static card, a checkbox labeled "Ich habe die Daten mit einem amtlichen Lichtbildausweis abgeglichen" should be shown, and only once this checkbox is checked, it should show "Nürnberg-Pass ist gültig".)
Should we do this in another issue? I guess it shouldn't block this PR.

frontend/lib/identification/verification_workflow/verification_qr_code_processor.dart
sarahsporck: So is it called pepper when there is no `totpSecret` and `hashSecretBase64` when there is a `totpSecret`?

michael-markl: no, it is always called papper :)

sarahsporck:
I would prefer to add the checkbox in a separate ticket, as I am really not familiar with how to build nice flutter ui and someone else can do this in half the time I'd need. Therefore I'll create an issue for this.

maxammann:
Is there already a task?

PR #758, opened 2023-01-26
757: fix category mapping

f1sh1918:
- adjusted category mapping by using global categoryIds from the csv


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/nuernberg/constants/Categories.kt
```diff
@@ -0,0 +1,15 @@
+package app.ehrenamtskarte.backend.stores.importer.nuernberg.constants
+
+val categoryMapping =
```
f1sh1918, line 3: just kept the mapping with default to ensure that wrong categoryIds will mapped to defaultValue `9`

maxammann (approved):

PR #774, opened 2023-02-03
Fix map on android

maxammann:
This updates to a verion of maplibre flutter which includes https://github.com/m0nac0/flutter-maplibre-gl/pull/194

It also fixes the output of a strack trace in the console.

fixes #774


michael-markl:
Very nice! Work on my Android 11 machine.

michael-markl (approved):

PR #777, opened 2023-02-03
Test in browsers

michael-markl:
This PR switches our testing framework for administration from Jest@Node to Jasmine@ChromeHeadless through Karma.
It is always better to test on the actual platform we deploy to.
Moreover, I'll need this later on, to test the hashing algo (which needs browser's crypto APIs).


maxammann:
I'm not entirely sure if we need to run our tests in browsers at this point in time.
The current crypto APIs we use are supported by Node and even deno:
* https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey#browser_compatibility
* https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues#browser_compatibility

Maybe you tested this though and it did not work.

I think the added overhead of running our non-ui unit tests in the browser is not really worth it.

michael-markl:
Hmm... I guess you are right with "Node has these APIs built in" (where these APIs are the Web Crypto API, and the TextEncoder and TextDecoder). Actually, my problem seems to have been Jest and, in particular, JSDOM.
Jest is destroying the global environment after every test (I guess) and creates its own global environment and tries to emulate the browser. While doing so, they actually ignore these "newer" (but not really new) APIs, see:
https://github.com/jsdom/jsdom/issues/2524
https://github.com/jsdom/jsdom/issues/1612

It's quite easy to hack around these issues, though...

> I think the added overhead of running our non-ui unit tests in the browser is not really worth it.

I would actually not agree with that. "Overhead" in terms of performance is negligible (I feel like the browser tests are actually faster than with Jest@Node^^). Also, testing on the actual platform seems to be much more reasonable (especially now that I know how badly Jest is trying to "emulate a browser") :D
The config overhead seems to be also very small (imo). To me it's quite fascinating how easy it was to set this up (with proper debugging support from within Intellij :D )

But, now that I know how to hack around Jest / JSDom, we can also close this PR without merging. Up to you :)

maxammann:
I feel like actually including jsdom as a dependency is maybe overkill.
I would not write any tests for functionality which involves the DOM. What we are left then is with APIs which are probably supported well by node these days.

So exceptionally, instead of testing only some parts of our webapp we could say that we are testing a module of our webapp which runs in node and web browsers.

If we structure it that way jest does not need jsdom at all and we can remove even more dependencies.

michael-markl:
> I feel like actually including jsdom as a dependency is maybe overkill.

Agree (to the extend that I don't think jsdom should exist :D)

> I would not write any tests for functionality which involves the DOM

I would agree that we should not write tests involving the DOM for now. Who knows, what the future will bring though.

> If we structure it that way jest does not need jsdom at all and we can remove even more dependencies.

Imo, Jest+jsdom is always the wrong decision :D

> So exceptionally, instead of testing only some parts of our webapp we could say that we are testing a module of our webapp which runs in node and web browsers.

I think, ultimately, this is a philosophical question. Imo, we should not test on the node platform, as the production will never run on node. Instead we should test these really important functions - the hashing we execute on browsers and write their output to the DB - on the right platforms (probably Chrome *and* Firefox). I guess there may always be slight differences in the implementations of the APIs, so we should make sure that we don't run into platform-specific errors.

administration/package.json
```diff
@@ -35,36 +35,47 @@
...
+    "karma-webpack": "^5.0.0",
     "prettier": "^2.7.1",
+    "ts-loader": "^9.4.2",
     "typescript": "^4.8.3"
   },
   "scripts": {
     "start": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh); react-scripts start",
     "build": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh); react-scripts build",
-    "test": "react-scripts test",
+    "test": "karma start --single-run",
```
maxammann, line 62: We should eject sooner or later. I think we are still depending on jest through react-scripts.

maxammann (approved):

administration/package.json
michael-markl: Yeah, I agree... We are slowly but surely moving away from react-scripts :)

michael-markl:
Okay, let's go with this approach for now. If we find it's too cumbersome in some way, we can always switch back to the alternative in #778 .

PR #780, opened 2023-02-05
Fix extensions in administration and make them more type safe

michael-markl:
Extensions have not been applied properly.

### DIFF
--- a/administration/src/cards/CardBlueprint.ts
+++ b/administration/src/cards/CardBlueprint.ts
@@ -1,6 +1,7 @@
 import { CardExtensions, CardInfo, DynamicActivationCode, QrCode, StaticVerificationCode } from '../generated/card_pb'
 import { dateToDaysSinceEpoch } from './validityPeriod'
 import { ExtensionHolder } from './extensions'
+import { PartialMessage } from '@bufbuild/protobuf'
 
 const MAX_NAME_LENGTH = 50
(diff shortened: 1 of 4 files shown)


administration/src/cards/CardBlueprint.ts
```diff
@@ -63,13 +64,14 @@ export class CardBlueprint {
...
-    for (const state of this.extensionHolders) {
-      if (!state) {
-        throw new Error('Tried to add invalid extension')
+    for (const holder of this.extensionHolders) {
+      if (holder.state === null) {
+        // We allow to skip invalid extensions to enable computing the protobuf size.
+        continue
       }
-      state.extension.setProtobufData(state, extension_message)
+      holder.extension.setProtobufData(holder.state, extensionsMessage)
```
michael-markl, line 74: This was the main cause of the error.

administration/src/cards/cardBlueprints.ts
```diff
@@ -1,36 +1,23 @@
...
-    {
-      state: bavaria_card_type.initialState,
-      extension: bavaria_card_type,
-    },
-    {
-      state: {
-        region_id: region.id,
-      },
-      extension: region_extension,
-    },
```
michael-markl, line 17: This was not type checked properly

maxammann (approved):
Fixed this already on another branch, but your fix actually improves the types. So let's merge this one!

PR #787, opened 2023-02-06
New hash concept

### LINKED ISSUE #673: (Placeholder) Design a way to reliably create stable hashes for cards

**Additional context**

### LINKED ISSUE #699: Add a version for hashes
We should introduce a version for hashes. That way we can decide during hash comparision whether the comparision failed because of the hashed data or an invalid hasing algorithm/method.

This version should be included in the DB and also in the QR codes.

This is related to #673 

None

**Additional context**
None

michael-markl:
Closes #673 . 
I implemented the hash concept defined in the comment https://github.com/digitalfabrik/entitlementcard/issues/673#issuecomment-1407695219.

As versioning with this approach seems very likely unnecessary, this also closes #699 .


frontend/lib/identification/qr_code_utils.dart
```diff
@@ -26,20 +28,41 @@ class QrCodeUtils {
...
 
-    final expirationDay = cardInfo.expirationDay;
-    data.setUint32(offset, expirationDay, Endian.little);
-    offset += 4;
-    data.setInt32(offset, cardInfo.extensions.extensionBavariaCardType.cardType.value, Endian.little);
-    offset += 4;
-    data.setInt32(offset, cardInfo.extensions.extensionRegion.regionId, Endian.little);
-    offset += 4;
+  Map<String, dynamic> messageToJsonObject(GeneratedMessage message) {
+    if (message.unknownFields.isNotEmpty) throw ArgumentError("Unknown field");
```
michael-markl, line 37: Probably we should return a more distinct error here and catch it, and show something like "Your app is likely out of date.".

administration/package.json
```diff
@@ -16,6 +16,7 @@
     "@mui/material": "^5.10.9",
     "@zxing/library": "^0.19.1",
     "apollo-upload-client": "^17.0.0",
+    "canonicalize": "^1.0.8",
```
maxammann, line 19: This dependency seems unmaintained. Should we just copy its code over as its not much code? https://github.com/erdtman/canonicalize/blob/master/lib/canonicalize.js

Then we can also do a pure TS implementation.

michael-markl: tbh, I don't think it's unmaintained, it's just some finished software :D But yeah, I can also copy paste it...

michael-markl: (also it has >150k weekly downloads https://www.npmjs.com/package/canonicalize)

michael-markl: Replaced it.

frontend/lib/util/json_canonicalizer.dart
```diff
@@ -0,0 +1,49 @@
...
+        }
+        next = true;
+        // Property names are strings - Use ES6/JSON
+        stringBuffer.write(json.encode(element));
+        stringBuffer.write(':');
+        // Property value - Recursive expansion
+        _serialize(jsonObject[element], stringBuffer);
+      }
+      stringBuffer.write('}');
+    }
```
michael-markl, line 47: Probably, we should add another `else` block and throw in this case.

michael-markl: Done.

administration/src/cards/hashCardInfo.ts
```diff
@@ -0,0 +1,105 @@
...
+        }
+        result[field.no.toString()] = messageToJsonObject(subMessage)
+        break
+      }
+      case 'enum':
+      case 'scalar': {
+        const value = message[field.localName]
+        if (value === undefined) {
+          // The field is not present.
+          continue
```
maxammann, line 58: I guess we do not need to encode the absence of fields explicitly, right?

administration/src/util/serializeToCanonicalJson.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Taken & adjusted from https://github.com/erdtman/canonicalize/blob/master/lib/canonicalize.js
+ * Under Apache 2.0 License
+ *
+ * Modifications:
+ * 1) Throw, if undefined is passed to the function (instead of returning undefined).
+ * 2) Throw, if an item of an array is undefined or if it is a Symbol (instead of serializing null).
+ * 3) Throw, if the value of an object is undefined or if it is a Symbol (instead of skipping the entry).
+ */
+function serializeToCanonicalJson(object: unknown): string {
```
maxammann, line 20: I think we should port over the simple tests here: https://github.com/erdtman/canonicalize/blob/master/test/simpletests.js

maxammann (approved):
Looks great!

administration/src/cards/hashCardInfo.ts
michael-markl: No, we should *not* encode absent fields. Otherwise, we won't be able to add new fields later.

administration/src/util/serializeToCanonicalJson.ts
michael-markl: Done.

PR #788, opened 2023-02-07
Upgrade JDK to 17 LTS

maxammann:
This PR finally defines our go-to JDK version which is Java 17 LTS.

You can use a newer version like Java 19 on your local machine. I added instructions how to tell IntelliJ to use whatever version you have (above 17 ofc)


docs/development-setup.md
```diff
@@ -34,12 +34,14 @@ This short guide focuses on setting up the project using IntelliJ instead of And
...
-11. Take a look at the backend: [http://localhost:7000](http://localhost:7000) (The public version is available at
+6. Install at least JDK 17 LTS
+7. Open the IntelliJ "Project Structure" and setup the required SDK called "entitlementcard-jdk" and point it to your JDK installation.
+   ![SDK/JDK setup](./img/intellij-sdk-setup.png)
+8. Run the backend: `cd backend && ./gradlew run --args="execute"` or `.\backend\gradlew.bat run --args="execute"` on Windows
+9. Create an admin account using `./gradlew run --args="create-admin <email> <password>"`
+10. Take a look at the martin endpoints: [http://localhost:5002/tiles/accepting_stores/index.json](http://localhost:5002/tiles/accepting_stores/index.json) and [http://localhost:5002/tiles/accepting_stores/rpc/index.json](http://localhost:5002/tiles/accepting_stores/rpc/index.json). The data shown on the map is fetched from a hardcoded url and is not using the data from the local martin!
+11. Take a look at the style by viewing the test map: [http://localhost:5002](http://localhost:5002)
+12. Take a look at the backend: [http://localhost:7000](http://localhost:7000) (The public version is available at
```
michael-markl, line 44:
```suggestion
12. Take a look at the backend: [http://localhost:8000](http://localhost:8000) (The public version is available at
```

michael-markl (approved):
Is there a technical reason why we should perform this upgrade *now*?

maxammann:
> Is there a technical reason why we should perform this upgrade _now_?

I was not able to compile with Java 19.
Changing the compile target should not have any implications if everything still compiles fine.

We were already using Java 11 in production (I guess nobody was aware or knew). So going to Java 17 should be a easy step.

maxammann:
There is the small change of something breaking. But I think the benefits of finally having defined properly which version we support is good :D

michael-markl:
👍🏽 I was just wondering.

Now, I am even more wondering why it would break with Java 19 :D

michael-markl:
... or do you mean Java 1.9? :D

maxammann:
> 👍🏽 I was just wondering.
> 
> Now, I am even more wondering why it would break with Java 19 :D

Gradle did not support Java 19 until the latest update.
I would say everyone is free to use whatever java version you want (above 17) for development.

That makes setup easier because you can probably just use whatever java you have installed.

- Debian Java 17
- Arch Java 19
- macOS Java 19

michael-markl:
17 is LTS, so I'd recommend that :D

maxammann:
I'll fix the entitlmentcard server after merge

.circleci/config.yml
```diff
@@ -264,7 +264,7 @@ jobs:
       _JAVA_OPTIONS: "-Xmx3g"
       GRADLE_OPTS: "-Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2"
     docker:
-      - image: cimg/openjdk:11.0.16-node
+      - image: cimg/openjdk:19.0.2-node
```
michael-markl, line 267: shouldn't we use 17 here as well though?

maxammann: Makes sense!

Issue #794, opened 2023-02-10
Bayern: Change contact data "Herausgeber" und "Haftung" with general data


dkehne:
Replaces the Contact details of Mr. Diener under Publisher and Liability by general data 
 ([Poststelle@stmas.bayern.de](mailto:Poststelle@stmas.bayern.de) and Durchwahl -01

maxammann:
Forgot to do this in #801

michael-markl:
@dkehne Should this also be adjusted in the datenschutzerklärung? https://bayern.ehrenamtskarte.app/data-privacy-policy

dkehne:
@michael-markl please everywhere where he is mentioned.

PR #801, opened 2023-02-14
Release ehrenamtskarte without ausweisen

maxammann:


frontend/lib/home/home_page.dart
```diff
@@ -27,6 +28,16 @@ class _HomePageState extends State<HomePage> {
   @override
   void initState() {
     super.initState();
+    final List<AppFlow> additionalTabs = buildConfig.featureFlags.verification
```
f1sh1918, line 31: i think it should just be called `verificationTab`

f1sh1918 (approved):
tested on ios

frontend/lib/home/home_page.dart
```diff
@@ -43,12 +54,7 @@ class _HomePageState extends State<HomePage> {
         "Suche",
         GlobalKey<NavigatorState>(debugLabel: "Search tab key"),
       ),
-      AppFlow(
-        const IdentificationPage(title: "Ausweisen"),
-        Icons.remove_red_eye_outlined,
-        "Ausweisen",
-        GlobalKey<NavigatorState>(debugLabel: "Auth tab key"),
-      ),
+      ...additionalTabs,
```
michael-markl, line 57: Can't you do an if clause inside the array declaration? 
Would be a lot easier to read.
https://github.com/dart-lang/language/blob/master/accepted/2.3/control-flow-collections/feature-specification.md

michael-markl (approved):
Would be nice if we could fix https://github.com/digitalfabrik/entitlementcard/issues/794 before releasing.

frontend/lib/home/home_page.dart
maxammann: I google for this feature but failed 😂

PR #804, opened 2023-02-15
766: rethinking card activation

sarahsporck:
Already tested for Static and Dynamic Qr codes. And for dev and real code. Feel free to test again.
I also refactored the hash functionality to an extension in dart, as (imo) it improves readability.


backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -23,10 +27,15 @@ class CardMutationService {
             if (!Authorizer.mayCreateCardInRegion(user, targetedRegionId)) {
                 throw UnauthorizedException()
             }
-            val totpSecret = if (card.totpSecretBase64 != null) Base64.getDecoder().decode(card.totpSecretBase64) else null
+            val activationSecret =
+                if (card.activationSecretBase64 != null)
+                    Base64.getDecoder().decode(card.activationSecretBase64)
+                else null
```
michael-markl, line 33: A bit more compact, but maybe little bit less readable:
```suggestion
            val activationSecret = card.activationSecretBase64?.let { Base64.getDecoder().decode(it) }
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -23,10 +27,15 @@ class CardMutationService {
             if (!Authorizer.mayCreateCardInRegion(user, targetedRegionId)) {
```
michael-markl, line 27: We should check at the beginning of this function, whether the CardGenerationModel is valid (ie. if dynamic, then activation secret is present, if static, then it's absent).
For invalid input, throw an exception.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
```diff
@@ -21,9 +21,18 @@ object CardRepository {
...
-    fun insert(cardInfoHash: ByteArray, totpSecret: ByteArray?, expirationDay: Long?, regionId: Int, issuerId: Int, codeType: CodeType) =
+    fun findByHashModelAndActivationSecret(project: String, hashModel: ByteArray, activationSecret: ByteArray): CardEntity? {
+        val query = (Projects innerJoin Regions innerJoin Cards)
+            .slice(Cards.columns)
+            .select { Projects.project eq project and (Cards.cardInfoHash eq hashModel) and (Cards.activationSecret eq activationSecret) }
+            .singleOrNull()
+        return if (query == null) null else CardEntity.wrapRow(query)
+    }
+
+    fun insert(cardInfoHash: ByteArray, activationSecret: ByteArray?, totpSecret: ByteArray?, expirationDay: Long?, regionId: Int, issuerId: Int, codeType: CodeType) =
```
michael-markl, line 32: The totpSecret should always be null when inserting a new card, so we can omit the argument.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
```diff
@@ -21,9 +21,18 @@ object CardRepository {
...
     }
 
-    fun insert(cardInfoHash: ByteArray, totpSecret: ByteArray?, expirationDay: Long?, regionId: Int, issuerId: Int, codeType: CodeType) =
+    fun findByHashModelAndActivationSecret(project: String, hashModel: ByteArray, activationSecret: ByteArray): CardEntity? {
+        val query = (Projects innerJoin Regions innerJoin Cards)
+            .slice(Cards.columns)
+            .select { Projects.project eq project and (Cards.cardInfoHash eq hashModel) and (Cards.activationSecret eq activationSecret) }
+            .singleOrNull()
+        return if (query == null) null else CardEntity.wrapRow(query)
+    }
```
michael-markl, line 30:
```suggestion
    fun findByHashAndActivationSecret(project: String, cardInfoHash: ByteArray, activationSecret: ByteArray): CardEntity? {
        val query = (Projects innerJoin Regions innerJoin Cards)
            .slice(Cards.columns)
            .select { Projects.project eq project and (Cards.cardInfoHash eq cardInfoHash) and (Cards.activationSecret eq activationSecret) }
            .singleOrNull()
        return if (query == null) null else CardEntity.wrapRow(query)
    }
```

(If you want you can also adjust the name of the function above to `findByHash` or `findByCardInfoHash`.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
```diff
@@ -32,4 +41,8 @@ object CardRepository {
             this.revoked = false
             this.codeType = codeType
         }
+
+    fun reactivate(card: CardEntity, totpSecret: ByteArray) {
```
michael-markl, line 45: Maybe `activate`?

michael-markl: We should probably check that CardEntity is actually a dynamic card?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +44,33 @@ class CardMutationService {
         }
         return true
     }
+
+    @GraphQLDescription("Activate a digital EAK")
+    fun activateCard(
+        project: String,
+        card: CardGenerationModel,
```
michael-markl, line 51: I think we should not reuse the CardGenerationModel here.
Instead, we should ask only for the things we need: cardInfoHashBase64 and activationSecretBase64.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +44,33 @@ class CardMutationService {
...
+        overwrite: Boolean
+    ): CardActivationResultModel {
+        if (card.codeType == CodeType.static) {
+            return CardActivationResultModel(ActivationState.failed)
+        }
+
+        val cardHash = Base64.getDecoder().decode(card.cardInfoHashBase64)
+        val activationSecret = Base64.getDecoder().decode(card.activationSecretBase64)
+        val activatedCard = transaction { CardRepository.findByHashModelAndActivationSecret(project, cardHash, activationSecret,) }
```
michael-markl, line 60: What does this trailing comma do after activationSecret? :D
We should probably call it `card` rather than `activatedCard`.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +44,33 @@ class CardMutationService {
...
+
+        if (activatedCard == null || activatedCard.activationSecret == null) {
+            return CardActivationResultModel(ActivationState.failed)
+        }
+
+        if (!overwrite) {
```
michael-markl, line 66: I think we're missing the check that the totp secret already exists:
```suggestion
        if (activatedCard.totpSecret != null && !overwrite) {
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +44,33 @@ class CardMutationService {
...
+        if (card.codeType == CodeType.static) {
+            return CardActivationResultModel(ActivationState.failed)
+        }
+
+        val cardHash = Base64.getDecoder().decode(card.cardInfoHashBase64)
+        val activationSecret = Base64.getDecoder().decode(card.activationSecretBase64)
+        val activatedCard = transaction { CardRepository.findByHashModelAndActivationSecret(project, cardHash, activationSecret,) }
+
+        if (activatedCard == null || activatedCard.activationSecret == null) {
+            return CardActivationResultModel(ActivationState.failed)
```
michael-markl, line 63: I think we can omit the check for `activatedCard.activationSecret == null` here, as we already used the `findByHashAndActivationSecret` function.

We should probably add a log (similar to when logging into administration fails) that the activation failed. (In order to block IPs that try to activate a card by brute forcing it :D )

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/types/CardGenerationModel.kt
```diff
@@ -4,7 +4,7 @@ import app.ehrenamtskarte.backend.verification.database.CodeType
 
 data class CardGenerationModel constructor(
     val cardInfoHashBase64: String,
-    val totpSecretBase64: String?,
+    val activationSecretBase64: String?,
     // Using Long because UInt is not supported, and a protobuf uin32 is not representable by a Kotlin Int
```
michael-markl, line 8:
```suggestion
    // Using Long because UInt is not supported, and a protobuf uint32 is not representable by a Kotlin Int
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +44,33 @@ class CardMutationService {
         }
         return true
     }
+
+    @GraphQLDescription("Activate a digital EAK")
```
michael-markl, line 48:
```suggestion
    @GraphQLDescription("Activate a dynamic entitlement card")
```

administration/src/cards/CardBlueprint.ts
```diff
@@ -97,13 +97,13 @@ export class CardBlueprint {
     // https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HTOP)
     // https://tools.ietf.org/html/rfc4226#section-4 - R6 (How long should a shared secret be? -> 160bit)
     // https://tools.ietf.org/html/rfc4226#section-7.5 - Random Generation (How to generate a secret? -> Random)
```
michael-markl, line 99: These comments should be moved to the backend now, as we generate the totp secret there.

administration/src/cards/CardBlueprint.ts
```diff
@@ -97,13 +97,13 @@ export class CardBlueprint {
...
-    const totpSecret = new Uint8Array(TOTP_SECRET_LENGTH)
-    crypto.getRandomValues(totpSecret)
+    const activationSecret = new Uint8Array(TOTP_SECRET_LENGTH)
```
michael-markl, line 100: We should probably define a new constant `ACTIVATION_SECRET_LENGTH`. How long should it be (cc @maxammann )?

michael-markl (changes requested):

frontend/lib/identification/activation_workflow/activation_code_scanner_page.dart
```diff
@@ -100,4 +87,43 @@ class ActivationCodeScannerPage extends StatelessWidget {
...
+    final activationResult = await activateCode(
+      client: client,
+      projectId: projectId,
+      code: activationCode,
+      overwriteExisting: true,
+    );
+
+    switch (activationResult.activationState) {
+      case ActivationState.success:
+        if (activationResult.totpSecret != null) {
```
michael-markl, line 108: I think we should show an error if `activationResult.totpSecret` is null.

frontend/lib/identification/activation_workflow/activation_code_scanner_page.dart
```diff
@@ -100,4 +87,43 @@ class ActivationCodeScannerPage extends StatelessWidget {
...
+        break;
+      case ActivationState.failed:
+        await QrParsingErrorDialog.showErrorDialog(
+          context,
+          "Der eingescannte Code ist ungültig.",
+        );
+        break;
+      case ActivationState.didNotOverwriteExisting:
+      // TODO: ask user to overwrite code
+      default:
```
michael-markl, line 126: I think we should show an error here to the user in these two cases.

frontend/lib/identification/qr_code_utils.dart
```diff
@@ -1,41 +1,28 @@
...
-    final json = messageToJsonObject(cardInfo);
+  List<int> _toBinary() {
+    final json = messageToJsonObject();
     final canonicalizedJsonString = JsonCanonicalizer().canonicalize(json);
     return utf8.encode(canonicalizedJsonString);
   }
 
-  Map<String, dynamic> messageToJsonObject(GeneratedMessage message) {
-    if (message.unknownFields.isNotEmpty) throw ArgumentError("Unknown field");
+  Map<String, dynamic> messageToJsonObject() {
```
michael-markl, line 22: maybe we should rename this function then to `toCanonicalJsonObject`?

frontend/lib/about/dev_settings_view.dart
```diff
@@ -179,6 +164,53 @@ class DevSettingsView extends StatelessWidget {
...
+          totpSecret: totpSecret,
+        );
+        provider.setCode(userCode);
+      }
+
+      messengerState.showSnackBar(
+        const SnackBar(
+          content: Text("Card activated."),
+        ),
+      );
```
michael-markl, line 203: This should be translated to German + whitelabeled.
(
 Nürnberg => "Nürnberg-Pass aktiviert."
 Bayern => "Ehrenamtskarte aktiviert."
)

frontend/lib/about/dev_settings_view.dart
```diff
@@ -179,6 +164,53 @@ class DevSettingsView extends StatelessWidget {
...
+        overwriteExisting: true,
+      );
+
+      if (activationResult.activationState == ActivationState.failed) {
+        throw QrCodeParseException("Code could not be activated from the server");
+      }
+      if (activationResult.activationState == ActivationState.didNotOverwriteExisting) {
+        throw QrCodeParseException("Did not overwrite existing QrCode.");
+      }
+      if (activationResult.totpSecret != null) {
```
michael-markl, line 189: If activationResult.totpSecret is null, an error should be shown to the user.

frontend/lib/about/dev_settings_view.dart
```diff
@@ -179,6 +164,53 @@ class DevSettingsView extends StatelessWidget {
...
+        code: activationCode,
+        overwriteExisting: true,
+      );
+
+      if (activationResult.activationState == ActivationState.failed) {
+        throw QrCodeParseException("Code could not be activated from the server");
+      }
+      if (activationResult.activationState == ActivationState.didNotOverwriteExisting) {
+        throw QrCodeParseException("Did not overwrite existing QrCode.");
+      }
```
michael-markl, line 188: Maybe we should invert the logic here:
if (activationResult.activationState !== ActivationState.success) { show error }

frontend/lib/about/dev_settings_view.dart
```diff
@@ -179,6 +164,53 @@ class DevSettingsView extends StatelessWidget {
...
+
+      final activationResult = await activateCode(
+        client: client,
+        projectId: projectId,
+        code: activationCode,
+        overwriteExisting: true,
+      );
+
+      if (activationResult.activationState == ActivationState.failed) {
+        throw QrCodeParseException("Code could not be activated from the server");
```
michael-markl, line 184:
```suggestion
        throw QrCodeParseException("Code ist ungültig.");
```

frontend/lib/about/dev_settings_view.dart
```diff
@@ -179,6 +164,53 @@ class DevSettingsView extends StatelessWidget {
...
+      }
+      if (activationResult.activationState == ActivationState.didNotOverwriteExisting) {
+        throw QrCodeParseException("Did not overwrite existing QrCode.");
```
michael-markl, line 187:
```suggestion
        throw QrCodeParseException("Fehler: Der Code wurde bereits genutzt und die zugehörige App wurde nicht deaktiviert.");
```

frontend/lib/about/dev_settings_view.dart
```diff
@@ -3,19 +3,23 @@ import 'dart:developer';
...
 import 'package:flutter/material.dart';
+import 'package:graphql_flutter/graphql_flutter.dart';
 import 'package:provider/provider.dart';
 
 // this data includes a Base32 encoded random key created with openssl
 // for testing, so this is intended
-final sampleActivationCodeBavaria = DynamicActivationCode(
+final sampleActivationCodeBavaria = DynamicUserCode(
   info: CardInfo(
     fullName: "Jane Doe",
```
michael-markl, line 24:
```suggestion
    fullName: "Erika Mustermann",
```
Closing #805 :D

frontend/lib/about/dev_settings_view.dart
```diff
@@ -30,7 +34,7 @@ final sampleActivationCodeBavaria = DynamicActivationCode(
   totpSecret: base32.decode("MZLBSF6VHD56ROVG55J6OKJCZIPVDPCX"),
 );
 
-final sampleActivationCodeNuernberg = DynamicActivationCode(
+final sampleActivationCodeNuernberg = DynamicUserCode(
   info: CardInfo(
     fullName: "Jane Doe",
```
michael-markl, line 39:
```suggestion
    fullName: "Erika Mustermann",
```
Closing https://github.com/digitalfabrik/entitlementcard/issues/805 :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +44,33 @@ class CardMutationService {
...
+        overwrite: Boolean
+    ): CardActivationResultModel {
+        if (card.codeType == CodeType.static) {
+            return CardActivationResultModel(ActivationState.failed)
+        }
+
+        val cardHash = Base64.getDecoder().decode(card.cardInfoHashBase64)
+        val activationSecret = Base64.getDecoder().decode(card.activationSecretBase64)
+        val activatedCard = transaction { CardRepository.findByHashModelAndActivationSecret(project, cardHash, activationSecret,) }
+
```
michael-markl, line 61: Please check whether the card is actually still valid: If it is expired or if it was revoked, we should return failed.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
sarahsporck: hmm. but I think actual error handling should not happen in the CardRepository itself, but before that. Adding an if/else still won't hurt though 🤷🏼‍♀️

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
sarahsporck: error should be thrown either way because of database constraint. I'll still add the check as it's probably better practice to not rely on constraints.

frontend/lib/about/dev_settings_view.dart
sarahsporck: so we should not consider ActivationState.did_not_overwrite_existing?

frontend/lib/about/dev_settings_view.dart
sarahsporck: This is in the dev_settings_view. Are you sure we need an error here? In the actual activationWorkflow I *do* show an error.

frontend/lib/about/dev_settings_view.dart
sarahsporck: This is still the dev_settings_view. I can change the text, but this should never be visible to the user.

michael-markl: Agree, sorry I somehow overlooked that this is only the dev settings view...  Should not be whitelabeled :D
Also, probably not important to change the text here...

frontend/lib/about/dev_settings_view.dart
michael-markl: It would still be good to throw an error in this case. Alternatively, just remove the `if` check and let it throw automatically when you dereference `activationResult.totpSecret!`. It's better not to show a success snackbar, if there was no success :)

frontend/lib/about/dev_settings_view.dart
michael-markl: It's just safer to invert the logic here.
If we ever add another entry to the enum in ActivationState, we might not be in the `success` state after checking for `didNotOverwriteExisting` and `failed`.

frontend/lib/about/dev_settings_view.dart
michael-markl: Not sure about this...

frontend/lib/about/dev_settings_view.dart
michael-markl: Not sure about this...

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
michael-markl: Hmmm depends :D I guess the repositories should also be responsible for the entities' coherence/validation.
They do the manipulation, so they should check whether that is an allowed manipulation I guess... The concepts / responsiblities are quite blurry in our setup though....

frontend/lib/identification/activation_workflow/activation_code_scanner_page.dart
sarahsporck: not sure why `didNotOverwriteExisting` should show an exception.

michael-markl: As it is not the expected outcome (the expected outcome would be `success`).
Once we implement the "ping-pong" (by asking the user whether the code should be overwritten) we can adjust this behavior. Until then, we should show an error...

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
sarahsporck: so you think if/else is enough of a check or should we actually throw an error?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardActivator.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.eatthepath.otp.TimeBasedOneTimePasswordGenerator
+import javax.crypto.KeyGenerator
+import javax.crypto.Mac
+
+object CardActivator {
+
+    public fun generateTotpSecret(): ByteArray {
+        // https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HTOP)
+        //      // https://tools.ietf.org/html/rfc4226#section-4 - R6 (How long should a shared secret be? -> 160bit)
+        //           // https://tools.ietf.org/html/rfc4226#section-7.5 - Random Generation (How to generate a secret? -> Random))))
```
michael-markl:
```suggestion
        // https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HTOP)
        // https://tools.ietf.org/html/rfc4226#section-4 - R6 (How long should a shared secret be? -> 160bit)
        // https://tools.ietf.org/html/rfc4226#section-7.5 - Random Generation (How to generate a secret? -> Random))))
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
sarahsporck: will be done in a separate task.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
```diff
@@ -14,13 +14,15 @@ import org.jetbrains.exposed.sql.or
 
 const val CARD_INFO_HASH_LENGTH = 32 // Using SHA256-HMAC
 const val TOTP_SECRET_LENGTH = 20
+const val ACTIVATION_SECRET_LENGTH = 20
```
michael-markl, line 17: @maxammann Please suggest some length :P

frontend/lib/about/dev_settings_view.dart
sarahsporck: i copied the logic from the normal activation workflow

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +45,40 @@ class CardMutationService {
...
+            return CardActivationResultModel(ActivationState.did_not_overwrite_existing)
+        }
+
+        val totpSecret = CardActivator.generateTotpSecret()
+        val encodedTotpSecret = Base64.getEncoder().encodeToString(totpSecret)
+        transaction { CardRepository.activate(card, totpSecret) }
+        return CardActivationResultModel(ActivationState.success, encodedTotpSecret)
+    }
+
+    private fun verifyNewCard(card: CardGenerationModel): Boolean {
```
michael-markl, line 81:
```suggestion
    private fun validateNewCard(card: CardGenerationModel): Boolean {
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -35,4 +45,40 @@ class CardMutationService {
         }
         return true
     }
+
+    @GraphQLDescription("Activate a dynamic EAK")
```
michael-markl: EAK = Ehrenamtskarte.
We should also consider other entitlement cards (like the Nürnberg Pass) in this description.

```suggestion
    @GraphQLDescription("Activate a dynamic entitlement card")
```

frontend/lib/identification/activation_workflow/activation_code_scanner_page.dart
```diff
@@ -100,4 +72,51 @@ class ActivationCodeScannerPage extends StatelessWidget {
...
+    final provider = Provider.of<UserCodeModel>(context, listen: false);
+    final activationSecretBase64 = const Base64Encoder().convert(activationCode.activationSecret);
+    final cardInfoBase64 = activationCode.info.hash(activationCode.pepper);
+
+    final activationResult = await activateCode(
+      client: client,
+      projectId: projectId,
+      activationSecretBase64: activationSecretBase64,
+      cardInfoHashBase64: cardInfoBase64,
+      overwriteExisting: false,
```
michael-markl, line 91: I think until we implement the ping-pong, we want to always overwrite the existing activation?
```suggestion
      overwriteExisting: true,
```

michael-markl (approved):

frontend/lib/identification/activation_workflow/activation_code_scanner_page.dart
sarahsporck: Sure. I'm fine with both.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
maxammann: Definitely reasonable 👌🏼 

It's like a password with 160bit entropy, which is comparable to a 25 char PW.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardActivator.kt
```diff
@@ -0,0 +1,21 @@
+package app.ehrenamtskarte.backend.verification.service
+
+import app.ehrenamtskarte.backend.verification.database.TOTP_SECRET_LENGTH
+import com.eatthepath.otp.TimeBasedOneTimePasswordGenerator
+import javax.crypto.KeyGenerator
+
+object CardActivator {
+
+    public fun generateTotpSecret(): ByteArray {
```
maxammann, line 9: Maybe try to wrap this into: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/synchronized.html

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt
```diff
@@ -14,13 +14,15 @@ import org.jetbrains.exposed.sql.or
...
 const val TOTP_SECRET_LENGTH = 20
+const val ACTIVATION_SECRET_LENGTH = 30
 
 enum class CodeType {
     static,
     dynamic
 }
 
 object Cards : IntIdTable() {
+    val activationSecret = binary("activationSecret", ACTIVATION_SECRET_LENGTH).nullable()
```
michael-markl, line 25: Probably we should name this 
```suggestion
    val activationSecretHash = binary("activationSecretHash", ACTIVATION_SECRET_LENGTH).nullable()
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardActivator.kt
```diff
@@ -16,6 +19,8 @@ object CardActivator {
         val keyGenerator = KeyGenerator.getInstance(algorithm)
         keyGenerator.init(TOTP_SECRET_LENGTH * 8)
 
-        return keyGenerator.generateKey().encoded
+        return hashKey(keyGenerator.generateKey().encoded)
```
michael-markl, line 22: I think we should not hash the TOTP secret

We should only hash the activation secret when writing to DB, and use the `Bcrypt.verifyer().verify(rawActiviationSecret, activationSecretHash)` method when a user is trying to activate his card.

(My explanation may be confusing -- I'm a bit sluggish today^^)

michael-markl (changes requested):

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/service/CardActivator.kt
sarahsporck: Got you! I was kind of wondering whether I did this correct in first place.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -26,48 +26,54 @@ class CardMutationService {
...
+        val activationSecretHash = card?.activationSecretHash
+
+        System.out.println(activationSecretHash?.size)
+        System.out.println(rawActivationSecret.size)
+        if (card == null || activationSecretHash == null) {
+            return CardActivationResultModel(ActivationState.failed)
         }
 
-        if (card == null) {
+        if (CardActivator.verifyActivationSecret(rawActivationSecret, activationSecretHash)) {
```
michael-markl, line 76: Should be inverted I think?

sarahsporck: Ahh. Makes sense. Then just invert and it should work.

michael-markl (approved):

PR #811, opened 2023-02-22
Upgrade react-router-dom

michael-markl:
This PR, we can finally (? :D) upgrade to a recent version of react-router-dom. The problem was that the usePrompt hook (as well as other parts of the API) was removed in a intermediate version of react-router-dom v6. With https://github.com/remix-run/react-router/issues/8139#issuecomment-1021457943 the feature was added back in a recent version.


michael-markl:
> PS: I hate Router library in JS.

PR #818, opened 2023-02-23
722: create cards in a batch

sarahsporck:
Not sure how to display the correct error message. More exactly which card was responsible for an error.


backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -14,24 +14,31 @@ import java.util.Base64
 @Suppress("unused")
 class CardMutationService {
     @GraphQLDescription("Stores a new digital EAK")
```
michael-markl, line 16:
```suggestion
    @GraphQLDescription("Stores a batch of new digital entitlementcards")
```

michael-markl (commented):
It would be better if we had a single call for both static and dynamic codes.

michael-markl (approved):

Issue #624, opened 2022-11-13
[Application] Auto-Select Region 


michael-markl:
When applying for the EAK, the application is directed to the Stadt/Landkreis in which the applicant's PLZ lies.
## Acceptance criteria:
* [ ] The Stadt/Landkreis is automatically detected while filling the application
* [ ] The applicant can see which Stadt/Landkreis is selected (and why it is selected).
* [ ] If the Stadt/Landkreis does not participate in the EAK program, a friendly message is shown with a link to https://www.lbe.bayern.de/engagement-anerkennen/ehrenamtskarte/antraege/index.php

Source of the csv for region information
https://downloads.suche-postleitzahl.org/v2/public/zuordnung_plz_ort.csv

PR #824, opened 2023-02-24
624: Auto select region

f1sh1918:
1. Go http://localhost:3000/beantragen and type a 5 digits postalCode, region should be auto-selected
2. You can also manually set the Region. When you reload and the region is manually selected it shouldn't bet auto-selected again
3. Since i moved the policy logic, please also check if that works

resolves https://github.com/digitalfabrik/entitlementcard/issues/624


administration/src/application/components/forms/OrganizationForm.tsx
```diff
@@ -15,16 +15,42 @@ import {
...
-    'Tierschutz',
-    'Sport',
-    'Bildung',
-    'Umwelt-/Naturschutz',
-    'Kultur',
-    'Gesundheit',
-    'Katastrophenschutz/Feuerwehr/Rettungsdienst',
-    'Kirchen',
-    'Andere',
+    {
```
f1sh1918, line 18: for region i need `item: { label: string, value: string}` so i also adjusted here. we could also provide different Option types

administration/src/application/components/forms/AddressForm.tsx
```diff
@@ -68,7 +68,7 @@ const AddressForm: Form<State, Options, ValidatedInput, AdditionalProps> = {
         </div>
         <SubForms.country.Component
           state={state.country}
-          setState={useUpdateStateCallback(setState, 'postalCode')}
+          setState={useUpdateStateCallback(setState, 'country')}
```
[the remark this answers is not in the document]
michael-markl: No, not intended:D thanks for fixing!

administration/src/application/components/forms/RegionForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const getOptions = (regions: Region[]): SelectItem[] =>
+  regions.map(region => {
+    return { label: getOptionsLabel(region.prefix, region.name), value: region.id.toString() }
+  })
+
+const RegionForm: Form<State, Options, ValidatedInput, AdditionalProps> = {
+  initialState: { ...createCompoundInitialState(SubForms) },
+  getArrayBufferKeys: createCompoundGetArrayBufferKeys(SubForms),
+  validate: createCompoundValidate(SubForms, { region: { items: [] } }),
```
f1sh1918, line 34: i'm not sure how to provide proper validation data, since i get the regionData that i need via additional props and not hardcoded as in `OrganizationForm` i even would know why we can't deactivate the validation for region according to options provided

administration/src/application/components/forms/RegionForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+  Component: ({ state, setState, regionData, postalCode }) => {
+    const projectId = useContext(ProjectConfigContext).projectId
+    const { enqueueSnackbar } = useSnackbar()
+    const {} = useGetRegionByPostalCodeQuery({
+      onError: () => {
```
f1sh1918, line 39: I think its nice for the user to have a indicator if region could be auto selected or not

administration/src/application/components/primitive-inputs/SelectForm.tsx
```diff
@@ -1,23 +1,27 @@
...
   initialState: { selectedText: '' },
   getArrayBufferKeys: () => [],
   validate: ({ selectedText }, options) => {
     if (selectedText.length === 0) return { type: 'error', message: 'Feld ist erforderlich.' }
-    if (!options.items.includes(selectedText))
-      return {
-        type: 'error',
-        message: `Wert muss einer der auswählbaren Optionen entsprechen.`,
-      }
+    // if (!options.includes(selectedText))
```
f1sh1918, line 20: why we need this?

administration/src/application/components/ApplyController.tsx
```diff
@@ -86,6 +100,10 @@ const ApplyController = () => {
             Über den Fortschritt Ihres Antrags werden Sie per E-Mail informiert.
             Sie können das Fenster jetzt schließen.`
 
+  if (!regions) {
+    return null
+  }
```
michael-markl, line 105: I guess we should show a loading spinner if it's loading and an error comp. if it's erroneous.

administration/src/application/components/primitive-inputs/SelectForm.tsx
michael-markl: In principle, I think it is not necessary. It's just another layer of validation (in case our `selectedText` -- which should probably now be called `selectedValue` -- was somehow changed to an invalid string by MUI's Select element).

administration/src/application/components/forms/RegionForm.tsx
michael-markl: As a rule of thumb: All the data that you need for validation (and that you cannot hardcode) should go into the `Options` slot, not in the `AdditionalProps` slot. Additional props are conceptually for styling rather than content stuff^^
You can access options both in the component (via the `options` prop) and inside the validate function (via the second argument).

We can also go over this together if you want :)

administration/src/application/components/primitive-inputs/SelectForm.tsx
michael-markl: If this is too much in your way, we can also drop it :)

administration/src/application/components/ApplyController.tsx
```diff
@@ -103,6 +121,7 @@ const ApplyController = () => {
             onSubmit={submit}
             loading={loading || loadingPolicy}
             privacyPolicy={policyData?.dataPolicy.dataPrivacyPolicy ?? ''}
+            regionData={regions}
```
michael-markl, line 124: Most likely, you should pass the regions inside the options prop of ApplicationForm and PersonalDataForm (and not inside their Additional Props)

administration/src/application/components/ApplyController.tsx
f1sh1918: sure it was just a draft :)

f1sh1918: added loadingSpinner and refetch

administration/src/application/components/forms/RegionForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+          })
+          setState(() => ({ region: { selectedValue: id.toString() } }))
+        }
+      },
+      variables: { postalCode: postalCode, projectId: projectId },
+      skip: postalCode.length !== 5,
+    })
+
+    return (
+      <SubForms.region.Component
```
michael-markl, line 59: Add a `<Typography>` tag with a text with something like:

In der Regel müssen Sie die Ehrenamtskarte in dem Landkreis oder der kreisfreien Stadt Ihres Hauptwohnsitzes beantragen. Weitere Informationen können Sie [hier einsehen](https://www.lbe.bayern.de/engagement-anerkennen/ehrenamtskarte/voraussetzungen/index.php).
Eine Liste der teilnehmenden Landkreise und kreisfreien Städte finden Sie unten.

And below add an alert (see https://mui.com/material-ui/react-alert/) with something like


{If erfolgreich} `<Alert sucess>`Die zuständige Behörde konnte anhand Ihrer Postleitzahl automatisch ermittelt werden.`</Alert>`
{If keine gültige PLZ} `<Alert info>`Bitte geben Sie oben eine 5-stellige Postleitzahl an, sodass die zuständige Behörde automatisch ermittelt werden kann.`</Alert>`
{If loading} `<CircularProgress />`
{Else:} `<Alert warn>`Die zuständige Behörde konnte nicht automatisch anhand Ihrer Postleitzahl ermittelt werden.`</Alert>`

administration/src/application/components/forms/RegionForm.tsx
michael-markl: I think it is a bit confusing that the snackbars appear with text that I don't yet understand. (I don't really know what it's talking about, when I am still entering my address).
I think it would be better to have explanations of what happened directly at the select form (I would suggest removing the snackbars in favor of the suggestions below). [Also note, that snackbars are not really accessible -- at least not in our current implementation I think. We should avoid using them too often]

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
```diff
@@ -33,4 +37,28 @@ object RegionsRepository {
...
+            val csvInput = FileReader("src/main/resources/import/plz_ort_bayern.csv")
+            val records: Iterable<CSVRecord> = CSVFormat.RFC4180.parse(csvInput)
+            val postalCodes: MutableMap<String, String> = getPostalCodes(records)
+            val regionIdentifier = postalCodes[postalCode] ?: throw Exception("Region couldn't be found")
+            val regionId = RegionEntity.find { Regions.regionIdentifier eq regionIdentifier and (Regions.projectId eq projectId) }.single().id
+            return RegionEntity[regionId]
+        } catch (e: Exception) {
+            throw Exception("Couldn't read CSV")
+        }
+    }
```
michael-markl, line 52: Reading the CSV file should be done only once when starting the backend. You can probably pass the `regionIdentifierByPostalCode` map into the webservice endpoint using the GraphQLContext.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
```diff
@@ -33,4 +37,28 @@ object RegionsRepository {
     fun updateDataPolicy(region: RegionEntity, dataPrivacyText: String) {
         region.dataPrivacyPolicy = dataPrivacyText
     }
+
+    fun findRegionByPostalCode(postalCode: String, projectId: EntityID<Int>): RegionEntity {
+        try {
+            val csvInput = FileReader("src/main/resources/import/plz_ort_bayern.csv")
```
michael-markl, line 43: I think there's a better way to access files from resources. Maybe take a look at other places where we have done this (eg. for the backend configuration).

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/webservice/schema/RegionsQueryService.kt
```diff
@@ -28,7 +30,13 @@ class RegionsQueryService {
     @GraphQLDescription("Returns region data for specific region.")
     fun regionByRegionId(regionId: Int): Region = transaction {
         val regionEntity = RegionsRepository.findRegionById(regionId)
+        Region(regionEntity.id.value, regionEntity.prefix, regionEntity.name, regionEntity.regionIdentifier, regionEntity.dataPrivacyPolicy)
+    }
 
+    @GraphQLDescription("Returns region data by postal code.")
+    fun regionByPostalCode(postalCode: String, projectId: String): Region? = transaction {
```
michael-markl, line 37:
```suggestion
    @GraphQLDescription("Returns region by postal code. Works only for the EAK project in which each region has an appropriate regionIdentifier.")
    fun regionByPostalCode(postalCode: String, project: String): Region? = transaction {
```

administration/src/graphql/regions/getRegionByPostalCode.graphql
```diff
@@ -0,0 +1,7 @@
+query getRegionByPostalCode($postalCode: String!, $projectId: String!) {
+  region: regionByPostalCode(postalCode: $postalCode, projectId: $projectId) {
```
michael-markl, line 2:
```suggestion
query getRegionByPostalCode($postalCode: String!, $project: String!) {
  region: regionByPostalCode(postalCode: $postalCode, project: $project) {
```

administration/src/application/components/ApplyController.tsx
```diff
@@ -3,16 +3,19 @@ import '@fontsource/roboto/400.css'
...
 import ApplicationForm from './forms/ApplicationForm'
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useContext, useMemo, useState } from 'react'
 import { SnackbarProvider, useSnackbar } from 'notistack'
 import styled from 'styled-components'
 import DiscardAllInputsButton from './DiscardAllInputsButton'
 import ApplicationErrorBoundary from '../ApplicationErrorBoundary'
+import { ProjectConfigContext } from '../../project-configs/ProjectConfigContext'
+import ErrorHandler from '../../ErrorHandler'
+import { Spinner } from '@blueprintjs/core'
```
michael-markl, line 18: Use  CircularProgress from mui instead. (Also add some margin/padding :) )

administration/src/application/components/ApplyController.tsx
```diff
@@ -50,11 +53,28 @@ const ApplyController = () => {
...
+    refetch: refetchPolicy,
+  } = useGetDataPolicyQuery({
     variables: { regionId: regionId },
-    // TODO: Add proper error handling and a refetch button when regionId query is implemented
     onError: () => enqueueSnackbar('Datenschutzerklärung konnte nicht geladen werden', { variant: 'error' }),
   })
+  const projectId = useContext(ProjectConfigContext).projectId
+  const {
+    loading: loadingRegions,
+    error,
```
michael-markl, line 68:
```suggestion
    error: errorRegions,
```

administration/src/application/components/ApplyController.tsx
```diff
@@ -69,7 +89,10 @@ const ApplyController = () => {
   }
 
   const submit = () => {
-    const validationResult = ApplicationForm.validate(state)
+    if (!regions) {
+      return
+    }
```
michael-markl, line 94:
```suggestion
```
I think, if you move the definition of `submit` below the loading/error/data-presence checks, we should not need this.

administration/src/application/components/ApplyController.tsx
```diff
@@ -28,7 +31,7 @@ const SuccessContent = styled.div`
   margin-bottom: 1rem;
 `
```
michael-markl, line 33: There's a variable `regionId` above, which should not longer be necessary. Would you mind deleting it?

michael-markl (changes requested):
Bug:
1. Enter a valid PLZ for which the region (region A) can be successfully detremined, 
2. The region is succesfully determined
3. Select a different region B
4. Go to the next step of the form
5. Go back

Expected: Region B is still selected
Actual: Region A was automatically reselected

[The same bug occurs, when you reload the page instead of going forth and back]

We should not overwrite a manual input of the user.
For that, we probably have to memorize whether the input was made by the user or not whenever it changes...
This involves a bit more fiddeling with the forms I guess :/ ... Let me know if we should take a look together

administration/src/application/components/ApplyController.tsx
f1sh1918: ok i think the query for dataPrivacyPolicy has to be moved to a position where we have the regionId

administration/src/application/components/ApplyController.tsx
f1sh1918: ok i just centered the spinner

administration/src/graphql/regions/getRegionByPostalCode.graphql
f1sh1918: i renamed this but we even use projectId in the context
`const project = useContext(ProjectConfigContext).projectId`

michael-markl: Yeah, the naming is not consistent throughout the components. However all our graphql endpoints use it as I suggested, so we should at least be consistent with our inconstancies :D

administration/src/application/components/forms/RegionForm.tsx
f1sh1918: implemented the alerts

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
f1sh1918: moved to the `common/webservice`

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
f1sh1918: used `Classloader.getSystemResources`

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
```diff
@@ -33,4 +35,21 @@ object RegionsRepository {
     fun updateDataPolicy(region: RegionEntity, dataPrivacyText: String) {
         region.dataPrivacyPolicy = dataPrivacyText
     }
+
+    fun findRegionByPostalCode(postalCode: String, projectId: EntityID<Int>, postalCodes: Map<String, String>): RegionEntity {
```
michael-markl, line 39:
```suggestion
    fun findRegionByPostalCode(postalCode: String, projectId: EntityID<Int>, regionIdentifierByPostalCode: Map<String, String>): RegionEntity {
```

michael-markl (approved):

PR #828, opened 2023-02-27
Properly encode URL parameters in mails

### LINKED ISSUE #827: Password reset key is not picked up correctly
**Describe the bug**
The password reset key is currently base64 encoded and acts as a URL parameter.
It ends with two "=" signs.
However, it seems that some email clients / browsers ignore the last two equal signs.
As a consequence, the user is unable to reset their password, because the reset keys do not match.

The solution should probably come up with a better method of how the reset key is encoded in the URL.

michael-markl:
Fixes #827 .


f1sh1918 (approved):
looks fine to me :)
I wonder whether we maybe should include an api call when opening the `reset-password` page and check if the resetKey exists in database and is not expired before the user does anything. If i open the same resetLink even after i already set my password there is no indicator for the user that its not valid anymore.

This could also be done in a separate task for sure

michael-markl:
>  I wonder whether we maybe should include an api call when opening the reset-password page and check if the resetKey exists in database and is not expired before the user does anything.

Alternatively, better error message would help a lot here as well. Definitely a different task though.

PR #833, opened 2023-03-02
826: Modify index on email to be lowercased

sarahsporck:
Expression indices are [not yet supported in exposed](https://github.com/JetBrains/Exposed/issues/895) therefore the manual creation. If you have any idea where to create the index or whether there might be a better solution let me no.

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
@@ -11,9 +11,10 @@ import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.javatime.datetime
 import org.jetbrains.exposed.sql.or
+import org.jetbrains.exposed.sql.transactions.TransactionManager
 
 object Administrators : IntIdTable() {
-    val email = varchar("email", 100).uniqueIndex()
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
```diff
@@ -11,9 +11,10 @@ import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.javatime.datetime
 import org.jetbrains.exposed.sql.or
+import org.jetbrains.exposed.sql.transactions.TransactionManager
 
 object Administrators : IntIdTable() {
-    val email = varchar("email", 100).uniqueIndex()
+    val email = varchar("email", 100)
```
michael-markl, line 17: Hmm.... In the optimal scenario, we would probably also remove the index that was previously created?

michael-markl: Also according to [stackoverflow](https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address) we should up our limit from 100 to at least 254... (a different task for sure)
I feel like we are slowly getting into migrations territory :D ... We should probably think about some proper way to do DB migrations in general in the future...

michael-markl (approved):

PR #841, opened 2023-03-03
839: Setup DB migrations

sarahsporck:
Flyway seems to be the most popular tool in regard to db migrations. There is a community and an enterprise version and I think the community edition provides enough for our regards. 
Setup seems to be quite simple.

### DIFF
--- a/backend/src/main/resources/db/migrations/V1__remove_old_email_index.sql
+++ b/backend/src/main/resources/db/migrations/V1__remove_old_email_index.sql
@@ -1,0 +1,1 @@
+alter table administrators drop constraint administrators_email_unique;
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -44,15 +44,19 @@ dependencies {
     // Use the Kotlin JUnit integration.
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit")
 
+    // Database
     implementation("org.jetbrains.exposed:exposed-core:$exposedVersion")
     implementation("org.jetbrains.exposed:exposed-dao:$exposedVersion")
     implementation("org.jetbrains.exposed:exposed-jdbc:$exposedVersion")
     implementation("org.jetbrains.exposed:exposed-java-time:$exposedVersion")
+    implementation("net.postgis:postgis-jdbc:2.5.0")
     implementation("org.postgresql:postgresql:42.5.1")
+
+    // Database Migrations
+    implementation("org.flywaydb:flyway-core:9.15.1")
+
     implementation("com.kohlschutter.junixsocket:junixsocket-core:2.6.1")
     implementation("com.kohlschutter.junixsocket:junixsocket-common:2.6.1")
-
-    implementation("net.postgis:postgis-jdbc:2.5.0")
 
     implementation("io.ktor:ktor-client-core:2.1.2")
     implementation("io.ktor:ktor-client-cio:2.1.2")
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/database/Database.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/database/Database.kt
@@ -3,6 +3,7 @@ package app.ehrenamtskarte.backend.common.database
 import app.ehrenamtskarte.backend.auth.database.repos.AdministratorsRepository
 import app.ehrenamtskarte.backend.auth.webservice.schema.types.Role
 import app.ehrenamtskarte.backend.config.BackendConfiguration
+import org.flywaydb.core.Flyway
 import org.jetbrains.exposed.sql.Database.Companion.connect
 import org.jetbrains.exposed.sql.StdOutSqlLogger
 import org.jetbrains.exposed.sql.addLogger
@@ -66,6 +67,17 @@ class Database {
                 setupDatabaseForApplication()
                 setupDatabaseForAuth()
             }
+
+            migrate(config)
+        }
+
+        fun migrate(config: BackendConfiguration) {
+            val flyway = Flyway.configure()
+                .baselineOnMigrate(true) // remove after first production release
+                .dataSource(config.postgres.url, config.postgres.user, config.postgres.password)
+                .locations("classpath:db/migrations")
+                .load()
+            flyway.migrate()
         }
     }
 }


michael-markl:
I feel like I cannot really review this right now as I don't have much time and not a lot of expertise on this matter... There's probably some more considerations to take regarding DB migrations:
* Proper documentation of the workflow
* When do we do DB migrations? At the start of every backend run? Manually using another CLI command?
* How do we make sure, the backend runs in sync with the DB schema? (Migrating on start would work; Alternatively:  On backend start: Verify that all migrations found in the backend code have been applied)
* Backup / Recovery strategy: Before every DB migration, we should probably do a backup, try to migrate and verify everything worked out as expected. If not, roll back.

Maybe the CMS team has more insights how to handle / setup DB migrations?

backend/src/main/resources/db/migrations/V1__remove_old_email_index.sql
```diff
@@ -0,0 +1 @@
+alter table administrators drop constraint administrators_email_unique;
```
timoludwig, line 1: It might be different for this specific setup, but in all migration frameworks I know the initial migration starts with an empty database and creates all tables and columns to allow a deterministic outcome (or is this done as part of the `setupDatabase*` functions?). Otherwise, this might even throw an error on a new setup where the `administrators_email_unique` constraint doesn't exist yet?

Of course this would normally cause an error on all systems which already have initialized the database, so they have to be [baselined](https://documentation.red-gate.com/fd/baseline-184127456.html) (or the new initial migration would have to be applied with the [skipExecutingMigrations](https://documentation.red-gate.com/fd/skip-executing-migrations-184127537.html)) option to tell flyway that it should mark the initial migration as applied without really changing anything.
This file would then be v2 and could be applied on top of the initial one.

timoludwig (commented):
> > I feel like I cannot really review this right now as I don't have much time and not a lot of expertise on this matter...
> 
> Same... I already worked with migrations, but not in any critical environment.
> 

Django takes care of most critical issues completely automatically, so with my comments I'm just throwing another unqualified opinion into the ring :sweat_smile: 

> > 
> > * Proper documentation of the workflow
> 
> I also thought of this. especially when changing branches and stuff the database might not match the current state.

For our CMS, Django provides an easy way of undoing migrations in such cases... if I understood correctly, flyway [only supports this](https://documentation.red-gate.com/fd/undo-184127463.html) as part of their "Teams" plan, and also since the migrations are not autogenerated, you have to manually write an "undo" migration for every migration you write, so nearly twice the work? If you don't need to persist data between branch changes, dropping the db and re-creating it with the new schema could be sufficient for the development workflow? Honestly, this is also what we're doing most of the time, since re-creating is faster than figuring out which migrations to undo :smile: 

> > * When do we do DB migrations? At the start of every backend run? Manually using another CLI command?
> 
> It's possible to create a gradle task for this. But for production I think it is better to execute it every time we start the backend. Already applied migrations will not be reapplied.

Depending on how often the server restarts, this could be a bit overkill... I mean it's really only required when the code base changes, so on every new deployment? Since it's automated anyway, maybe just add the migration step to the [autoupdate](https://git.tuerantuer.org/DF/salt/src/branch/master/states/entitlementcard/eak-autoupdate) cronjob?

> > * How do we make sure, the backend runs in sync with the DB schema? (Migrating on start would work; Alternatively:  On backend start: Verify that all migrations found in the backend code have been applied)

Maybe the [validate](https://documentation.red-gate.com/fd/validate-184127464.html) command can be used on every server start?
On the other hand, if the deployment and migration steps are fully automated, there isn't much potential for errors anymore (we should only set up notifications in case the migration itself fails)...

> > * Backup / Recovery strategy: Before every DB migration, we should probably do a backup, try to migrate and verify everything worked out as expected. If not, roll back.
> 
> The framework tries to apply the migrations in a single transaction. If anything fails, the database will roll back. Regardless I think backing up the db becomes very important.

[Database backups](https://git.tuerantuer.org/DF/salt/src/branch/master/pillars/backup/org/tuerantuer/entitlementcard.sls) seem to exist on our server infra? But restoring them should probably be tested at least once in a while to make sure we didn't miss anything important.

michael-markl:
On another note:
Maybe we can use exposed's [statementsRequiredToActualizeScheme](https://github.com/JetBrains/Exposed/blob/master/exposed-core/src/main/kotlin/org/jetbrains/exposed/sql/SchemaUtils.kt#L353) (or related methods in this file) to validate that our backend with the schema defined in Exposed actually run against a DB that supports that schema?

backend/src/main/resources/db/migrations/V1__remove_old_email_index.sql
sarahsporck: Jep you're completely right. It must be at least V2 to be properly applied. Tested that again today. The setupDatabase* functions will create all the tables. Correct. But as this is coupled into the kotlin code and nondeterministic, meaning there can be a different outcome when applying the setupDatabase function e.g. initially and on an already existing Database, As the tables are created in kotlin directly, and columns and constraints are added automatically when we add them in java. See [jetbrains exposed](https://github.com/JetBrains/Exposed).

sarahsporck:
So we could also consider using a simple native implementation for migrations. In that case we could take the following project as a basis: https://gitlab.com/andreas-mausch/exposed-migrations/-/tree/master/
I'll also try implementing this "simple" implementation in the following weeks. then we can compare.

f1sh1918 (approved):
i think the V1 file should be removed since this was just an example and we start with V2, no?
I feel fine with this solution and tested some sql and kotlin migrations on it

backend/src/main/resources/db/migrations/V1__remove_old_email_index.sql
f1sh1918, line 1:
```suggestion
alter table administrators drop constraint if exists administrators_email_unique;
```

michael-markl: @timoludwig If you want, you can also add your opinion to #906 :D

sarahsporck:
I feel like the custom approach comes with higher flexibility and serves the same features as flyway. So I'd go with that solution and close this PR.

PR #845, opened 2023-03-03
Small backend fixes

michael-markl:
This PR fixes a few small annoyances
* Order regions and admins in queries
* Start webservice only once all handlers are configured (helpful if their configuration errors)
* Log all SQL queries in dev mode


backend/src/main/kotlin/app/ehrenamtskarte/backend/common/database/Database.kt
```diff
@@ -51,14 +51,16 @@ class Database {
...
+                } else {
+                    DatabaseConfig.invoke {
+                        this.sqlLogger = StdOutSqlLogger
+                    }
+                },
             )
-
             transaction {
-                if (!config.production) {
-                    addLogger(StdOutSqlLogger)
```
maxammann, line 59: As this is removed, how are queries logged?

maxammann (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/database/Database.kt
michael-markl: I removed logging for this specific transaction, but added logging for all transactions by adding to the DatabaseConfig

Issue #814, opened 2023-02-23
Indicate no applications in UI


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Display a text that there are currently no applications, becaus an empty white screen could also indicate an error.

michael-markl:
Closed by #846

PR #846, opened 2023-03-03
Add application verifications to administration

### LINKED ISSUE #835: Reflect the application verification state in administration frontend
An application can have multiple states depending on whether referenced applications already have verified an application. 

As a Mitarbeiter in der Kommune I would like to see whether an organization already verified (or rejected) an application. 

For now a very simple icons for ❌ (rejected), ✅ (accepted) and 🕐 (pending) should be fine. #586 will then improve on this presentation.

More improved version of this is discussed in #586, which can be implemented after the release.

**Additional context**

Currently blocked by #834

michael-markl:
@sarahsporck Tagging you (in case you would be running into conflicts with this)

Closes #835 and #814


backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/dataloader/ApplicationDataLoader.kt
```diff
@@ -0,0 +1,64 @@
...
+            transaction {
+                ApplicationRepository.findByIds(ids).map { it?.let { ApplicationView.fromDbEntity(it) } }
+            }
+        }
+    }
+}
+
+const val VERIFICATIONS_BY_APPLICATION_LOADER_NAME = "VERIFICATIONS_BY_APPLICATION_LOADER"
+val verificationsByApplicationLoader: DataLoader<Int, List<ApplicationVerificationView>?> =
+    DataLoaderFactory.newDataLoader { ids ->
```
sarahsporck, line 32: > As [BatchLoader](https://javadoc.io/static/com.graphql-java/java-dataloader/2.2.3/org/dataloader/BatchLoader.html) batch functions are executed the resulting futures are cached using a cache implementation of choice, so they will only execute once. Individual cache keys can be cleared, so they will be re-fetched when referred to again.

We probably need to clear the cached result, when an application verification is accepted or rejected. When testing the icons by directly editing the database entry, the query will still return the old value until I restarted the backend service.

michael-markl: > We probably need to clear the cached result, when an application verification is accepted or rejected. When testing the icons by directly editing the database entry, the query will still return the old value until I restarted the backend service.

Agreed. The question is: should we (A) disable the cache entirely for all/specific dataloaders, or should we (B) invalidate their caches for specific ids when we expect a change (the "faster" option)?
(A) is the "safer" option because we can't forget invalidating ids, but
(B) is the "faster" option, because we don't need to query unchanged data.

sarahsporck: I feel like (A) should be the right answer, as the UI should reflect the state of the DB at some point. In that regard I didn't find anything about the cache expiring after any time. But I am also unsure of how many Applications we might be talking about here. 🤔

michael-markl: I created a different task for that (as it will affect more code than I want to touch this PR): https://github.com/digitalfabrik/entitlementcard/issues/851
We should nevertheless implement this before Release in Bayern.

sarahsporck (approved):

PR #853, opened 2023-03-08
851: Refactor data loaders and disable caching

### LINKED ISSUE #851: Disable Caching on DataLoaders
See [here](https://github.com/digitalfabrik/entitlementcard/pull/846#discussion_r1128093841) for a discussion why we should disable caching on dataloaders in the backend for now.

michael-markl:
Closes #851. Refactoring of the data loaders:
* Disable caching on all data loaders
* Make registering and naming of data loaders less boilerplaty and less error-prone
* Avoid name conflicts of data loaders
* Make them more type safe

Also, I rewrote the data loader for physicalStoreByStoreIdLoader which makes the `searchAcceptingStores` GraphQL query  a lot faster: Instead of ~20 single SQL queries (one for each store) we now make a single SQL query.

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/webservice/schema/types/AcceptingStore.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/webservice/schema/types/AcceptingStore.kt
@@ -1,25 +1,27 @@
 package app.ehrenamtskarte.backend.stores.webservice.schema.types
 
-import app.ehrenamtskarte.backend.stores.webservice.dataloader.CATEGORY_LOADER_NAME
-import app.ehrenamtskarte.backend.stores.webservice.dataloader.CONTACT_LOADER_NAME
-import app.ehrenamtskarte.backend.stores.webservice.dataloader.PHYSICAL_STORE_BY_STORE_ID_LOADER_NAME
+import app.ehrenamtskarte.backend.common.webservice.fromEnvironment
+import app.ehrenamtskarte.backend.stores.webservice.dataloader.categoryLoader
+import app.ehrenamtskarte.backend.stores.webservice.dataloader.contactLoader
+import app.ehrenamtskarte.backend.stores.webservice.dataloader.physicalStoreByStoreIdLoader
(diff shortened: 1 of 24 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/webservice/schema/types/AcceptingStore.kt
```diff
@@ -1,25 +1,27 @@
...
 
-    fun category(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<Category> =
-        dataFetchingEnvironment.getDataLoader<Int, Category>(CATEGORY_LOADER_NAME).load(categoryId)
+    fun category(environment: DataFetchingEnvironment): CompletableFuture<Category> =
+        categoryLoader.fromEnvironment(environment).load(categoryId).thenApply { it!! }
 
-    fun physicalStore(dataFetchingEnvironment: DataFetchingEnvironment): CompletableFuture<PhysicalStore?> =
-        dataFetchingEnvironment.getDataLoader<Int, PhysicalStore?>(PHYSICAL_STORE_BY_STORE_ID_LOADER_NAME).load(id)
+    fun physicalStore(environment: DataFetchingEnvironment): CompletableFuture<PhysicalStore?> =
+        physicalStoreByStoreIdLoader.fromEnvironment(environment).load(id)
```
sarahsporck, line 26: It seems you forgot sth. here.
```suggestion
        physicalStoreByStoreIdLoader.fromEnvironment(environment).load(id).thenApply { it!! }
```

sarahsporck (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/webservice/schema/types/AcceptingStore.kt
michael-markl: Actually not: We designed our API to support AcceptingStores that do not have a PhyiscalStore, (so this is a 0..1 connection). Hence, the `physicalStore` function here should return a nullable value.

Issue #854, opened 2023-03-09
URI not hierarchical - region file can't be loaded


f1sh1918:
**Describe the bug**
When we try to load the region file inside the java bundle not in the IDE it's failing the the error below
Maybe connected to this issue: https://stackoverflow.com/questions/10144210/java-jar-file-use-resource-errors-uri-is-not-hierarchical

**To Reproduce**
1. Start the backend on server

**Expected behavior**
CSV file gets loaded while backend starts

**Additional context**
```js
backend@entitlementcard:~$ /opt/ehrenamtskarte/backend/bin/backend execute
[main] INFO app.ehrenamtskarte.backend.config.BackendConfiguration - Loading backend configuration from /etc/entitlementcard/config.yml.
Exception in thread "main" java.lang.Exception: Couldn't read CSV
    at app.ehrenamtskarte.backend.regions.utils.PostalCodesLoader.loadRegionIdentifierByPostalCodeMap(PostalCodesLoader.kt:21)
    at app.ehrenamtskarte.backend.common.webservice.GraphQLHandler.<init>(GraphQLHandler.kt:36)
    at app.ehrenamtskarte.backend.common.webservice.WebService.start(WebService.kt:50)
    at app.ehrenamtskarte.backend.Execute.run(EntryPoint.kt:114)
    at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:198)
    at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:211)
    at com.github.ajalt.clikt.parsers.Parser.parse(Parser.kt:18)
    at com.github.ajalt.clikt.core.CliktCommand.parse(CliktCommand.kt:400)
    at com.github.ajalt.clikt.core.CliktCommand.parse$default(CliktCommand.kt:397)
    at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:415)
    at com.github.ajalt.clikt.core.CliktCommand.main(CliktCommand.kt:440)
    at app.ehrenamtskarte.backend.EntryPointKt.main(EntryPoint.kt:119)
Caused by: java.lang.IllegalArgumentException: URI is not hierarchical
    at java.base/java.io.File.<init>(File.java:420)
    at app.ehrenamtskarte.backend.regions.utils.PostalCodesLoader.loadRegionIdentifierByPostalCodeMap(PostalCodesLoader.kt:10)
    ... 11 more
[...]

PR #855, opened 2023-03-09
854: Fix uri not hierarchical

f1sh1918:
Looks like you can not use `File()` for internal resources instead we have to use `InputStream`
https://coderanch.com/t/750338/java/URI-hierarchical-DropBox-Upload-Code -> second post
resolves https://github.com/digitalfabrik/entitlementcard/issues/854


michael-markl:
Maybe, for testing this, you can try and do the same bundling/packing that the CI does? (I *think* it does the bundling)

f1sh1918:
> Maybe, for testing this, you can try and do the same bundling/packing that the CI does? (I _think_ it does the bundling)

@michael-markl i found finally some articles about the cause of this problem so i'm quite sure it should be solved now

maxammann (approved):
Looks good. Not tested though. Merging this now though so the next deployment does not update the backend.

PR #857, opened 2023-03-10
837: verification view for applications

sarahsporck:
I think technically the main functionality has been implemented. 
What's still missing is 
- [x] authentication with the accessKey and 
- [x] feedback for successfully submitting the form,

which I'd prefer to implement in separate PRs, as this one already feels quite bloated.

[EDIT] It feels like forever that I've implemented some UI. Feel free to be harsh 😀

### DIFF
--- a/administration/src/application-verification/VerificationController.tsx
+++ b/administration/src/application-verification/VerificationController.tsx
@@ -1,0 +1,151 @@
+import { Button, Card, H1, H4, NonIdealState, Spinner } from '@blueprintjs/core'
+import React, { useContext, useState } from 'react'
+import { useParams } from 'react-router-dom'
+import styled from 'styled-components'
+import JsonFieldView from '../components/applications/JsonFieldView'
+import ErrorHandler from '../ErrorHandler'
+import {
+  useGetApplicationByApplicationVerificationAccessKeyQuery,
+  useVerifyOrRejectApplicationVerificationMutation,
+} from '../generated/graphql'
+import { ProjectConfigContext } from '../project-configs/ProjectConfigContext'
+import { useAppToaster } from '../components/AppToaster'
+import { CARD_PADDING } from '../components/applications/ApplicationsOverview'
+import { format } from 'date-fns'
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 100%;
+  max-width: 800px;
+  margin: auto;
+  justify-content: center;
+  padding: ${CARD_PADDING}px;
+`
+
+const CenteredMessage = styled(NonIdealState)`
+  margin: auto;
+`
+
+const ButtonContainer = styled.div`
+  display: flex;
+  width: inherit;
+  flex-direction: row;
+  justify-content: space-around;
+`
+
+const Text = styled.p`
+  margin: 20px 0 10px 0;
+`
+
+const Info = styled.aside`
+  background-color: #f8f8f8;
+  width: inherit;
+  border-left: solid 3px;
+  border-color: gray;
+  padding: 5px;
+  margin: 10px;
+`
+
+type ApplicationVerificationProps = {
+  applicationVerificationAccessKey: string
+}
+
+const ApplicationVerification = ({ applicationVerificationAccessKey }: ApplicationVerificationProps) => {
+  const [verificationFinised, setVerificationFinished] = useState(false)
+  const config = useContext(ProjectConfigContext)
+  const appToaster = useAppToaster()
+  const showErrorToaster = () => appToaster?.show({ intent: 'danger', message: 'Etwas ist schief gelaufen.' })
+  const [verifyOrRejectApplicationVerification, { loading: verificationLoading }] =
+    useVerifyOrRejectApplicationVerificationMutation({
+      onError: error => {
+        console.error(error)
+        showErrorToaster()
+      },
+      onCompleted: ({ result }, clientOptions) => {
+        if (!result) {
+          console.error('Verify operation returned false.')
+          showErrorToaster()
+        } else {
+          setVerificationFinished(true)
+        }
+      },
+    })
+
+  const submitApplicationVerification = (verified: boolean) => {
+    verifyOrRejectApplicationVerification({
+      variables: {
+        accessKey: applicationVerificationAccessKey,
+        verified,
+      },
+    })
+  }
+
+  const { loading, error, data, refetch } = useGetApplicationByApplicationVerificationAccessKeyQuery({
+    variables: { applicationVerificationAccessKey },
+  })
+
+  if (loading) return <Spinner />
+  else if (error || !data) return <ErrorHandler refetch={refetch} />
+  if (data.verification.rejectedDate || data.verification.verifiedDate)
+    return <CenteredMessage title='Sie haben diesen Antrag bereits bearbeitet.' />
+  if (verificationFinised)
+    return (
+      <CenteredMessage
+        title={'Ihre Eingaben wurden erfolgreich gespeichert.'}
+        description='Vielen Dank für Ihre Mithilfe. Sie können das Fenster jetzt schließen.'
+      />
+    )
+
+  const { jsonValue, createdDate: createdDateString, id } = data.application
+  const createdDate = new Date(createdDateString)
+  const jsonField = JSON.parse(jsonValue)
+  const baseUrl = `${process.env.REACT_APP_API_BASE_URL}/application/${config.projectId}/${id}`
+  return (
+    <Container>
+      <H1>{config.name}</H1>
+      <Text>
+        Guten Tag {data.verification.contactName},
+        <br />
+        <br />
+        Sie wurden gebeten, die Angaben eines Antrags auf Ehrenamtskarte zu bestätigen. Die Antragsstellerin oder der
+        Antragssteller hat Sie als Kontaktperson der Organisation {data.verification.organizationName} angegeben. Im
+        Folgenden können Sie den zugehörigen Antrag einsehen. Wir bitten Sie, die enthaltenen Angaben, welche die
+        Organisation {data.verification.organizationName} betreffen, zu bestätigen. Falls Sie denken, die Angaben wurden
+        fälschlicherweise gemacht, bitten wir Sie, den Angaben zu widersprechen.
+      </Text>
+      <Card>
+        <H4>Antrag vom {format(createdDate, 'dd.MM.yyyy, HH:mm')}</H4>
+        <JsonFieldView jsonField={jsonField} baseUrl={baseUrl} hierarchyIndex={0} />
+      </Card>
+      <Text>
+        Können Sie die Angaben, welche die Organisation <b>{data.verification.organizationName}</b> betreffen,
+        bestätigen?
+      </Text>
+      <Info>
+        <b>Wichtig!</b> Die Eingabe kann nicht rückgängig gemacht werden.
+      </Info>
+      <ButtonContainer>
+        <Button icon='cross' loading={verificationLoading} onClick={() => submitApplicationVerification(false)}>
+          Widersprechen
+        </Button>
+        <Button icon='tick' loading={verificationLoading} onClick={() => submitApplicationVerification(true)}>
+          Bestätigen
+        </Button>
+      </ButtonContainer>
+    </Container>
+  )
+}
+
+const ApplicationVerificationController = () => {
+  const { applicationVerificationAccessKey } = useParams()
+
+  if (!applicationVerificationAccessKey) {
+    return <CenteredMessage title='Nicht gefunden' description='Die Seite konnte nicht gefunden werden.' />
+  }
+
+  return <ApplicationVerification applicationVerificationAccessKey={applicationVerificationAccessKey} />
+}
+
+export default ApplicationVerificationController
--- a/administration/src/graphql/applications/getApplication.graphql
+++ b/administration/src/graphql/applications/getApplication.graphql
@@ -1,0 +1,15 @@
+query getApplicationByApplicationVerificationAccessKey($applicationVerificationAccessKey: String!) {
+  application: getApplicationByApplicationVerificationAccessKey(
+    applicationVerificationAccessKey: $applicationVerificationAccessKey
+  ) {
+    id
+    createdDate
+    jsonValue
+  }
+  verification: getApplicationVerification(accessKey: $applicationVerificationAccessKey) {
+    contactName
+    organizationName
+    verifiedDate
+    rejectedDate
+  }
+}
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
@@ -25,6 +25,7 @@ import org.jetbrains.exposed.sql.transactions.transaction
 import java.io.File
 import java.nio.file.Paths
 import java.security.SecureRandom
+import java.time.LocalDateTime
 import java.util.Base64
 
 object ApplicationRepository {
@@ -92,12 +93,62 @@ object ApplicationRepository {
         }
     }
 
+    fun getApplicationByApplicationVerificationAccessKey(applicationVerificationAccessKey: String): ApplicationView {
+        return transaction {
+            (Applications innerJoin ApplicationVerifications)
+                .slice(Applications.columns)
+                .select { ApplicationVerifications.accessKey eq applicationVerificationAccessKey }
+                .single()
+                .let {
+                    ApplicationView.fromDbEntity(ApplicationEntity.wrapRow(it))
+                }
+        }
+    }
+
+    fun getApplicationVerification(accessKey: String): ApplicationVerificationEntity {
+        return transaction {
+            ApplicationVerificationEntity.find { ApplicationVerifications.accessKey eq accessKey }
+                .single()
+        }
+    }
+
+    private fun isAlreadyVerified(applicationVerification: ApplicationVerificationEntity): Boolean {
+        return applicationVerification.verifiedDate != null || applicationVerification.rejectedDate != null
+    }
+
+    fun verifyApplicationVerification(accessKey: String): Boolean {
+        return transaction {
+            val applicationVerification = getApplicationVerification(accessKey)
+            if (isAlreadyVerified(applicationVerification)) {
+                false
+            } else {
+                applicationVerification.verifiedDate = LocalDateTime.now()
+                true
+            }
+        }
+    }
+
+    fun rejectApplicationVerification(accessKey: String): Boolean {
+        return transaction {
+            val applicationVerification = getApplicationVerification(accessKey)
+            if (isAlreadyVerified(applicationVerification)) {
+                false
+            } else {
+                applicationVerification.rejectedDate = LocalDateTime.now()
+                true
+            }
+        }
+    }
+
     fun delete(applicationId: Int, graphQLContext: GraphQLContext): Boolean {
         return transaction {
             val application = ApplicationEntity.findById(applicationId)
             if (application != null) {
-                val project = (Projects innerJoin Regions).select { Regions.id eq application.regionId }.single()
-                    .let { ProjectEntity.wrapRow(it) }
+                val project =
+                    (Projects innerJoin Regions)
+                        .select { Regions.id eq application.regionId }
+                        .single()
+                        .let { ProjectEntity.wrapRow(it) }
                 val applicationDirectory =
                     Paths.get(graphQLContext.applicationData.absolutePath, project.project, application.id.toString())
                 ApplicationVerifications.deleteWhere { ApplicationVerifications.applicationId eq applicationId }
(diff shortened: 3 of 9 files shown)


sarahsporck:
1. create an application
2. go to localhost:3000/antrag-verifizieren/{id}?accessKey={accessKey} and get the id and access key from the database

f1sh1918:
> 2\. antrag-verifizieren

@sarahsporck i tried this f.e. and i get "Ein Fehler ist aufgetreten"
http://localhost:3000/antrag-verifizieren/6?accessKey=aGmV8Wv2w7QcZ2_oJJjz_92yH-IVY0_Y1fFB_5qFUeX4XaYNKBEZ6HYxsqkOcXVOTzzsMAQBIwdrIUWEHaiqJQ==

administration/src/components/NotFoundView.tsx
```diff
@@ -0,0 +1,13 @@
+import { NonIdealState } from '@blueprintjs/core'
+import styled from 'styled-components'
+
+const NotFound = styled(NonIdealState)`
+  margin: auto;
+`
+const NotFoundView = () => {
```
f1sh1918, line 7: you can use shorthand
```js
const NotFoundView = () => <NotFound title='Nicht gefunden' description='Diese Seite konnte nicht gefunden werden.' />
```

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,110 @@
...
+  margin: 10px;
+`
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 100%;
+  justify-content: center;
+  padding: ${CARD_PADDING}px;
```
f1sh1918, line 25: if you just use it once, i think you can just use
```suggestion
  padding: 20px;
```

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,110 @@
+import { Button, Card, H1, Spinner } from '@blueprintjs/core'
```
f1sh1918, line 1: i think everything connected to the application process should be responsible and use material components @michael-markl ?

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const appToaster = useAppToaster()
+  const showErrorToaster = () => appToaster?.show({ intent: 'danger', message: 'Etwas ist schief gelaufen.' })
+  const [verifyApplicationVerification, { loading: verificationLoading }] = useVerifyApplicationVerificationMutation({
+    onError: error => {
+      console.error(error)
+      showErrorToaster()
+    },
+  })
+
+  const submitApplicationVerification = useCallback(
```
f1sh1918, line 51: if you don't use the fkt in a lifecycle hook i think you can spare out the `useCallback`

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const { loading, error, data, refetch } = useGetApplicationQuery({
+    variables: { applicationVerificationId, accessKey: applicationAccessKey },
+  })
+
+  if (loading) return <Spinner />
+  else if (error || !data) return <ErrorHandler refetch={refetch} />
+
+  const { jsonValue, id, verifications } = data.application
+  const verification = verifications.length >= 0 ? verifications[0] : null
```
f1sh1918, line 71:
```suggestion
  const verification = verifications.length > 0 ? verifications[0] : null
```
i think the length shouldn't be 0, no? I even dunno why it can have multiple verifacations? :(

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const jsonField = JSON.parse(jsonValue)
+  const baseUrl = `${process.env.REACT_APP_API_BASE_URL}/application/${config.projectId}/${id}`
+  return (
+    <Container>
+      <H1>{config.name}</H1>
+      <span>
+        Können Sie die Angaben, die die Organisation <b>{verification?.organizationName}</b> betreffen bestätigen?
```
f1sh1918, line 78: should there be a check
```js
if(!verification) {
return null
}
```
I mean if there is no organization to verify who should verify then? ;)

administration/src/components/NotFoundView.tsx
```diff
@@ -0,0 +1,13 @@
...
+  return (
+      <NotFound title='Nicht gefunden' description='Diese Seite konnte nicht gefunden werden.' />
```
f1sh1918, line 9: i think the message should be a little bit more specific .f.e "Es konnte kein Antrag gefunden werden". If you create a separate component which means it can be reused i would tend to pass title and description or just use the `<NotFound>` directly in `VerificationController`

f1sh1918 (commented):
i think some minor improvements tbd and the attachment download is not working (can also be solved in separate task)

administration/src/components/verification/VerificationController.tsx
sarahsporck: I tried to stay consistent with the already existing layout, and therefore copied the constant. On second thought it makes much more sense to export & import the value than to copy it.

administration/src/components/verification/VerificationController.tsx
sarahsporck: Jep. it is possible and a typo as well. Further, I now added, that the list is filtered for the organisation, that opened the link to make it more clear.

administration/src/components/verification/VerificationController.tsx
sarahsporck: The verification should always exist, as otherwise the application could not be loaded. However I could still add the check if you insist :)

michael-markl (commented):
... I think I made a mistake:
We should probably not make the internal verification ids public. Thus, the access key was designed (at least from my side^^) to be both a unique access key (for retrieving the application) and an authentication key (that is only known to the organization).
Hence, the URL would be `https://asdf/antrag-verifizieren/{access-key}` (alternatively with a query param), leaving out the verification id. Do you think that would still be reasonable?

administration/src/components/verification/VerificationController.tsx
michael-markl: @f1sh1918 : In the Antrag, you can specify multiple organizations that you have been volunteering at (over the last years). That's why multiple organizations can verify the application.

f1sh1918: ok but if we have multiple organization, every organization has to receive an email and you have to specify the organization by some identifier maybe also in the link and resolve it here instead of using `verifications[0]`

sarahsporck: this snippet is already outdated. see the updated code :)

sarahsporck:
> We should probably not make the internal verification ids public. Thus, the access key was designed (at least from my side^^) to be both a unique access key (for retrieving the application) and an authentication key (that is only known to the organization).

But, although very unlikely the access key actually is not unique, isn't it? It is just a random byte string that is then base64 encrypted. Besides from that I don't see any pain points here.

michael-markl:
> But, although very unlikely the access key actually is not unique, isn't it? It is just a random byte string that is then base64 encrypted. Besides from that I don't see any pain points here.

Afaik, the DB column has a uniqueness constraint. Maybe, it just throws an error when trying to persist the application in this very very unlikely case. Not sure, if it's worth doing anything about it.

administration/src/graphql/applications/getApplication.graphql
```diff
@@ -0,0 +1,14 @@
+query getApplication($applicationVerificationId: Int!, $accessKey: String!) {
```
f1sh1918, line 1: Maybe `getApplicationVerification` would be a better naming, since we need a `getApplication` query for the user

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationQueryService.kt
```diff
@@ -30,4 +30,16 @@ class EakApplicationQueryService {
...
         }
     }
+
+    @GraphQLDescription("Queries an application by application verificationId")
+    fun getApplication(
+        dfe: DataFetchingEnvironment,
+        applicationVerificationId: Int,
+        accessKey: String
+    ): ApplicationView {
+        val context = dfe.getContext<GraphQLContext>()
```
f1sh1918, line 40: context is not needed

administration/src/components/NotFoundView.tsx
sarahsporck: I moved this one into the VerificationController for now.

f1sh1918 (approved):
Tested works fine :)

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -63,4 +63,18 @@ class EakApplicationMutationService {
             ApplicationRepository.delete(applicationId, context)
         }
     }
+
+    @GraphQLDescription("Verifies an application verification")
+    fun verifyApplicationVerification(
```
michael-markl, line 68: I feel like we should adjust the naming... But I don't really have a better one off the top of my head :D
... Maybe `VerifyOrRejectApplication` ?

backend/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ plugins {
     // Apply the Kotlin JVM plugin to add support for Kotlin.
     id("org.jetbrains.kotlin.jvm") version "1.8.0"
     id("org.jlleitschuh.gradle.ktlint") version "11.0.0"
+    id("org.jetbrains.kotlin.android") version "1.8.0" apply false
```
michael-markl, line 16: What does that do? :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -92,12 +94,62 @@ object ApplicationRepository {
...
+                }
+        }
+    }
+
+    fun verifyApplicationVerification(accessKey: String): Boolean {
+        return transaction {
+            val applicationVerification = ApplicationVerificationEntity.find { ApplicationVerifications.accessKey eq accessKey }.single()
+            if (applicationVerification != null) {
+                applicationVerification.rejectedDate = null
+                applicationVerification.verifiedDate = LocalDateTime.now()
```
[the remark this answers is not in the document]
sarahsporck: In the frontend we currently do not allow to verify an application twice. So we could also disallow it in the backend. So I'll remove setting null

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -92,12 +94,62 @@ object ApplicationRepository {
...
+                .let {
+                    ApplicationVerificationView.fromDbEntity(it)
+                }
+        }
+    }
+
+    fun verifyApplicationVerification(accessKey: String): Boolean {
+        return transaction {
+            val applicationVerification = ApplicationVerificationEntity.find { ApplicationVerifications.accessKey eq accessKey }.single()
+            if (applicationVerification != null) {
```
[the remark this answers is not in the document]
sarahsporck: you're right. I tried copying how we used single at other places. 🤔

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
sarahsporck: I was also struggling with this and considered creating two separate endpoints. But it seems like an unnecessary overhead. I'll use your proposition.

backend/build.gradle.kts
sarahsporck: I tried resolving my issue with `Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.8.0, expected version is 1.6.0.`. But this only helped in some places, so I'll remove it again.

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const { jsonValue, id } = data.application
+  const jsonField = JSON.parse(jsonValue)
+  const baseUrl = `${process.env.REACT_APP_API_BASE_URL}/application/${config.projectId}/${id}`
+  return (
+    <Container>
+      <H1>{config.name}</H1>
+      <span>
+        Können Sie die Angaben, welche die Organisation <b>{data.verification.organizationName}</b> betreffen
+        bestätigen?
```
michael-markl, line 97: I think some more context is missing:
Proposal:

Guten Tag {data.verification.contactName},

Sie wurden gebeten, die Angaben eines Antrags auf Ehrenamtskarte zu bestätigen.
Die Antragsstellerin oder der Antragssteller hat Sie als Kontaktperson der Organisation {data.verification.organizationName} angegeben.
Im Folgenden können Sie den zugehörigen Antrag einsehen.
Wir bitten Sie, die enthaltenen Angaben, welche die Organisation {data.verification.organizationName} betreffen, zu bestätigen. Falls Sie denken, die Angaben wurden fälschlicherweise gemacht, bitten wir Sie, den Angaben zu widersprechen.

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,124 @@
...
+      </span>
+      <ApplicationViewCard>
+        <JsonFieldView jsonField={jsonField} baseUrl={baseUrl} hierarchyIndex={0} />
+      </ApplicationViewCard>
+      <ButtonContainer>
```
michael-markl, line 102: Maybe, we should repeat the question here:
`Können Sie die Angaben, welche die Organisation <b>{data.verification.organizationName}</b> betreffen, bestätigen?`

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,124 @@
...
+    )
+
+  const { jsonValue, id } = data.application
+  const jsonField = JSON.parse(jsonValue)
+  const baseUrl = `${process.env.REACT_APP_API_BASE_URL}/application/${config.projectId}/${id}`
+  return (
+    <Container>
+      <H1>{config.name}</H1>
+      <span>
+        Können Sie die Angaben, welche die Organisation <b>{data.verification.organizationName}</b> betreffen
```
michael-markl, line 96:
```suggestion
        Können Sie die Angaben, welche die Organisation <b>{data.verification.organizationName}</b> betreffen,
```

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <ApplicationViewCard>
+        <JsonFieldView jsonField={jsonField} baseUrl={baseUrl} hierarchyIndex={0} />
+      </ApplicationViewCard>
+      <ButtonContainer>
+        <Button icon='cross' loading={verificationLoading} onClick={() => submitApplicationVerification(false)}>
+          Widersprechen
+        </Button>
+        <Button icon='tick' loading={verificationLoading} onClick={() => submitApplicationVerification(true)}>
+          Bestätigen
+        </Button>
```
michael-markl, line 108: Maybe, we should add an info box that tells the user that this choice cannot be changed?

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  if (loading) return <Spinner />
+  else if (error || !data) return <ErrorHandler refetch={refetch} />
+  if (data.verification.rejectedDate || data.verification.verifiedDate)
+    return <CenteredMessage title='Dieser Antrag wurde bereits bearbeitet' />
+  if (verified !== 'pending')
+    return (
+      <CenteredMessage
+        title={`Der Antrag wurde Erfolgreich ${verified === 'verified' ? 'verifiziert' : 'abgelehnt'}`}
+        description='Sie können das Fenster jetzt schließen.'
```
michael-markl, line 85:
```suggestion
        title={`Ihre Eingaben wurden erfolgreich gespeichert.`}
        description='Vielen Dank für Ihre Mithilfe. Sie können das Fenster jetzt schließen.'
```

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const jsonField = JSON.parse(jsonValue)
+  const baseUrl = `${process.env.REACT_APP_API_BASE_URL}/application/${config.projectId}/${id}`
+  return (
+    <Container>
+      <H1>{config.name}</H1>
+      <span>
+        Können Sie die Angaben, welche die Organisation <b>{data.verification.organizationName}</b> betreffen
+        bestätigen?
+      </span>
+      <ApplicationViewCard>
```
michael-markl, line 99: Can you maybe add a title inside the card with something like 
`Antrag vom 26.02.2023, 21:47` (just as in the administration view)?

michael-markl (approved):

administration/src/components/verification/VerificationController.tsx
```diff
@@ -0,0 +1,124 @@
+import { Button, Card, H1, NonIdealState, Spinner } from '@blueprintjs/core'
```
michael-markl, line 1: Although our file structure of the administration frontend is super messy (and should be refactored at some point), I think we should avoid putting the VerificationController into the `src/components` folder. Maybe, we should rather put it into the `src/application` folder as it's more related to the application stuff (and should use material ui instead of blueprintjs)? Alternatively into a new folder `src/application-verification` or so?

administration/src/components/verification/VerificationController.tsx
f1sh1918: hm i think this text should be part of the email. i think. Do you think we need it twice?

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -92,12 +93,50 @@ object ApplicationRepository {
...
+    fun getApplicationVerification(accessKey: String): ApplicationVerificationEntity {
+        return transaction {
+            ApplicationVerificationEntity.find { ApplicationVerifications.accessKey eq accessKey }
+                .single()
+        }
+    }
+
+    fun verifyApplicationVerification(accessKey: String): Boolean {
+        return transaction {
+            val applicationVerification = getApplicationVerification(accessKey)
```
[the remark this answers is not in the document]
sarahsporck: The error is throw once again by the database, due to the constraint. But, as always I'll add the check to ensure that this is also reflected in the code.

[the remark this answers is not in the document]
michael-markl: I think, here we should be even more strict and disallow verifying if already verified. (Because we also disallow it in the application frontend)

administration/src/components/verification/VerificationController.tsx
michael-markl: I think we should definitely add more context to the website as well.

administration/src/components/verification/VerificationController.tsx
sarahsporck: I went with a separate folder, as `src/application` already encompasses the whole application form.

administration/src/components/verification/VerificationController.tsx
sarahsporck: I think there should be a text to introduce what the application is all about. Preferably it should not be the same text as in the email. For now I used the proposed text by @michael-markl to finish this pr.

administration/src/graphql/applications/getApplication.graphql
sarahsporck: Naming was adjusted to `getApplicationByApplicationVerificationAccessKey`

administration/src/components/verification/VerificationController.tsx
sarahsporck: So not using a width for the Json View already improved the responsiveness a lot. However it is probably best to improve this by using mui components.

michael-markl: I agree... Also afaik, JsonView does not use blueprint a lot (I think only for the Attachment Download Button). We could probably reuse a lot of it if we switch to MUI.

administration/src/application-verification/VerificationController.tsx
```diff
@@ -0,0 +1,151 @@
...
+      <Card>
+        <H4>Antrag vom {format(createdDate, 'dd.MM.yyyy, HH:mm')}</H4>
+        <JsonFieldView jsonField={jsonField} baseUrl={baseUrl} hierarchyIndex={0} />
+      </Card>
+      <Text>
+        Können Sie die Angaben, welche die Organisation <b>{data.verification.organizationName}</b> betreffen,
+        bestätigen?
+      </Text>
+      <Info>
+        <b>Wichtig!</b> Die Eingabe kann nicht Rückgängig gemacht werden.
```
michael-markl, line 127:
```suggestion
        <b>Wichtig!</b> Die Eingabe kann nicht rückgängig gemacht werden.
```

PR #861, opened 2023-03-15
856: Add health check endpoint

sarahsporck:
Test with: `curl http://localhost:8000/health --verbose`

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/HealthHandler.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/HealthHandler.kt
@@ -1,0 +1,25 @@
+package app.ehrenamtskarte.backend.application.webservice
+
+import app.ehrenamtskarte.backend.config.BackendConfiguration
+import app.ehrenamtskarte.backend.projects.database.ProjectEntity
+import app.ehrenamtskarte.backend.projects.database.Projects
+import io.javalin.http.Context
+import org.jetbrains.exposed.exceptions.ExposedSQLException
+import org.jetbrains.exposed.sql.transactions.transaction
+
+class HealthHandler(private val config: BackendConfiguration) {
+    fun handle(ctx: Context) {
+        try {
+            ctx.status(200)
+            transaction {
+                val projectIds = config.projects.map { it.id }
+                val projects = ProjectEntity.find { Projects.project inList projectIds }
+                if (projects.empty()) {
+                    ctx.status(502)
+                }
+            }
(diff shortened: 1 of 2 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/HealthHandler.kt
```diff
@@ -0,0 +1,28 @@
...
+        try {
+            transaction {
+                val projectIds = config.projects.map { it.id }
+                val projects = ProjectEntity.find { Projects.project inList projectIds }
+                if (projects.empty()) {
+                    ctx.status(502)
+                }
+            }
+            ctx.status(200)
+        } catch (exception: ExposedSQLException) {
```
sarahsporck, line 24: It could also make sense to change this to `exception: Exception`, to increase the resilience of this endpoint.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/HealthHandler.kt
```diff
@@ -0,0 +1,28 @@
...
+import java.net.ConnectException
+
+class HealthHandler(private val config: BackendConfiguration) {
+    fun handle(ctx: Context) {
+        try {
+            transaction {
+                val projectIds = config.projects.map { it.id }
+                val projects = ProjectEntity.find { Projects.project inList projectIds }
+                if (projects.empty()) {
+                    ctx.status(502)
```
michael-markl, line 20: Should we return after this? Otherwise we might still send 200?

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/HealthHandler.kt
michael-markl: As you want. Probably any other exception will also return some status != 200.

michael-markl (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/HealthHandler.kt
sarahsporck: ah. you're right 👍🏼

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/HealthHandler.kt
sarahsporck: you're right. Otherwise it will return 500 Internal Server Error

Issue #696, opened 2023-01-10
Disable applications for Nürnberg (server-side)


maxammann:
**Is your feature request related to a problem? Please describe.**
Right now it is possible to create applications in nürnberg, even though nürnberg does not use the feature.

**Describe the solution you'd like**
Add a flag to the config of the backend to disable the feature.

**Describe alternatives you've considered**
None

**Additional context**
None

maxammann:
@michael-markl I created this issue. I think it makes sense, right?

related to #695

michael-markl:
Yes, thank you!

Issue #324, opened 2021-03-20
(Placeholder) Do not send verbose errors to frontend


maxammann:
Currently the backend sends the plain error objects to the frontend when an error occurs. This is not a good behavior, because it can leak internal data.
![image](https://user-images.githubusercontent.com/905221/111870258-f2133c00-8983-11eb-865d-66f1f7d4e312.png)

TODO: Add a solution.

Schedulaar:
@maxammann I think this is not a bug in production, is it?

maxammann:
Probably not, but I'd leave this task open until someone checked a crash like this again.

maxammann:
I just observed this again on the current deployment

michael-markl:
A solution to this issue should introduce a general concept of how to treat errors in the GraphQL backend (& interface)

maxammann:
Yes absolutely! Now that we know why these messages show up in the frontend this is a technical task.

maxammann:
Very much related to #659

sarahsporck:
Should be resolved due to #873, as all backend exceptions have their own error message now.

PR #870, opened 2023-03-16
836: Send verification emails

sarahsporck:
If one email cannot be sent (, e.g. because of invalid from address) then the application is not persisted.


backend/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
michael-markl, line 1: This file should be removed.

michael-markl (approved):
> If one email cannot be sent (, e.g. because of invalid from address) then the application is not persisted.

Hmm... Not sure if this is a good approach. I think it would be a good approach, if the user got a proper error message when trying to submit the application. Without a proper error message, the user would have to guess what the problem is...

sarahsporck:
> > If one email cannot be sent (, e.g. because of invalid from address) then the application is not persisted.
> 
> Hmm... Not sure if this is a good approach. I think it would be a good approach, if the user got a proper error message when trying to submit the application. Without a proper error message, the user would have to guess what the problem is...

I think we generally are really wonky with our error handling. Imo the graphql interface should always return sth like .
```json
{
    "status": <http_status_like>,
    "message": <maybe_optional_because_of_translations>,
    "result": <only_for_queries>
}
```

I'll open an issue for this. 
In this case though I think it is often logical to fail here. But you are right. A better error message should be displayed, although I'm unsure what to return in an error case.

michael-markl:
> I think we generally are really wonky with our error handling. Imo the graphql interface should always return sth like .

We are super wonky, we never introduced a concept for handling errors^^

Related: https://github.com/digitalfabrik/entitlementcard/issues/324

The alternative to your suggestion is to use the graphql philosophy and use/improve GraphQL errors: https://www.apollographql.com/docs/react/data/error-handling/

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -18,29 +22,63 @@ class EakApplicationMutationService {
     fun addEakApplication(
         regionId: Int,
         application: Application,
+        project: String,
         dfe: DataFetchingEnvironment,
     ): Boolean {
         val context = dfe.getContext<GraphQLContext>()
+        val backendConfig = context.backendConfiguration
+        val projectConfig = backendConfig.projects.first { it.id == project }
```
michael-markl, line 30: We should make sure that `regionId` stems from a region related to that project.

Also related: https://github.com/digitalfabrik/entitlementcard/issues/696

sarahsporck:
> If one email cannot be sent (, e.g. because of invalid from address) then the application is not persisted.

After thinking about this a while I think this is pretty difficult to handle. consider two emails already have been sent and one failed. In that case the organization mitarbeiter will still receive emails with an (invalid) link then, which feels much like spam to me. Therefore I reconsidered that we do not roll back here and let the mitarbeiter of the region handle the situation which would go like this in the optimal situation.
An application with three organisations has been created. And sending the mail to two of the organisations has worked fine. However, one failed. The two organisations will verify the application successfully, but the one who has not received the email address has not. In that case the mitarbeiter of the entitlementcard will contact the mitarbeiter at the organization him:herself after maybe 2 weeks. 
Showing an error to the applicant in this case has no benefit I think, as he:she cannot interfere or improve the situation. I think the best solution regarding "usability" would be to set the verification in an error state in the db, to let the mitarbeiter of the entitlementcard know he:she has to interfere.

Issue #877, opened 2023-03-20
Add link checker for links send via mail


f1sh1918:
As a user i want to get a feedback if the link i open from my e-mail ist valid

**Acceptance Criteria:**

- add a function that checks the accessKey/Id of the link against the database
- give feedback if the id does not exists
- send the plain text of the link additionally in the mails

**Implementations**
- [ ] reset password
- [ ] verification
- [ ] applicant view

Issue #873, opened 2023-03-20
Improve error handling


sarahsporck:
**Is your feature request related to a problem? Please describe.**
We are currently very inconsistent, how we handle errors in both backend and frontend.
A consistent error culture can improve the user experience and avoid any unwanted side effects (as errors shown in the frontend)

**Describe the solution you'd like**
- Consistent usage of Graphql errors: https://www.apollographql.com/docs/react/data/error-handling/ 
- Consistent error handling (not ignoring errors...)

**Describe alternatives you've considered**
Error return messages: 
```json
{
    "status": <http_status_like>,
    "message": <maybe_optional_because_of_translations>,
    "result": <only_for_queries>
}
```

**Additional context**
There are several related tickets for this: 
- #324 
- #836 
- #659 
- #830

f1sh1918:
@sarahsporck do you think this is possible to implement within this week. I see that its import but i think maybe we have more urgent tasks, f.e. to add a backend switch to testing system, since we should be able to start testing very soon

sarahsporck:
I think error handling **can** in fact also wait till after the release, but should be the issue we tackle directly after all other important tasks have been done. Such as e.g. a backend switch to a testing system. Not sure if i'll be able to finish this this week as this concerns backend, frontend and administration in the same way.

Issue #825, opened 2023-02-27
Don't require inserting the email address in the password reset form


michael-markl:
**Is your feature request related to a problem? Please describe.**
Currently, when the user resets their password (or sets it for the first time), they receive an email with a password reset link.
Clicking on that link, the user has to enter his email address. However, this should not be necessary. Instead, we can include the email address in the reset link as a URL param.

*Background*: If the user misspells his/her email, we cannot give him a proper error message. Avoiding these mistakes, can reduce a lot of confusion.

f1sh1918:
implemented with https://github.com/digitalfabrik/entitlementcard/pull/878

PR #878, opened 2023-03-20
877: Link checker

f1sh1918:
Just added the link checker to reset password that checks if resetKey exists and the link is not expired.
Added for applicant and verification view an empty collection check

resolves https://github.com/digitalfabrik/entitlementcard/issues/877


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordMutationService.kt
```diff
@@ -53,6 +53,9 @@ class ResetPasswordMutationService {
             Sie können Ihr Passwort unter dem folgenden Link zurücksetzen:
             $administrationBaseUrl/reset-password/${URLEncoder.encode(key, StandardCharsets.UTF_8)}
             
+            Falls ihr Browser den Link nicht korrekt auflösen kann, kopieren Sie bitte folgenden Link in Ihren Browser:
+            $administrationBaseUrl/reset-password/$key
+
```
michael-markl, line 58: Let's not do this. The above link is always the correct one. Not URL encoding $key produces an invalid URL, and it was okay, that Outlook disrespected those = chars.
Adding this comment would just add confusion imo.
```suggestion
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordQueryService.kt
```diff
@@ -0,0 +1,22 @@
...
+import app.ehrenamtskarte.backend.auth.database.Administrators
+import app.ehrenamtskarte.backend.auth.webservice.schema.types.Administrator
+import com.expediagroup.graphql.generator.annotations.GraphQLDescription
+import org.jetbrains.exposed.sql.transactions.transaction
+import java.time.LocalDateTime
+
+@Suppress("unused")
+class ResetPasswordQueryService {
+    @GraphQLDescription("Verify password reset link")
+    fun checkPasswordResetLink(resetKey: String): Administrator {
```
michael-markl, line 13: We should also have a `project` argument as input, and make sure the admin is in the correct project.

administration/src/components/InvalidLink.tsx
```diff
@@ -0,0 +1,15 @@
...
+import React, { ReactElement } from 'react'
+
+const CenteredAlert = styled(Alert)`
+  margin: auto;
+`
+
+const InvalidLink = (): ReactElement => (
+  <CenteredAlert severity='info'>
+    <AlertTitle>Ihr Link ist ungültig</AlertTitle>
+    Ihr Link konnte nicht korrekt aufgelöst werden. Bitte kopieren Sie den Link manuell aus Ihrer E-Mail.
```
michael-markl, line 11:
```suggestion
    Ihr Link ist ungültig. Versuchen Sie, den Link manuell aus Ihrer E-Mail in die Addresszeile Ihres Browsers zu kopieren.
```

administration/src/components/applications/ApplicationApplicantController.tsx
```diff
@@ -25,6 +26,7 @@ const ApplicationApplicantController = (props: { providedKey: string }) => {
   })
 
   if (loading) return <CircularProgress style={{ margin: 'auto' }} />
+  else if (error?.message.includes('Collection is empty')) return <InvalidLink />
```
michael-markl, line 29: Let's not rely on internal error messages. We should probably define a new GraphQLError (currently we only have InvalidEmailError as an example I think) . Also relates to our error handling issue #873 .

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordQueryService.kt
```diff
@@ -0,0 +1,27 @@
...
+    @GraphQLDescription("Verify password reset link")
+    fun checkPasswordResetLink(project: String, resetKey: String): Administrator {
+        return transaction {
+            val projectId = ProjectEntity.find { Projects.project eq project }.single().id.value
+            val admin = AdministratorEntity
+                .find { Administrators.passwordResetKey eq resetKey and (Administrators.projectId eq projectId) }.single()
+            if (admin.passwordResetKeyExpiry!!.isBefore(LocalDateTime.now())) {
+                throw Exception("Password reset key has expired.")
+            }
+            Administrator.fromDbEntity(admin)
```
michael-markl, line 24: I would rather not return the administrator given a reset key.
People might try to brute force any reset keys with this endpoint, and if they ever find one (which is -- I admit -- quite unlikely) they have full access to the account.

To solve your "usability" issue (#825), I would suggest to also encode the email in the reset link, e.g.
`/reset-password?email={email}&token={resetKey}` 

Then, in the administration frontend, we can autofill the email field from the URL parameter.

Building on top of that, we can also add the email as input to your `checkPasswordResetLink` query.
Then, we look up the administrator with that specific email, check whether the reset key matches (*we should actually store a hash in the DB instead of the raw key!*) and return TRUE if it was successfull, otherwise FALSE.

Related: https://security.stackexchange.com/questions/86913/should-password-reset-tokens-be-hashed-when-stored-in-a-database

f1sh1918: hm there are very different opinions in this thread. Is it really necessary to hash a reset key thats only valid for 24h? Well then we can create a separate task because thats not really connected to the link checker

f1sh1918: https://github.com/digitalfabrik/entitlementcard/issues/887

f1sh1918: adjusted pw reset link to use SearchParams

administration/src/components/auth/ResetPasswordController.tsx
```diff
@@ -33,15 +46,24 @@ const ResetPasswordController = () => {
...
   const warnMessage = validateNewPasswordInput(newPassword, repeatNewPassword)
   const isDirty = newPassword !== '' || repeatNewPassword !== ''
 
+  if (error) {
+    console.log(getMessageFromApolloError(error))
+    return (
+      <CenteredMessage title={getMessageFromApolloError(error).title}>
+        {getMessageFromApolloError(error).description}
+      </CenteredMessage>
+    )
```
michael-markl, line 64:
```suggestion
    const { title, description } = getMessageFromApolloError(error)
    return (
      <CenteredMessage title={title}>
        {description}
      </CenteredMessage>
    )
```

administration/src/components/applications/ApplicationApplicantController.tsx
michael-markl: Hmm.. yeah I agree... It feels like it's getting messy quickly. 🤔 
Maybe, for this PR we could keep the complexity low, by simply returning False (in the graphql endpoint) if the link is invalid (where invalid includes "Link is expired"). Then, we show the user something like 
```md
# Der Link ist ungültig.

Der Link zum Zurücksetzen Ihres Passwortes ist ungültig. Mögliche Gründe:
* Der Link ist abgelaufen. Sie können [hier](/reset-password) einen neuen Link erhalten.
* Der Link wurde fehlerhaft in den Browser übertragen. Versuchen Sie, den Link manuell aus der Email in die Adresszeile Ihres Browsers zu kopieren.
* Sie haben Ihr Passwort mithilfe des Links bereits zurückgesetzt
* Sie haben einen weiteren Link angefordert. Es ist immer nur der aktuellste Link gültig.
```

Writing this up made me release that we cannot really know which of these cases we're in, so imo it makes sense to simplify our system by returning a `false` boolean instead of throwing several errors. What do you think?

f1sh1918: hm but this endpoint returns application data not a boolean. So we need a separate function that just returns a boolean right?

f1sh1918: Maybe for this case we could just check if the error codes are equal
```js
const codesEqual = error.graphQLErrors.every((value, index, array)=>value.extensions.code === array[0].extensions.code);
```

f1sh1918: hm but then we need a separate Query that just returns a boolean (valid/invalid) right now we getting an ApplicationView

administration/src/application-verification/VerificationController.tsx
```diff
@@ -70,8 +72,17 @@ const ApplicationVerification = ({ applicationVerificationAccessKey }: Applicati
     variables: { applicationVerificationAccessKey },
   })
 
+  // TODO handle two error messages of graphQL queries
```
michael-markl, line 75: What does that mean?

administration/src/application-verification/VerificationController.tsx
```diff
@@ -70,8 +72,17 @@ const ApplicationVerification = ({ applicationVerificationAccessKey }: Applicati
...
+
   if (loading) return <CircularProgress style={{ margin: 'auto' }} />
-  else if (error || !data) return <ErrorHandler refetch={refetch} />
+  else if (error)
+    return (
+      <InvalidLink
+        title={getMessageFromApolloError(error).title}
+        description={getMessageFromApolloError(error).description}
+      />
+    )
```
michael-markl, line 84:
```suggestion
    const { title, description}  = getMessageFromApolloError(error)
    return <InvalidLink title={title} description={description} />
```

Also: It is weird to me that we show an InvalidLink component for any error. Couldn't there a different error?
What happens on connection / server errors? I think we should show a refetch button there?

administration/src/components/applications/ApplicationApplicantController.tsx
```diff
@@ -24,7 +26,14 @@ const ApplicationApplicantController = (props: { providedKey: string }) => {
...
 
   if (loading) return <CircularProgress style={{ margin: 'auto' }} />
-  else if (error || !data) return <ErrorHandler refetch={refetch} />
+  else if (error) {
+    return (
+      <InvalidLink
+        title={getMessageFromApolloError(error).title}
+        description={getMessageFromApolloError(error).description}
+      />
+    )
```
michael-markl, line 35: Same as above.

administration/src/components/auth/ResetPasswordController.tsx
```diff
@@ -33,15 +46,20 @@ const ResetPasswordController = () => {
...
+        passwordResetKey: queryParams.get('token')!,
       },
     })
 
   const warnMessage = validateNewPasswordInput(newPassword, repeatNewPassword)
   const isDirty = newPassword !== '' || repeatNewPassword !== ''
 
+  if (error) {
+    const { title, description } = getMessageFromApolloError(error)
+    return <CenteredMessage title={title}>{description}</CenteredMessage>
```
michael-markl, line 60: Similar to above: On some errors, we should probably have a retry button?

michael-markl (approved):
More and more, I have the feeling that using the graphql errors (and extending them to more endpoints) leads to unintuitive, intransparent code (especially in the error handling in the frontends, but also in the backend)...

We can merge this PR, but I feel like in #873 we should analyze whether proper return types might be a better fit for us.

administration/src/application-verification/VerificationController.tsx
f1sh1918: yes you are right about the refetch. Probably we shouldn't call it `InvalidLink` but extend the `ErrorHandler` with a optional description

administration/src/components/applications/InvalidLinkTemplate.tsx
```diff
@@ -1,21 +1,13 @@
...
-      </li>
       <li>
         Der Link wurde fehlerhaft in den Browser übertragen. Versuchen Sie, den Link manuell aus der Email in die
-        Adresszeile Ihres Browsers zu kopieren
+        Adresszeile Ihres Browsers zu kopieren.
       </li>
-      <li>Sie haben Ihr Passwort mithilfe des Links bereits zurückgesetzt</li>
-      <li>Sie haben einen weiteren Link angefordert. Es ist immer nur der aktuellste Link gültig.</li>
+      <li>Sie haben Ihr Passwort mithilfe des Links bereits zurückgesetzt. (Passwort Reset)</li>
+      <li>Sie haben einen weiteren Link angefordert. Es ist immer nur der aktuellste Link gültig. (Passwort Reset)</li>
```
michael-markl, line 10: Hmm... not the best UX :D Also, this component should probably not be inside an `applications` folder?

f1sh1918: I mean I can throw different InvalidLink Errors and handle them separatly

michael-markl: Would be better imo.

Issue #886, opened 2023-03-22
Application createDate timestamp wrong on production


f1sh1918:
**Describe the bug**
On production system the createDate timestamp is UTC instead of UTC+1

**To Reproduce**
1.https://bayern.ehrenamtskarte.app/beantragen
2. Submit an application
3. Login with a region manager of the region that you submitted the application
4. Go to eingehende Anträge
5. Check if the creation date is correct

**Expected behavior**
When i create an application in UTC+1 it should be reflected in the database.
Note: It works locally. 
At 10:17 this creationDate was set in productionDB
`2023-03-22 09:17:58.083568`

f1sh1918:
we should use a defined timezone for creation date

maxammann:
I think the actual data stored in the DB is independent of timezones. We just need to evaluate the time according to our timezone (which is set in the project settings).

I think by default the DB connector tells the DB which timezone to choose. Or it just uses the system timezone.

PR #895, opened 2023-03-28
886: Datetime UTC

f1sh1918:
- use utc for localTime in backend
- add timezone for project in administration projectConfig
- display date using project specific timezone in administration

resolves https://github.com/digitalfabrik/entitlementcard/issues/886


michael-markl:
I just realized that we use LocalDateTime in our DB entities. It would make much more sense if we could use Instant instead (see 
https://stackoverflow.com/a/32443004). Not sure, how easy it would be to adjust that.

administration/package.json
```diff
@@ -17,6 +17,7 @@
     "@zxing/library": "^0.19.2",
     "apollo-upload-client": "^17.0.0",
     "date-fns": "^2.29.3",
+    "date-fns-tz": "^2.0.0",
```
michael-markl, line 20: I think we don't need this dependency and we can just use [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat). We don't need to support IE11 :D

administration/src/util/getDateFromUTC.ts
```diff
@@ -0,0 +1,2 @@
+const getDateFromUTC = (dateString: string): Date => new Date(`${dateString}Z`)
```
michael-markl, line 1: It would be better if the GraphQL endpoint would send this `Z` suffix already. I believe changing from LocalDateTime to Instant should fix this as well.

f1sh1918: yes indeed i check this `Instant.now()` but couldn't figure out how to set this with `defaultExpression`

administration/src/project-configs/bayern/config.ts
```diff
@@ -15,6 +15,7 @@ const config: ProjectConfig = {
   createEmptyCard: createEmptyBavariaCard,
   dataPrivacyHeadline: dataPrivacyBaseHeadline,
   dataPrivacyContent: DataPrivacyBaseText,
+  timezone: 'UTC',
```
michael-markl, line 18: Why did you choose UTC here? Should be 'Europe/Berlin' I guess?

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
```diff
@@ -19,9 +20,10 @@ class WebService {
         val host = config.server.host
         val port = Integer.parseInt(config.server.port)
         val dataDirectory = config.server.dataDirectory
-
         val applicationData = File(dataDirectory, "applications")
 
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"))
+
```
michael-markl, line 26: Can you check whether this is still necessary? Maybe we should remove it, if it's not necessary. (Just because we probably shouldn't fiddle with globals)

michael-markl (commented):
Did not test it yet

administration/src/project-configs/bayern/config.ts
f1sh1918: just by mistake, used for testing. thx for the find

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
f1sh1918: this is necessary

michael-markl:
@f1sh1918 I pushed another commit that fixes the default CurrentTimestamp() expression (and should make the rest more robust as well):
* I made the DB session use UTC as well to be in sync with the rest of the application.
* I made the timezone overwrite the very first line of our backend, to make sure it always runs (not just when starting the webservice)

Overall, I believe this is an Exposed bug (when used in combination with postgres) to be honest. Imo, it should use a `timestamp with timezone` column for timestamp, but oh well...

michael-markl (approved):

Issue #871, opened 2023-03-16
CORS blocks login to showcase


f1sh1918:
When i want to login to: https://druckerei.entitlementcard.app/
I receive login fehlgeschlagen.

```js
Access to fetch at 'https://api.entitlementcard.app/' from origin 'https://druckerei.entitlementcard.app' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

f1sh1918:
we do not use the showcase but the staging environment, so its not really relevant anymore

PR #896, opened 2023-03-29
888: Deactivate beantragen

f1sh1918:
Additionally to the nginx redirect, i added the possibility to reach the application form internally via user-settings

### DIFF
--- a/administration/src/components/user-settings/ApplicationLinkCard.tsx
+++ b/administration/src/components/user-settings/ApplicationLinkCard.tsx
@@ -1,0 +1,23 @@
+import { Button, Card, H2 } from '@blueprintjs/core'
+import React, { ReactElement } from 'react'
+import { useNavigate } from 'react-router-dom'
+
+const ApplicationLinkCard = (): ReactElement => {
+  const navigate = useNavigate()
+  return (
+    <div style={{ display: 'flex', justifyContent: 'center', marginTop: '20px' }}>
+      <Card style={{ maxWidth: '500px' }}>
+        <H2>Ehrenamtskarte beantragen</H2>
+        <p>
+          Aktuell ist das neue Beantragungsformular der Ehrenamtskarte nur intern für Testzwecke zugänglich. Hier
+          gelangen sie zum Beantragungsformular der Ehrenamtskarte.
+        </p>
+        <div style={{ textAlign: 'right', padding: '10px 0' }}>
+          <Button text='Beantragen' intent='primary' onClick={() => navigate('/beantragen')} />
+        </div>
+      </Card>
+    </div>
+  )
+}
+
+export default ApplicationLinkCard
--- a/administration/src/components/user-settings/UserSettingsController.tsx
+++ b/administration/src/components/user-settings/UserSettingsController.tsx
@@ -6,7 +6,10 @@ import ChangePasswordForm from './ChangePasswordForm'
 import validatePasswordInput from '../auth/validateNewPasswordInput'
 import { WhoAmIContext } from '../../WhoAmIProvider'
 
+import ApplicationLinkCard from './ApplicationLinkCard'
+
 const UserSettingsController = () => {
+  const { applicationFeatureEnabled } = useContext(ProjectConfigContext)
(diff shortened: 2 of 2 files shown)


f1sh1918:
> Do regions really want to use this for internal testing?
> 
> If no, I think we don't need these changes. If we want to test it ourselves on the production system, we can still go to https://druckerei.entitlementcard.app/beantragen and set the localstorage variable `project-override` to `bayern.ehrenamtskarte.app` (after fixing CORS #871 :D )
> 
> If yes, I'll do another review :D

I think they want to test this. I'm really not sure if everyone will get access to the test system. Somehow they should be able to test the process

michael-markl:
>  I'm really not sure if and when everyone will get access to the test system.

Not sure if I understand correctly: I don't think that actual admins of regions will ever get access to the test system. Afaik, the test system is only for our internal testing purposes?

f1sh1918:
> > I'm really not sure if and when everyone will get access to the test system.
> 
> Not sure if I understand correctly: I don't think that actual admins of regions will ever get access to the test system. Afaik, the ~test~ staging system is only for our internal testing purposes?

Okay i dunno. I thought that maybe some key users could get access. But then even more we have to provide this. As i got daniel right the service user should get familiar with the process

administration/src/components/user-settings/ApplicationLinkCard.tsx
```diff
@@ -0,0 +1,23 @@
...
+const ApplicationLinkCard = (): ReactElement => {
+  const navigate = useNavigate()
+  return (
+    <div style={{ display: 'flex', justifyContent: 'center', marginTop: '20px' }}>
+      <Card style={{ maxWidth: '500px' }}>
+        <H2>Ehrenamtskarte beantragen</H2>
+        <p>
+          Aktuell ist das neue Beantragungsformular der Ehrenamtskarte nur intern für Testzwecke zugänglich. Hier
+          gelangen sie zum Beantragungsformular der Ehrenamtskarte.
+        </p>
```
michael-markl, line 14:
```suggestion
        <H2>Antragsformular testen</H2>
        <p>
          Aktuell ist das neue Antragsformular der Ehrenamtskarte nur für Testzwecke zugänglich. Hier
          gelangen Sie zum Antragsformular der Ehrenamtskarte.
        </p>
```

administration/src/components/user-settings/ApplicationLinkCard.tsx
```diff
@@ -0,0 +1,23 @@
...
+        <div style={{ textAlign: 'right', padding: '10px 0' }}>
+          <Button text='Beantragen' intent='primary' onClick={() => navigate('/beantragen')} />
```
michael-markl, line 16:
```suggestion
        <H2>Antragsformular testen</H2>
        <p>
          Aktuell ist das neue Antragsformular der Ehrenamtskarte nur für Testzwecke zugänglich. Hier
          gelangen Sie zum Antragsformular der Ehrenamtskarte.
        </p>
        <div style={{ textAlign: 'right', padding: '10px 0' }}>
          <Button text='Zum Antrag' intent='primary' onClick={() => navigate('/beantragen')} />
```

administration/src/components/user-settings/UserSettingsController.tsx
```diff
@@ -52,18 +55,22 @@ const UserSettingsController = () => {
...
+        setNewPassword={setNewPassword}
+        repeatNewPassword={repeatNewPassword}
+        setRepeatNewPassword={setRepeatNewPassword}
+        submitDisabled={!valid}
+        warnMessage={isDirty ? warnMessage : null}
+        loading={loading}
+        submit={submit}
+      />
+      {/* TODO #897: [Application] Remove Redirect for bayern */}
+      {applicationFeatureEnabled && <ApplicationLinkCard />}
```
michael-markl, line 72: I think it would make more sense, if it was on the `ApplicationsOverview` tab somewhere (eg. above everything else).

michael-markl (approved):

michael-markl:
>I thought that maybe some key users could get access. 

I guess that makes sense

PR #906, opened 2023-03-29
911: Setup migrations

sarahsporck:
So this PR is basically the counterpart to #841. I did not use an external migration tool, but only created a migration table, as it was also suggested by @maxammann . 

Pros: 
- we do not need an external closed source dependency
- missing features such as rollback can be added later
- migrations can be written in *.java

Cons: 
- missing experience on how migrations work and best practices
- potentially additional maintenance
- migrations cannot be written in sql (only indirect by passing the sql string to exposed)

While originally, I was pretty convinced that flyway is the better solution I now think that using a custom migration implementation has the same value, and is probably also implemented very similar. 

Please, leave you're opinions on this in a timely manner so we can decide, which way we want to go :)


backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
```diff
@@ -115,5 +117,14 @@ class Execute : CliktCommand(help = "Starts the webserver") {
     }
 }
 
+class Migrate : CliktCommand(help = "Migrates the database") {
+    private val config by requireObject<BackendConfiguration>()
+
+    override fun run() {
+        val db = Database.setup(config)
+        runAllMigrations(db, Clock.systemDefaultZone())
```
f1sh1918, line 125: Maybe we should set UTC here explicitly since we use it for all our timestamps. SystemDefault can vary?

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+    val sorted = migrations.map { it.version }.sorted()
+    if ((1..migrations.size).toList() != sorted) {
+        throw IllegalStateException("List of migrations version is not consecutive: $sorted")
+    }
+}
+
+private fun shouldRun(latestVersion: Int?, migration: Migration): Boolean {
+    val run = latestVersion?.let { migration.version > it } ?: true
+    if (!run) {
+        logger.debug("Skipping migration version ${migration.version}: ${migration.name}")
```
f1sh1918, line 87: i think a `logger.info` would be nice that shows that migrations will be skipped

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V1_RemoveEmailIndex.kt
```diff
@@ -0,0 +1,16 @@
...
+import app.ehrenamtskarte.backend.migration.Migration
+import org.jetbrains.exposed.sql.transactions.TransactionManager
+import org.jetbrains.exposed.sql.transactions.transaction
+
+internal class V1_RemoveEmailIndex : Migration() {
+    val indexName = "email_lower_idx"
+
+    override fun run() {
+        transaction {
+            val dropStatement = "drop index $indexName"
```
f1sh1918, line 12: when i run the migration script the index should be removed, right?
Maybe i started sth wrong :D
`run --args="migrate"` not sure if thats fine

f1sh1918 (commented):
Well i checked both PRs now and i feel good with flyway to be honest i just created some basic migration scripts in sql and kotlin and everything worked fine. For our concerns it should be fine. Since i have no real experience in database migration i would tend to use a library but if you feel comfortable with a custom migration its also ok for me.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/database/Schema.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.jetbrains.exposed.dao.IntEntity
+import org.jetbrains.exposed.dao.IntEntityClass
+import org.jetbrains.exposed.dao.id.EntityID
+import org.jetbrains.exposed.dao.id.IdTable
+import org.jetbrains.exposed.sql.javatime.CurrentTimestamp
+import org.jetbrains.exposed.sql.javatime.timestamp
+
+object Migrations : IdTable<Int>() {
+    override val id = integer("version").entityId()
+    override val primaryKey = PrimaryKey(id)
```
michael-markl, line 12: Not sure, but if you're fiddling with id and primaryKey, maybe you could instead just use a normal `Table`?
I think IdTable<Int> does not anything more than that:
https://github.com/JetBrains/Exposed/blob/master/exposed-core/src/main/kotlin/org/jetbrains/exposed/dao/id/IdTable.kt#L41

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/database/Schema.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    val name = varchar("name", length = 400).index()
```
michael-markl, line 14: Not sure, if we need an index on name.
Do we need a uniqueness constraint? 

```suggestion
    val name = varchar("name", length = 400).index()
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+                this.name = "Initialize"
+                this.executedAt = now()
+            }
+        }
+    }
+}
+
+private fun checkVersions(migrations: List<Migration>) {
+    val sorted = migrations.map { it.version }.sorted()
+    if ((1..migrations.size).toList() != sorted) {
```
michael-markl, line 79: Does this check for duplicates too?

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.jetbrains.exposed.sql.exists
+import org.jetbrains.exposed.sql.transactions.transaction
+import org.jetbrains.exposed.exceptions.ExposedSQLException
+import org.slf4j.LoggerFactory
+import org.postgresql.util.PSQLException
+
+private val logger = LoggerFactory.getLogger(Migrations::class.java)
+
+fun runAllMigrations(database: Database, clock: Clock) {
+    runMigrations(database, listOf(V1_RemoveEmailIndex()), clock)
```
michael-markl, line 20: Should we return a bool indicating whether migration was successfull. It would be useful to set the correct return type.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+}
+
+fun runMigrations(database: Database, migrations: List<Migration>, clock: Clock) {
```
michael-markl, line 23: Can we wrap everything here inside a transaction? That way we would not do anything to the DB (I hope) and can roll back to the last backend code.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+}
+
+private fun checkVersions(migrations: List<Migration>) {
+    val sorted = migrations.map { it.version }.sorted()
+    if ((1..migrations.size).toList() != sorted) {
+        throw IllegalStateException("List of migrations version is not consecutive: $sorted")
+    }
+}
```
michael-markl, line 82: I think it would be a bit better not to sort (so we make sure the migration list has been written in the correct order).
And to make it is more readable, rename it,
And to make it consistent return a bool and do some `logger.error; return` in the `runMigrations` function

```suggestion
private fun versionsAreConsecutiveAndDuplicateFree(migrations: List<Migration>)
    = migrations.map { it.version } == (1..migrations.size).toList()
```

michael-markl (commented):
Personally (and somewhat surprisingly for me :D), I think I like this approach better as it has quite a little footprint, is easy to understand (no need to fight through flyway docs), and super flexible.

I think in both PRs you are missing one important aspect:
V1 should always create all tables at some baseline (I would suggest baseline=current tables in this PR / in main). Otherwise it depends on our production server.

This means, we'll have to create a lot of duplication in the V1 migration file (as we should not use the schema definitions in the "production code", as they can change in the future), but that's fine.
That being said, consequently, we should remove all other setupDatabase functions. We never want files other than migration files to alter the DB schema.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+            create(Migrations)
+        }
+
+        val numMigrations = MigrationEntity.all().count()
+
+        if (numMigrations == 0L) {
+            MigrationEntity.new(0) {
+                this.name = "Initialize"
+                this.executedAt = now()
+            }
```
michael-markl, line 72: I think this should not be necessary?

michael-markl (commented):
In a next step / issue, we could set up our CI to run all migrations starting from an empty table, just to have some confidence on push.

f1sh1918:
> Personally (and somewhat surprisingly for me :D), I think I like this approach better as it has quite a little footprint, is easy to understand (no need to fight through flyway docs), and super flexible.
> 
> I think in both PRs you are missing one important aspect: V1 should always create all tables at some baseline (I would suggest baseline=current tables in production).
> 
> This means, we'll have to create a lot of duplication in the V1 migration file (as we should not use the schema definitions in the "production code", as they can change in the future), but that's fine. That being said, consequently, we should remove all other setupDatabase functions. We never want files other than migration files to alter the DB schema.

OK so as i got that right, for every adjustment i make for the schema like adding a column, i have to write a migration.
Hm so it would make sense also to be able to execute sql files?!

michael-markl:
> OK so as i got that right, for every adjustment i make for the schema like adding a column, i have to write a migration.

That's my understanding of DB migrations, yes.

> Hm so it would make sense also to be able to execute sql files?!

I'm not entirely sure what your questions refers to, but you can also execute arbitrary SQL statements through exposed.

>  Maybe a stupid question but can we not just run the setupDatabase functions once in the first migration file to create the baseline?

Not a stupid question. But: I think, what DB migrations usually want to achieve is, that if you start with an empty DB and apply all migrations, you end up in a deterministic state. This means, that once you have written a migration script (and executed it on the prod system), you should usually never change it. Otherwise, you cannot replicate the production schema by simply running all migrations again.

So let's assume, we would just use the setupDatabase functions in the V1 migration step. At some point we want to add/drop a column from a table. Then you would adjust the schema in the production code (that we call from our webservice). But with this adjustment, you also change the V1 migration step.

What we could do, I think, is redefine our "first" schema inside the V1 migration file (basically copy pasting every Schema we have into the migration file) and call exposed's SchemaUtils.create on these schemas.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/database/Setup.kt
```diff
@@ -0,0 +1,10 @@
+package app.ehrenamtskarte.backend.migration.database
+
+import org.jetbrains.exposed.sql.SchemaUtils
+import app.ehrenamtskarte.backend.migration.database.Migrations
+
+fun setupDatabase() {
+    SchemaUtils.create(
+        Migrations,
+    )
+}
```
timoludwig, line 10: This step does not make sense anymore, since the schema already has been created in the `createTableIfNotExists()` function during migrations, right?

timoludwig (commented):
Looks very promising :+1:

I don't have a strong preference towards one of the two solutions, but I'd give this one a slight edge since having the possibility to use the exposed ORM in migrations can be a crucial advantage, especially when dealing with data migrations where you not only alter the schema of a table, but also its contents.

I noticed there is a fork of the initial code base, [Suwayomi/exposed-migrations](https://github.com/Suwayomi/exposed-migrations), which seems to be more or less maintained and has a few improvements, so I guess we could also import that library and provide all our suggestions as upstream contributions.

Other than that, I agree with everything @michael-markl said :grin:

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V1_RemoveEmailIndex.kt
sarahsporck: jep that should have been fine. Maybe you can try again with the changes @michael-markl applied. Three migrations should be executed here.

sarahsporck:
@michael-markl did you create the baseline migration from the production db?

michael-markl:
> @michael-markl did you create the baseline migration from the production db?

No, that still needs to be done (see Todo in V2 step). I don't have access to the backend so I couldn't do it 😅

sarahsporck:
@maxammann or @f1sh1918 could one of you send me the current production sql schemes? I think you just need to run `pg_dump --schema-only`. Correct me if I'm wrong @michael-markl 😅

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+                MigrationEntity.all().maxBy { it.version }.version.value
+            }
+
+    logger.info("Database version before migrations: $latestVersion")
+    try {
+    migrations.sortedBy { it.version }.filter { shouldRun(latestVersion, it) }.forEach {
+        logger.info("Running migration version ${it.version}: ${it.name}")
+
+            transaction(database) {
+                it.run()
```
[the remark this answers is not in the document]
sarahsporck: So we created a separate command to allow skipping the baseline setup. Is there any other situation this could be useful?

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
```diff
@@ -0,0 +1,90 @@
...
+}
+
+private fun checkVersions(migrations: List<Migration>) {
+    val sorted = migrations.map { it.version }.sorted()
+    if ((1..migrations.size).toList() != sorted) {
+        throw IllegalStateException("List of migrations version is not consecutive: $sorted")
+    }
+}
+
+private fun shouldRun(latestVersion: Int?, migration: Migration): Boolean {
```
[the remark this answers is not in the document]
sarahsporck: The second solution is kind of already implemented now. Therefore, I'd not implement skipping or overwriting versions for now. But might be necessary in the future. Who knows :)

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/database/Schema.kt
sarahsporck: Probably not on the name. And I guess the index could also be removed.

f1sh1918:
> @maxammann or @f1sh1918 could one of you send me the current production sql schemes? I think you just need to run `pg_dump --schema-only`. Correct me if I'm wrong @michael-markl 😅 

I've tried several times now. No idea why it didn't work. No error and nothing.. Just couldn't find any  created schema sql file on the server. Gonna try to get it tommorow again...

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V1_InitializeMigrations.kt
```diff
@@ -0,0 +1,15 @@
+package app.ehrenamtskarte.backend.migration.migrations
+
+import app.ehrenamtskarte.backend.migration.Migration
+import app.ehrenamtskarte.backend.migration.Statement
+
+internal class V1_InitializeMigrations : Migration() {
```
timoludwig, line 6: In my opinion, the migration table should be created by the migration framework and not by a migration itself.
It just does not make sense design-wise, since if you change the schema of the migration table and its usage in `MigrationUtils`, these changes need to be present from the first migration on (e.g. when setting up a new system), but this first migration is supposed to be immutable. So if you define your migration schema inside a migration, you will never be able to change it, which kind of defies the point of a migration.
So I would suggest to use `SchemaUtils.create()` somewhere inside the migration utils in this specific case. Most likely you will never change the migration schema anyways, but just in case you do, you are flexible enough to handle the transition in the code.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/runMigrations.kt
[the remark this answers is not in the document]
timoludwig: No, not necessarily. Maybe if something goes wrong and the migration table ends up in an inconsistent state. But I guess it's also possible to fix it manually in this case...

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V1_InitializeMigrations.kt
michael-markl: I think I agree :D So the changes to the migrations table should probably be handled separately.

> So I would suggest to use SchemaUtils.create() somewhere inside the migration utils in this specific case. Most likely you will never change the migration schema anyways, but just in case you do, you are flexible enough to handle the transition in the code.

I think it is reasonable to create the migrations table and do any other Migrations-Table-Migrations (:D)  just before we apply any actual migrations (but still inside the transaction, so if migration fails, we can safely go back to the old backend code (with the old migrations table))

timoludwig: OK, but I would not handle migration-schema changes inside separate migrations, because then we would have to create a migrations-migration table to keep track of which migration-migrations are already applied :joy:

michael-markl: Yep! I guess, changes to the migration schema will happen rarely enough to handle them in the code directly without keeping track of them in another table :D
Also, once these changes have been applied in prod, we can drop the code for migrating to new new migration schema :D

sarahsporck:
@michael-markl is there anything holding this back to be merged? I also tested this locally and did not encounter any issues.

michael-markl:
No, imo it can be merged. I am excited for this😻

f1sh1918:
> 

@sarahsporck 
yes we should have waiting for deactivation of auto update
i wanted to test it now and i get
```js
Exception in thread "main" app.ehrenamtskarte.backend.migration.DatabaseOutOfSyncException: 
Latest migration versions do not match: Version on DB null - Code Version 2
```

sarahsporck:
> > 
> 
> @sarahsporck 
> yes we should have waited for deactivation of auto update on production
> i wanted to test it now and i get with an empty database
> ```js
> Exception in thread "main" app.ehrenamtskarte.backend.migration.DatabaseOutOfSyncException: 
> Latest migration versions do not match: Version on DB null - Code Version 2
> ```

Locally? And you did execute just the migrate command?

sarahsporck:
Currently migratuons are not yet run in the ci. But can only executed locally. So as long as you don't accidentally migrate on the live system everything should be fine.

michael-markl:
Since #916 is merged, they are run in PRs and on merge.

Issue #890, opened 2023-03-27
Add card activationDate [Statistics]


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a service user i want to know how much cards were activated in specific period

**Describe the solution you'd like**

- when a card was activated in the app, an activationDate timestamp should be written in the database

PR #907, opened 2023-03-29
890: Add cardActivation date

f1sh1918:
- activate verification feature flag in build config

resolves https://github.com/digitalfabrik/entitlementcard/issues/890


michael-markl (commented):
I think for stats purposes a `firstActivationDate` would make a bit more sense? (ie. persisting the date when a card was first activated).
Should we change it to `firstActivationDate`?

If not, the graph that I imagine could change in hindsight, which would be weird.

f1sh1918:
> I think for stats purposes a `firstActivationDate` would make a bit more sense? (ie. persisting the date when a card was first activated). Should we change it to `firstActivationDate`?
> 
> If not, the graph that I imagine could change in hindsight, which would be weird.

Okay that makes sense i think. So i adjust the naming and only set it if its null

michael-markl:
Sounds great!

sarahsporck (approved):
Tested on Android device. Strangely I had to recreate the db for this. I thought exposed creates columns and tables, that do not exist... I hope this does not mess up the Nürnberg card...

f1sh1918:
> 

this could be a good use case for database migrations :)
I also had to recreate the card table. I think there is no real card data in there but we should try to avoid it

sarahsporck (approved):
This kind of fixes this, although I could not use an old QR-Code for the card activation. Not sure why though...

f1sh1918:
> This kind of fixes this, although I could not use an old QR-Code for the card activation. Not sure why though...

alright no hurry to merge this. I gonna check this next week. Probably it makes more sense to wait until db migrations are ready and write a migration for this instead of using the `schemaUtils`

michael-markl:
I think we should also wait until after the migration PR is merged.

f1sh1918:
> I think we should also wait until after the migration PR is merged.

i updated this with my first migration :)

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0003_AddFirstActivationDate.kt
```diff
@@ -0,0 +1,26 @@
...
+        exec(
+            """
+            ALTER TABLE cards ADD "firstActivationDate" TIMESTAMP NULL;
+            """.trimIndent()
+        )
+        exec(
+            """
+            UPDATE cards
+            SET "firstActivationDate" = current_timestamp
+            where "totpSecret" is not null
```
michael-markl, line 22: I am not entirely sure, whether semicolon is necessary here.

```suggestion
            where "totpSecret" is not null;
```

michael-markl (approved):
Can you merge in the latest main into this PR before merging? The CI check is on main but not on this PR yet.

Issue #908, opened 2023-03-29
Create missing tables and columns in database setup


f1sh1918:
**Is your feature request related to a problem? Please describe.**
If we add new columns to a database we want to to created when we the backend restarts

- implement `SchemaUtils.createMissingTablesAndColumns()` in every `setupDatabase()`

f1sh1918:
Should we add this to every database setup or only if needed?

michael-markl:
This should not be necesary once migrations are set up.

f1sh1918:
will be handle with migrations #911

PR #909, opened 2023-03-29
908: Create missing columns in tables

f1sh1918:
- create missing columns in database when new columns were added to the schema

resolves https://github.com/digitalfabrik/entitlementcard/issues/908

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/Setup.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/Setup.kt
@@ -7,4 +7,5 @@ fun setupDatabase() {
         Applications,
         ApplicationVerifications,
     )
+    SchemaUtils.createMissingTablesAndColumns(Applications, ApplicationVerifications)
 }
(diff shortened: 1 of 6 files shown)


f1sh1918:
Probably i can close this pr since we gonna update our schema with the migrations?!

sarahsporck:
Depends. When we want to rollout the access key, activation code, etc. in the next backend release (before the migration pr has finished) then we may still need this. Depending on how much the production and development db is diverged rn. I'll remove the schema update from the card schema as well in the other branch.

michael-markl:
Weirdly, applications.accessKey, applications.withdrawalDate is already in the prod DB according to the pg_dump.
Which `activation code` do you mean?

As I think, the migration PR should be mergable in the next day(s), I think we can close this (especially as these changes are not *super* urgent, and we would need to do another pg_dump).

michael-markl:
> I thought this changes are not needed at all, since we want to alter the schema via migrations?

Once the migrations PR is merged, this is correct :)

f1sh1918:
not needed after migrations

PR #911, opened 2023-03-31
Suggestion for DB migrations

michael-markl:
@sarahsporck : I've played around a little bit with the migrations, because I wanted to test whether we can actually implement some schema validation (other than based on the migration version):
So what this PR does is:
* On every start validate, that the "production" schema definitions, ie. the Exposed Table definitions, match the existing DB schema.
* If the schema does not match, it suggests migration statements.
* [Note that this is still independent of the choice between the two migration approaches]

But then, I got a little side-tracked with minor adjustments that are all inside this PR 😁 Sorry about that^^
This includes:
 * Creating the migrations table itself is now V1
 * The baseline migration is now V2
 * All other migrations are V3+
 * I added a command for skipping the baseline step for prod

Feel free to cherry-pick whatever you think is reasonable into your PR, or merge the whole branch into yours, if you want / if it makes sense. :D

It also addresses some of Timo's comments in your PR


sarahsporck:
Nice 👍🏼 
I think I'm gonna take everything you recommended.

PR #912, opened 2023-03-31
Create directory for responsive components

sarahsporck:
This PR is a lot about moving around components, so best checkout the branch and see how you like the other folder structure.

I'd like to make these changes, to 

1. simplify the project structure
2. prepare for the error handling (as we'll need two error handlers: one for responsive screens and one for the normal web screens)

---

As an overview here is what has changed:

- separate folder for all `mui-components` (feel free to recommend a different name)
- a directory `hooks` in the root of the project. But I'm not really sure whether this makes sense, as hooks can also contain components, which then must be separated.
- `ErrorHandler` for web screens
- moved the `globalArrayBuffersManager.ts` to `utils`
- moved all non-UI files in application into a `application/util` directory


michael-markl (approved):
I generally like the refactoring.

Another idea to improve overall structure is to not structure by "type of file" (components vs hooks vs utils) at the top, but by "module" and then further down by "type of file" if necessary.

However, as it is quite important to separate mui and bp (blueprint), we could have `mui-modules` and `bp-modules` folders and maybe several "shared" UI-agnostic modules at the top.

- src
  - mui-modules
    - application
      - components
      - hooks
      - ...
    - application-verification
      - components
      - hooks
      - ...
  - bp-modules
    - applications-overview
      - components
      - hooks
      - ...
    - users
      - components
      - hooks
      - ...
    - ... 
  - errors
    - components
    - hooks
    - ...
  - ... 


I'm not saying this is necessarily a better structure, but that would have been my first idea if I had to refactor :D

-------------


> moved the globalArrayBuffersManager.ts to utils

I think that one belongs to the application module (it's only used there afaik)

f1sh1918 (approved):
i'm fine with that, even i would prefer removing blueprint at one point and only use responsive mui components but i think thats way out of scope :)

michael-markl:
> i'm fine with that, even i would prefer removing blueprint at one point and only use responsive mui components but i think thats way out of scope :)

It's worth trying I think. I don't think the effort for migrating completely to mui would be huge, but the priority is quite low I think...
It would definitely reduce unnecessary complexity.

sarahsporck:
> I generally like the refactoring.
> 
> Another idea to improve overall structure is to not structure by "type of file" (components vs hooks vs utils) at the top, but by "module" and then further down by "type of file" if necessary.
> 
> However, as it is quite important to separate mui and bp (blueprint), we could have `mui-modules` and `bp-modules` folders and maybe several "shared" UI-agnostic modules at the top.

I think what you suggest is similar to how we had it in the integreat project some time ago. And back then we decided to move to a more "flat" project structure, as it became really hard to find specific components just by checking out the file structure. 
However I do understand, why it can make sense here, as the current structure is already modular oriented.

> 
> > moved the globalArrayBuffersManager.ts to utils
> 
> I think that one belongs to the application module (it's only used there afaik)

That's probably an argument for a modular filestructure.

michael-markl:
Yep, I also don't like the "deepness" of my suggestion :D
In my opinion, the last folder structure (ie. 
 *  components
 *  hooks
 * ...
)
is often not necessary. So I would (personally!) probably rather flat the structure on that end rather than the opposite end :D

But I guess that's up to you ^^

sarahsporck:
So I decided to not use any `components` folder and instead keep components as flat as possible (but sorted by module). And for hooks and other utilities folders are created as locally as possible, but only if necessary. 
This results in the following structure:

(with * marked entries are optional and should be omitted if no member present)
- bp-modules
    - module A
        - {...components}
        - util*
        - hooks*
    - module B
        - ...
- mui-modules
    - module A
        - {...components}
        - util*
        - hooks*
    - module B
        - ...
- errors
- hooks
- util
- {...components}


Further I tried moving all current elements into the correct folder. 
And lastly, as the imports are quite all over the place I added a prettier plugin (which we already use in integreat) to sort the imports.
Once again I think it is best to checkout the code and evaluate yourself whether you find everything :)

michael-markl (approved):
Navigation could also be moved to src/bp-modules (and maybe renamed to NavigationBar)

I feel like useUpdateStateCallback.ts and useVersionLocallyStoredState.ts **could** be moved to mui-modules/application/hooks as they are exclusively used there. 

I would not move ApplicationErrorBoundary into `util` as it's a "standard" React component

The rest LGTM.

**Edit:** Documentation of the structure would be nice, maybe in administration/README.md?

administration/README.md
```diff
@@ -0,0 +1,36 @@
+# Administration (Digitale Druckerei)
+
+The administration project contains the react frontend for card and region management. As well as public pages to verify or withdraw card applications.
```
michael-markl, line 3:
```suggestion
The administration project contains the React frontends for three audiences:
* Applicants can apply for an entitlementcard
* Organizations can verify or reject application details
* Administrators can manage applications, cards, and regions.
```

PR #915, opened 2023-04-01
Another suggestion for Migrations

michael-markl:
@sarahsporck Same rules apply: Take what you want into your PR :D

I applied Timo's suggestions as well as changed a few lines here and there when testing everything
* Use nested transactions: Sometimes exposed commits a transaction (e.g. if you do SchemaUtils.create). Using no nested transactions actually commits these transactions onto the DB, which we don't want in the case of failure.
*  4-digit migration versions: Good for sorting by file name
* Reverted your changes with IntTable: I guess your initial implementation was just fine. I didn't know Exposed requires an `id` attribute for most its logic.
* Apply the actual baseline and adjust the sample migrations. 

I think with that we're pretty much set up. What's missing:
* On deploy, first run migrate, then run start. (and of course the manual first-deployment with migrate-skip-baseline :D)


sarahsporck:
Very nice! Thank you @michael-markl. I thought about loading sql files directly in the migrations, but realized that modern IDEs don't struggle with hilighting strings as sql. 
So... 👍🏼:)

Issue #920, opened 2023-04-04
[Staging] ProjectId not set


f1sh1918:
**Describe the bug**
Application can not be send since projectId was not found, since the hostname is used for projectId

**To Reproduce**
1. https://staging.bayern.ehrenamtskarte.app/beantragen
2. Try to send application
3. Check console error

**Expected behavior**
"bayern.ehrenamtskarte.app" is used as projectId and error does not occur

PR #921, opened 2023-04-04
920: Staging projectId

f1sh1918:
resolves https://github.com/digitalfabrik/entitlementcard/issues/920

### DIFF
--- a/administration/src/project-configs/ProjectConfigContext.tsx
+++ b/administration/src/project-configs/ProjectConfigContext.tsx
@@ -1,8 +1,7 @@
 import { createContext, ReactNode } from 'react'
 import getProjectConfig, { ProjectConfig } from './getProjectConfig'
 
-const STAGING_PREFIX = 'staging.'
-const projectConfig = getProjectConfig(window.location.hostname.replace(STAGING_PREFIX, ''))
+const projectConfig = getProjectConfig(window.location.hostname)
 
 export const ProjectConfigContext = createContext<ProjectConfig>(projectConfig)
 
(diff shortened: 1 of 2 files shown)


administration/src/project-configs/ProjectConfigContext.tsx
```diff
@@ -1,7 +1,8 @@
 import { createContext, ReactNode } from 'react'
 import getProjectConfig, { ProjectConfig } from './getProjectConfig'
 
-const projectConfig = getProjectConfig(window.location.hostname)
+const STAGING_PREFIX = 'staging.'
```
f1sh1918, line 4: maybe we can put this also in the `build.configs`

administration/src/project-configs/ProjectConfigContext.tsx
```diff
@@ -1,7 +1,8 @@
...
+const projectConfig = getProjectConfig(window.location.hostname.replace(STAGING_PREFIX, ''))
```
michael-markl, line 5: I think it would be nicer if we list the different hostnames inside `getProjectConfig`:

```ts
const getProjectConfig = (hostname: string): ProjectConfig => {
  switch (window.localStorage.getItem('project-override') ?? hostname) {
    case 'bayern.ehrenamtskarte.app':
    case 'staging.bayern.ehrenamtskarte.app':
      return bayernConfig
    case 'nuernberg.sozialpass.app':
    case 'staging.nuernberg.sozialpass.app':
      return nuernbergConfig
    default:
      console.debug('Falling back to showcase.')
      return showcaseConfig
  }
}
```

michael-markl (approved):

Issue #923, opened 2023-04-04
[Staging] Add CI workflow for staging environment


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As we want to have a staging system, we need a separate workflow that builds and deploys staging builds on the staging server

**Describe the solution you'd like**

- add env variable for staging api
- create separate workflow that will be triggered by merging in main
- adjust prod workflow to be only executed when a release branch was pushed

**Describe alternatives you've considered**
- trigger prod workflow manually

michael-markl:
I think the prod deployment still needs to be tested?

PR #924, opened 2023-04-04
923: CI staging environment

f1sh1918:
- added env file with api url for staging
- adjusted normal backend workflow to build with this env file and deploy on staging server
- added backend-production workflow to be triggered when a git tag was set

resolves https://github.com/digitalfabrik/entitlementcard/issues/923


administration/package.json
```diff
@@ -57,8 +58,9 @@
     "typescript": "^4.9.5"
   },
   "scripts": {
-    "start": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh); react-scripts start",
-    "build": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh); react-scripts build",
+    "start": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh);react-scripts start",
+    "build": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh);react-scripts build",
+    "build:staging": "set -ae; REACT_APP_APPLICATION_COMMIT=$(./application_commit.sh);env-cmd -f .env.staging react-scripts build",
```
michael-markl, line 63: Hmm... I feel like it's a bit overkill to have a new npm command (and a new dev dependency) just for setting the api url.
Can't we infer the API url from the administration's hostname?
(Similar to what you do in https://github.com/digitalfabrik/entitlementcard/pull/921)?

f1sh1918: Sure we can. But we can just assume that it's `staging. *`

.circleci/config.yml
```diff
@@ -67,7 +69,66 @@ workflows:
...
+          context:
+            - entitlementcard-apt
+          requires:
+            - backend-pack
+            - administration-pack
+            - martin-pack
+            - meta-pack
+          filters:
+            tags:
+              only: /.*/
```
michael-markl, line 131: I'm not a CircleCI expert, but maybe instead of duplicating the workflow just because we want to run one additional job of it for tags, we could maybe use a single `backend` workflow (as before), set the filters of each job to
```yaml
          filters:
            tags:
              only: /.*/
```
(which should make them run on every branch and on every tag)
and have two jobs of the kind `deploy`:
```yaml
      - deploy:
          name: deploy-staging
          server: staging.entitlementcard.app
          context:
            - entitlementcard-apt
          requires:
            - backend-pack
            - backend-health-check
            - administration-pack
            - martin-pack
            - meta-pack
          filters:
            branches:
              only:
                - main
      - deploy:
          name: deploy-production
          server: entitlementcard.app
          context:
            - entitlementcard-apt
          requires:
            - backend-pack
            - administration-pack
            - martin-pack
            - meta-pack
          filters:
            branches:
              ignore:
                - /.*/
            tags:
              only: /.*/
```

Again, I'm not sure if that works, but that was my idea when reading https://circleci.com/docs/workflows/#executing-workflows-for-a-git-tag

michael-markl (approved):

.circleci/config.yml
f1sh1918: no you are probably right. that makes sense. More expert than me :)

michael-markl:
Staging deployment failed unfortunately, https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/2461/workflows/72fd78b4-8eed-432f-92a9-2171fd3bcfe1/jobs/21619

PR #929, opened 2023-04-05
Fix deployment in CircleCI pipeline

svenseeberg:


f1sh1918 (approved):
looks fine,thx. would like to know why the other hostnames don't work with sftp

PR #931, opened 2023-04-06
add-postgis-extension

f1sh1918:
I'm not sure if this the right place but if i drop all tables and try to run the baseline migration i get the following error mentioned here:
https://stackoverflow.com/questions/6850500/postgis-installation-type-geometry-does-not-exist


f1sh1918:
not needed

PR #932, opened 2023-04-06
Upgrade mobile scanner to 3.1.1

michael-markl:
This needs testing (only the scanner screen) on Android and iOS.

### DIFF
--- a/frontend/pubspec.lock
+++ b/frontend/pubspec.lock
@@ -649,7 +649,7 @@ packages:
       path: "."
       ref: f35709343e03439746857a9fbbd23dbb9bbafafc
       resolved-ref: f35709343e03439746857a9fbbd23dbb9bbafafc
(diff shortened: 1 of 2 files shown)


f1sh1918 (approved):
Tested on real devices android and ios. I'm just wondering why this change in the `pubspec` is not reflecting in the `podfile.lock` when `pod install` is triggered. It would be nice if i could commit it

michael-markl:
> Tested on real devices android and ios.

Great, thanks!

> I'm just wondering why this change in the pubspec is not reflecting in the podfile.lock when pod install is triggered. It would be nice if i could commit it

Hmm not sure, maybe the podfile.lock doesn't change for this minor update? (I don't really know when exactly pods change...) If there are no changes on your machine, then probably nothing changed?
Otherwise, feel free to commit & push.

Please merge when you're ready :)

PR #933, opened 2023-04-07
add timestamp to apt pkg

f1sh1918:
add apt pkg timestamp


f1sh1918:
just test

.circleci/config.yml
```diff
@@ -66,10 +66,7 @@ workflows:
             - administration-pack
             - martin-pack
             - meta-pack
-          filters:
-            branches:
-              only:
-                - main
```
michael-markl, line 72: I think this should not be removed?

PR #934, opened 2023-04-11
873: Convert exceptions to graphqlerrors

sarahsporck:
Sorry, again a PR with a bunch of changes.
Summarized, I tried to use GraphqlErrorException for all Exceptions that make sense to be passed to the frontend (to display a descriptive message). In that sense be critical when reviewing. If you feel like an Exception does not make sense I am very open for suggestions. 
For errors we don't expect in the frontend we use IllegalArgumentExceptions for errors during json parsing and error codes for other cases.

- [401] UnauthorizedException
- [403] ForbiddenException
- [404] NotFoundException and ProjectNotFoundException
- [500] IOException and ExecutionException


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordQueryService.kt
```diff
@@ -2,41 +2,28 @@ package app.ehrenamtskarte.backend.auth.webservice.schema
...
 
-class PasswordResetKeyExpiredException() : GraphqlErrorException(
-    newErrorException().extensions(
-        mapOf(
-            Pair("code", "PASSWORD_RESET_KEY_EXPIRED")
-        )
-    )
-)
-
-class InvalidPasswordResetLinkException() : GraphqlErrorException(
```
f1sh1918, line 22: this shouldn't be deleted since there is a separate template using this code

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordQueryService.kt
```diff
@@ -2,41 +2,28 @@ package app.ehrenamtskarte.backend.auth.webservice.schema
...
     fun checkPasswordResetLink(project: String, resetKey: String): Boolean {
         return transaction {
-            val projectId = ProjectEntity.find { Projects.project eq project }.single().id.value
+            val projectEntity = ProjectEntity.find { Projects.project eq project }.firstOrNull() ?: throw ProjectNotFoundException(project)
+            val projectId = projectEntity.id.value
             val admin = AdministratorEntity
                 .find { Administrators.passwordResetKey eq resetKey and (Administrators.projectId eq projectId) and not(Administrators.deleted) }.singleOrNull()
             if (admin == null) {
-                throw InvalidPasswordResetLinkException()
+                throw InvalidLinkException()
```
f1sh1918, line 26: please undo

f1sh1918 (commented):
Looks fine to me. Just the one exception is needed for giving a better feedback to the user

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/JwtService.kt
```diff
@@ -13,7 +14,7 @@ object JwtService {
     private val secret by lazy {
         System.getenv("JWT_SECRET")
             ?: readJwtSecretFile(System.getenv("JWT_SECRET_FILE"))
-            ?: throw NoSuchElementException("Environment variable JWT_SECRET not set")
+            ?: throw MissingAuthorizationException()
```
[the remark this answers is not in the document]
sarahsporck: Hmm. I tried to keep this behavior the same as before. An `UnauthorizedException` would lead to 401, while the `NoSuchElementException` would be forwarded to the frontend. But maybe you're right. I'll change it to `UnauthorizedException`

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordQueryService.kt
sarahsporck: Jep. You're right. That change was actually intended. When working with HTTP clients you also only work with a fixed amount of response codes, which you use for all endpoints. Depending on the endpoint the error code can have a different meaning. That's what I tried to do here. 
Of course the frontend code needs to be adjusted for this as well, which will be done in the follow up PR. 

I'll move that change in the other PR and revert here. This way I'll not accidentally introduce an error here.

sarahsporck:
As I started on implementing the exception handling in the frontend I realiyed that some Exceptions didn't really make much sense, so I removed/refactored them. 
Instead of EmptyInput, TooLongInput, TooZoung and InvalidDateFormat there is now an InvalidJson Exception.

f1sh1918 (approved):
looks fine

PR #935, opened 2023-04-11
884: Implement backend switch

### LINKED ISSUE #884: Implement backend switch
As a tester i want to test new features on a staging environment. Therefore i need the option to switch between testing and production backend in the mobile app.

**Acceptance Criteria**

- Implement a hidden backend switch protected by a password

f1sh1918:
Testing:
Go to About
Tap 10x Icon
Insert password
Switch Endpoint

Open again to check current endpoint

resolves #884


michael-markl (commented):
A few remarks:

1. The mapstyle url should probably also switch to the staging one.
2. We should probably delete all local state when switching backend. Otherwise we might run into weird state.
3. Probably we want to persist the fact that we switched the backend to staging, so that we can also test the app start 
4. If the app is using "staging", then we should probably add a textual hint somewhere, so that the tester knows which environment he's currently testing against. Optimally, we would show something like a transparent watermark above the whole app :D
5. I am not sure how well our previous "app environment" model fits your changes and the current situation with staging: Currently, we have an "app environment" in {production, showcase, local} (see frontend/lib/configuration/definitions.dart). Imo, "showcase" does not really make sense any longer? For me, "showcase" is just a sample project that we did not do anything with. I would suggest to remove "showcase" from our "app environments" and add a "staging" environment. (However @maxammann was mostly responsible for introducing the app environments, so he should know :D )

f1sh1918:
> 

1. this makes sense
2. what do you mean by local state? Can you give me an example
3. Ah okay so i think we could write this into settings and check it while initState
4. Gonna see what i can do
5. i agree. Hm the question is if we want to show features that are not yet released (staging). Else staging could be the new showcase. For the current showcase we need some more configuration and i don't see thats worth to do it

michael-markl:
> what do you mean by local state? Can you give me an example

Basically everything stored on the device, like settings, activated card, etc. (I think that's already it).

frontend/lib/configuration/configuration.dart
```diff
@@ -1,6 +1,23 @@
...
+
+  @override
+  bool updateShouldNotify(covariant _Configuration oldWidget) =>
+      configuration.mapStyleUrl != oldWidget.configuration.mapStyleUrl ||
+      configuration.graphqlUrl != oldWidget.configuration.graphqlUrl ||
+      configuration.projectId != oldWidget.configuration.projectId ||
+      configuration.showDevSettings != oldWidget.configuration.showDevSettings;
+}
+
+class Configuration extends StatefulWidget {
```
michael-markl, line 19: I think it would be better not to make this a Stateful Widget.
Instead I think, we should probably simply update "the props" (React speak) of the `Configuration InheritedWidget` based on a settings change.
Currently, the "Configuration" widget is the first widget in our app widget tree. I think, we could make the SettingsProvider the first widget, and then add a stateless widget (maybe call it `ConfigurationProvider`) below it that watches the `settings.stagingEnabled` property and passes the correct values of `graphqlUrl` and `mapStyleUrl` to `Configuration`. [Currently, these values are determined in main.dart. This should be moved to the mentioned ConfigurationProvider]

That way we wouldn't have two places where the URLs are determined and we can reduce overall complexity as we don't have duplicate state in the settings and a `Stateful Configuration` widget

frontend/lib/home/home_page.dart
```diff
@@ -109,6 +122,27 @@ class HomePageState extends State<HomePage> {
...
+    if (SchedulerBinding.instance.schedulerPhase != SchedulerPhase.idle) {
+      // wait for the end of that frame.
+      await SchedulerBinding.instance.endOfFrame;
+      if (!mounted) return false;
+    }
+
+    final config = Configuration.of(context);
+    config.switchEndpoint(settings.graphqlUrl, settings.mapStyleUrl);
+    return true;
+  }
```
michael-markl, line 145: I think this hack should not be necessary if  you apply the above suggestions.

michael-markl:
> is that sufficient?

Banner looks nice :D

frontend/lib/configuration/configuration.dart
f1sh1918: reverted configuration class and changed order of providers.

frontend/lib/configuration/settings_model.dart
```diff
@@ -16,12 +17,17 @@ class SettingsModel extends ChangeNotifier {
...
       _locationFeatureEnabled = await loadLocationFeatureEnabled();
+      _staging = await loadStaging();
     } on Exception {
       _preferences?.clear();
     }
     return this;
   }
 
+  Future<void> clearPreferences() async {
+    _preferences?.clear();
```
michael-markl, line 28: I think we should 
* Reset all local fields
* notify the listeners
* await the `clear` promise

frontend/lib/about/backend_switch_dialog.dart
```diff
@@ -0,0 +1,109 @@
...
+  void switchBackendUrl(BuildContext context) {
+    final config = Configuration.of(context);
+    final String endpoint = config.graphqlUrl;
+    switch (endpoint) {
+      case 'https://api.entitlementcard.app':
+        {
+          updateSettings(stagingEnabled: true);
+          break;
+        }
+      case 'https://api.staging.entitlementcard.app':
```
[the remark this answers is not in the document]
f1sh1918: we can :)

frontend/lib/configuration/settings_model.dart
```diff
@@ -60,8 +77,10 @@ class SettingsModel extends ChangeNotifier {
 
   bool get locationFeatureEnabled => _locationFeatureEnabled;
 
+  bool get staging => _staging;
```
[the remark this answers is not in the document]
f1sh1918: i prefer `enableStaging`  tbh

frontend/lib/configuration/settings_model.dart
f1sh1918: added `notify` and await not sure what else to reset is missing?

michael-markl: I think we need to reset
      _firstStart 
      _hideVerificationInfo 
      _locationFeatureEnabled 
     _staging

f1sh1918: But isn't this done by the clear function?

f1sh1918: Okay i'm probably lacking some basic concepts of flutter...
Every time the `notifyListeners()` was called the settings gonna be `initialize()` because the `FutureBuilder<Settings>` recognizes that.
So when i `setEnableStaging` the `SettingsModel` gonna get `initialize()` so even if i reset these fields you describe they gonna be overwritten immediately in the `initialize` fkt and set to whatever in the loadFunctions is marked as default value . So i don't get the point of doing that tbh

frontend/lib/configuration/settings_model.dart
```diff
@@ -16,12 +17,22 @@ class SettingsModel extends ChangeNotifier {
...
       _preferences?.clear();
     }
     return this;
   }
 
+  Future<void> clearSettings() async {
+    _firstStart = true;
+    _hideVerificationInfo = false;
+    _locationFeatureEnabled = false;
+    _enableStaging = false;
```
f1sh1918, line 31: or can i just create
`SettingsModel()` to reset the values?

frontend/lib/configuration/settings_model.dart
michael-markl: > Okay since the notifyListeneres triggers another initialize of the SettingsModel

I'm surprised about this, and I believe/hope this should not be true.

> why the default values and the fallback values differ.

That's weird indeed. Imo, the default values are not relevant (as we initialize them) . We might be able to get rid of them by using the `late` keyword.

michael-markl (approved):

PR #938, opened 2023-04-12
Fix application version

michael-markl:
This fixes the issue that the application version would not increase any longer after the recent refactor in #912.

I added another check to the bash script verifying that the application source folder still exists.


f1sh1918 (approved):
Version was updated during test :)

PR #939, opened 2023-04-12
Minor improvements for printing

michael-markl:
@f1sh1918 : When testing #937 I noticed a few minor quirks:
1. When canceling the print dialog, I could not restart it (on the same application).
2. Minor visuals: Everything centered, weird boxshadow, wrong text for attachments

This was my attempt to fixing these on the fly. I also refactored very slightly. Feel free to take these suggestions into your PR.


f1sh1918:
> @f1sh1918 : When testing #937 I noticed a few minor quirks:
> 
> 1. When canceling the print dialog, I could not restart it (on the same application).
> 2. Minor visuals: Everything centered, weird boxshadow, wrong text for attachments
> 
> This was my attempt to fixing these on the fly. I also refactored very slightly. Feel free to take these suggestions into your PR.

1. Alright this is way easier thx. Seems like the event listener is only working for chrome, because i faced the same issue before. but in this case not really needed, nice.

gonna merge it

PR #946, opened 2023-04-18
639: Invalidate card frontend

### LINKED ISSUE #639: Invalidate cards in frontend after they expired
As a user I want to see that a card is no longer valid in the app. For example after 2 years a card expires and is no longer valid. 

If the card is no longer valid (regardless of the reason), then disable the card in the frontend.

**Only take into account the expiration date (from the anmeldecode)**

## Acceptance Criteria:
 - [ ] Add a expiration check to the frontend. If the card expired, then indicate that in the frontend

### LINKED ISSUE #677: Check the validity of a card against the backend regularly

Disable cards which become invalid because of their expiration date or because they have been revoked, or because the same Anmeldecode was scanned twice.
Check if cards are still valid by verifying them against the backend.

An invalid card should stay in the app but it should be grayed out or a thick text should be displayed above it.

If the card becomes valid again then enable it again.

- [ ] Depends on: #17 

TBD

None

**Additional context**
None

f1sh1918:
Testing:
1. Change `expirationDay: 19446,` in dev_settings_view
2. Set invalid card in dev settings

resolve #639 
resolves #677


frontend/lib/identification/util/card_info_utils.dart
```diff
@@ -19,3 +19,10 @@ extension Hashing on CardInfo {
     return utf8.encode(canonicalizedJsonString);
   }
 }
+
+bool isCardExpired(CardInfo cardInfo) {
+  final expirationDay = cardInfo.hasExpirationDay() ? cardInfo.expirationDay : null;
+  return expirationDay == null
+      ? true
```
michael-markl, line 26: Should be false I guess

f1sh1918: yes

sarahsporck (commented):
Shouldn't we check with the backend if the card is still valid as it could not only have been exbired but there is also the possibility it's been revoked?

frontend/lib/identification/card_detail_view/more_actions_dialog.dart
```diff
@@ -19,37 +19,39 @@ class MoreActionsDialog extends StatelessWidget {
...
-            title: Text(localization.verifyTitle),
-            subtitle: Text(localization.verifyDescription),
-            leading: const Icon(Icons.verified, size: 36),
-            onTap: () {
-              Navigator.pop(context);
-              startVerification();
-            },
-          ),
-        ],
+      content: SingleChildScrollView(
```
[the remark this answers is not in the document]
f1sh1918: because the text is overflowing on some small screens

f1sh1918:
> #639

good point. I just read the criteria in the issue which said. Take the expirationDate from the Anmeldecode.
The point is right now we have no functionality that revokes cards, so i think its not needed yet and should be implemented when this functionality is provided

frontend/lib/identification/card_detail_view/card_detail_view.dart
```diff
@@ -107,6 +108,7 @@ class RichQrCode extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final bool isExpired = isCardExpired(userCode.info);
```
michael-markl, line 111: Minor: I think `isExpired` should be a "prop" for `RichQrCode`. Then we only evaluate `isCardExpired` once in `CardDetailView`.

michael-markl:
> Shouldn't we check with the backend if the card is still valid as it could not only have been exbired but there is also the possibility it's been revoked?

I think we should do both.
* Check locally if the card has expired (which would be quite annoying to work around (for the user) by changing the system time)
* Check via API request if it is still valid (can be worked around rather easily by using flight mode)

Of course, we could also make the app API-validate the card on every start, and if it this failed (because of connection errors) for two weeks (which again can be tricked by changing system time) we don't show the QR code.

We should probably API-validate the card on app-start anyways, as the card might even get invalidated by scanning the anmeldecode on another device. Then, it would be great if the user could see this right away, after relaunching the app on the old device.

f1sh1918:
> > Shouldn't we check with the backend if the card is still valid as it could not only have been exbired but there is also the possibility it's been revoked?
> 
> I think we should do both:
> 
> * Check locally if the card has expired (which would be quite annoying to work around (for the user) by changing the system time)
> * Check via API request if it is still valid (can be worked around rather easily by using flight mode)
> 
> Of course, we could also make the app API-validate the card on every start, and if it this failed (because of connection errors) for two weeks (which again can be tricked by changing system time) we don't show the QR code. In that case, we can probably omit the local expiration date check.
> 
> We should probably API-validate the card on app-start anyways, as the card might even get invalidated by scanning the anmeldecode on another device. Then, it would be great if the user could see this right away, after relaunching the app on the old device.

Not sure if you read my comment above. I'm aware these checks are need at some point and therefore is another ticket but for now we don't have any functionality for  "scanning anmeldecode on other device", that it gets invalid yet (currently you can use it on x devices at least as i tested). I also mentioned this in the other issue #677. 
The revoke functionality has a lot of issues to consider and is not implemented yet. Until there is no possibility to revoke the card there is no need to check this in my opinion because the value will stay `false` and the `expirationDate` is also immutable in the backend

michael-markl:
In this whole conversation we use "Anmeldecode" synonomously to "Aktivierungscode"? 

> [...] for now we don't have any functionality for "scanning anmeldecode on other device", that it gets invalid yet (currently you can use it on x devices at least as i tested)

Afaik, this is (or at least should be) wrong. Once you scan an "Aktivierungscode", the app makes an API-request ("activate card"), the backend generates a new TOTP secret and overwrites the old TOTP secret (thus invalidating the old activation).

f1sh1918:
> In this whole conversation we use "Anmeldecode" synonomously to "Aktivierungscode"?
> 
> > [...] for now we don't have any functionality for "scanning anmeldecode on other device", that it gets invalid yet (currently you can use it on x devices at least as i tested)
> 
> Afaik, this is (or at least should be) wrong. Once you scan an "Aktivierungscode", the app makes an API-request ("activate card"), the backend generates a new TOTP secret and overwrites the old TOTP secret (thus invalidating the old activation).

Yes we do. Well i tested the same code on two devices locally. I can repeat that if needed

michael-markl:
> [...] So we need a visible feedback on the first device to also look invalid :)

Right, that's why we need to API-validate on a regular basis, to actually detect this.

> Then this was a missunderstanding i expected the activation code not to work on a second device and that the user has to revoke first on the old device (which is not yet implemented).

Just to clarify: We never planned to allow the user to revoke his card; only administrators should be able to `revoke` a card. (Also once revoked, it cannot be activated).

f1sh1918:
Alright just as a summary:

- Check regulary against the CurrentTimestamp of the backend if the card is valid
- check the topt secret
- we could save the last check date in the device settings
- if no check was possible since 1 week or sth f.e due to no Internet, we could deactivate the qr code and tell the User to enable Internet and sync with the backend

What do you think? Did I forget sth?

michael-markl:
> What do you think

sounds good to me.

f1sh1918:
Ok i updated the pr. Now additionally to the frontend check i validate:

- on each app start when the "Ausweisen" Tab was opened its getting checked for validity and a timestamp of this check will be written in the settings.
- if this check wasn't possible for more than 7 days (no internet) the qrCode will be hidden with a message to enable internet and to resync the card.
- if the verification process returns invalid due to card is `revoked` or old `totp` was sent, the card will be removed from the device.
Note: There is no alert yet if you use the same Anmeldecode for another device #949

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/types/CardVerificationResultModel.kt
```diff
@@ -0,0 +1,6 @@
+package app.ehrenamtskarte.backend.verification.webservice.schema.types
+
+data class CardVerificationResultModel(
+    val valid: Boolean,
+    val lastCheck: String
```
michael-markl, line 5:
```suggestion
    val verificationTimeStamp: String
```

frontend/lib/identification/card_detail_view/verification_code_view.dart
```diff
@@ -57,32 +65,69 @@ class VerificationCodeViewState extends State<VerificationCodeView> {
...
-                        foregroundColor: Theme.of(context).textTheme.bodyMedium?.color,
-                        gapless: false,
+            return isVerificationExpired
+                ? IconButton(
+                    icon: const Icon(Icons.refresh),
+                    color: Theme.of(context).appBarTheme.backgroundColor,
+                    onPressed: () {
+                      verifyCard(otpCode, userCode, context);
+                    },
+                  )
```
michael-markl, line 75: I haven't built the app and looked at it yet, but I suppose, we need some textual explanation of what's going on here.

frontend/lib/identification/card_detail_view/verification_code_view.dart
```diff
@@ -57,32 +65,69 @@ class VerificationCodeViewState extends State<VerificationCodeView> {
...
+    final DynamicVerificationCode qrCode = DynamicVerificationCode(
+      info: userCode.info,
+      pepper: userCode.pepper,
+      otp: otpCode.code,
+    );
+    final cardVerification = await queryDynamicServerVerification(client, projectId, qrCode);
+    if (cardVerification.valid) {
+      updateLastCardVerification(cardVerification.lastCheck);
+    } else {
+      removeInvalidCard();
```
michael-markl, line 120: I think we should not remove the invalid card automatically.
In the case of an implementation/configuration error on the backend (say we accidentally always return `invalid` because we're connected to an empty DB) we don't want to automatically delete all cards in every user's app.)

Instead, we should probably persist, that the card is invalid. Probably we should persist both the timestamp and the `valid` bool.

frontend/lib/identification/util/card_info_utils.dart
```diff
@@ -19,3 +19,14 @@ extension Hashing on CardInfo {
...
+
+bool isCardExpired(CardInfo cardInfo) {
+  final expirationDay = cardInfo.hasExpirationDay() ? cardInfo.expirationDay : null;
+  return expirationDay == null
+      ? false
+      : DateTime.fromMillisecondsSinceEpoch(0).add(Duration(days: expirationDay)).isBefore(DateTime.now());
+}
+
+bool isVerificationCheckExpired(String lastCheck) {
+  return DateTime.now().toUtc().isAfter(DateTime.parse(lastCheck).add(Duration(days: 7)));
```
michael-markl, line 31: We should probably also check that `now` is greater or equal lastVerificationTimeStamp

michael-markl:
> Does it make sense to keep the card in the app if invalid like revoked or old totp?

As I explained in one of my comments above: YES

>  Maybe then we should send back different VerificationStates for each case

We cannot detect whether the otp code sent along with the request belongs to an old totp secret or is simply invalid (as we don't keep old totp secrets). I think it's enough to just show some text like "this card is invalid and these are the possible reasons: [possible reasons]".

frontend/lib/identification/card_detail_view/verification_code_view.dart
f1sh1918: i propose a text above with last commit

frontend/lib/identification/card_detail_view/verification_code_view.dart
f1sh1918: i updated the this in the last commit. thats why i converted it to draft yesterday. But in any case thx for the comments. 

I just have to move the verification variables from settings to the  `UserCode` Provider

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardQueryService.kt
```diff
@@ -5,23 +5,26 @@ import app.ehrenamtskarte.backend.exception.service.ProjectNotFoundException
...
 import graphql.schema.DataFetchingEnvironment
+import java.time.Instant
 import java.util.Base64
 
 @Suppress("unused")
 class CardQueryService {
-    @GraphQLDescription("Returns whether there is a card in the given project with that hash registered for that this TOTP is currently valid")
-    fun verifyCardInProject(project: String, card: CardVerificationModel, dfe: DataFetchingEnvironment): Boolean {
+    @GraphQLDescription("Returns whether there is a card in the given project with that hash registered for that this TOTP is currently valid and a timestamp of the last check")
+    fun verifyCardInProject(project: String, card: CardVerificationModel, dfe: DataFetchingEnvironment): CardVerificationResultModel {
```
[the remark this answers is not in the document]
f1sh1918: Hm okay and you think its necessary to get the timestamp from the backend instead of using the `DateTime` from the client app. In any case we can not avoid manipulation of the system time. So we could also use the client side timestamp

frontend/lib/configuration/settings_model.dart
```diff
@@ -32,6 +32,25 @@ class SettingsModel extends ChangeNotifier {
...
+    }
+    return obj;
+  }
+
+  String get lastCardVerification => _getString("lastCardVerification") ?? '1970-01-01T00:00:00.000000Z';
+
+  Future<void> setLastCardVerification({required String lastVerification}) async {
+    await _preferences?.setString("lastCardVerification", lastVerification);
+    notifyListeners();
+  }
```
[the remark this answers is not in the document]
f1sh1918: So the `DynamicUserCode` should be extended by `verificationTimestamp` and `cardValid`?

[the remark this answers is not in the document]
michael-markl: Adding protobuf fields does not break. All fields are optional by default. usually.

frontend/lib/identification/util/card_info_utils.dart
```diff
@@ -19,3 +19,10 @@ extension Hashing on CardInfo {
     return utf8.encode(canonicalizedJsonString);
   }
 }
+
+bool isCardExpired(CardInfo cardInfo) {
+  final expirationDay = cardInfo.hasExpirationDay() ? cardInfo.expirationDay : null;
+  return expirationDay == null
+      ? false
+      : DateTime.fromMillisecondsSinceEpoch(0).add(Duration(days: expirationDay)).isBefore(DateTime.now());
+}
```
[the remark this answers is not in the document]
f1sh1918: alright i understood the point and added an `toleranceInHours` and a comment

frontend/lib/identification/card_detail_view/verification_code_view.dart
f1sh1918: Now the card is shown as invalid if revoked or activated on other device

frontend/lib/identification/util/card_info_utils.dart
f1sh1918: i updated this to check and added a tolerance of 
```js
// 7days x 24h + 1day for expiration day and 12h for UTC+12 = 192
int cardValidationExpireHours = 192;
```

sarahsporck (approved):
Tested and works fine. 
Bildschirmfoto 2023-04-27 um 15 16 46
However I find the revalidation doesn't look very pleasing. Especially the icon followed by "Weitere Aktionen". Maybe putting the icon above the text or also adding a text to the icon can help. Maybe @michael-markl has an idea.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardQueryService.kt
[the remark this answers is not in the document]
michael-markl: I think it would be slightly better to use a timestamp created in the backend: Then whenever the app *does* have internet connection, the timestamp is updated, and the user would have to have set his system date/time to the backend time. It is a bit harder to workaround.
I'm not sure, if it's worth the increased complexity though. You have to decide :D

frontend/lib/identification/card_detail_view/verification_code_view.dart
```diff
@@ -27,6 +34,8 @@ class VerificationCodeViewState extends State<VerificationCodeView> {
   void initState() {
     super.initState();
     _otpCode = widget._otpGenerator.generateOTP(_resetQrCode);
+    // On every app start when this widget will be initialized, we verify with the backend if the card is valid
```
michael-markl, line 37:
```suggestion
    // On every app start when this widget will be initialized, we verify with the backend if the card is valid, in order to detect if one of the following events happened:
    // - the card was activated on another device
    // - the card was revoked
    // - the card expired (on backend's system time)
```

frontend/lib/identification/activation_workflow/activation_code_scanner_page.dart
```diff
@@ -97,12 +98,13 @@ class ActivationCodeScannerPage extends StatelessWidget {
...
-          totpSecret: totpSecret,
-        );
-        provider.setCode(userCode);
+
+        provider.setCode(DynamicUserCode(
+            info: activationCode.info,
+            pepper: activationCode.pepper,
+            totpSecret: totpSecret,
+            cardVerification:
+                CardVerification(cardValid: true, verificationTimeStamp: daysSinceEpoch(DateTime.now().toUtc()))));
```
michael-markl, line 107: We should optimally also use a backend created timestamp here.
The good thing here: We would not break the API, if we add a field `activationTimeStamp` to `activationResult`.

specs/card.proto
```diff
@@ -58,6 +59,11 @@ message CardInfo {
   optional CardExtensions extensions = 3;
 }
 
+message CardVerification {
+  optional bool cardValid = 1;
+  optional uint32 verificationTimeStamp = 2;
```
michael-markl, line 64: I think we should store it as secondsSinceEpoch (as this is the more-or-less default to store time stamps). And we don't have to think about how we round to days.

```suggestion
  
  optional uint64 verification = 2;
```

michael-markl (changes requested):
Except one comment, I'm fine with these changes.

Good stuff  🚀

specs/card.proto
f1sh1918: @michael-markl but the need a separate package for dart? Or how to deal with `Int64` ?
https://pub.dev/packages/fixnum

michael-markl: I think this dependency is Okay, it's at least from flutter.dev

f1sh1918:
> Tested and works fine. Bildschirmfoto 2023-04-27 um 15 16 46 However I find the revalidation doesn't look very pleasing. Especially the icon followed by "Weitere Aktionen". Maybe putting the icon above the text or also adding a text to the icon can help. Maybe @michael-markl has an idea.

i think this case is happening very rarely. I added a label to the button. But to change order means to change at lot in the widgets which is not really necessary imo
image

Added deprecation message, v2 and the adjustment of the data type to uint64

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardQueryService.kt
```diff
@@ -5,23 +5,26 @@ import app.ehrenamtskarte.backend.exception.service.ProjectNotFoundException
...
         if (card.codeType == CodeType.STATIC) {
-            return card.totp == null && CardVerifier.verifyStaticCard(project, cardHash, projectConfig.timezone)
+            return CardVerificationResultModel(card.totp == null && CardVerifier.verifyStaticCard(project, cardHash, projectConfig.timezone), timestamp)
         } else if (card.codeType == CodeType.DYNAMIC) {
-            return card.totp != null && CardVerifier.verifyDynamicCard(project, cardHash, card.totp, projectConfig.timezone)
+            return CardVerificationResultModel(card.totp != null && CardVerifier.verifyDynamicCard(project, cardHash, card.totp, projectConfig.timezone), timestamp)
         }
-        return false
+        return CardVerificationResultModel(false, timestamp)
     }
```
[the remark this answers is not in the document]
f1sh1918: check last commit

frontend/lib/identification/util/canonical_json.dart
```diff
@@ -24,6 +24,8 @@ extension CanonicalJson on GeneratedMessage {
         map[field.tagNumber.toString()] = value;
       } else if (value is int) {
         map[field.tagNumber.toString()] = value.toString();
+      } else if (value is bool) {
+        map[field.tagNumber.toString()] = value.toString();
```
michael-markl, line 28: Why do we need this? I think we should revert it.

frontend/lib/util/date_utils.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:fixnum/fixnum.dart';
+
+Int64 secondsSinceEpoch(DateTime date) {
+  return Int64(date.difference(DateTime(1970)).inSeconds);
```
michael-markl, line 4: I think it would be better to use (date.toUtc().millisecondsSinceEpoch / 1000).round() or similar. I think DateTime(1970) creates a datetime in local time zone, but not sure.

specs/card.proto
```diff
@@ -58,6 +59,11 @@ message CardInfo {
   optional CardExtensions extensions = 3;
 }
 
+message CardVerification {
+  optional bool cardValid = 1;
+  optional uint64 verificationTimeStamp = 2;
```
michael-markl, line 64: Please add a comment about unit.

frontend/lib/identification/util/canonical_json.dart
f1sh1918: i couldn't write and read from `SecureStorage` without this check, because i added a bool to the protobuf

frontend/lib/util/date_utils.dart
f1sh1918: this should also do the trick `DateTime(1970).toUtc()` without any further calculation?

specs/card.proto
```diff
@@ -58,6 +59,16 @@ message CardInfo {
   optional CardExtensions extensions = 3;
 }
 
+message CardVerification {
+  optional bool cardValid = 1;
+
+  // Verification timestamp in seconds (calculated from 1970-01-01).
```
michael-markl, line 65:
```suggestion
  // Verification timestamp in seconds (calculated from 1970-01-01 UTC).
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardQueryService.kt
```diff
@@ -5,13 +5,15 @@ import app.ehrenamtskarte.backend.exception.service.ProjectNotFoundException
...
 import app.ehrenamtskarte.backend.verification.webservice.schema.types.CardVerificationModel
+import app.ehrenamtskarte.backend.verification.webservice.schema.types.CardVerificationResultModel
 import com.expediagroup.graphql.generator.annotations.GraphQLDescription
 import graphql.schema.DataFetchingEnvironment
 import java.util.Base64
 
 @Suppress("unused")
 class CardQueryService {
-    @GraphQLDescription("Returns whether there is a card in the given project with that hash registered for that this TOTP is currently valid")
+    @Deprecated("Deprecated in favor of CardVerificationResultModel that return a current timestamp", ReplaceWith("verifyCardInProjectV2"))
```
michael-markl, line 15:
```suggestion
    @Deprecated("Deprecated since May 2023 in favor of CardVerificationResultModel that return a current timestamp", ReplaceWith("verifyCardInProjectV2"))
```

frontend/lib/identification/util/canonical_json.dart
michael-markl: Probably it was because you added the fields to cardInfo instead of userCode.

frontend/lib/util/date_utils.dart
michael-markl: I think not: DateTime(1970).toUtc() would create a DateTime at 1.1.1970 00:00 **local time zone** and then convert it to UTC, resulting in, depending on the local time zone, 31.12.1969 22:00 UTC. You can however use `DateTime.utc(1970)` to initialize a date at UTC timezone. So the following would work:
```dart
  return Int64(date.difference(DateTime.utc(1970)).inSeconds)
```

michael-markl (approved):
After addressing the remaining unresolved comments, you can merge imo 🚀

PR #947, opened 2023-04-18
884: Proposal: implement backend switch

michael-markl:
@f1sh1918. After taking another look, I think the SettingsModel was overly complicated. All those private fields are not necessary as _preferences.get is synchronous (and uses an in-memory lookup table). So the initialization boils down to calling SharedPreferences.getInstance() which reads the settings from disk.
The fact that `notifyListeners` triggered an `initialize` was because the EntryWidget actually subscribed to the SettingsModel and called the initialize future every time it was notified. I've refactored this, so that the SettingsProvider (a new Widget) takes care of initializing the SettingsModel before handing it over to the ChangeNotifierProvider. Then, the rest of the "EntryWidget" seemed to be trivial stuff which I moved to App to clean everything up (and finally to remove this weird EntryWidget)


f1sh1918:
> @f1sh1918. After taking another look, I think the SettingsModel was overly complicated. All those private fields are not necessary as _preferences.get is synchronous (and uses an in-memory lookup table). So the initialization boils down to calling SharedPreferences.getInstance() which reads the settings from disk. The fact that `notifyListeners` triggered an `initialize` was because the EntryWidget actually subscribed to the SettingsModel and called the initialize future every time it was notified. I've refactored this, so that the SettingsProvider (a new Widget) takes care of initializing the SettingsModel before handing it over to the ChangeNotifierProvider. Then, the rest of the "EntryWidget" seemed to be trivial stuff which I moved to App to clean everything up (and finally to remove this weird EntryWidget)

thx for the nice explanation and refactoring. At least i understand now this concept of FutureBuilder :)

frontend/lib/configuration/settings_model.dart
```diff
@@ -1,91 +1,67 @@
...
-    _hideVerificationInfo = enabled;
+    await _preferences?.setBool("hideVerificationInfo", enabled);
     notifyListeners();
-    await _preferences?.setBool(hideVerificationInfoPropertyName, enabled);
   }
 
-  Future<bool> loadLocationFeatureEnabled() async {
-    return _preferences?.getBool("location") ?? false;
-  }
+  bool get locationFeatureEnabled => _getBool("location") ?? true;
```
f1sh1918, line 56: shouldn't this be `false` per default?

f1sh1918 (approved):
Nice refactoring!

frontend/lib/configuration/settings_model.dart
michael-markl: Yes, that should be false. Can you maybe adjust this? Also feel free to merge it into your branch

PR #948, opened 2023-04-20
901: Activity log

### LINKED ISSUE #901: [Sozialpass] Implement Activity Log
As a user I want to get an overview of my activity during my session in the "digitale druckerei"

**Acceptance Criteria**
- the activity log can be found in "Benutzereinstellungen"
- the user can see which cards were created in this session

**Technical Tasks**
- show the activity log in the "Benutzereinstellungen " for sozialpass only
- create activity log entry when cards were created including (CreationDate, CardExpireDate, Passnumber, Birthday)
- delete the activityLog after logout (auto logout after 60min and manual logout)

f1sh1918:
- I decided to save all `CardBlueprint` data in the `SessionStorage` to be more generic for other project. Since its local on client and session based it should be fine.
- created a `ActivityLogConfig` in the `buildConfig` to define the `columnNames` of the table and to render a project specific `ActivityLogEntry` which can be adjusted easily for other projects

**Note:** i think for logout we shouldn't clear the storage, since the most common use case imo would be the user gets interrupted / idle -> auto logout and then doesn't know what was created before. So the log will only be cleared if the tab/browser window was closed

resolves #901


sarahsporck (approved):
Tested in Firefox.
Looks Nice, Works nice 👍🏼 
If we ever need to implement this for another client we might need to generalize some more, but for now this looks fine

PR #954, opened 2023-05-02
905: Send automatic mail for new applications and verifications

sarahsporck:
I had to rename useQueryHandler as it is in fact not a hook (Kinda messed this up in the other pr) and therefore eslint would not allow to use it in other hooks. Renaming fixed this.

Additionally I moved the mails that are sent into the mailer (but I did not change the messages, so no need to review this :)). This way we have a much better overview of what default sentences and structure we use in the mails.

Due to the rename there were still quite a few files changed. Sorry for this 🙈


backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -28,18 +39,142 @@ object Mailer {
...
+        - ${projectConfig.administrationName}
+        """.trimIndent()
+        for (recipient: AdministratorEntity in recipients) {
+            try {
+                sendMail(
+                    backendConfig,
+                    projectConfig.smtp,
+                    projectConfig.administrationName,
+                    recipient.email,
+                    "Ein neuer Antrag ist eingegangen",
```
f1sh1918, line 123:
```suggestion
                    "Ein neuer Antrag wurde verifiziert",
```
Falscher Betreff, wenn ein neuer Antrag verifiziert wurde

f1sh1918 (approved):
Everythings works and looks fine. Just some mail templating.
And local mail sending doesn't work with the check you added
You could add this small issue into your pr #960

PR #958, opened 2023-05-03
Refactor extensions into classes

sarahsporck:
The basic idea is to get rid of create emptyCardInProject and to keep cards as abstract as possible, while the typechecker is still working. 
The solution kind of follows a factory pattern, where the classes set in the project config act as factories and instances are created when the classes are invoked in the CreateCardsForm.

This should also work better with the implemented CSV import functionality, as the column headers could be static properties of these extensions.


administration/src/cards/extensions/extensions.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { BavariaCardTypeExtension } from './BavariaCardTypeExtension'
+import { BirthdayExtension } from './BirthdayExtension'
+import { NuernbergPassNumberExtension } from './NuernbergPassNumberExtension'
+import { RegionExtension } from './RegionExtension'
+
+export abstract class Extension<T, R> {
+  public state: T | null = null
+  abstract setInitialState(args: R): void
+  abstract isValid(): boolean
+  abstract createForm(onChange: () => void): React.ReactElement | null
```
f1sh1918, line 13: i think its better to import the `ReactElement from react` explicitly instead importing the whole library

f1sh1918 (approved):
Really nice refactoring. Way easier to read and adjust

administration/src/cards/extensions/extensions.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { CardExtensions } from '../../generated/card_pb'
+import BavariaCardTypeExtension from './BavariaCardTypeExtension'
+import BirthdayExtension from './BirthdayExtension'
+import NuernbergPassNumberExtension from './NuernbergPassNumberExtension'
+import RegionExtension from './RegionExtension'
+
+export interface JSONExtension<T> {
+  name: string
+  state: T | null
+}
```
sarahsporck, line 13: This allows to correctly type the Activity Log

sarahsporck (commented):
@f1sh1918 I needed to make some small adjustments to make this work with the ActivityLog, but I'm quite happy how it came out. Still maybe you want to check out the last commit :)

steffenkleinle (approved):
Looks good

PR #974, opened 2023-05-19
891: Test card creation

sarahsporck:
I decided to use jest instead of Karma and jasmine here, as especially jest component tests are quite troublesome with jasmine. Jasmine only uses spies and has no other mocking implementation and Karma seems to swallow some error messages. To keep the tests performant though I did not use `ts-jest` but swc, which reduces the test run duration drastically.


f1sh1918 (approved):
Awesome work :) and faast! Nice

PR #977, opened 2023-05-23
900: add field nuernberg pass id & add new pdf template

sarahsporck:


administration/src/project-configs/getProjectConfig.ts
```diff
@@ -12,7 +12,11 @@ export interface PdfConfig {
   title: string
   templatePath: string | null
   issuer: string
-  infoToDetails: (info: CardInfo, region: Region, shorten: boolean) => string
+  elements?: {
+    staticQrCodes?: PdfQrCodeElementRenderer[]
+    dynamicQrCodes: PdfQrCodeElementRenderer[]
```
michael-markl, line 17: Maybe make the names of the qr codes more specific? 

```suggestion
    staticVerificationQrCodes?: PdfQrCodeElementRenderer[]
    dynamicActivationQrCodes: PdfQrCodeElementRenderer[]
```

administration/src/cards/pdf/PdfElements.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export function mmToPt(mm: number) {
+  return (mm / 25.4) * 72
+}
+
+export type PdfElementRenderer<T extends Record<string, any>> = (info: T) => void
+export type PdfElement<P extends Record<string, any>, T extends Record<string, any>> = (
+  options: P
+) => PdfElementRenderer<T>
```
michael-markl, line 13: I am wondering whether it would be more appealing to just specify the `PdfElement` options (`P`) in the project configs instead of `PdfElement(options: P)`.
I think we could then make the API a bit less confusing / more explicit:

```suggestion
export type PdfElementRenderer<ConfigOptions extends Record<string, any>, DynamicOptions extends Record<string, any>> = (
  configProps: ConfigOptions , dynamicProps: DynamicOptions
) => void
```

Then, the `PdfQrCodeElement.ts` would become something like:
```ts
export type PdfQrCodeConfigOptions = {
  size: number
} & Coordinates

type QrCodeDynamicOptions = {
  page: PDFPage
  qrCode: PdfQrCode
}

const PdfQrCodeRenderer: PdfElementRenderer<PdfQrCodeConfigOptions, QrCodeDynamicOptions> = ({ size, x, y }, { page, qrCode }) => {
  const qrCodeSizePdf = mmToPt(size)
  const qrCodeXPdf = mmToPt(x)
  const qrCodeYPdf = page.getSize().height - qrCodeSizePdf - mmToPt(y)

  const qrCodeContent = new QrCode({
    qrCode: qrCode,
  }).toBinary()

  drawQRCode(qrCodeContent, qrCodeXPdf, qrCodeYPdf, qrCodeSizePdf, page, false)
}
```

michael-markl (approved):
Have not tested

f1sh1918 (changes requested):
Really nice refactoring. Looks so much cleaner. But there is an error in the pdf. One bigger and one minor thing.

1.The pass-id should be on the back of the print card and the pass-number on the front.

image

2. The label in the form is wrong here (Nürnberg)
image

f1sh1918 (approved):
Fixed some wrong import. the rest is fine :)

steffenkleinle:
Does this fix https://github.com/digitalfabrik/entitlementcard/issues/783?

PR #990, opened 2023-05-25
976: prevent password pasting

### LINKED ISSUE #976: Prevent password pasting for new password
Currently by setting a new password the user can copy paste in the input fields.
In the "Neues Password bestätigen" field it should be disabled to force the user typing the password.

Catch the "paste" event for this field and disable pasting a password

sarahsporck:
fixes #976

### DIFF
--- a/administration/src/bp-modules/user-settings/ChangePasswordForm.tsx
+++ b/administration/src/bp-modules/user-settings/ChangePasswordForm.tsx
@@ -65,7 +65,7 @@ const ChangePasswordForm = () => {
         }}>
         <PasswordInput label='Aktuelles Passwort' value={currentPassword} setValue={setCurrentPassword} />
         <PasswordInput label='Neues Passwort' value={newPassword} setValue={setNewPassword} />
-        <PasswordInput label='Neues Passwort bestätigen' value={repeatNewPassword} setValue={setRepeatNewPassword} />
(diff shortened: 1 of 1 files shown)


steffenkleinle:
Do we really want that? If you use a password manager with autogenerated and little longer password with special characters this is really annoying :see_no_evil:

steffenkleinle (approved):

michael-markl:
Personally, I would also be super annoyed by this (and I don't see this very often at all in practice).
Here's also some officially looking blog post that supports my opinion^TM :D 
https://www.ncsc.gov.uk/blog-post/let-them-paste-passwords

We could also ask our security expert @maxammann  :D

sarahsporck:
I don't really have a super strong opinion on this. 
I am also very fine with not doing it, but maybe let's see what @maxammann says.

Small disclaimer. `onPaste={e => e.preventDefault()}` does not stop the user from drag'n'dropping a text into the field. Further it seems that especially managers that already run in the browser set the text on the dom element directly, as the firefox password manager still works.

maxammann:
I'm usually very annoyed if a website forces me to retype a password :D

I think if a user knows copy paste that well, then it is fine to do that.

maxammann:
Apart from that the solution is sufficient I think.

PR #991, opened 2023-05-26
Upgrade flutter to 3.10

michael-markl:
This was a much more boring upgrade than I thought :D

It needs testing on iOS though.

Here's a changelog: 
https://docs.flutter.dev/release/whats-new#10-may-2023-google-io-2023-310-release

### DIFF
--- a/frontend/.fvm/fvm_config.json
+++ b/frontend/.fvm/fvm_config.json
@@ -1,4 +1,4 @@
 {
-  "flutterSdkVersion": "3.7.6",
+  "flutterSdkVersion": "3.10.2",
   "flavors": {}
 }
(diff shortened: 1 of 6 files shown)


sarahsporck:
Tested on ios simulator (everything but QRCode)

f1sh1918:
@michael-markl did you test protobuf with this version?
I can not run "build config" and for me it looks like the protoc plugin does only support dart sdk until `>3.0.0`

michael-markl:
> Update: okay had to delete .dart_tool folder

Interesting, I didn't have to do that 🤔  CI seems happy as well... Not sure what the actual issue was for you 😕

f1sh1918 (approved):
Happened also for sarah. But no matter.

Tested on ios real device. activation works

PR #1005, opened 2023-06-03
Release 3.1.1

f1sh1918:


administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -48,6 +48,7 @@ class BirthdayExtension extends Extension<BirthdayState, null> {
             if (e.target.value !== null) {
               const millis = Date.parse(e.target.value)
               if (!isNaN(millis)) {
+                console.log(new Date(e.target.value))
```
sarahsporck, line 51: Should be removed.

sarahsporck (approved):
I think this should be properly fixed asap. But for a hotfix this solution should be alright.

PR #1008, opened 2023-06-06
Introduce PlainDate

### LINKED ISSUE #1006: Ensure correct UTC Date for CSV Import
Currently we use `daysSinceEpoch` to save the birthday. Therefore we need the date in UTC. Currently we get the date in GMT+1 from CSV which leads to the `currentDate-1`. As a workaround we add 2 hours to ensure the correct birthday.

- Find a proper way to get the date in UTC instead of adding hours
- remove workaround for birthday and expirationDate in csv import
- remove workaround in CSVCard.test.ts

michael-markl:
Fixes #1006 .

What needs to be tested:
Everything related to Geburtsdatum and Gültigkeitsdatum.
Does the PDF show the correct dates for both nürnberg and bayern?
Does the database have the correct expiration date?
Optimally: When scanning the QR Code, does the phone show the correct dates?


administration/src/cards/CSVCard.test.ts
```diff
@@ -36,8 +37,7 @@ describe('CSVCard', () => {
     card.setValue('Kartentyp', 'Goldkarte')
 
     expect(card.fullName).toBe('Thea Test')
-    // TODO #1006: Ensure correct UTC Date for CSV Import, also remove the hours here
-    expect(card.expirationDate).toEqual(new Date(2022, 3, 3, 5))
```
michael-markl, line 40: I believe, there is still a bug in prod (before this PR)?
I think we should not `expect(card.expirationDate).toEqual(new Date(2022, 3, 3, 5)` when we previously do `card.setValue('Ablaufdatum', '03.04.2022')`.

f1sh1918: But the Set value temporär adds 5 hours. I added this to avoid failing tests

michael-markl: Yep, but the month actually differs: `Date(2022, 3, 3, 5)` vs `03.04.2022`

administration/src/bp-modules/cards/AddCardForm.tsx
```diff
@@ -32,11 +32,15 @@ const ExtensionForm = ({ extension, onUpdate }: ExtensionFormProps) => {
...
-  isBefore(expirationDate, Date.now()) || isAfter(expirationDate, add(Date.now(), { years: maxCardValidity }))
+const maxCardValidity = { years: 99 }
+const hasCardExpirationError = (expirationDate: PlainDate): boolean => {
+  const today = PlainDate.fromLocalDate(new Date())
+
+  return expirationDate.isBefore(today) || expirationDate.isAfter(today.add(maxCardValidity))
+}
 
 const CreateCardForm = ({ cardBlueprint, onRemove, onUpdate }: CreateCardsFormProps) => {
+  const today = PlainDate.fromLocalDate(new Date())
```
f1sh1918, line 43: i think you could create `today` once before the `hasCardExpirationError` and also use it there since its just the date and not datetime

administration/src/cards/CSVCard.test.ts
michael-markl: Ah, nvmd, I was confused again, that `new Date` uses a monthIndex vs an actual month.
So this is less suspicious.

f1sh1918: hm thats true, but it works on production. Anyway i would like to merge your pr and create a new release before they start creating the sozialpass cards

administration/src/cards/CardBlueprint.ts
```diff
@@ -102,9 +104,7 @@ export class CardBlueprint implements JSONCardBlueprint {
     return new CardInfo({
       fullName: this.fullName,
       expirationDay:
-        expirationDate !== null && expirationDate.getTime() > 0 && !this.hasInfiniteLifetime()
```
f1sh1918, line 105: `expirationDate.getTime()` this check is required. If you type in the dateInput field and start typing the year it could be `0002` for the first digit which leads to negative time and protobuf fails
`invalid uint 32: -24681`

f1sh1918 (changes requested):
The negative time check is missing check comment which leads to an error while typing the year.
Also user is blocked by typing in the year if an error occurs

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -37,19 +39,21 @@ class BirthdayExtension extends Extension<BirthdayState, null> {
...
+            min: minBirthday.toString(),
             style: { fontSize: 14, padding: '6px 10px' },
           }}
           onChange={e => {
             if (e.target.value !== null) {
-              const millis = Date.parse(e.target.value)
-              if (!isNaN(millis)) {
-                this.state = { birthday: dateToDaysSinceEpoch(new Date(e.target.value)) }
+              try {
+                const date = PlainDate.from(e.target.value)
```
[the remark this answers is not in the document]
michael-markl: Hmm... I think it is okay to disallow arbitrary text here.
I think browsers should automagically do the right thing when using `type='date'` on the input. Did you have a problem with a specific browser?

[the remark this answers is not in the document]
michael-markl: (Also, the behavior you described should not change by this PR)

administration/src/bp-modules/cards/AddCardForm.tsx
michael-markl: I think it does not matter perf wise. Maybe some people are working during (mid-)night times, so we might only evalute `today` when we need it. (Or even worse: If you leave the tab open for a couple days, `today` would not change).

administration/src/cards/extensions/BirthdayExtension.tsx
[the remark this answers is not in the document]
f1sh1918: yes in chrome once you type a year which is invalid you can not type in a different year. This should be fixed imo

administration/src/util/PlainDate.ts
```diff
@@ -64,7 +66,10 @@ class PlainDate {
    * @example new PlainDate(1900, 1, 1).toUTCDate().toISOString() == "1900-01-01T00:00:00.000Z"
    */
   toUTCDate(): Date {
-    return new Date(Date.UTC(this.isoYear, this.isoMonth - 1, this.day))
+    const date = new Date(Date.UTC(this.isoYear, this.isoMonth - 1, this.day))
+    // We need to setUTCFullYear, as Date.UTC adds 1900 years if year is between 0 and 99 inclusive.
```
f1sh1918, line 70: nice find!

f1sh1918 (approved):
Nice work!

What i've tested:
- Create single card & csv for sozialpass and entitlementcard
- Checked dates of csv import, table and pdf print to be equal
- Checked activityLog
- Activated card for nuernberg has correct expiry and birthday
- verify printed barcode shows correct expiry and birthday

For me its fine to merge. Then i can release another version tonight

PR #1015, opened 2023-06-12
1014: Limit fullname length

### LINKED ISSUE #1014: [Druckerei] Limit fullname length
Currently we allow 50 chars for the fullname which is the protobuf limit.
However the limit for a single line on the digital card is currently 30chars respective the device width.
Everything more than that, breaks the cards view.
We should consider that some chars are wider than others.

- limit the max char length of the "Name"input  to 30 chars
- Add a check for the csv import

f1sh1918:
resolves #1014 

-limit is 30 to also ensure that names with wide chars can be displayed on the digital card and printed card in a single row
-added fitbox widget to ensure rendering of the fullname content

### DIFF
--- a/administration/src/cards/CardBlueprint.ts
+++ b/administration/src/cards/CardBlueprint.ts
@@ -9,7 +9,7 @@ import RegionExtension from './extensions/RegionExtension'
 import { Extension, ExtensionInstance, JSONExtension } from './extensions/extensions'
 import { PEPPER_LENGTH } from './hashCardInfo'
 
-const MAX_NAME_LENGTH = 50
+const MAX_NAME_LENGTH = 30
 const ACTIVATION_SECRET_LENGTH = 20
 
 export interface JSONCardBlueprint<E = ExtensionInstance> {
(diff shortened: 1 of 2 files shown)


administration/src/cards/CardBlueprint.ts
```diff
@@ -54,7 +54,7 @@ export class CardBlueprint {
 
   isFullNameValid(): boolean {
     const encodedName = new TextEncoder().encode(this.fullName)
-    return this.fullName.length > 0 && encodedName.length < MAX_NAME_LENGTH
+    return this.fullName.length > 0 && encodedName.length <= MAX_NAME_LENGTH
```
sarahsporck, line 57: We also add this as a db constraint. Or is it already too late for that?

sarahsporck (commented):
Tested on Android emulator. Works fine :)

administration/src/cards/CardBlueprint.ts
f1sh1918:
```js
A check constraint does check the values already in the table and cannot be created if there are rows violating the condition. That's the point of a constraint, assert that the data meets the condition.
```

sarahsporck: hmm okay. Then I guess we have to leave it as it is. Otherwise we'd need to remove/change entries in the db.

sarahsporck (approved):

Issue #795, opened 2023-02-13
Change wording "Anmeldecode" in PDF


f1sh1918:
**Is your feature request related to a problem? Please describe.**
In our application we use "Aktivierungscode einscannen" in the PDF we use "Anmeldecode"

**Describe the solution you'd like**
We should agree on a wording.

dkehne:
I have checked the marketing material from STMAS but the wording was not used somewhere. 

Use Aktivierungscode.

hauf-toni:
Updated PDF can be found [here](https://nextcloud.tuerantuer.org/index.php/f/4338814).

steffenkleinle:
> Updated PDF can be found [here](https://nextcloud.tuerantuer.org/index.php/f/4338814).

Thanks @hauf-toni. Do you have the same for the Bayerische Ehrenamtskarte? Couldn't find it in nextcloud.

michael-markl:
> Updated PDF can be found [here](https://nextcloud.tuerantuer.org/index.php/f/4338814).

The instructions on the left still have two occurences of "Anmeldecode". Could you replace these too?

hauf-toni:
The updated PDF for Nürnberg can be found [here](https://nextcloud.tuerantuer.org/index.php/f/4338814), for Bayern [here](https://nextcloud.tuerantuer.org/index.php/f/4496444)

sarahsporck:
will be done in #903 as pdf updates either way in the corresponding pr

PR #1017, opened 2023-06-12
903: Add support for pdf form field

sarahsporck:
Also implements #795 .


f1sh1918 (approved):
Just renamed the PdfTextElement to uppercase to fix imports.
Really nice work.

Tested cards creation for sozialpass & ehrenamtskarte
Both single card and csv import

@sarahsporck 
please add the source location of the templates in nextcloud to the documentation and refer to the issue here

PR #1023, opened 2023-06-15
1022: Name length special chars

### LINKED ISSUE #1022: Special caracters lower the name limit
**Describe the bug**
Special characters lower the 30 chars limit 
seems to me like checking bytes not characters

**To Reproduce**
1. create a CSV with 29 chars name
2. append special character 'ä'
3. import file
4. observe wrong name

**Expected behavior**
The limit should count characters regardles of encoding

**Additional context**
Also as leading and trailing spaces are stripped, multiple spaces in between count to the length of name, but will be replaced by a single space. This leads to wrong marked names in the list with less than 30 character.

f1sh1918:
resolves #1022

### DIFF
--- a/administration/src/bp-modules/cards/ImportCardsTable.tsx
+++ b/administration/src/bp-modules/cards/ImportCardsTable.tsx
@@ -19,6 +19,11 @@ const TableContainer = styled.div`
   flex-grow: 1;
 `
 
+const StyledCell = styled(Cell)`
+  font-size: 0.85rem;
+  white-space: break-spaces;
+`
+
 const CardImportTable = ({ headers, cardBlueprints }: CardImportTableProps) => {
   const cellRenderer = useCallback(
     (rowIndex: number, columnIndex: number) => {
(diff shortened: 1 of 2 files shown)


sarahsporck:
Please also improve how the cell renders white spaces. They should not be collapsed.
I already did some digging into that and this should work: 
```javascript
        <Cell
          wrapText={true}
          key={rowIndex + '-' + columnIndex}
          style={{ fontSize: '0.85rem', whiteSpace: 'break-spaces' }}
          tooltip={!valid ? 'Validierungsfehler' : undefined}
          intent={!valid ? 'danger' : 'none'}>
          <TruncatedFormat2 detectTruncation={true} preformatted={true}>{!!value ? value : '-'}</TruncatedFormat2>
        </Cell>
```

f1sh1918:
> Please also improve how the cell renders white spaces. They should not be collapsed. I already did some digging into that and this should work:
> 
> ```js
>         <Cell
>           wrapText={true}
>           key={rowIndex + '-' + columnIndex}
>           style={{ fontSize: '0.85rem', whiteSpace: 'break-spaces' }}
>           tooltip={!valid ? 'Validierungsfehler' : undefined}
>           intent={!valid ? 'danger' : 'none'}>
>           <TruncatedFormat2 detectTruncation={true} preformatted={true}>{!!value ? value : '-'}</TruncatedFormat2>
>         </Cell>
> ```

thx for the hint. added commit @sarahsporck

sarahsporck (approved):

PR #1025, opened 2023-06-24
Fix CI: Migrate constructor syntax due to protoc_plugin 21

michael-markl:
The CI broke due to an update of protoc_plugin.
Please also run `dart pub global activate protoc_plugin` locally and verify if version 21 is installed (You might have to upgrade your global flutter/dart installation).

See https://pub.dev/packages/protoc_plugin/changelog#2102 for the relevant changes of protoc_plugin.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,7 @@ commands:
       - run:
           name: Install Dart
           command: |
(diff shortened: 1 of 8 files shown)


f1sh1918 (approved):
Tested on ios simulator seems fine

PR #1036, opened 2023-06-29
1035: Notification sent to all users

### LINKED ISSUE #1035: [Bavaria]Notification of applications sends to all users
**Describe the bug**
When a new application arrives a notification will be send to users of this region that have activated this in user settings.
But only the users of this specific region should get a notification

**To Reproduce**
1. Enable notifications in user settings
2. Create an application for another region than yours
3. Check your mails

**Expected behavior**
Notifications will only be send if the user belongs to the region of the application

f1sh1918:
resolves #1035

- Login as region admin/manager: activate notifications
- Send application to your region -> should receive notification
- send application to a different region -> shouldn't receive notification
- test the same for application verification


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/repos/AdministratorsRepository.kt
```diff
@@ -122,17 +122,17 @@ object AdministratorsRepository {
...
+                (Projects.project eq project) and (Administrators.notificationOnApplication eq true) and (Administrators.regionId eq regionId)
             }.let { AdministratorEntity.wrapRows(it) }.toList()
         }
 
-    fun getNotificationRecipientsForVerification(project: String): List<AdministratorEntity> =
+    fun getNotificationRecipientsForVerification(project: String, regionId: Int): List<AdministratorEntity> =
         transaction {
             (Administrators innerJoin Projects).select {
-                (Projects.project eq project) and (Administrators.notificationOnVerification eq true)
+                (Projects.project eq project) and (Administrators.notificationOnVerification eq true) and (Administrators.regionId eq regionId)
```
michael-markl, line 135: Are we missing a check for "deleted" as well?

f1sh1918: Good point!

michael-markl (approved):
Have not tested.

PR #1047, opened 2023-07-05
1043: Allow legacy blau and gold as card types

### LINKED ISSUE #1043: Allow "blau" and "gold" as card type for mass card creation via csv
By using the data from existing data bases e.g. Freinet the card types are defined as "blau" and "gold" but we define it as "Standard" and "Goldkarte". Either switch to "blau" and "gold" or make the two additional strings valid for csv imports.

steffenkleinle:
Fixes #1043


f1sh1918 (approved):
Tested and works fine :)

PR #1052, opened 2023-07-10
1051: No feedback pw reset

### LINKED ISSUE #1051: No Feedback for password reset for unknown mail address
**Describe the bug**
If the pw reset function was used with an unknown mail address there is no error message/ feedback to the user.
**To Reproduce**
1. https://staging.bayern.ehrenamtskarte.app/forgot-password
2. Type a mail address that is not existing in the database
3. click on "Password zurücksetzen
4. Check that nothing happens. Only console error

**Expected behavior**
The user will get a feedback that the mail was sent to the mail address if it exists

image

f1sh1918:
resolves #1051

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordMutationService.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordMutationService.kt
@@ -25,7 +25,7 @@ class ResetPasswordMutationService {
         val projectConfig = backendConfig.projects.first { it.id == project }
         transaction {
             val user = Administrators.innerJoin(Projects).slice(Administrators.columns)
-                .select((Projects.project eq project) and (LowerCase(Administrators.email) eq email.lowercase()))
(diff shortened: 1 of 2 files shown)


michael-markl (approved):
LGTM, not tested

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ResetPasswordMutationService.kt
```diff
@@ -26,10 +26,12 @@ class ResetPasswordMutationService {
         transaction {
             val user = Administrators.innerJoin(Projects).slice(Administrators.columns)
                 .select((Projects.project eq project) and (LowerCase(Administrators.email) eq email.lowercase()))
-                .single().let { AdministratorEntity.wrapRow(it) }
-
-            val key = AdministratorsRepository.setNewPasswordResetKey(user)
-            Mailer.sendResetPasswodMail(backendConfig, projectConfig, key, email)
+                .singleOrNull()?.let { AdministratorEntity.wrapRow(it) }
```
michael-markl, line 29: We should (above) probably also check that the Admin is not deleted?

PR #1060, opened 2023-07-21
Validate cardInfoHash length

sarahsporck:


backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -99,8 +104,12 @@ class CardMutationService {
...
     }
 
-    private fun validateNewCard(card: CardGenerationModel): Boolean {
+    private fun validateCodeType(card: CardGenerationModel): Boolean {
         return (card.codeType == CodeType.STATIC && card.activationSecretBase64 == null) ||
             (card.codeType == CodeType.DYNAMIC && card.activationSecretBase64 != null)
     }
+
+    private fun validateCardHash(cardHash: ByteArray): Boolean {
+        return cardHash.size != 32
```
michael-markl, line 113: We should probably add a comment here that we expect a SHA-256 hash, and thus the byte array should be of size 256/8=32.

michael-markl (approved):
Have not tested.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
michael-markl, line 113: I think the function validateCardHash should not return true, if card hash is invalid.

Imo, it would be better to name the function `isCardHashValid` to avoid any confusion. (same for validateCodeType)

michael-markl (changes requested):

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -36,17 +37,21 @@ class CardMutationService {
                 if (!Authorizer.mayCreateCardInRegion(user, targetedRegionId)) {
                     throw ForbiddenException()
                 }
-                if (!validateNewCard(card)) {
+                if (!isCodeTypeValid(card)) {
                     throw InvalidCodeTypeException()
                 }
+                val decodedCardInfoHash = Base64.getDecoder().decode(card.cardInfoHashBase64)
+                if (isCardHashValid(decodedCardInfoHash)) {
```
michael-markl, line 44:
```suggestion
                if (!isCardHashValid(decodedCardInfoHash)) {
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -99,8 +104,13 @@ class CardMutationService {
...
 
-    private fun validateNewCard(card: CardGenerationModel): Boolean {
+    private fun isCodeTypeValid(card: CardGenerationModel): Boolean {
         return (card.codeType == CodeType.STATIC && card.activationSecretBase64 == null) ||
             (card.codeType == CodeType.DYNAMIC && card.activationSecretBase64 != null)
     }
+
+    private fun isCardHashValid(cardHash: ByteArray): Boolean {
+        // we expect a SHA-256 hash, thus the byte array should be of size 256/8=32
+        return cardHash.size != 32
```
michael-markl, line 114:
```suggestion
        return cardHash.size == 32
```

backend/bin/main/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -0,0 +1,116 @@
+package app.ehrenamtskarte.backend.verification.webservice.schema
```
michael-markl, line 1: Where does this files come from ? :D

sarahsporck: argh. unsure. I think my vscode kotlin plugin generates these files. will remove this again ^^

michael-markl (approved):

f1sh1918 (approved):
looks fine to me

PR #1061, opened 2023-07-21
Fail CI when pubspec.lock changes

sarahsporck:
Any other files that might change unvoluntarily?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,7 +183,7 @@ jobs:
       - install-protobuf-linux
       - run:
           name: Install Flutter Packages
-          command: fvm flutter pub get
+          command: fvm flutter pub get --enforce-lockfile
       - run:
           name: Check Formatting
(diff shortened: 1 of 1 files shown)


michael-markl:
We should probably (instead and/or additionally) replace `flutter pub get` by `flutter pub get --enforce-lockfile`: https://dart.dev/tools/pub/cmd/pub-get#--enforce-lockfile

sarahsporck:
> We should probably (instead and/or additionally) replace `flutter pub get` by `flutter pub get --enforce-lockfile`: https://dart.dev/tools/pub/cmd/pub-get#--enforce-lockfile

Nach so was hatte ich tatsächlich gesucht ^^ Danke schön :) Habs geändert

michael-markl:
There seem to be 2 more occurences of `flutter pub get`. Could you add the option there too?

michael-markl (approved):

PR #1068, opened 2023-08-02
1066: Log card activation

### LINKED ISSUE #1066: Add logging for card activation
Since we are facing some activation issue, we should implement more logging for card activation in the backend to have more information to resolve the issues

- log card id and debug information for the several card activation states.

Additionally we should collect sentry information

f1sh1918:
resolves #1066 

We can discuss what else should we log or if this is too much

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
@@ -86,7 +86,7 @@ class CardMutationService {
         }
 
         if (!CardActivator.verifyActivationSecret(rawActivationSecret, activationSecretHash)) {
-            logger.info("${context.remoteIp} failed to activate card with id:${card.id}")
+            logger.info("${context.remoteIp} failed to activate card with id:${card.id} and overwrite: $overwrite")
             return CardActivationResultModel(ActivationState.failed)
         }
 
@@ -102,7 +102,7 @@ class CardMutationService {
         val totpSecret = CardActivator.generateTotpSecret()
         val encodedTotpSecret = Base64.getEncoder().encodeToString(totpSecret)
         transaction { CardRepository.activate(card, totpSecret) }
-        logger.info("Card with id:${card.id} was activated from ${context.remoteIp}")
+        logger.info("Card with id:${card.id} and overwrite: $overwrite was activated from ${context.remoteIp}")
         return CardActivationResultModel(ActivationState.success, encodedTotpSecret)
     }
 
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -86,7 +86,7 @@ class CardMutationService {
         }
 
         if (!CardActivator.verifyActivationSecret(rawActivationSecret, activationSecretHash)) {
-            logger.info("${context.remoteIp} failed to activate entitlement card")
+            logger.info("Card with id:${card.id} from ${context.remoteIp} failed to activate entitlement card")
```
steffenkleinle, line 89:
```suggestion
            logger.info("${context.remoteIp} failed to activate card with id:${card.id}")
```

steffenkleinle (approved):
Can we also log if a card did overwrite a card on another device (or add some more information in that case)?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -95,12 +95,14 @@ class CardMutationService {
...
 
         if (!overwrite && card.totpSecret != null) {
+            logger.info("Card with id:${card.id} did not overwrite card from ${context.remoteIp}")
             return CardActivationResultModel(ActivationState.did_not_overwrite_existing)
         }
 
         val totpSecret = CardActivator.generateTotpSecret()
         val encodedTotpSecret = Base64.getEncoder().encodeToString(totpSecret)
         transaction { CardRepository.activate(card, totpSecret) }
+        logger.info("Card with id:${card.id} was activated from ${context.remoteIp}")
```
michael-markl, line 105: What's the reason for logging the IP addresses? I think, we don't really need them for debugging purposes. It might be more interesting whether `overwrite` is true or not at this point.

sarahsporck: The main reason to log the ip adress can be to block certain ip adresses if they spam our server. I think for the activation this could only make sense for failed activations, but feel free to disagree :)

sarahsporck (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
f1sh1918: i added the overwrite value

michael-markl:
@sarahsporck I am not entirely sure, whether our data privacy terms allow this sort of ip-address-collection (more severly, with these logs we can actually combine ip-address and card-id to find out which card-id belongs to which person (roughly)). We should probably have Daniel / Sven confirm this change @svenseeberg & @dkehne

Issue #1065, opened 2023-08-02
Card can't be activated


f1sh1918:
**Describe the bug**
Some users report that they can not activate a card with the pdf. They receive the message: "Ihre Karte ist ungültig Sie wurde entweder widerrufen oder auf einem anderen Gerät aktiviert"

**To Reproduce**
Really difficult to reproduce because it seem to depend on the device or android version. We have to check if we get strange backend logs

**Expected behavior**
Card gets activated

Example card can be found in mattermost where the logs can be checked
https://chat.tuerantuer.org/digitalfabrik/pl/jkrgej4zot8zzru9uyggi5bbjc

f1sh1918:
may be related to #1056

Issue #1049, opened 2023-07-06
Golden EAK shows as invalid despite being valid


michael-markl:
**Describe the bug**

A volunteer (Landkreis Kulmbach, EAK) activated its GOLDEN card by scanning an activation code.

The "Ausweisen"-Tab shows their golden card including their name, but below, there is the following error message:
"Ihre Karte ist ungültig. Sie wurde entweder widerrufen oder auf einem anderen Gerät aktiviert."

On a different device (without some specific security setting) activating the card worked without problems.

Unclear:
* Did the error message show directly after scanning the Ausweisen Tab
* What kind of seurity setting is activated?

**To Reproduce**

**TODO:** Describe steps to reproduce the behavior.

**Expected behavior**

There should not be an error message, but instead a QR Code after succesfully activating the code.

**Smartphone:**
 - Device: ?
 - OS: Android
 - Version ?

f1sh1918:
@michael-markl can we get the device and platform version? specific android version?

PR #1069, opened 2023-08-04
Harden queryServerVerification

michael-markl:
Apply some hardening to `queryServerVerification`.
I believe, this PR does *not* fix the recent problems with card activations, where cards could successfully be activated, but were then shown as invalid.

Related:
#1049, #1065, #1056


sarahsporck (approved):
Seems reasonable. Thanks for the PR!

PR #1070, opened 2023-08-04
Show error if activation with override fails

michael-markl:
Previously, errors that happened after the user confirms overriding an existing device, were not shown to the user.


michael-markl:
You could try to activate the same card twice.
Then, once the dialog appears about asking to override the previous activation, shut down the backend server. Then confirm the dialog, and check whether an error is shown to the user.

f1sh1918 (approved):
works fine. tested on android

PR #1071, opened 2023-08-04
Untangle card status in app

michael-markl:
This PR untangles how the app determines the status of a persisted card.
It should be clearer to see what is shown in the "Ausweisen" tab.

The only functional change:
* We initiate self-verification in any status (previously, we did not self-verify if the card expired (according to local clock) or if it the backend ever responded with `invalid`). 
* Try to detect race-conditions in self-verification (possibly conflicting with activation): If the UserCode was set while doing the server request, then ignore the result.

Minor change:
* VerificationCodeView sets its own timer (instead of OTPGenerator setting it), in order to be able to cancel it in case it is being disposed. Previously, we had an error in the console, when the VerificationCodeView was unmounted.

### DIFF
--- a/frontend/lib/identification/card_detail_view/card_detail_view.dart
+++ b/frontend/lib/identification/card_detail_view/card_detail_view.dart
@@ -1,14 +1,18 @@
 import 'package:ehrenamtskarte/configuration/configuration.dart';
 import 'package:ehrenamtskarte/graphql/graphql_api.dart';
 import 'package:ehrenamtskarte/identification/card_detail_view/more_actions_dialog.dart';
-import 'package:ehrenamtskarte/identification/card_detail_view/verification_code_view.dart';
+import 'package:ehrenamtskarte/identification/card_detail_view/self_verify_card.dart';
 import 'package:ehrenamtskarte/identification/id_card/id_card.dart';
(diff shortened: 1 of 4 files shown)


frontend/lib/identification/card_detail_view/card_detail_view.dart
```diff
@@ -94,49 +128,79 @@ class CardDetailView extends StatelessWidget {
...
-      this.compact = false,
-      required this.isExpired});
+  const QrCodeAndStatus(
+      {super.key, required this.onMoreActionsPressed, required this.onSelfVerifyPressed, required this.userCode});
 
   @override
   Widget build(BuildContext context) {
-    final wasCardVerifiedLately = cardWasVerifiedLately(userCode.cardVerification);
-    final isCardInvalid = !userCode.cardVerification.cardValid;
+    final CardStatus status = CardStatus.from(userCode);
```
sarahsporck, line 172: <3 very pretty.

sarahsporck (approved):
Tested on android emulator.

f1sh1918 (approved):
really nice improvement :)
Tested for revoking and change expirationDay in the database (even that won't happen)

PR #1072, opened 2023-08-04
Avoid DB race condition when activating

michael-markl:
To avoid race conditions when activating a card (e.g. if the app somehow requests activation too often), this PR introduces DB transaction isolation for the activation transaction.


sarahsporck (approved):
I guess this is hard to actually test. But normal activation still works as intended.

f1sh1918 (approved):
looks fine to me

michael-markl:
I actually tested this by adding some thread sleeping inside the transaction and activated twice the same card.
That threw an error (as expected) for the second attempt. Without the change it didn't throw the error.

michael-markl:
I might test this once more to check if it also throws if we activate different cards at the same time. That should hopefully not be the case (but I'll check it once more)

michael-markl:
I confirmed that activating different in parallel cards does not error out.

PR #1079, opened 2023-08-16
1074: add applicant name to verification Mail

sarahsporck:
As always, feel free to critique the format. Not my strong point ^^

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
@@ -148,7 +148,7 @@ object Mailer {
             projectConfig.smtp,
             projectConfig.administrationName,
             applicationVerification.contactEmailAddress,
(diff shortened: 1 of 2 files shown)


steffenkleinle (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -148,7 +148,7 @@ object Mailer {
             projectConfig.smtp,
             projectConfig.administrationName,
             applicationVerification.contactEmailAddress,
-            "Bestätigung notwendig: Antrag auf Bayerische Ehrenamtskarte",
+            "Bestätigung notwendig: Antrag auf Bayerische Ehrenamtskarte [${applicationVerification.contactName}]",
```
michael-markl, line 151: I guess we don't want to send the contact name but the applicant name ?

sarahsporck: Jep you're right. Once again a splendid review. Will change that.

PR #1083, opened 2023-08-16
1073: add second attachment

sarahsporck:
Not sure about the visualisation. With the way I implemented this it is not immediately clear, that you can upload two attachments. What do you think? We could also display both "upload"-buttons and validate for either of them to be set.


f1sh1918 (commented):
Works nicely. I just was thinking about why not showing the filename?
I think it would be more intuitive for the user to know which attachment was added/deleted

```js
setState(() => ({ MIMEType: fileType, filename: file.name, arrayBufferKey: key }))
```

image

f1sh1918 (approved):
Works fine :)

PR #1088, opened 2023-08-16
1076: Note non unique postal code

### LINKED ISSUE #1076: Show note for postal code without unique region
As a user i want to be informed if my postal code can be allocated with multiple regions (f.e. 84030) in the application process
Currently the first region that matches will be auto assigned in the dropdown.

- show a note if more than one region can be allocated to the postal code with a message like
'Ihre PLZ ist nicht eindeutig zuzuordnen. Bitte wählen Sie Option 1 oder Option 2... aus dem Auswahlfeld'
- don't auto assign the first matching region

f1sh1918:
resolves #1076 

Here are some postal codes to test: 84030, 85540

If you have better idea for design or wording let me know

### DIFF
--- a/administration/src/graphql/regions/getRegionsByPostalCode.graphql
+++ b/administration/src/graphql/regions/getRegionsByPostalCode.graphql
@@ -1,0 +1,8 @@
+query getRegionsByPostalCode($postalCode: String!, $project: String!) {
+  regions: regionsByPostalCode(postalCode: $postalCode, project: $project) {
+    id
+    name
+    prefix
+    activatedForApplication
+  }
+}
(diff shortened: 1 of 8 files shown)


administration/src/graphql/regions/getRegionsByPostalCode.graphql
```diff
@@ -0,0 +1,8 @@
+query getRegionsByPostalCode($postalCode: String!, $project: String!) {
```
f1sh1918, line 1: renamed

sarahsporck (approved):
Tested in firefox. Looks great!

PR #1090, opened 2023-08-17
1080: Add start day

### LINKED ISSUE #1080: [Sozialpass] Add start date
As a user i want to have a start date on my sozialpass id card to see when the validity period starts. Furthermore the service worker is able to create cards that are not valid yet

## Acceptance Criteria:
 - the create card form has a date input field start date
 - the start date will be send with the barcode
 - the start date will be shown on the digital and printed card (see screenshot)
 - the card will be shown as invalid when the start date isn't reached
 - the passId moves next to the birthday
 
## Technical Tasks:
 - add start date extension for nürnberg
 - add a date input field to the create card form
 - add a date input field for the csv import
 - adjust card validation including the start date (card should be activated but be invalid if start date isn't reached)
 - adjust the date label on card and pdf
 - add startDate in protobuf cardInfo (ensure backwards compatibility)
 - Test if qrCodes without startDates still work

startDate

f1sh1918:
resolves #1080


frontend/lib/identification/card_detail_view/card_detail_view.dart
```diff
@@ -144,7 +145,9 @@ enum CardStatus {
   // The card was verified lately by the server and it responded that the card is invalid.
   invalid,
   // In any other case, we assume the card is valid and show the dynamic QR code
-  valid;
+  valid,
+  // The validity period didn't start yet
+  stillInvalid;
```
sarahsporck, line 150: how about `notYetValid`? stillInvalid can also mean that the card was invalid at some point and is still invalid.

sarahsporck (approved):
Tested on android emulator.

administration/src/project-configs/nuernberg/pdf.ts
```diff
@@ -17,10 +17,11 @@ const renderPdfDetails = ({ info, cardBlueprint }: InfoParams) => {
   const passId = findExtension(cardBlueprint.extensions, NuernbergPassIdExtension)?.state?.nuernbergPassId
   const expirationDate = PlainDate.fromDaysSinceEpoch(expirationDay)
   const birthdayDate = PlainDate.fromDaysSinceEpoch(info.extensions?.extensionBirthday?.birthday ?? 0)
+  const startDate = PlainDate.fromDaysSinceEpoch(info.extensions?.extensionStartDay?.startDay ?? 0)
   return `${info.fullName}
 Pass-ID: ${passId ?? ''}
 Geburtsdatum: ${birthdayDate.format('dd.MM.yyyy')}
-Gültig bis: ${expirationDate.format('dd.MM.yyyy')}`
+Gültig: ${startDate.format('dd.MM.yyyy')} bis ${expirationDate.format('dd.MM.yyyy')}`
```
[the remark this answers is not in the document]
f1sh1918: But after this pr you can not create cards without a Start date. I think it's just importante for the app when you activate old Barcodes?

specs/card.proto
```diff
@@ -30,11 +30,21 @@ message BavariaCardTypeExtension {
...
+  optional uint32 start_day = 1;
+}
+
 message CardExtensions {
   optional RegionExtension extension_region = 1;
   optional BirthdayExtension extension_birthday = 2;
   optional NuernbergPassNumberExtension extension_nuernberg_pass_number = 3;
   optional BavariaCardTypeExtension extension_bavaria_card_type = 4;
+  optional StartDayExtension extension_start_day = 5;
 }
```
michael-markl, line 48: Please add a test that checks the generated hash of a cardinfo using this extension to stay stable.
Please add a test both to frontend and administration

michael-markl (changes requested):
Imo, the start date must also be added to the DB table and checked in the backend.

specs/card.proto
f1sh1918: You mean because you could change the time/date on the local device to make the card valid?

michael-markl: No I imagine a unit test that checks whether the hashed card info with the newly added extension stays stable over time (... for when we modify the code / protobuf in the future).

There are already some tests involving this hash stability, but another one with the added extension would be beneficial.

f1sh1918: Oh sry I wanted to reply to your first comment that it should be added to db. The rest was clear.

michael-markl:
> You mean because you could change the time/date on the local device to make the card valid?
> Hm couldn't we just add general check if the device date/timestamp fits to the backend timestamp?
> Instead of checking each timestamp we add @michael-markl

Not sure what you mean by "each timestamp we add". I don't think there will be yet another timestamp in the future regarding the validity period.

Also I think, we should make sure that the backend has the (ultimate) capability of checking whether a card is valid or not (simply looking by the row in the DB table). (Might also be interesting to have a statistic like "Currently valid cards" which we wouldn't be able to accurately determine if we don't add it to the DB). More importantly (imo), the `verifyCard` endpoint should always be correct and, hence, should also reflect whether the validity period has already started.

Implementing some validity checks on the frontend (of the verifier's device) and some validity checks on the backend seems a bit arbitrary to me from a design perspective.

frontend/test/card_info_util_test.dart
```diff
@@ -41,5 +41,18 @@ void main() {
...
+        ..fullName = 'Max Mustermann'
+        ..expirationDay = 365 * 40 // Equals 14.600
+        ..extensions = (CardExtensions()
+          ..extensionRegion = (RegionExtension()..regionId = 93)
+          ..extensionBirthday = (BirthdayExtension()..birthday = -365 * 10)
+          ..extensionNuernbergPassNumber = (NuernbergPassNumberExtension()..passNumber = 99999999)
+          ..extensionStartDay = (StartDayExtension()..startDay = 365 * 2));
+      final pepper = const Base64Decoder().convert("MvMjEqa0ulFDAgACElMjWA==");
+      expect(cardInfo.hash(pepper), '1ChHiAvWygwu+bH2yOZOk1zdmwTDZ4mkvu079cyuLjE=');
+    });
```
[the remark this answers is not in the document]
f1sh1918: i did not rerequest your review because i wasn't finished :D 
The card hash did not update in `administration` after adding extension so i removed the test temp.
I have problems running karma tests. So i decided to wait for sarahs prs.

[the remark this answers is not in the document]
f1sh1918: ok now everything works, gonna update soon :)

administration/src/cards/extensions/StartDayExtension.tsx
```diff
@@ -0,0 +1,96 @@
...
+  setInitialState() {
+    this.state = { startDay: initialMinStartDayDate.toDaysSinceEpoch() }
+  }
+
+  hasValidStartDayDate(startDay?: number): boolean {
+    if (startDay === undefined) {
+      return false
+    }
+    const date = PlainDate.fromDaysSinceEpoch(startDay)
+    return !date.isBefore(initialMinStartDayDate)
```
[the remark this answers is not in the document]
f1sh1918: i introduced an optional `hasFormDependencyError` to extensions

frontend/lib/identification/util/card_info_utils.dart
```diff
@@ -40,3 +39,13 @@ bool cardWasVerifiedLately(CardVerification cardVerification) {
       : DateTime.now().toUtc().isBefore(DateTime.fromMillisecondsSinceEpoch(0)
           .add(Duration(seconds: lastVerificationTimestamp.toInt() + cardValidationExpireSeconds)));
 }
+
+bool isCardNotYetValid(CardInfo cardInfo) {
```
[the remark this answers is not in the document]
f1sh1918: added some tests, but not sure how to cover timezone issues

frontend/test/card_info_util_test.dart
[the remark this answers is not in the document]
f1sh1918: added a test for administration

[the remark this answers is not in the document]
michael-markl: > i did not rerequest your review because i wasn't finished :D

Ah sorry :D Was too curious :) Will wait next time

administration/src/cards/extensions/StartDayExtension.tsx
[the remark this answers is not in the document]
michael-markl: Hmm... I feel like we should probably check it in the CardBlueprint class. I think the AddCardForm checks do not apply for the CSV import?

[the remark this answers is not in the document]
f1sh1918: thats true but i think to show the error for the particular form we have also to pass a prop to the template /StartDayExtension?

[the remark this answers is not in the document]
michael-markl: Hmm... I still think we should implement this startDate/expiryDate check. A card that is never valid should not be creatable and we should hint the administration user that he did something wrong.
Maybe, we can also add a very simple check to the backend which simply throws an error, and open an issue to make the user interface better later.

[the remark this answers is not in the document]
sarahsporck: We could add this in the cardBlueprint 

> Hmm... I feel like we should probably check it in the CardBlueprint class. I think the AddCardForm checks do not apply for the CSV import?

They should do.
You could add this ccheck in ´isExpirationDateValid` in the CardBlueprint class.

frontend/test/card_info_util_test.dart
```diff
@@ -41,5 +41,55 @@ void main() {
...
+      final cardInfo = CardInfo()
+        ..fullName = 'Max Mustermann'
+        ..expirationDay = 365 * 40 // Equals 14.600
+        ..extensions = (CardExtensions()
+          ..extensionRegion = (RegionExtension()..regionId = 93)
+          ..extensionBirthday = (BirthdayExtension()..birthday = -365 * 10)
+          ..extensionNuernbergPassNumber = (NuernbergPassNumberExtension()..passNumber = 99999999)
+          ..extensionStartDay = (StartDayExtension()..startDay = 365 * 70));
+      expect(isCardNotYetValid(cardInfo), true);
+    });
```
michael-markl, line 70: Maybe, you could use https://pub.dev/packages/clock to test how `isCardNotYetValid` behaves when the time is very close to the `startDay`.
I think, ultimately, we should properly solve the timezone logic in the app. For the expiration day, we did some weird tolerance stuff.
Maybe, this can also be done in a seperate PR / issue though. I believe your current logic is "good enough" for now.

frontend/lib/identification/id_card/card_content.dart
michael-markl, line 1: I guess, this slightly changes the layout for Bayern as well. Is this okay?

administration/src/cards/extensions/StartDayExtension.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+class StartDayExtension extends Extension<StartDayState, null> {
+  public readonly name = StartDayExtension.name
+
+  setInitialState() {
+    const today = PlainDate.fromLocalDate(new Date())
+    this.state = { startDay: today.toDaysSinceEpoch() }
+  }
+
+  createForm(onUpdate: () => void, hasFormDependencyError?: boolean) {
```
michael-markl, line 19: I think we should remove this `hasFormDependencyError` again for now?

frontend/lib/identification/id_card/card_content.dart
f1sh1918: it still shows `Gültig bis: <Date>`
I had to reduce the padding a little bit because there were spacing issue in the verification idCard Modal. But thats just minor.

frontend/test/card_info_util_test.dart
f1sh1918: https://github.com/digitalfabrik/entitlementcard/issues/1096

administration/src/cards/extensions/StartDayExtension.tsx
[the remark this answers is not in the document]
f1sh1918: ok i added a check for the form and the csv @michael-markl

michael-markl (approved):
Not tested

frontend/lib/identification/card_detail_view/card_detail_view.dart
```diff
@@ -157,6 +160,8 @@ enum CardStatus {
       return CardStatus.timeOutOfSync;
     } else if (!code.cardVerification.cardValid) {
       return CardStatus.invalid;
+    } else if (isCardNotYetValid(code.info)) {
+      return CardStatus.notYetValid;
```
michael-markl, line 164: Sorry, but one more issue I think:

Shouldn't these two cases be swapped?
If the card is not yet valid, then the server would respond with `invalid`, thus the isCardNotYetValid case would (usually) never occur.

frontend/lib/identification/card_detail_view/card_detail_view.dart
```diff
@@ -213,6 +218,9 @@ class QrCodeAndStatus extends StatelessWidget {
             CardStatus.valid => [
                 _PaddedText('Mit diesem QR-Code können Sie sich bei Akzeptanzstellen ausweisen:'),
                 Flexible(child: VerificationCodeView(userCode: userCode))
+              ],
+            CardStatus.notYetValid => [
```
michael-markl, line 222: Could you move this before the `invalid` case?

frontend/lib/identification/card_detail_view/card_detail_view.dart
```diff
@@ -157,6 +160,8 @@ enum CardStatus {
       return CardStatus.timeOutOfSync;
     } else if (!code.cardVerification.cardValid) {
       return CardStatus.invalid;
+    } else if (isCardNotYetValid(code.info)) {
+      return CardStatus.notYetValid;
     } else {
```
michael-markl, line 165: Sorry, but one more issue I think:

Shouldn't these two cases be swapped?
If the card is not yet valid, then the server would respond with `invalid`, thus the isCardNotYetValid case would (usually) never occur.



```suggestion
    } else if (isCardNotYetValid(code.info)) {
      return CardStatus.notYetValid;
    } else if (!code.cardVerification.cardValid) {
      return CardStatus.invalid;
    } else {
```

administration/src/bp-modules/cards/AddCardForm.tsx
```diff
@@ -33,10 +33,17 @@ const ExtensionForm = ({ extension, onUpdate }: ExtensionFormProps) => {
...
+    return false
+  }
   const today = PlainDate.fromLocalDate(new Date())
-
-  return expirationDate.isBefore(today) || expirationDate.isAfter(today.add(maxCardValidity))
+  return (
+    expirationDate.isBefore(today) ||
+    expirationDate.isAfter(today.add(maxCardValidity)) ||
+    !cardBlueprint.isExpirationDateValid()
+  )
```
sarahsporck, line 46: I think all of these checks can be moved into the CardBlueprint. Into the is ExpirationDateValid()

administration/src/cards/CardBlueprint.ts
```diff
@@ -83,9 +84,20 @@ export class CardBlueprint {
     )
   }
 
+  isStartDayBeforeExpirationDay = (expirationDate: PlainDate): boolean => {
+    const startDayExtension = this.extensions.find(ext => ext.name === 'StartDayExtension') as StartDayExtension
```
sarahsporck, line 88:
```suggestion
    const startDayExtension = findExtension(this.extensions, StartDayExtension)
```

sarahsporck (commented):
I left some minor comments. Please rerequest review once you have updated the last changes, then I'll look over it (hopefully) one last time

administration/src/bp-modules/cards/AddCardForm.tsx
f1sh1918: moved please rereview :). @sarahsporck

Issue #1075, opened 2023-08-07
Create card from application


f1sh1918:
As a service user i want to create cards directly from application. So a manual filling is not needed

## Acceptance Criteria:
 - add a button "Karte erstellen"
 - with a click on the button you will be forwarded to the "Einzelne Karten erstellen" page
 - input fields will be filled with the application data

## Technical Tasks:
 - add a button "Karte erstellen" to the bottom of the application
 - button will be shown only if logged in as a service user
 - click on the button forwards to the 'cards/add' route and passes attributes via queryParams
 - the input fields should be filled with the values of the queryParams
 - queryParams should be optional, test that Sozialpass will not break

sarahsporck:
Blocked by #1082

sarahsporck:
@hauf-toni I think we need a design for this. Currently the application is quite cluttered with lot's of buttons and stuff.
For context: 
As a 'Mitarbeiter einer Stadt' I want to view, export and delete entitlementcard applications. To ease the workflow from application to a created card, we add a functionality to create a card from an application. This is what this ticket is about.

However the button bar is now quite full and there is too much information at once at the bottom of the Application. Could you come up with a better design for this?

<img width="800" alt="Bildschirmfoto 2023-08-28 um 10 07 03" src="https://github.com/digitalfabrik/entitlementcard/assets/28013368/0afeda0a-b94b-43bf-bc65-70b50cd99a9e">
<img width="200" alt="Bildschirmfoto 2023-08-28 um 10 07 48" src="https://github.com/digitalfabrik/entitlementcard/assets/28013368/8f60c067-2e99-4872-92e6-cf4b5dbdf95a">
<img width="586" alt="Bildschirmfoto 2023-08-28 um 10 09 42" src="https://github.com/digitalfabrik/entitlementcard/assets/28013368/013c226d-ac33-4c3e-a97e-340797d03c35">

sarahsporck:
Maybe moving the expand into the header is an option (check out https://m2.material.io/components/cards#behavior). And last info: this view is only used in web so no reason to stress about mobile for now :)

hauf-toni:
👋 design proposal can be found [here](https://www.figma.com/file/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?type=design&node-id=31%3A1226&mode=design&t=uV7kx23srnS0TirO-1)

f1sh1918:
nice design proposal @hauf-toni 
[...]

PR #1091, opened 2023-08-18
1075: Create card from application

sarahsporck:
Little bit ugly, as we don't actually know the parsed json structure in the frontend. But works :)


administration/src/project-configs/bayern/config.ts
```diff
@@ -1,13 +1,40 @@
...
+  }
+
+  if (cardType) {
+    const cardTypeExtensionIdx = config.card.extensions.findIndex(ext => ext === BavariaCardTypeExtension)
+    const value = cardType.value === 'Goldene Ehrenamtskarte' ? 'Goldkarte' : 'Standard'
+    query.set(config.card.extensionColumnNames[cardTypeExtensionIdx] ?? '', value)
+  }
+
+  return '?' + query.toString()
+}
```
michael-markl, line 30: I think it would be good, if we could handle the application json a bit more carefully: 
The original reason for the unstructured data was that we might change the application form in the future (as requirements change), but we also need to keep old applications using the old version of the application form.
Hence, there might be applications of different "format" at the same time.

Imo, it would be best if we could explicitly model the error-path here by having `applicationJsonToCardQuery` return `null` if the card query cannot be created (e.g. because some important property like name or card type cannot be determined) and then disable the button / link.

More concretely, imo, we should not use `as JsonField<'xyz'>` here. Instead, afaik, you can use `if (jsonfield.type !== 'xyz') return null`  to check if the field has the correct type and TS does its type narrowing magic.

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -137,17 +160,30 @@ const ApplicationView: FunctionComponent<{
...
+            <ButtonGroup fill alignText='left'>
+              {height > COLLAPSED_HEIGHT ? (
+                <PrintAwareButton onClick={() => setCollapsed(true)} icon='caret-up'>
+                  Weniger anzeigen
+                </PrintAwareButton>
+              ) : null}
+              <PrintAwareButton onClick={() => printApplicationById(id)} intent='primary' icon='print'>
+                PDF exportieren
+              </PrintAwareButton>
+              <Tooltip
```
f1sh1918, line 172: in which case will this happen? If we show this tooltip it should explain why this can't be done.
Shouldn't all applications have sufficient information?
I would either explain why the card can't be created from application or even not show the button at all

administration/src/project-configs/bayern/config.ts
```diff
@@ -1,13 +1,40 @@
...
+  }
+
+  const forenames = findValue(personalData, 'forenames', 'String')
+  const surname = findValue(personalData, 'surname', 'String')
+
+  if (!forenames || !surname || !cardType) {
+    return null
+  }
+
+  query.set(config.card.nameColumnName, forenames.value + ' ' + surname.value)
```
f1sh1918, line 24:
```suggestion
  query.set(config.card.nameColumnName, `${forenames.value} ${surname.value}`)
```

f1sh1918 (commented):
Works fine. Nicely done.
I just don't like the design of this "ActionBar"
Maybe you can consult toni for some improvements.
At least we should not use the same color for two action buttons next to it and have the first gray and the last red. That looks very confusing.

image

administration/src/bp-modules/applications/ApplicationsOverview.tsx
sarahsporck: This was mainly a follow up on the remark @michael-markl made. Whe an application cannot be mapped to a card then this tooltip is displayed and the button is disabled. I'll thik about how the text could be improved :). Currently(!) this should never be the case. But what I definitely forgot was to not shouw this button for Nürnberg. I'll add this.

sarahsporck:
> Works fine. Nicely done. I just don't like the design of this "ActionBar" Maybe you can consult toni for some improvements. At least we should not use the same color for two action buttons next to it and have the first gray and the last red. That looks very confusing. 

I also thought about creating a ticket for toni. But in that case I think we need to change more than just the 'create card' button. I'll create a ticket for this. Should I wait for @hauf-toni to merge this pr then?

> Do the applicant and verifier view look the same?

Nope. In the verifier view the buttons are different.

hauf-toni:
I see @f1sh1918 's point here, feel free to create a ticket providing some context for the toolbar to improve the usability & design

f1sh1918:
> > Works fine. Nicely done. I just don't like the design of this "ActionBar" Maybe you can consult toni for some improvements. At least we should not use the same color for two action buttons next to it and have the first gray and the last red. That looks very confusing.
> 
> I also thought about creating a ticket for toni. But in that case I think we need to change more than just the 'create card' button. I'll create a ticket for this. Should I wait for @hauf-toni to merge this pr then?
> 
> > Do the applicant and verifier view look the same?
> 
> Nope. In the verifier view the buttons are different.

Maybe a short call would be sufficient with toni. I was thinking about sth like a "Dropdown" Button or a [Split Button](https://mui.com/material-ui/react-button-group/#split-button) where all "Weiterverarbeitende Aktionen" will be bundled as a primary Button and keep the "Antrag löschen" as it is

hauf-toni:
Is there any information that I need missing from this [thread](https://github.com/digitalfabrik/entitlementcard/issues/1075#issuecomment-1695250597)? Otherwise, I would start working on a design proposal for this theme and get back to you if I run into problems.

sarahsporck:
> Is there any information that I need missing from this [thread](https://github.com/digitalfabrik/entitlementcard/issues/1075#issuecomment-1695250597)? Otherwise, I would start working on a design proposal for this theme and get back to you if I run into problems.

Check out this issue for more reference: https://github.com/digitalfabrik/entitlementcard/issues/1075

f1sh1918:
@sarahsporck nice implementation :)

I think we should improve the ui of the verification a little bit by adding dividers and increase the margin between the elements

Example:
image

sarahsporck:
The columns could become very long, when there are more than one organisation verification. Although this could be addressed by allowing more column breaks, the layout would then become very full. Therefore, I had another exchange with @hauf-toni and decided to remove the columns in favor of more collapsibles.

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -205,26 +63,33 @@ const ApplicationsOverview = (props: { applications: Application[] }) => {
   const [updatedApplications, setUpdatedApplications] = useState(props.applications)
   const { applicationIdForPrint, printApplicationById } = usePrintApplication()
   const sortedApplications = useMemo(() => sortApplications(updatedApplications), [updatedApplications])
-
+  console.log(sortedApplications)
```
f1sh1918, line 66: rm log

f1sh1918 (changes requested):
Looks really nice. Not sure if "Persönliche Daten" shouldn't be collapsed initial.

On printView all collapsables should be collapsed. Probably you need a controlled state for this
image

administration/package.json
```diff
@@ -26,6 +26,7 @@
     "react-dom": "^18.2.0",
     "react-flip-move": "^3.0.5",
     "react-router-dom": "^6.14.0",
+    "react-to-print": "^2.14.13",
```
f1sh1918, line 29: couldn't find where this is in use? Can be removed?

f1sh1918 (approved):
Super nice works fine :)
Just one probably unused pkg

administration/package.json
sarahsporck: Jap. Missed that 👍🏼

PR #1094, opened 2023-08-22
Show time out of sync error

### LINKED ISSUE #1065: Card can't be activated

michael-markl:
This should resolve #1065, #1056, #1049

When during self-verification, the time of the device is out of sync with the sever, the device would generate a wrong otp code and, thus, the server would respond that the card is not valid. The current app-store version (<= 3.1.3) of the app had the additional problem, that if the server responded once with "invalid", the app would not try to self-verify again the next time the app launches. This was already fixed by #1071 . This means, in the current dev-version (also before this PR), if you landed in this screen, resynced your local time, and restarted the app, the error would go away.

The solution I implemented here, is to simply analyze whether the time of the local device is out of sync at the time of self-verification. If so, we show a simple error like "Die Zeit deines Geräts scheint nicht zu stimmen. Synchronisiere deine Uhrzeit in den Geräte-Einstellungen."

### DIFF
--- a/frontend/lib/about/dev_settings_view.dart
+++ b/frontend/lib/about/dev_settings_view.dart
@@ -9,6 +9,7 @@ import 'package:ehrenamtskarte/graphql/graphql_api.graphql.dart';
 import 'package:ehrenamtskarte/identification/activation_workflow/activate_code.dart';
 import 'package:ehrenamtskarte/identification/activation_workflow/activation_code_parser.dart';
 import 'package:ehrenamtskarte/identification/activation_workflow/activation_exception.dart';
+import 'package:ehrenamtskarte/identification/card_detail_view/self_verify_card.dart';
 import 'package:ehrenamtskarte/identification/qr_code_scanner/qr_code_processor.dart';
 import 'package:ehrenamtskarte/identification/qr_code_scanner/qr_parsing_error_dialog.dart';
 import 'package:ehrenamtskarte/identification/user_code_model.dart';
@@ -50,6 +51,8 @@ class DevSettingsView extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final settings = Provider.of<SettingsModel>(context);
+    final client = GraphQLProvider.of(context).value;
+    final userCodeModel = Provider.of<UserCodeModel>(context);
     return Padding(
       padding: const EdgeInsets.all(15.0),
       child: Column(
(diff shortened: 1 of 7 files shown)


frontend/lib/about/dev_settings_view.dart
```diff
@@ -74,6 +77,10 @@ class DevSettingsView extends StatelessWidget {
             title: const Text('Set expired last card verification'),
             onTap: () => _setExpiredLastVerification(context),
           ),
+          ListTile(
+            title: const Text('Trigger self-verification'),
+            onTap: () => selfVerifyCard(userCodeModel, Configuration.of(context).projectId, client),
+          ),
```
sarahsporck, line 83: This was already implemented in the 'eine andere Karte verifizieren' screen. However, I also see no pain in having it in another place as well :)

sarahsporck (approved):
You genius :) works like charm! Very nice 👍🏼

frontend/lib/identification/verification_workflow/query_server_verification.dart
```diff
@@ -3,19 +3,30 @@ import 'package:ehrenamtskarte/identification/util/card_info_utils.dart';
...
+  final timeBeforeRequest = DateTime.now();
+  final stopWatch = Stopwatch()..start();
+  final result = await _queryServerVerification(
     client,
     projectId,
     hash,
     verificationCode.otp,
     CodeType.kw$DYNAMIC,
   );
+  stopWatch.stop();
```
f1sh1918, line 22: i think it would be nice to add a comment here

frontend/lib/about/dev_settings_view.dart
michael-markl: I think in the "eine andere Karte verifizieren", it does not trigger the same process. Here, it triggers the self-verification process which also persists the card verification

michael-markl:
> You genius :) works like charm! Very nice 👍🏼

Credit goes to @f1sh1918 for discovering that time-offset is the actual issue. :D Let's hope that this actually fixes all of the problems.

f1sh1918:
> > You genius :) works like charm! Very nice 👍🏼
> 
> Credit goes to @f1sh1918 for discovering that time-offset is the actual issue. :D Let's hope that this actually fixes all of the problems.
> 

Hm not Sure if this also fixes the case when no activationTimestamp and totpSecret were written in the database but it's a super important fix :)

michael-markl:
> Hm not Sure if this also fixes the case when no activationTimestamp and totpSecret were written in the database but it's a super important fix :)

True! I'm also not sure what happened there yet. That might actually be a different problem still :/

PR #1095, opened 2023-08-22
1063: print cardInfoHash on pdf

### LINKED ISSUE #1063: [Druckerei] Print card info hash on pdf
As a service user i want to know to be able to identify a particular card to be able to let it revoke in the database. Therefore the card info hash should be printed on the pdf
- show the first 16 chars  (base64) of the card info hash below the qr code in small letters

f1sh1918:
resolves #1063 

i didn't see the need to shorten the hash? Any opinions?


michael-markl:
I'm fine with keeping the whole hash, but I think it is not perfectly aligned and a bit too small. Please consider merging in https://github.com/digitalfabrik/entitlementcard/pull/1097 :)

f1sh1918:
> I'm fine with keeping the whole hash, but I think it is not perfectly aligned and a bit too small. Please consider merging in #1097 :)

thx :) i merged please review my pr here

michael-markl (approved):

PR #1103, opened 2023-08-28
904: Add i18n for translations

### LINKED ISSUE #904: [Sozialpass] Provide the app in english
As a user i want to be able to switch the language to english

## Acceptance Criteria:
 - implement a button to switch the language from german to english and back

## Technical Tasks:
 - Implement i18n
 - implement a button switch to change the language
 - provide labels in english and german

steffenkleinle:
### Short description

This PR extracts all user visible text constants to a new [arb](https://github.com/google/app-resource-bundle) file asset and adds general support for i18n.

### Proposed changes

- [l10n/localization](https://docs.flutter.dev/ui/accessibility-and-localization/internationalization) is setup
- All text constants are extracted to a new arb file
- All former text constants are now translated using l10n

Fixes #904.

---


frontend/assets/bayern/translations/de.json
```diff
@@ -0,0 +1,130 @@
+{
+  "publisher": "Herausgeber",
```
steffenkleinle, line 2: @f1sh1918 @sarahsporck @michael-markl do you think it makes sense to streamline these constants here and perhaps introduce namespaces? I would stick close to the folder structure and use `about`, `identification`, `map`, `navigation`, `permission`, `search`, `store` and `category`. What do you think?

f1sh1918: Yes i think namespaces would be good to gain a better overview. I think you can do it here

michael-markl:
Just a general comment: Did you consider using the "standard flutter" approach of localization as described in their docs?
https://docs.flutter.dev/ui/accessibility-and-localization/internationalization

steffenkleinle:
> Just a general comment: Did you consider using the "standard flutter" approach of localization as described in their docs? https://docs.flutter.dev/ui/accessibility-and-localization/internationalization

@michael-markl tbh just because I am used to react-native where you need a library for everything, so I didn't even check.

However, one thing that does not work with the standard flutter approach: It uses .arb files intstead of normal .json files which doesn't support namespaces. What would works is keys like `categories_culture` or similar with underscores.

Opinions on whether we should use the standard way and work around the namespace issue or stick to the current library? @michael-markl @f1sh1918 @sarahsporck

michael-markl:
@steffenkleinle According to the ARB spec (see https://github.com/google/app-resource-bundle/wiki/ApplicationResourceBundleSpecification#arb-namespace-and-registration) there might be some namespace support, however, I don't know whether the Flutter implementation supports this.

michael-markl:
One benefit of the ARB-way would be "type"-safety (in the sense that you cannot misspell translation keys) as special Dart code is generated from the ARB file (if I understand correctly).

(It's not a recommendation from me, but I am kind of interested in these ARB files. I like that (a) the translation files have some proper spec, and (b) that translation keys can also hold meta data like a description or a screenshot).

steffenkleinle:
> One benefit of the ARB-way would be "type"-safety (in the sense that you cannot misspell translation keys) as special Dart code is generated from the ARB file (if I understand correctly).
> 
> (It's not a recommendation from me, but I am kind of interested in these ARB files. I like that (a) the translation files have some proper spec, and (b) that translation keys can also hold meta data like a description or a screenshot).

Agreeing with all the points. One drawback would be that the tools we use in the integreat project to handle constants won't work anymore/would need some adjustments. I think I would still give those ARB files a try.

steffenkleinle:
A few notes, while in general the localization works pretty nice (typesafety :tada:), it seems like no features are really implemented at the moment and there are a few problems for us:
- No nesting in an arb file -> Not possible to use nesting like we do in integreat. Instead I know used nesting identifiers separated by an underscore, e.g. `about_title`. Let me know if you don't like that.
- No namespaces or multiple files per locale possible even though its in the .arb specs -> might be a problem for translating the build config (even though there are hacky workarounds, e.g. including both `de` and `en` in the build config, won't scale though): https://github.com/flutter/flutter/issues/107157
- No dynamic translations possible, which is why the refactoring of the categoryAssets with context was necessary: https://github.com/flutter/flutter/issues/105672

All in all it seems like a great tool in general but not that actively maintained/improved. It might work for us for now but also might have too many limitations in the future.

steffenkleinle:
Tasks:
- [ ] Actually translate to English
- [ ] Disable English for bayern
- [ ] Translate build config

steffenkleinle:
This should work so far, if you change something in `app_en.arb`, it should show up in the app (if your system language is set to English).

f1sh1918 (approved):
Tested on android and ios seem to work fine for both even they recommend in the docs to add this into `Info.plist`.
Seem not to be require anymore.

```js
<key>CFBundleLocalizations</key>

<array>

    <string>en</string>

    <string>ar</string>

</array>
```

Even tested using different system languages on device

frontend/lib/about/license_page.dart
```diff
@@ -51,15 +53,15 @@ class CustomLicensePage extends StatelessWidget {
 
           return CustomScrollView(
             slivers: <Widget>[
-              const CustomSliverAppBar(title: 'Lizenzen'),
+              CustomSliverAppBar(title: context.l10n.about_licenses),
```
sarahsporck, line 56: I guess you removed the constant constructor as the value can now change?

frontend/lib/category_assets.dart
```diff
@@ -30,157 +29,159 @@ class CategoryAsset {
...
+      id: 16,
+      name: context.l10n.category_digitalParticipation,
+      shortName: context.l10n.category_digitalParticipation,
+      icon: 'assets/category_icons/16.svg',
+      detailIcon: 'assets/detail_headers/16_teilhabe.svg',
+      color: Color(0xFFFA0000),
+    ),
+    CategoryAsset(
+      id: 17,
+      name: context.l10n.category_sports,
```
sarahsporck, line 172: Should be category_sportsLong? -> `'Sport/Bewegung/Tanz'`

frontend/lib/about/license_page.dart
steffenkleinle: Yes, exactly.

frontend/lib/category_assets.dart
```diff
@@ -30,157 +29,159 @@ class CategoryAsset {
...
+      id: 11,
+      name: context.l10n.category_clothingLong,
+      shortName: context.l10n.category_clothing,
+      icon: 'assets/category_icons/11.svg',
+      detailIcon: 'assets/detail_headers/11_kleidung.svg',
+      color: Color(0xFF6EBE00),
+    ),
+    CategoryAsset(
+      id: 12,
+      name: context.l10n.category_cultureLongAlternative,
```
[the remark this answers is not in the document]
steffenkleinle: Because the categories 3 and 12 are both culture:
``` dart
  CategoryAsset(
    id: 3,
    name: 'Bildung/Kultur/Unterhaltung',
    shortName: 'Kultur',
  ), 
  CategoryAsset(
    name: 'Kultur/Museen/Freizeit',
    shortName: 'Kultur',
```

If you have other ideas let me know.

frontend/lib/category_assets.dart
steffenkleinle: Good catch, thanks!

frontend/lib/identification/verification_workflow/verification_qr_scanner_page.dart
```diff
@@ -75,49 +77,44 @@ class VerificationQrScannerPage extends StatelessWidget {
...
       await _onError(
         context,
-        'Der eingescannte Code ist bereits am ${dateFormat.format(e.expiry)} abgelaufen.',
+        context.l10n.identification_codeExpired(expirationDate),
         e,
       );
     } on QrCodeParseException catch (e) {
       await _onError(
         context,
-        'Der Inhalt des eingescannten Codes kann nicht verstanden '
```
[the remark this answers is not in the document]
steffenkleinle: No, it doesn't lead to the same linebreaks anymore. Let me know if I should readd them. Personally I think its perhaps better to let flutter handle it depending on the screen size?

frontend/lib/category_assets.dart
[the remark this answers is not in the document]
sarahsporck: The one with id 12 is a Nürnberg category. So we could call it `cultureLongNuernberg` for better understanding.

frontend/lib/identification/id_card/card_content.dart
```diff
@@ -46,11 +48,11 @@ class CardContent extends StatelessWidget {
   const CardContent(
       {super.key, required this.cardInfo, this.region, required this.isExpired, required this.isNotYetValid});
 
-  String get _formattedExpirationDate {
+  String _getFormattedExpirationDate(BuildContext context) {
     final expirationDay = cardInfo.hasExpirationDay() ? cardInfo.expirationDay : null;
     return expirationDay != null
         ? DateFormat('dd.MM.yyyy').format(DateTime.fromMillisecondsSinceEpoch(0).add(Duration(days: expirationDay)))
```
sarahsporck, line 54: Would it be possible to use localized date formats? Not necessary for this PR, but maybe you've read sth about this already.

sarahsporck (approved):
Looks good!

frontend/lib/identification/id_card/card_content.dart
steffenkleinle: Yes, that is possible. See https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization#messages-with-dates.

I will create an issue :)

frontend/lib/identification/verification_workflow/verification_qr_scanner_page.dart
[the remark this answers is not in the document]
steffenkleinle: Will keep it as it is :)

Issue #1001, opened 2023-06-02
Replace deprecated blueprintjs components


sarahsporck:
**Is your feature request related to a problem? Please describe.**
We use a few deprecated blueprintjs components. This could lead to bugs unnoticed.

**Describe the solution you'd like**
Replace all deprecated components with the 'newer' counterparts.

Deprecated Components:
- (DateInput)
- IResizeEntry
- Select

**Describe alternatives you've considered**
/

**Additional context**
The deprecated blueprintjs date input was removed, because it didn't work properly. The newer component might fix this (we currently use dateinput from material, as it just did not work properly). #1003 

Updating date input is blocked by https://github.com/palantir/blueprint/issues/5699

michael-markl:
I cannot access https://chat.tuerantuer.org/digitalfabrik/pl/qzsn9itzwtf1pfgndyjg3jopqw
Would be nice, if you could list the deprecated blueprintjs components in the issue description

PR #1104, opened 2023-08-28
Extract linting rules into separate file

sarahsporck:
Added rule `prefer-template` and use `react-app/jest` instead of `jest/recommended` and `jest/style`.


administration/src/bp-modules/cards/CreateCardsButtonBar.tsx
```diff
@@ -16,7 +15,7 @@ const CreateCardsButtonBar = ({ cardBlueprints, generateCards, goBack }: CreateC
   return (
     <ButtonBar>
       <Button icon='arrow-left' text='Zurück zur Auswahl' onClick={goBack} />
-      <Tooltip2
+      <Tooltip
```
f1sh1918, line 18: why did you switch to tooltip?

f1sh1918 (commented):
image

Sometimes the tooltip doesn't disappear on chrome

administration/src/bp-modules/cards/CreateCardsButtonBar.tsx
sarahsporck: Actually, I only switched because the tests had problems at first, and I assumed it was related to the deprecated Tooltip2 component. I can change it back in this PR, but at some point we (again) might want to move away from deprecated components (-> #1001 )

f1sh1918: ok i'm fine. dunno if the disappearing problem also occurs with tooltip2

f1sh1918 (approved):
Looks fine to me

steffenkleinle (approved):
LGTM, not testd

PR #1106, opened 2023-08-28
1105: Negative int value start day

### LINKED ISSUE #1105: StartDay Invalid uint 32 negative number
**Describe the bug**
When i add the startDay by typing into the input field i get "invalid uint 32" error with a negative value

**To Reproduce**
1. Go to Sozialpass
2. Add cards
3. Type in the required fields (name has to be filled at least)
4. Type a startDate. By reaching the year this error occurs

**Expected behavior**
This error doesn't occur and the value will be validated as usual

f1sh1918:
resolves #1105 

Since we use uint32 in protobuf we have to ensure that no negative values will be passed to the protobuf size check while typing the date, so i pass a placeholder date instead of negative value

I decided to add a minStartDay. I discussed this earlier with @michael-markl but i think it should not be possible to set a random startDate like `0020-01-01` even protobuf would fail so i use this placeholder date.

### DIFF
--- a/administration/src/cards/extensions/StartDayExtension.tsx
+++ b/administration/src/cards/extensions/StartDayExtension.tsx
@@ -8,12 +8,22 @@ import { Extension } from './extensions'
 
 type StartDayState = { startDay: number }
 
+const minStartDay = new PlainDate(2020, 1, 1)
+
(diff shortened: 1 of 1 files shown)


michael-markl (approved):
Not tested

PR #1109, opened 2023-08-28
1108: Prevent auto install entitlementcard

### LINKED ISSUE #1108: [CI] Prevent entitlementcard apt updates to be installed auto on production
Currently the apt updates for the entitlementcard will be installed via autoupdate every friday on production. This should only happen for staging not production

- Either put the apt packages in a directory where they will not be installed auto. Then move the packages with the `eak-update` script before installing them.
- Or deploy the packages not when a release branch was created but with a separate job

f1sh1918:
resolves #1108

script was deployed and tested on staging server

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,8 +69,8 @@ workflows:
           filters:
             branches:
               only:
-                - main
                 - /release.*/
+                - /1108.*/
       - deploy:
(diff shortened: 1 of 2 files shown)


steffenkleinle (approved):
not tested but LGTM

PR #1112, opened 2023-09-01
Upgrade backend dependencies

michael-markl:


sarahsporck (approved):
Roughly tested with frontend application.

PR #1113, opened 2023-09-04
Collect Statistics from Matomo

sarahsporck:
Credentials for matomo can be found in passbolt. For now I tracked:
- search
- card activation
- card verification
- card creation

in each request I tried to include:
- the user ip address
- user agent (for device information)
- language (especially interesting after we implement i18next)
- region (if possible)

Furthermore I decided to implement tracking completely in the backend, as this is where all information is present. This also means we don't have to implement tracking in frontend AND administration. 

What's still missing? 
- option to disable tracking in the frontend and probably also in administration
- a matomo health endpoint as errors are currently only logged, so it could go unnoticed if matomo is down or behaves strangely (@svenseeberg or does this already exist?)
- production credentials


administration/src/cards/createCards.ts
```diff
@@ -22,7 +22,7 @@ export class CreateCardsError extends Error {
   }
 }
 
-async function createCards(client: ApolloClient<object>, activationCodes: Codes, region: Region) {
+async function createCards(client: ApolloClient<object>, project: string, activationCodes: Codes, region: Region) {
```
sarahsporck, line 25: @f1sh1918  I realized that I added the project here. Is it okay to change endpoints in the administration, as we ship changes in web directly this might only mess with caching I guess. The other option would be to omit adding the project and instead extract it from the region. Any opinion on this?

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.matomo.java.tracking.MatomoRequest
+import org.matomo.java.tracking.MatomoRequestBuilder
+import org.matomo.java.tracking.MatomoTracker
+import org.slf4j.LoggerFactory
+import java.io.IOException
+import java.util.concurrent.ExecutionException
+
+enum class EventCategory {
+    Cards
+}
```
sarahsporck, line 20: We could also change the category. I guess adding `Cards` as category is kind of meaningless

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -0,0 +1,159 @@
...
+
+object Matomo {
+    val logger = LoggerFactory.getLogger(Matomo::class.java)
+    private const val ACTION_NAME = "Request"
```
sarahsporck, line 24: Without an action name the request is not displayed properly in matomo.

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -0,0 +1,159 @@
...
+            val response = tracker.sendBulkRequestAsync(matomoRequests)
+            val httpResponse = response.get()
+            if (httpResponse.statusLine.statusCode > 399) {
+                logger.debug("Request failed with status code ${httpResponse.statusLine.statusCode}")
+            }
+        } catch (e: Exception) {
+            when (e) {
+                is IOException -> logger.debug("Could not send request to Matomo")
+                is ExecutionException, is InterruptedException ->
+                    logger.debug("Error while getting response")
```
sarahsporck, line 72: Any recommendation on better loggings are welcome :)

frontend/lib/graphql/configured_graphql_provider.dart
```diff
@@ -1,13 +1,56 @@
...
+    // If the widget was removed from the tree while the asynchronous platform
+    // message was in flight, we want to discard the reply rather than calling
+    // setState to update our non-existent appearance.
+    if (!mounted) return;
+
+    setState(() {
+      _platformVersion = platformVersion;
+    });
+  }
+
```
sarahsporck, line 51: copy/pasted from [here](https://pub.dev/packages/fk_user_agent/example)

backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
```diff
@@ -23,14 +23,16 @@ data class MapConfig(val baseUrl: String)
 data class GeocodingConfig(val enabled: Boolean, val host: String)
 data class CsvWriterConfig(val enabled: Boolean)
 data class SmtpConfig(val host: String, val port: Int, val username: String, val password: String)
+data class MatomoConfig(val siteId: Int, val url: String, val accessToken: String)
```
michael-markl, line 26: Maybe we should do some sanity check that two different projects do not have the exact same matomo config?
Might prevent data corruption when new projects are added with copy pasta mistakes :D

administration/src/cards/createCards.ts
michael-markl: Should be okay in this case.

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -0,0 +1,159 @@
...
+                .eventCategory(EventCategory.Cards.toString())
+                .eventValue(if (successful) 1 else 0)
+                .customTrackingParameters(if (card != null) mapOf("dimension1" to card.regionId) else emptyMap())
+                .also { attachRequestInformation(it, request) }
+        )
+    }
+
+    fun trackSearch(projectConfig: ProjectConfig, request: HttpServletRequest, query: String, params: SearchParams, numResults: Int) {
+        if (projectConfig.matomo === null) return
+        if (params.categoryIds != null) return
```
michael-markl, line 149: Why do you reteurn when categoryIds is not null?
Maybe, we should use some special encoding for the matomo search query?

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -0,0 +1,159 @@
...
+
+        val matomoRequest = requestBuilder
+            .actionName(ACTION_NAME)
+            .siteId(siteId)
+            .authToken(matomoConfig.accessToken)
+            .build()
+
+        try {
+            val response = tracker.sendRequestAsync(matomoRequest)
+            val httpResponse = response.get()
```
michael-markl, line 39: I feel like, we should not await this future here, to not unnecessarily increase the delay the response of our API.
We should do the error logging asynchronously too.

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -0,0 +1,159 @@
...
+        val tracker = MatomoTracker(url)
+        val matomoRequests = requestBuilder.map {
+            it.actionName(ACTION_NAME)
+            it.siteId(siteId)
+            it.authToken(matomoConfig.accessToken)
+            it.build()
+        }
+        try {
+            val response = tracker.sendBulkRequestAsync(matomoRequests)
+            val httpResponse = response.get()
```
michael-markl, line 64: Same as above.

michael-markl (commented):
Just some drive-by comments;

I didn't pay attention to what's collected and if we can deduce all desired statistics from that.

administration/src/cards/createCards.ts
f1sh1918: as long as you don't use the endpoint in the flutter app thats fine. we ship backend and administration together

f1sh1918 (commented):
Hm i can not find matomo credential in passbolt.
Maybe i don't have sufficient rights?
Please provide me access to test :)

michael-markl:
On a side note: It *might*  be beneficial to also add a matomo docker instance to the dev setup, so we can more easily test the tracking locally. I am not sure how complicated that would be, though.

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
sarahsporck: See https://github.com/matomo-org/matomo-java-tracker/issues/133. I decided to add the categories in the search query. Do you think it makes more sense to add the plain text category names here? Would be a little bit more effort, but easier to evaluate.

sarahsporck:
> On a side note: It _might_ be beneficial to also add a matomo docker container to the dev setup, so we can more easily test the tracking locally. I am not sure how complicated that would be, though.

Added fancy docker setup for this :) Besides there is a dev site in our live matomo where we can test against the live system.

> Hm i can not find matomo credential in passbolt. Maybe i don't have sufficient rights? Please provide me access to test :)

Don't have the rights to do this @svenseeberg

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
sarahsporck: Never mind. I tested this again. Should work fine without the action name

docs/development-setup.md
```diff
@@ -57,6 +57,14 @@ adb reverse tcp:8000 tcp:8000
 12. Take a look at the style by viewing the test map: [http://localhost:5002](http://localhost:5002)
 13. Take a look at the backend: [http://localhost:8000](http://localhost:8000) (The public version is available at
     api.entitlementcard.app)
+14. Set up the matomo instance [http://localhost:5003](http://localhost:5003) (The public version is available at https://matomo-entitlementcard.tuerantuer.org)
+15. (optional) Add your matomo config to the backend config
+```yaml
+matomo:
```
f1sh1918, line 63: shouldn't this be added to config-local.yaml?
Even i do that i can not start the backend since
```js
Exception in thread "main" com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "matomo" (class app.ehrenamtskarte.backend.config.BackendConfiguration), not marked as ignorable (7 known properties: "server", "production", "projects", "geocoding", "map", "postgres", "csvWriter"])
```
Where exactly do i have to add this @sarahsporck

docs/development-setup.md
```diff
@@ -57,6 +57,14 @@ adb reverse tcp:8000 tcp:8000
...
+  siteId: 1
+  url: http://localhost:5003
+  accessToken: <matomo-access-token>
```
f1sh1918, line 66: how can i find the token?

docs/development-setup.md
sarahsporck: you need to add this to the project config. that's in the config.local.yml e.g.
```yaml
projects:
  - id: nuernberg.sozialpass.app
  ...
  matomo:
    ...
```

docs/development-setup.md
sarahsporck: visit `localhost:5003` and go to the settigns. Either search for "token" or go to "Persönlich->Sicherheit"

docs/development-setup.md
f1sh1918: so not in the first level? probably this was my problem. i got an error when i started the backend with this additional parameters

f1sh1918: could you please update the docs that its easier to know where to add matomo

docs/development-setup.md
f1sh1918: i had to adjust the `config.ini.php` on the docker container to changed the trusted hosts and add the port
Not sure if that is just my problem
```js
[General]
salt = "36941bec3f510a9c9411000c89ee712b"
trusted_hosts[] = "localhost:5003"
```

f1sh1918 (commented):
For me everything looks fine i just could need some help how to configure matomo locally to proper test this pr. The docker instance and everything works so far :)
And don't we have to put the matomo config for the backend in the `config.yml`  for the different stages?

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
f1sh1918: for me its fine

docs/development-setup.md
f1sh1918: @sarahsporck please add how to findthe `matomo-access-token` and update the `url` to `http://localhost:5003/matomo.php`

f1sh1918 (approved):
works fine. just some updates to docs are missing :)
Amazing work!

frontend/lib/graphql/configured_graphql_provider.dart
```diff
@@ -1,14 +1,57 @@
...
+
+
```
steffenkleinle, line 54:
```suggestion
```

frontend/lib/graphql/configured_graphql_provider.dart
```diff
@@ -1,14 +1,57 @@
...
+class ConfiguredGraphQlProvider extends StatefulWidget {
   final Widget child;
 
   const ConfiguredGraphQlProvider({super.key, required this.child});
 
+  @override
+  State<StatefulWidget> createState() => ConfiguredGraphQlProviderState();
+}
+
+class ConfiguredGraphQlProviderState extends State<ConfiguredGraphQlProvider> {
```
steffenkleinle, line 20: What are we using this whole thing for? What events/things are we tracking from the frontend? How is the tracking with matomo connected to graphql?

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
steffenkleinle: Do you plan to remove this then?

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -0,0 +1,165 @@
...
+import kotlinx.coroutines.launch
+import org.jetbrains.exposed.sql.transactions.transaction
+import org.matomo.java.tracking.MatomoRequest
+import org.matomo.java.tracking.MatomoRequestBuilder
+import org.matomo.java.tracking.MatomoTracker
+import org.slf4j.LoggerFactory
+import java.io.IOException
+import java.util.concurrent.ExecutionException
+
+object Matomo {
```
steffenkleinle, line 21: Optional: I think in this file there is some duplicated stuff which might be reduced.

f1sh1918:
@sarahsporck when do you expect to finish/merge this pr?
I would like to create a release this week

f1sh1918 (approved):
for me its fine to merge @sarahsporck

PR #1116, opened 2023-09-05
[fix]: add partial index for admin email

sarahsporck:
Should fix migrations not working properly on main.

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
@@ -10,6 +10,7 @@ import org.jetbrains.exposed.dao.id.IntIdTable
 import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.javatime.timestamp
+import org.jetbrains.exposed.sql.lowerCase
 import org.jetbrains.exposed.sql.or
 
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
```diff
@@ -35,6 +36,7 @@ object Administrators : IntIdTable() {
             (deleted eq Op.TRUE and (role eq Role.NO_RIGHTS.db_value)) or
                 (deleted eq Op.FALSE and (role neq Role.NO_RIGHTS.db_value))
         }
+        uniqueIndex(functions = listOf(email.lowerCase()))
```
michael-markl, line 39: To get rid of the remaining warning as well, use a custom index name:
```suggestion
        uniqueIndex(customIndexName = "email_lower_idx", functions = listOf(email.lowerCase()))
```

michael-markl (commented):
I do not think, this fixes any error, it only gets rid of a warning

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
sarahsporck: Nice ^^ for me there was no remaining warning.

sarahsporck:
> I do not think, this fixes any error, it only gets rid of a warning

Jep. You're right.

f1sh1918 (approved):

PR #1117, opened 2023-09-06
1114: Sort applications by status

### LINKED ISSUE #1114: Sort applications by status
As a user i want to see the approved and rejected applications first to be faster in finding processable applications (create cards or delete application)

- sort applications by status
- fully approved -> fully rejected -> rest
- each status itself should be sorted date asc

f1sh1918:
resolves #1114 

Please properly test the sorting by creating some applications with different status.
Applications that are not fully approved or rejected will have the status "waiting" and put in the end, since they are not actionable

### DIFF
--- a/administration/package-lock.json
+++ b/administration/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "administration",
-  "version": "3.1.5",
+  "version": "3.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "administration",
-      "version": "3.1.5",
+      "version": "3.2.0",
       "hasInstallScript": true,
       "dependencies": {
         "@apollo/client": "^3.7.16",
--- a/administration/src/bp-modules/applications/ApplicationsOverview.tsx
+++ b/administration/src/bp-modules/applications/ApplicationsOverview.tsx
@@ -178,13 +178,8 @@ export class ApplicationViewComponent extends React.Component<{
   }
 }
 
-const sortByStatus = (a: number, b: number): number => {
-  return a - b
-}
-
-const sortByDate = (a: Date, b: Date): number => {
-  return b.getTime() - a.getTime()
-}
+const sortByStatus = (a: number, b: number): number => a - b
+const sortByDateAsc = (a: Date, b: Date): number => a.getTime() - b.getTime()
 
 const getVerificationStatus = (status: number[]): VerificationStatus => {
   if (status.every(val => val === VerificationStatus.Verified)) return VerificationStatus.Verified
@@ -192,11 +187,11 @@ const getVerificationStatus = (status: number[]): VerificationStatus => {
   return VerificationStatus.Awaiting
 }
 
-// Applications will be sorted by unique status which means fully verified/rejected and within this status by creation date
+// Applications will be sorted by unique status which means fully verified/rejected and within this status by creation date asc
 const sortApplications = (applications: Application[]): Application[] =>
   applications
     .map(application => ({ ...application, status: getVerificationStatus(application.verifications.map(getStatus)) }))
-    .sort((a, b) => sortByStatus(a.status, b.status) || sortByDate(new Date(a.createdDate), new Date(b.createdDate)))
+    .sort((a, b) => sortByStatus(a.status, b.status) || sortByDateAsc(new Date(a.createdDate), new Date(b.createdDate)))
 
 const ApplicationsOverview = (props: { applications: Application[] }) => {
   const [updatedApplications, setUpdatedApplications] = useState(props.applications)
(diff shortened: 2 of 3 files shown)


administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -178,13 +178,28 @@ export class ApplicationViewComponent extends React.Component<{
   }
 }
 
+const sortByStatus = (a: number, b: number): number => a - b
+const sortByDateAsc = (a: Date, b: Date): number => a.getTime() - b.getTime()
+
+const getVerificationStatus = (status: number[]): VerificationStatus => {
```
michael-markl, line 184: I feel like it would be better to do define the function like
```ts
const getApplicationStatus = (application: Application): ApplicationStatus => {
```
and introduce a new `ApplicationStatus` enum (which would currently hold the same entries as VerificationStatus).

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -178,13 +178,28 @@ export class ApplicationViewComponent extends React.Component<{
...
+
+// Applications will be sorted by unique status which means fully verified/rejected and within this status by creation date asc
+const sortApplications = (applications: Application[]): Application[] =>
+  applications
+    .map(application => ({ ...application, status: getVerificationStatus(application.verifications.map(getStatus)) }))
+    .sort((a, b) => sortByStatus(a.status, b.status) || sortByDateAsc(new Date(a.createdDate), new Date(b.createdDate)))
+
 const ApplicationsOverview = (props: { applications: Application[] }) => {
   const [updatedApplications, setUpdatedApplications] = useState(props.applications)
   const { applicationIdForPrint, printApplicationById } = usePrintApplication()
```
michael-markl, line 199: Maybe you can use `useMemo` to create the sorted array only when the set of applications changes, and not every time this component is rendered. Something like this:

```ts
    const sortedApplications = useMemo(() => sortApplications(applications), [updatedApplications])
```

administration/src/bp-modules/applications/ApplicationsOverview.tsx
f1sh1918: yes i already thought about that. :)

administration/src/bp-modules/applications/ApplicationsOverview.tsx
f1sh1918: ah okay i thought if i define the function outside of the component it will not be rerendered every time. Where i defined the function there is no `props.applications` i think these hook are useful inside the component when they depend on state and/or lifecycle hooks and you can not define them outside @michael-markl . But maybe i got that wrong

administration/src/bp-modules/applications/ApplicationsOverview.tsx
f1sh1918: i renamed them for better understanding to `fullyRejected`, `fullyApproved`, `ambiguous`

administration/src/bp-modules/applications/ApplicationsOverview.tsx
michael-markl: React does (and can) not know if the sorting happens within an external function or not. Any call to an external function will be done whenever the component is rendered (ie. when the function `ApplicationsOverview` is exectued).

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -178,13 +184,38 @@ export class ApplicationViewComponent extends React.Component<{
...
+          .map(application => ({
+            ...application,
+            status: getApplicationStatus(application.verifications.map(getStatus)),
+          }))
+          .sort(
+            (a, b) =>
+              sortByStatus(a.status, b.status) || sortByDateAsc(new Date(a.createdDate), new Date(b.createdDate))
+          ),
+    [props.applications]
+  )
```
michael-markl, line 214: This now memoizes the function which sorts the applications.
This is not what we want; we want to memoize the actual sorted array.

f1sh1918: Ah okay I got it. Gonna change that

administration/package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "administration",
-  "version": "3.1.5",
+  "version": "3.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "administration",
-      "version": "3.1.5",
+      "version": "3.2.0",
```
michael-markl, line 9: Not sure if this change should be in this PR.

michael-markl (approved):
Not tested

PR #1118, opened 2023-09-06
1115: Trim search text

### LINKED ISSUE #1115: Enhance the search funktion by stripping trailing blanks
When searching for an acceptance point, typing the first letters and use the autocomplete function on ios or android the subject is not found because of a trailing blank added by the system. 

Strip occasionally trailing blanks before searching the database.

f1sh1918:
resolves #1115 

Test with: "Museum der Stadt Augsburg " -> now works
If you use auto complete on soft keyboard the additional empty space resulted in not getting a search result


sarahsporck (approved):
Tested on Android emulator

PR #1134, opened 2023-10-06
Update setup readme

ztefanie:
Updated some of the instructions to setup the project


docs/development-setup.md
```diff
@@ -50,14 +46,21 @@ adb reverse tcp:8000 tcp:8000
...
-12. Take a look at the style by viewing the test map: [http://localhost:5002](http://localhost:5002)
-13. Take a look at the backend: [http://localhost:8000](http://localhost:8000) (The public version is available at
+8. Clone the submodule inside `backend/ehrenamtskarte-maplibre-style`
+9. Run the backend migration: `run --args "migrate"`
+10. Run the backend: `cd backend && ./gradlew run --args="execute"` or `.\backend\gradlew.bat run --args="execute"` on Windows
+11. Create an admin account using `./gradlew run --args="create-admin <project> <role> <email> <password> <region>"`
+12. Take a look at the martin endpoints: [http://localhost:5002/tiles/accepting_stores/index.json](http://localhost:5002/tiles/accepting_stores/index.json) and [http://localhost:5002/tiles/accepting_stores/rpc/index.json](http://localhost:5002/tiles/accepting_stores/rpc/index.json). The data shown on the map is fetched from a hardcoded url and is not using the data from the local martin!
+13. Take a look at the style by viewing the test map: [http://localhost:5002](http://localhost:5002)
+14. Take a look at the backend: [http://localhost:8000](http://localhost:8000) (The public version is available at
```
ztefanie, line 55: Those links on ports 5002 and 8000 are not working for me. They all return `404 Not found` error, except `http://localhost:5002/tiles/accepting_stores/rpc/index.json` it returns `can not parse "accepting_stores" to a i16`. 

Is this also the case for you? If so, what would be the correct links? If not, how can I fix this?

sarahsporck: Okay, I did some evaluating. So the links currently don't work and that is correct, I guess. I guess @maxammann just forgot to remove them when he updated the `./docker/nginx*.config`. As in earlier days there was one config for the development setup and one for staging. Now, there is only one config which is used for both. Therefore, at least that's what I guess what happened here, [the map and style and other information have been stripped from nginx](https://github.com/digitalfabrik/ehrenamtskarte-maplibre-style/tree/22cc66f0a9a361372496fe2f2d29d4fb0ee70fdf) and the docs haven't been updated.
It is possible to readd the links. But for that I'd ask @maxammann or @f1sh1918, whether this would have other implications.

docs/development-setup.md
```diff
@@ -50,14 +46,21 @@ adb reverse tcp:8000 tcp:8000
...
     api.entitlementcard.app)
 
+# Administration
+
+1. Run `npm install`
+2. Create folder `generated` inside `administration/src`
```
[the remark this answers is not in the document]
f1sh1918: hm this def. shouldn't be needed. Maybe there are some issues with your protobuf version. Maybe you could add sth like if an error "<errorMessage>" occurs please create the folder xy

docs/development-setup.md
maxammann: Calling `http://localhost:8000` gives the root of the backend that is a 404 currently. This is expected. We could add a route in the backend to return some text or HTML.

A 404 on http://localhost:5002` is also expected. The root of the martin tile server returns a 404.

It is unexpected that getting the tilejson for the accepting stores return an error through this link `http://localhost:5002/tiles/accepting_stores/rpc/index.json`. I can not make sense of the error without digging deeper or setting it up.

Not sure about the changes in the docker setup. I think the URLs above should work as you need them for a local development setup.

sarahsporck: Then I'll propose a PR to readd support for those links and create another ticket for digging into the issue with `http://localhost:5002/tiles/accepting_stores/rpc/index.json`

ztefanie: So i will just add a hint to the docs that it is supposed these urls return a 404

sarahsporck: Keep it as it is, I'll fix it in #1141

sarahsporck: @ztefanie ahh. before I miss it. localhost:8000 is supposed to not work as this port is not even mirrored by docker in the docker compose. So you could just remove `14. ...`

sarahsporck (approved):

PR #1135, opened 2023-10-06
Remove optional keyword from card.proto

ztefanie:
When setting up the project I got this error:
"Explicit 'optional' labels are disallowed in the Proto3 syntax. To define 'optional' fields in Proto3, simply remove the 'optional' label, as fields are 'optional' by defaul"

The [docs](https://pub.dev/packages/protoc_plugin) say: "To compile a .proto file, you must use the protoc command which is installed separately. protoc 3.0.0 or above is required."

So i assume everybody has proto3, so it should be the best to remove the optional keyword. 

Which protoc version are you on and is this generation working for you?

### DIFF
--- a/specs/card.proto
+++ b/specs/card.proto
@@ -4,7 +4,7 @@ syntax = "proto3";
 message RegionExtension {
   // Using int32 instead of uint32, even though we do not expect negative values.
   // The reason is that int32 are better supported in GraphQL and SQL.
-  optional int32 regionId = 1;
+  int32 regionId = 1;
 }
 
 message BirthdayExtension {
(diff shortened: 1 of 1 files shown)


michael-markl:
I believe you are using a quite outdated version of protoc. The optional keywords in proto3 do have a semantic meaning, and they are used here intentionally.

https://protobuf.dev/programming-guides/proto3/#field-labels

michael-markl:
```sh
$ protoc --version
libprotoc 3.21.11
```

ztefanie:
Thanks for the hint
I was using `libprotoc 3.6.1` and updated now to `libprotoc 23.3`, which works fine

PR #1138, opened 2023-10-09
1137: generalize create admin queries

sarahsporck:


ztefanie (approved):
Tested and works as expected

Issue #1141, opened 2023-10-09
Add routes for map, styles, etc. for development


sarahsporck:
**Is your feature request related to a problem? Please describe.**
In our docs we link a few routes to see the currently configured map and styles. However they don't work anymore, as the support was dropped a few versions ago. Readd the support.

**Describe the solution you'd like**
Add the routes to the nginx config.

**Describe alternatives you've considered**
Dropping the support and removing the links from the docs.

**Additional context**
https://github.com/digitalfabrik/entitlementcard/pull/1134#discussion_r1348783148

PR #1143, opened 2023-10-09
1141: add route for map and styles to our reverse_proxy

sarahsporck:
### Short description

In #1134 we realized, that some urls did not work. This should be fixed in this PR.

### Proposed changes

- add route for map and styles to our nginx server
- mount submodule to nginx output

### Side effects

- there shouldn't be anythin
-

### Resolved issues

Fixes: #1141


ztefanie (approved):
Nice :tada: Works as expected

f1sh1918 (approved):
nicely done :)

Issue #1144, opened 2023-10-09
[Dev]Fix adding base64 cards


f1sh1918:
**Describe the bug**
Currently it is not possible to add cards with base64 which is especially important for ios because you can not scan qr-codes on simulators

**To Reproduce**
1. Go To Über
2. Follow instruction in documentation
3. Paste base64 code

**Expected behavior**
A valid card can be added to test the identification view

PR #1145, opened 2023-10-09
1144: Add base64 cards

f1sh1918:
### Short description

This pr enables an dev to easily add a card via base64 activationCode. Especially useful for ios testing

### Proposed changes

- print the activationCode
- adjust comments

### Resolved issues

Fixes: #1144


administration/src/cards/pdf/PdfQrCodeElement.ts
```diff
@@ -27,6 +28,11 @@ const pdfQrCodeElement: PdfElement<PdfQrCodeElementProps, PdfQrCodeElementRender
     qrCode: qrCode,
   }).toBinary()
 
+  // Log base64 activationCode for development purposes
+  if (qrCode.case === 'dynamicActivationCode' && window.location.hostname === 'localhost') {
```
sarahsporck, line 32: Maybe you can create a helper constant somewhere to reuse this. Sth like:
`const DEV_MODE = window.location.hostname === 'localhost'`

frontend/lib/about/dev_settings_view.dart
```diff
@@ -166,6 +166,7 @@ class DevSettingsView extends StatelessWidget {
     );
   }
 
+  // You can either copy/paste from clipboard on ios or type 'adb shell input text <text>' in terminall for android
```
sarahsporck, line 169: I think this only applies for macOS. On linux I didn't have any problems with sharing the clipboard with the host system. @ztefanie @steffenkleinle ?
```sh
adb shell input text <text>
```

f1sh1918:
created new pr on MultiCards to simplify testing

frontend/lib/about/dev_settings_view.dart
ztefanie: Yes this is not needed on linux. Also I think this comment is unnecessary, as how to share code between hostsystem and emulator is not specific to our code and can be googled easily. On browserstack there is a special hack to copy from clipbord to emulators clipbord and maybe behaviour changes on ios / android / linux whatever in the future and we would have to update our code then.

Issue #1081, opened 2023-08-16
[Sozialpass] Switch PassId and Passnumber


f1sh1918:
As a user i don't need the passnumber anymore for the sozialticket but only the passId for the DE-Ticket so i only want to have this passID on the card to not be confused about different numbers/id.

**Note:** To ensure backward compatibility in the app we keep the passnumber extension that was already delivered in the QR-Code and just rename the labels for the druckerei user.

## Acceptance Criteria:
 - PassId form field is removed from create cards
 - The passId will be shown in the printed and digital card
 - The passNr will be removed from the back of the printed card

## Technical Tasks:
 - Remove passId extension for the sozialpass (!!!)
 - adjust the labels from passName to passId in the csv import and create card form (buildConfig)
 - ensure that the activity log has the correct table header PassID
 - remove passnumber in the csv import example file
 - adjust the pdf template to only show the passId (printed card)
 
 Blocked:**Should be deployed after 10th of november**

SteffiStuffel:
@f1sh1918 why is this blocked?

f1sh1918:
> @f1sh1918 why is this blocked?

We need a final clearance when this should be deployed due to some  Regulation with the PassNumber which was needed to get certain social aid but will be obsolet starting next year

Issue #797, opened 2023-02-13
Hint the user to update the app if unknown protobuf fields were found


michael-markl:
In the case, the user wants to verify a card and the parser (or rather the hash algorithm) finds unknown fields  in the protobuf, we might want to hint the user that either the scanned card is invalid or his app is outdated.

Especially, activating a card should be denied when it contains unknown fields, and we should check if an app update is available.

_Originally posted by @michael-markl in https://github.com/digitalfabrik/entitlementcard/pull/787#discussion_r1097483562_
            
We should be very cautious, though: It is not desirable, that an "attacker" can trigger this warning by simply generating a QR code with unknown fields.

f1sh1918:
we can implement this if we have an use case for this. For example a new import field gonna be added to protobuf

PR #1147, opened 2023-10-10
1081: Switch passID and passNumber

ztefanie:
### Short description
Removed passId extension and renamed passnumber labels

### Resolved issues

Fixes: #1081


sarahsporck (changes requested):
Apparently you can also change the protobuf extension fields (as long as the type and the tag (the field number)) don't change (as this is what the system will look at. It would be awesome if you could do this and refactor the card.proto NuernbergPassNumberExtension to NuernbergPassIdExtension.

f1sh1918:
> Apparently you can also change the protobuf extension fields (as long as the type and the tag (the field number)) don't change (as this is what the system will look at. It would be awesome if you could do this and refactor the card.proto NuernbergPassNumberExtension to NuernbergPassIdExtension.

Hm okay i didn't know that. But how does an old app version read the barcode data if there is not the `NuernbergPassNumberExtension` but the `NuernbergPassIdExtension` ?

f1sh1918:
@ztefanie i think due to my lack of knowledge about protobuf (what sarah described) my created issue was confusing.

As far as i understood, you can just remove the `PassNumberExtension` and set the data of the `PassIdExtension` with the `setProtobufData` function. On the pdf you can just the remove the back content of the printed pdf.
I think right now there are a lot of wrong changes in the `PassIdExtension` like length and checks, labelFor, etc. They should stay the same. Maybe a revert would make sense. Sorry for the confusion. I didn't know that deep how protobuf is working.

Maybe @sarahsporck could interrupt if i got it wrong :)

sarahsporck (approved):
Like it a lot. Thoroughly tested the protobuf. For context. Renaming fields is only possible when **not** using the protobuf json or text representation, as well as no field masks. As in those cases the field name is explicitly required. As we convert our protobuf to "[canonical json](https://github.com/digitalfabrik/entitlementcard/issues/673)" via our own implementations we have control over this. And in that implementation we don't use field names but tags for the json keys instead.

sarahsporck (changes requested):
Sorry selected wrong state. I think you need to keen the validity check for the NuerbergPassIdExtension :)

administration/src/cards/extensions/NuernbergPassIdExtension.tsx
```diff
@@ -1,29 +1,31 @@
...
   public readonly name = NuernbergPassIdExtension.name
 
   setInitialState() {}
   createForm(onUpdate: () => void) {
     return (
       <FormGroup
-        label='Nürnberg-Pass-ID'
-        labelFor='nuernberg-pass-id-input'
+        label='Nürnberg-Pass-Id'
+        labelFor='nuernberg-pass-number-input'
```
f1sh1918, line 18: these label and labelFor should be reverted

administration/src/cards/extensions/NuernbergPassIdExtension.tsx
```diff
@@ -1,29 +1,31 @@
...
+import { PartialMessage } from '@bufbuild/protobuf'
 
+import { CardExtensions } from '../../generated/card_pb'
 import { Extension } from './extensions'
 
-const nuernbergPassIdNumberLength = 10
+type NuernbergPassIdState = { passId: number }
 
-type NuernbergPassIdState = { nuernbergPassId: number }
+const nuernbergPassIdLength = 8
```
f1sh1918, line 9: this should be reverted to 10

michael-markl:
My reply to https://github.com/digitalfabrik/entitlementcard/issues/673#issuecomment-1759320210, @sarahsporck 
> @michael-markl can I shortly come back to this issue with you? I'm currently looking at https://github.com/digitalfabrik/entitlementcard/pull/1147, and there we change the naming of 2 protobuf fields and one protobuf message.
I was wondering if this could be an issue for us, but as I see it, we should be completely free to change the names of fields, as they are not used at all in our protobufToJson implementation, right?

Renaming fields / messages in proto definitions does only change the code text, but not the binary representation on the wire. Also, the hashing is not affected as it only uses the field numbers and not their names (just as the binary represenation).

However, if I understand correctly, nürnberg pass id and pass number are not the same.
One of them can be used for the deutschland ticket.
To be able distinguish between the two (on old vs on new cards), I suggest to rather add a separate extension (or a new field in the existing extension). We can then deprecate the old extension (see https://protobuf.dev/programming-guides/proto3/#options) (and keep the old extension unset for new cards).

sarahsporck:
> However, if I understand correctly, nürnberg pass id and pass number are not the same. One of them can be used for the deutschland ticket. To be able distinguish between the two (on old vs on new cards), I suggest to rather add a separate extension (or a new field in the existing extension). We can then deprecate the old extension (see https://protobuf.dev/programming-guides/proto3/#options) (and keep the old extension unset for new cards).

Not all cards have a Nürnberg-Pass-Id. Therefore, the idea was to use the (now deprecated) Nürnberg-Pass-Nummer in cases where the Id is missing. Therefore, the whole switcheroo.

michael-markl:
> Not all cards have a Nürnberg-Pass-Id.

You mean the cards that have already been generated so far, right?

> Therefore, the idea was to use the (now deprecated) Nürnberg-Pass-Nummer in cases where the Id is missing.

If this is the expected behavior, we can just rename the field/message (while I still don't think it's great) .
But: Not being to able to (easily) distinguish if the value is a pass-number or a pass-id in the app seems bad to me. Hence, I still suggest creating a new field/extension.

f1sh1918:
> > Not all cards have a Nürnberg-Pass-Id.
> 
> You mean the cards that have already been generated so far, right?
> 
> > Therefore, the idea was to use the (now deprecated) Nürnberg-Pass-Nummer in cases where the Id is missing.
> 
> If this is the expected behavior, we can just rename the field/message (while I still don't think it's great) .
> But: Not being to able to (easily) distinguish if the value is a pass-number or a pass-id in the app seems bad to me. Hence, I still suggest creating a new field/extension.

Yes these old cards don't have the PassId.

But if we use a next extension what happens with old app versions that scan New Barcodes with passId

michael-markl:
> But if we use a new extension what happens with old app versions that scan new Barcodes with passId?

I don't know :D in the best case, it simply does not show the pass number when it cannot find it in the card info. In the worst case it crashes. But that's okay (imo) as long as we provide an app update supporting the new codes optimally before we generate the new codes.

f1sh1918:
> > But if we use a new extension what happens with old app versions that scan new Barcodes with passId?
> 
> I don't know :D in the best case, it simply does not show the pass number when it cannot find it in the card info. In the worst case it crashes. But that's okay (imo) as long as we provide an app update supporting the new codes optimally before we generate the new codes.

Well the app will not crash. Basically there won't be an id displayed. 

OK then it seems like we gonna force the user to update.
I already discussed that long before with Sarah and that's why we came to the conclusion (that's why I created the issue like this) to simply change Labels 🙄

michael-markl:
I think I understand your problem now:

> There will be a mass creation of new cards next year and there may be tons of issues from user that didn't update the app that will be handled in support calls.

I did not know that. So the people who already have a nürnberg pass will get a new one? Thus, a lot of people will still have an old app version installed and try to activate their new pass. Then, they want to activate their Deutschland ticket but the ID is not visible because of their outdated app.

Another possibility would be:
* Apply your renaming strategy
* Add another enum field to the extension that tells us whether the number in the field is the new pass id. For older cards this field will not be set, and with that we can at least distinguish in the app, whether the field is populated with a pass id or with a pass number. In a future app update, we should then (probably?) hide the field if it's a pass number (and not an ID), so that people don't try to buy their Deutschlandticket with the pass number (which might happen a lot still^^).

Also related: https://github.com/digitalfabrik/entitlementcard/issues/797. If we had this already, we could more swiftly force the user to update their app when we encounter unknown/new fields. That's still relevant for future changes.

f1sh1918:
> I think I understand your problem now:
> 
> > There will be a mass creation of new cards next year and there may be tons of issues from user that didn't update the app that will be handled in support calls.
> 
> I did not know that. So the people who already have a nürnberg pass will get a new one? Thus, a lot of people will still have an old app version installed and try to activate their new pass. Then, they want to activate their Deutschland ticket but the ID is not visible because of their outdated app.
> 
> Another possibility would be:
> * Apply your renaming strategy
> * Add another enum field to the extension that tells us whether the number in the field is the new pass id. For older cards this field will not be set, and with that we can at least distinguish in the app, whether the field is populated with a pass id or with a pass number. In a future app update, we should then (probably?) hide the field if it's a pass number (and not an ID), so that people don't try to buy their Deutschlandticket with the pass number (which might happen a lot still^^).
> 
> Also related: https://github.com/digitalfabrik/entitlementcard/issues/797. If we had this already, we could more swiftly force the user to update their app when we encounter unknown/new fields. That's still relevant for future changes.

Hm I don't think the problem with the old Passnr is a big issue since the sozialpass is just valid for one year and then will be invalid and they get a new one with the  PassId and can delete the old one. But to have an indicator is a good idea

administration/src/cards/extensions/NuernbergPassIdExtension.tsx
```diff
@@ -1,29 +1,31 @@
...
     return (
       <FormGroup
-        label='Nürnberg-Pass-ID'
-        labelFor='nuernberg-pass-id-input'
+        label='Nürnberg-Pass-Id'
+        labelFor='nuernberg-pass-number-input'
         intent={this.isValid() ? undefined : Intent.DANGER}>
         <InputGroup
-          id='nuernberg-pass-id-input'
+          id='nuernberg-pass-number-input'
```
f1sh1918, line 21: should also be id. revert pls

specs/card.proto
```diff
@@ -42,7 +42,7 @@ message StartDayExtension {
 message CardExtensions {
   optional RegionExtension extension_region = 1;
   optional BirthdayExtension extension_birthday = 2;
-  optional NuernbergPassNumberExtension extension_nuernberg_pass_number = 3;
```
f1sh1918, line 45: adding here an enum would be nice whether its PassID or PassNr and set the passId for now

michael-markl: I think it would make sense to make PassNr the default (i.e. the one at position 0), as we cannot set it for older cards.

f1sh1918 (commented):
There are some small things to adjust.
But works fine so far.
Tested on android with old and new qrCode.
Tested the new QRCode also with an old app version

sarahsporck (approved):
Tested an Android emulator and seems to work pretty fine. Thank you for your patience through the discussion. Awesome work!

specs/card.proto
```diff
@@ -39,10 +35,20 @@ message StartDayExtension {
...
+  passNr = 0;
+  passId = 1;
+}
+
 message CardExtensions {
   optional RegionExtension extension_region = 1;
   optional BirthdayExtension extension_birthday = 2;
-  optional NuernbergPassNumberExtension extension_nuernberg_pass_number = 3;
+  optional NuernbergPassIdExtension extension_nuernberg_pass_id = 3;
+  optional NuernergPassIdentifier nuernberg_pass_identifier = 30;
```
f1sh1918, line 51: should this be a separate extension? I expected this to  be added to the `NuernbergPassIdExtension`

f1sh1918, line 51: will this be set somewhere @ztefanie?

f1sh1918 (changes requested):
Hm should this `NuernergPassIdentifier` be a separate extension?
I think it should even be set a value for it for the new cards

ztefanie:
> Hm should this `NuernergPassIdentifier` be a separate extension? I think it should even be set a value for it for the new cards

I do not get, what you want me to do. Please explain in more detail.

f1sh1918:
> > Hm should this `NuernergPassIdentifier` be a separate extension? I think it should even be set a value for it for the new cards
> 
> I do not get, what you want me to do. Please explain in more detail.

You basically created an extension and didn't add any information into it. (enum extension) So there should be added the enum `PassID` to indicate that its the PassId.
Furthermore this enum could be added in `NuernbergPassIdExtension` not sure whats the best way @michael-markl , @sarahsporck 
I even thought renaming the extension  NuernbergPassNumberExtension` to `NuernbergPassIdExtension` is not what we want to do? Thats why we added this enum to indicate whats behind

michael-markl:
> In the frontend when the protobuf is decoded... wait a second. How do we know what value we get here?

From the added `identifier` enum; that's the point of adding it. We also have to set its value to pass_id when generating new cards, of course. If `id_extension.identifier` is not set, we know that the value is a pass_number. Also, afaik, accessing a protobuf field that is not set in the binary representation, returns whatever "zero" stands for -- in our case pass_number.

Maybe, I also misinterpret your question; let me know :)

ztefanie:
> After that the missing steps are:
> 
> * If a new card is created in administration the `identifier` should be set to `passId` (1)
> * In the frontend when the protobuf is decoded... wait a second. How do we know what value we get here?
> 
> @ztefanie if you don't mind, I'd take over from here, especially as you'll also be on vacation next week. Maybe we'll still need the old field additionally, as otherwise we might not be able to differ the card type in the frontend.

@sarahsporck thanks for taking over, i am really confused here.

administration/src/cards/hashCardInfo.test.ts
```diff
@@ -183,8 +183,8 @@ describe('hashCardInfo', () => {
         extensionBirthday: new BirthdayExtension({
           birthday: -365 * 10,
```
michael-markl, line 184: We should also add another test that sets the new identifier to passid. (Also in the frontend)

michael-markl (commented):
Looks good to me, codewise! Sorry for any confusion I might have contributed here, but the solution is good I think :D

f1sh1918 (approved):
Tested on android with old and new barcodes. Tested new qrCode on ios.Works fine. nicely done :)

Issue #1151, opened 2023-10-11
[Druckerei] Increase timeout for logout


f1sh1918:
**Is your feature request related to a problem? Please describe.**
The users of the druckerei want to have a longer timeout for being logged out.

**Describe the solution you'd like**
- Increase the logout timeout to 60min (3600s)

f1sh1918:
not needed

PR #1162, opened 2023-10-12
1151: Increase logout time

f1sh1918:
The user will now auto logged out after 60mins (3600s)
Fixes: #1151

### DIFF
--- a/administration/src/AuthProvider.tsx
+++ b/administration/src/AuthProvider.tsx
@@ -24,7 +24,7 @@ const removeToken = () => window.localStorage.removeItem(LOCAL_STORAGE_KEY)
 
 const extractTokenPayload = (token: string): TokenPayload => {
   const payload: { exp: number; adminId: number } = JSON.parse(atob(token.split('.')[1]))
-  const expiry = new Date(payload.exp * 1000) // exp is in seconds, not milliseconds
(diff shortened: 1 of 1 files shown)


administration/src/AuthProvider.tsx
```diff
@@ -24,7 +24,7 @@ const removeToken = () => window.localStorage.removeItem(LOCAL_STORAGE_KEY)
 
 const extractTokenPayload = (token: string): TokenPayload => {
   const payload: { exp: number; adminId: number } = JSON.parse(atob(token.split('.')[1]))
-  const expiry = new Date(payload.exp * 1000) // exp is in seconds, not milliseconds
+  const expiry = new Date(payload.exp * 3600) // exp is in seconds, not milliseconds.
```
michael-markl, line 27: I feel like this is not the right fix. The "*1000" simply converts to milliseconds.

f1sh1918: Ok that thats probably true so the payload.exp should be the right one :)

michael-markl: Yes, the expiry is set in the backend somewhere.

f1sh1918: found it and its already 1hour :)

f1sh1918:
will be closed. nothing to do here

Issue #1154, opened 2023-10-12
Split up circleci config


ztefanie:
For better readability circleci config should be split up.

- Create the subfolder `src/commands` `src/jobs` and `src/workflows` inside `.circleci`
- put every command, job and workflow in a separate file
- create autogenerated header file
- add commands to merge file to a config file
- create a check if config file is up to date
- add the check to ci if config file is up to date

PR #1166, opened 2023-10-12
1154 split circleci config

ztefanie:
### Short description
Have a look at the second commit, as the first one is just the splitting up with no other changes.

### Proposed changes

- Splitted up config file
- added a check for valid config file
- added script to build config file
- added workflow "commit" for checks that should be done for every commit, as the current flow is spending a lot of circleci credits unnecessarily 

### TODO before merging
- [ ] adjust github settings to execute the commit workflow instead of `frontend` and `backend` workflows

### Resolved issues

Fixes: #1154


scripts/circleci-update-config.sh
```diff
@@ -0,0 +1 @@
+{ cat .circleci/autogenerated_header.yml; circleci config pack .circleci/src; } > .circleci/config.yml && circleci config validate
```
f1sh1918, line 1: i think this script won't work if you execute it from the `/scripts` folder

f1sh1918 (commented):
Looks good except of the not working update script for me
you can ignore that just got it work
`bash scripts/circleci-update-config.sh`

f1sh1918 (approved):
Looks fine to me

PR #1167, opened 2023-10-12
Fix build config

f1sh1918:
### Short description

There were some issue in the build config of the flutter app. Types were wrong and we don't need the features flag verification for android, i think. Also the bundleId was wrong for sozialpass.

### Proposed changes

- remove `featureFlags` from android
- rename to `buildFeatures` since it was misleading and using same name
- adjust bundleIdentifier


michael-markl:
I think there was / is? a feature flag `verification` that told the app whether the verification (Ausweisen) tab should be shown (and possibly if CAMERA permission should be requested).

f1sh1918:
> I think there was / is? a feature flag `verification` that told the app whether the verification (Ausweisen) tab should be shown (and possibly if CAMERA permission should be requested).

yea this is true but thats a common `featureFlag` for both platforms and i doesn't make sense to put it only to android.
Even your point with the camera permission couldn't be true but there is not check for permission.

I think we shouldn't use `featureFlags`  for android builds. I just found the `featureFlags` used in `app/build.gradle` an renamed them

frontend/android/app/build.gradle
```diff
@@ -49,13 +49,13 @@ ext.setupApplication = { ApplicationProductFlavor flavor, buildConfigName ->
...
 ext.setupVariant = { variant, buildConfigName ->
     var buildConfig = createBuildConfig(buildConfigName)
-    if (buildConfig.featureFlags.excludeX86) {
+    if (buildConfig.buildFeatures.excludeX86) {
         // Verify this using:
         // unzip -l frontend/build/app/outputs/apk/Bayern/debug/app-Bayern-debug.apk | grep x86
         // unzip -l frontend/build/app/outputs/apk/Nuernberg/debug/app-Nuernberg-debug.apk | grep x86
         variant.packaging.jniLibs.excludes.add('**/lib/x86**')
     }
-    if (buildConfig.excludeLocationPlayServices) {
```
f1sh1918, line 58: hm this should have thrown an error before

sarahsporck (approved):
Seems to do what it should do 👍🏼

sarahsporck:
I'm wondering whether we should hide the location button and the `In meiner Nähe suchen` button in Floss builds, as location won't work anyways.

Issue #1176, opened 2023-10-18
[Sozialpass] Configure available languages in build config


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
It should be configurable which language the app is translated in, e.g. nuernberg wants german and english while bayern only wants german atm.

**Describe the solution you'd like**
There should be an option as a list of locales in the build configs in which the app should be translated. Those should be used in [app.dart](https://github.com/digitalfabrik/entitlementcard/blob/f84710b44a9beba7834c2896912187392251bfa2/frontend/lib/app.dart#L74) to configure the available locales.

This should be:
- nuernberg: `['en', 'de']`
- bayern: `['de']`

**Describe alternatives you've considered**
Have the same languages for all whitelabels.

**Additional context**
Part of #1174.

PR #1180, opened 2023-10-18
1176: Configure locales in build config

steffenkleinle:
### Short description

Add the available locales to the build config and use them to configure the app locales.

### Proposed changes

- Add `appLocales` to build config
- Fix builder.dart script
- Use `appLocales` to configure the app locales

### Side effects

None.

### Resolved issues

Fixes: #1176.

### DIFF
--- a/frontend/build-configs/bayern/index.ts
+++ b/frontend/build-configs/bayern/index.ts
@@ -31,6 +31,7 @@ export const bayernCommon: CommonBuildConfigType = {
         showcase: "https://api.entitlementcard.app",
         local: "http://localhost:8000",
     },
+    appLocales: ['de'],
     cardBranding: {
         headerTextColor: "#008dc9",
         headerColor: "#F5F5FFF5",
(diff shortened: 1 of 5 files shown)


steffenkleinle:
To test this, you can set the system language to English and change one translation in `app_en.arb` to something else. For bayern this should not show up in the app while for nürnberg it should.

frontend/l10n.yaml
```diff
@@ -1,3 +1,3 @@
 arb-dir: lib/l10n
-template-arb-file: app_en.arb
+template-arb-file: app_de.arb
```
f1sh1918, line 2: why this change?

steffenkleinle: Basically it doesn't really matter as long as the translation files are identical/have the same keys. But since (at least atm) all our apps/white labels have german as a language but not every app has english I figured using german as a default scheme makes more sense.

Issue #1039, opened 2023-07-03
Can't upload csv in firefox live system


sarahsporck:
**Describe the bug**

**To Reproduce**
1. Go to bayern.ehrenamtskarte.app/cards/import
2. Click on Mehrere Karten erstellen
3. Lade die gegebene CSV hoch (*)
4. See error "Ungültiger Dateityp"

(*) CSV datei
```csv
Name,Ablaufdatum,Kartentyp
Thea Test,03.3.2026,Standard
Tilo Torvitz,,Goldkarte
Torben Trost,,Goldkarte
Tina Tuchel,12.10.2024,Standard
Tamara Tsaregorodtseva,05.05.2025,Standard
Tom Thomasson,05.05.2025,Standard
Torsten Tuchele,1.1.2025,Goldkarte
```
**Expected behavior**
File is loaded and entries are displayed

**Desktop (please complete the following information):**
 - Browser firefox
 - Version ?

**Additional context**

maybe check out the discussion in this stackoveflow answer
https://stackoverflow.com/a/11834872

ztefanie:
I cannot reproduce this. When I save the file as a .csv it works as expected. If I remove the file ending, uploading fails as expected. Also exporting a csv from libre-office works as expected, creating a correct csv that is recognized correctly.

I tried Firefox 118.0.2 (a very new version)
and Firefox 75.0 (release from april 2020)

sarahsporck:
I tested locally, if removing the "accept" value in the html changes anything and found that it doesn't, so I'd try to reopen #1059 and see if that changes anything. I'm pretty sure this will fix this.

PR #1181, opened 2023-10-19
1039: add mime type text/csv for firefox ubuntu

sarahsporck:
### Short description

I'm pretty sure, that this fixes the problem, although I'm not sure why it occurs in the first place 🤷🏼‍♀️ .

### Proposed changes

- add accept header for text/csv to input
-

### Side effects

- Technically nothing, but please still test wether the csv upload still works for you and this change will not have an implication on existing working functionality.

### Resolved issues

Fixes: #1039


f1sh1918 (approved):
not tested

Issue #1128, opened 2023-09-23
Log errors as ERRORS in backend


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Currently some errors are logged with log level info. To improve working with the backend log we could change this to log level ERROR.

**Describe the solution you'd like**
Use correct log type.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1185, opened 2023-10-20
1128: log errors as errors

sarahsporck:
### Short description

@svenseeberg requested that we should change the log level of errors to errors to improve filtering out error messages in the live system.

### Proposed changes

- log all thrown (and logged) exceptions on error level

### Resolved issues

Fixes: #1128


svenseeberg (commented):
Looks good to me :) Thanks

f1sh1918 (commented):
There are some `logger.debug` in matomo @sarahsporck 
Shouldn't this also be converted to error when we start using matomo for production?
In `FilterLbe.kt` is an catch block with exception that should be converted to error level

f1sh1918 (approved):

PR #1187, opened 2023-10-26
1029: proposal empty map auto screenshot

sarahsporck:
### Short description

I moved the commands into the fastfile, as it is pretty hard to write out the whole command you proposed in the command line. Besides that I have had issues with active support (which I didn't commit), didn't you have the same problem? -> https://github.com/fastlane/fastlane/issues/21585

### Proposed changes

- move commands into fastfile
- pass launch arguments when testing from xcode


f1sh1918 (approved):
very nice implementation.

Issue #1191, opened 2023-10-27
System mail does not visualize umlauts the right way


dkehne:
I have no further bug description except the fact that the person has on devices where the umlauts are not show right has also the problem that the link does not work when copied/pasted, even it works from other devices where the umlauts are shown correctly. 
<img width="802" alt="278582684-1c63d02d-2fa7-4fea-8284-352b5ea51216" src="https://github.com/digitalfabrik/entitlementcard/assets/4182469/68eff933-be28-494d-b52b-ae58ec074ec2">

ztefanie:
This is an encoding problem, the mail in the screenshot is displayed in `iso 8859-1`, but we have a header saying it should be displayed in `utf8`. I opened a PR change dit to `UTF-8` as this is the correct wording, as this can fix the error for some users, see:

> The charset value should be utf-8, not any other values such as utf8. Using utf8, for example, is a common mistake, and even though it is valid nowadays as the [specifications](https://encoding.spec.whatwg.org/#names-and-labels) and browsers now alias utf8 to utf-8, that wasn’t the case in the past, so things might break in [some older browsers](https://twitter.com/jacobrossi/status/591435377291866112). The same may be true for other agents (non-browsers) that may scan/get the content and may not have the alias.
[source](https://webhint.io/docs/user-guide/hints/hint-meta-charset-utf-8/#:~:text=The%20charset%20value%20should%20be,break%20in%20some%20older%20browsers.)

But this error can also occure when users force their mail client to use another encoding as we put in the header. So if this error does not disappear for the user, i recomment checking the settings or updating the client.

We also can make sure the link only contains ASCI chars, but I think it alread does, to at least ensure the link is always working.

PR #1194, opened 2023-10-31
1191: Fix mail content type header charset

ztefanie:
### Short description

Changed encoding

### More info

This is an encoding problem, the mail in the screenshot is displayed in `iso 8859-1`, but we have a header saying it should be displayed in `utf8`. I opened a PR change dit to `UTF-8` as this is the correct wording, as this can fix the error for some users, see:

> The charset value should be utf-8, not any other values such as utf8. Using utf8, for example, is a common mistake, and even though it is valid nowadays as the [specifications](https://encoding.spec.whatwg.org/#names-and-labels) and browsers now alias utf8 to utf-8, that wasn’t the case in the past, so things might break in [some older browsers](https://twitter.com/jacobrossi/status/591435377291866112). The same may be true for other agents (non-browsers) that may scan/get the content and may not have the alias.
[source](https://webhint.io/docs/user-guide/hints/hint-meta-charset-utf-8/#:~:text=The%20charset%20value%20should%20be,break%20in%20some%20older%20browsers.)

But this error can also occure when users force their mail client to use another encoding as we put in the header. So if this error will not disappear for the user by my changes, i recommend checking the settings or updating the client.

We also can make sure the link only contains ASCI chars, but I think it alread does, to at least ensure the link is always working.

### Resolved issues

Fixes: #1191

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
@@ -55,7 +55,7 @@ object Mailer {
                         .from(fromName, smtpConfig.username)
                         .withSubject(subject)
                         .withPlainText(message)
-                        .withHeader("Content-Type", "text/plain; charset=utf8")
+                        .withHeader("Content-Type", "text/plain; charset=UTF-8")
                         .buildEmail()
                 ).join()
         } catch (exception: MailException) {
@@ -160,6 +160,7 @@ object Mailer {
         personalData: PersonalData,
         accessKey: String
     ) {
+        println("MAIL SENT")
         val message = """
         Guten Tag ${personalData.forenames.shortText} ${personalData.surname.shortText},
 
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
@@ -160,7 +160,6 @@ object Mailer {
         personalData: PersonalData,
         accessKey: String
     ) {
-        println("MAIL SENT")
         val message = """
         Guten Tag ${personalData.forenames.shortText} ${personalData.surname.shortText},
 


backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -55,7 +55,7 @@ object Mailer {
                         .from(fromName, smtpConfig.username)
                         .withSubject(subject)
                         .withPlainText(message)
-                        .withHeader("Content-Type", "text/plain; charset=utf8")
+                        .withHeader("Content-Type", "text/plain; charset=UTF-8")
```
michael-markl, line 58: Good catch. Maybe we also need to wrap it in quotes? Most mails I receive personally have it wrapped... (not sure what the standard says though) 🤔 , like so:

```suggestion
                        .withHeader("Content-Type", "text/plain; charset=\"UTF-8\"")
```

michael-markl (commented):
Maybe, we should also set this `Content-Transfer-Encoding` header (it's at lost set for most of my incoming mails). I'm not entirely sure what it does, though.

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
ztefanie: I was also thinking about putting it into quotes, as e.g. google does it for mails they send, but the I kept it as it is, as the [RFC2045](https://www.ietf.org/rfc/rfc2045.txt) from 1992 already defined:

> the  following two forms
  Content-type: text/plain; charset=us-ascii (Plain text)
  Content-type: text/plain; charset="us-ascii"
  are completely equivalent.

So we should be fine with sticking to it without quote.

ztefanie:
> Maybe, we should also set this `Content-Transfer-Encoding` header (it's at least set for most of my incoming mails). I'm not entirely sure what it does, though.
> 
> _EDIT_: Maybe, it's also already set by our email library?

The simplejavamail package we use, sets it default to "quoted-printable", which is (or should be) completely fine with UTF-8 encoding.

sarahsporck (approved):
Very nice! Hopefully this will fix most of our mail encoding issues, we had in the past. 👍🏼

sarahsporck (approved):

f1sh1918 (approved):
Looks fine to me :)

PR #1204, opened 2023-11-13
1161: Deliver administration & backend beta

f1sh1918:
Whats has been done:
- [x] Split up `administration` & `backend` jobs
- [x] Split up build jobs in `check_` and `build_` 
- [x] Created workflows for `beta` and `production` for both jobs
- [x] Trigger webhooks for installation of debian packages
- [x] Renamed some job names that were wrongly named  
- [x] Use proper package name with version info for packaging
- [x] Pass version info to administration build
- [x] Created `commit_main` workflow for checking the builds and removed build checking from `commit`
- [x] Unhold apk packages from installation on Testserver (executed on testserver)

The production delivery will be done in a separate task #1271


f1sh1918:
Hey @ztefanie,
here is my first working draft for administration & backend jobs.
I'm just not sure how we want to handle "promotion" process because in integreat i think we just copy the beta artifacts.
If we deploy from a release branch/tag it might not make that big difference if we trigger that delivery_process again

.circleci/src/jobs/deploy_administration.yml
```diff
@@ -0,0 +1,34 @@
+docker:
```
ztefanie, line 1: We use the wording `deliver` not `deploy`, i would stick to this also for the job.

.circleci/src/jobs/deploy_administration.yml
```diff
@@ -0,0 +1,34 @@
...
+  - image: cimg/base:2022.09
+parameters:
```
ztefanie, line 3: I think it is better to pass a enum ['production' or 'beta'] an use this to evaluate which server and fingerprint should be used.

.circleci/src/jobs/deploy_administration.yml
```diff
@@ -0,0 +1,34 @@
...
+      - '|1|iikuvSrIo3wkj+EqUgLRMsAq6yk=|r9bSjkawWFa94b45qE/se5Oio5k= ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIED3MwobbCs+ENMLXdyqlJb3bJ26TuIYt977TA3NrN66'
+    type: enum
+steps:
+  - attach_workspace:
+      at: /tmp/workspace
+  - add_ssh_keys:
+      fingerprints:
+        - "a1:3f:a7:c3:ff:12:40:1d:85:de:a7:ab:12:3f:cc:05"
+  - run:
+      name: "SFTP upload package"
```
ztefanie, line 21: No need for quotes in the name

.circleci/src/jobs/deploy_administration.yml
```diff
@@ -0,0 +1,34 @@
...
+      command: |
+        echo "<< parameters.ssh-host-fingerprint >>" >> known_hosts
+        echo "Uploading: " /tmp/workspace/debs/administration/*.deb
+        sftp -b - -o UserKnownHostsFile=known_hosts ci@<< parameters.server >>:/local-apt-repository/ \<<< "put -r /tmp/workspace/debs/administration/*.deb"
+  - run:
```
ztefanie, line 26: I think it is better to put the install step to the beginning of the job.

.circleci/src/jobs/deploy_backend.yml
```diff
@@ -18,8 +18,17 @@ steps:
       fingerprints:
         - "a1:3f:a7:c3:ff:12:40:1d:85:de:a7:ab:12:3f:cc:05"
   - run:
-      name: "SFTP upload"
+      name: "SFTP upload packages"
```
ztefanie, line 21: Same as above, no need for quotes 
and file should be named "deliver backend"
and "install" should be moved to the top

.circleci/src/workflows/commit.yml
```diff
@@ -5,6 +5,10 @@ jobs:
   - pack_backend:
       requires:
         - build_backend
+  - build_administration
+  - pack_administration:
+      requires:
+        - build_administration
```
ztefanie, line 11: Why would we want to do this on every commit?

.circleci/src/workflows/deliver_beta_administration.yml
```diff
@@ -0,0 +1,15 @@
...
+  - build_administration
+  - pack_administration:
+      requires:
+        - build_administration
+  - deploy_administration:
+      name: deploy_staging
+      server: entitlementcard-test.tuerantuer.org
+      ssh-host-fingerprint: '|1|dkYQrdGB1QML0o+POL3QzAkBbek=|b4Tm0Ymh82UKyZPJfVKy4t+MFV8= ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIONuARu18Fktz+j4QosI+nRqgMnFOMgE7OZLuTOwgZ0k'
+      context:
+        - entitlementcard-apt
```
ztefanie, line 12: I think it would be more clear to put this env var into our `credentials-entitlementcard` context instead of creating a new context.

f1sh1918: this context existed before so i just added this var

.circleci/src/workflows/commit.yml
f1sh1918: i mean before we had build backend that included administration, so now i just readded it as a single step. Why shouldn't we just test if the webapp builds successfully?

.circleci/src/jobs/deploy_administration.yml
f1sh1918: hm i understand the point of readability but it requires way more code to implement the different conditions. not sure if thats an improvement in that case

.circleci/src/workflows/deliver_beta_administration.yml
f1sh1918: moved env var to proposed context

.circleci/src/jobs/deploy_administration.yml
f1sh1918: okay maybe then i  don't know how to proper implement this in circle ci
I thought i have to pass f.e. a boolean with `production=true`
Then i use a `when` condition and use there the particular values for production and the else condition for beta.
So if we take `deliver_administration` job as an example, i have to do the run steps `Upload package` and `install package` in both cases with different values and duplicate the lines of code

.circleci/src/workflows/commit.yml
ztefanie: Building is a step that only rarely does not succeed, but cost a lot of credits. So maybe the amout of credits we use for doing this is not worth the outcome.

michael-markl: `build_administration` does not only "build" it, but also runs linter, tests, etc. So this is likely to fail quickly. We could think of extracting the actual webpack build, but I think that the webpack build does not cost toooo much either.

ztefanie: extracting seems like a good option. 

Even if it is not "tooo much" costs, doing it unnecessarily for every commit sums up.

f1sh1918: should i also extract the gradle build for backend @ztefanie ?

ztefanie: `check-backend-health` has a success rate of 99% in the last 90 days. So I think it would be fine, to only run this check on commits on main or even only on delivery.

f1sh1918: alright. i created a `commit_main` where i do all the build jobs (just for bayern on frontend should be fine?)
https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/4043/workflows/0002504a-3868-4ba1-abac-70945adc5099 (~11min)
And in the normal `commit` are just checks
https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/4040/workflows/13bbb8fb-48fb-42fa-ba80-3b62e30f7b2b (~3min)

is that fine for you @ztefanie ?

.circleci/src/@common.yml
```diff
@@ -8,7 +8,19 @@ parameters:
   run_commit:
     default: true
     type: boolean
-  run_backend:
+  run_commit_main:
```
f1sh1918, line 11: @ztefanie do i have to register this somewhere to be executed on commit on main?
I think you've showed sth in your presentation but couldn't find it

.circleci/src/jobs/build_administration.yml
```diff
@@ -7,31 +7,22 @@ steps:
...
       keys:
-        - v1-node-modules-{{ checksum "package.json" }}-{{checksum "package-lock.json" }}
+        - v2-node-modules-{{ checksum "package.json" }}-{{checksum "package-lock.json" }}
   - install_dart_linux
   - install_protobuf_linux
   - run:
-      name: Install node dependencies
-      command: "[ ! -d node_modules ] && npm ci --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache."
+      name: Install node dependencies on root level
+      command: "cd .. && [ ! -d node_modules ] && npm ci --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache."
```
f1sh1918, line 15: needed for npm workspaces

.circleci/src/@common.yml
ztefanie: No it is enough to specify the condition in the beginning of the workflow.

.circleci/src/jobs/check_administration.yml
```diff
@@ -0,0 +1,31 @@
+docker:
+  - image: cimg/node:19.1.0
+environment:
+working_directory: ~/project/administration
+steps:
+  - checkout:
+      path: ~/project
+  - restore_cache:
+      keys:
+        - v2-node-modules-{{ checksum "package.json" }}-{{checksum "package-lock.json" }}
```
ztefanie, line 10: I think it would be nice, if the cache key also includes  `administration`-string, so we know this keys belong to the node_modules in the adminsitration directory.

.circleci/src/jobs/check_backend.yml
```diff
@@ -0,0 +1,32 @@
...
+      name: Test
+      command: |
+        ./gradlew test
+  - save_cache:
+      paths:
+        - .gradle/wrapper
+      key: v2-gradle-wrapper-{{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+  - save_cache:
+      paths:
+        - .gradle/caches
```
ztefanie, line 31: These two paths (`.gradle/wrapper` and `.gradle/caches`)are not set correctly. You have set the working direcotry for the whole job to `~project/backend` and the path for cache path is relative to the working_directory of your job. But these .gradle paths are in the home directory

.circleci/src/jobs/deliver_administration.yml
```diff
@@ -0,0 +1,44 @@
...
+    description: Whether builds are delivered to production or beta.
+    type: boolean
+steps:
+  - run:
+      name: Install Curl
+      command: |
+        sudo apt update
+        sudo apt install curl -y
+  - attach_workspace:
+      at: /tmp/workspace
```
ztefanie, line 14: This is the wrong directory for the workspace. we use `~/attached_workspace` for it. Just use the`prepare_workspace` command and we can always be sure to use the correct folder.

.circleci/src/jobs/deliver_administration.yml
```diff
@@ -0,0 +1,44 @@
...
+  - add_ssh_keys:
+      fingerprints:
+        - "a1:3f:a7:c3:ff:12:40:1d:85:de:a7:ab:12:3f:cc:05"
```
ztefanie, line 17: No need for quotes.

.circleci/src/jobs/deliver_administration.yml
```diff
@@ -0,0 +1,44 @@
...
+#              https://webhook.entitlementcard.tuerantuer.org/hooks/install-admin-$EAK_WEBHOOK
+  - unless:
+      condition: << parameters.production >>
+      steps:
+        - run:
+            name: SFTP upload package
+            command: |
+              echo "|1|dkYQrdGB1QML0o+POL3QzAkBbek=|b4Tm0Ymh82UKyZPJfVKy4t+MFV8= ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIONuARu18Fktz+j4QosI+nRqgMnFOMgE7OZLuTOwgZ0k" >> known_hosts
+              echo "Uploading: " /tmp/workspace/debs/administration/*.deb
+              sftp -b - -o UserKnownHostsFile=known_hosts ci@entitlementcard-test.tuerantuer.org:/local-apt-repository/ \<<< "put -r /tmp/workspace/debs/administration/*.deb"
```
ztefanie, line 40: If we write to the correct / default known host file, which is `~/. ssh/known_hosts` we can get rid of manually specifying the `UserKnownHostsFile`

In integreat, we just disable the host key checking. Why have you decided to add the checks here?

.circleci/src/jobs/deliver_backend.yml
```diff
@@ -0,0 +1,44 @@
...
+  production:
+    description: Whether builds are delivered to production or beta.
+    type: boolean
+steps:
+  - run:
+      name: Install Curl
+      command: |
+        sudo apt update
+        sudo apt install curl -y
+  - attach_workspace:
```
ztefanie, line 13: Same as above, please use the prepare_workspace command

.circleci/src/jobs/deliver_backend.yml
```diff
@@ -0,0 +1,44 @@
+docker:
```
ztefanie, line 1: The `deliver_backend` and `deliver_administraion` jobs are quite a lot of duplication. Especailly the duplication of the ssh keys is ugly.
We can create a job `deliver_to_server` that has a parameter, if we want to deploy administration or backed.

.circleci/src/jobs/pack_administration.yml
```diff
@@ -6,13 +6,15 @@ steps:
       path: ~/project
   - attach_workspace:
```
ztefanie, line 7: Please remove

.circleci/src/jobs/pack_administration.yml
```diff
@@ -6,13 +6,15 @@ steps:
...
+  - run: ~/project/scripts/pack_deb.sh -v "${NEW_VERSION_NAME}" -f /tmp/workspace/administration/build -d "Administration backend for the Ehrenamtskarte app" -n "eak-administration"
   - run: |
-      mkdir -p /tmp/artifacts/debs
-      mv *.deb /tmp/artifacts/debs
+      mkdir -p /tmp/artifacts/debs/administration
+      mv *.deb /tmp/artifacts/debs/administration
   - store_artifacts:
       path: /tmp/artifacts
   - persist_to_workspace:
       root: /tmp/artifacts
```
ztefanie, line 18: please use  `~/attached_workspace`

.circleci/src/jobs/pack_backend.yml
```diff
@@ -6,13 +6,15 @@ steps:
...
+  - run: ~/project/scripts/pack_deb.sh -v "${NEW_VERSION_NAME}" -t /tmp/workspace/backend/build/distributions/*.tar -s ~/project/scripts/eak-backend.service -d "Backend server for the Ehrenamtskarte app" -n "eak-backend" -c "openjdk-17-jre-headless"
   - run: |
-      mkdir -p /tmp/artifacts/debs
-      mv *.deb /tmp/artifacts/debs
+      mkdir -p /tmp/artifacts/debs/backend
+      mv *.deb /tmp/artifacts/debs/backend
   - store_artifacts:
       path: /tmp/artifacts
   - persist_to_workspace:
       root: /tmp/artifacts
```
ztefanie, line 18: Same please use ` ~/attached_workspace/`

.circleci/src/workflows/commit_main.yml
```diff
@@ -0,0 +1,58 @@
+when:
+  and:
+    - << pipeline.parameters.run_commit_main >>
+    - equal: [main, << pipeline.git.branch >>]
+jobs:
+  - bump_version:
+        prepare_delivery: false
+        context:
```
ztefanie, line 8: no need to pass the deliverino context, when prepare_delivery is false.

.circleci/src/workflows/deliver_beta_administration.yml
```diff
@@ -0,0 +1,22 @@
+when: << pipeline.parameters.run_deliver_beta_administration >>
+jobs:
+  - bump_version:
+        prepare_delivery: false
+        context:
```
ztefanie, line 5: prepare_delivery should be true here, shouldn't it?

.circleci/src/workflows/deliver_beta_backend.yml
```diff
@@ -0,0 +1,35 @@
+when: << pipeline.parameters.run_deliver_beta_backend >>
+jobs:
+  - bump_version:
+      prepare_delivery: false
```
ztefanie, line 4: Should be true?

.circleci/src/workflows/deliver_beta_backend.yml
```diff
@@ -0,0 +1,35 @@
...
+  - pack_meta:
+      requires:
+        - bump_version
+  - deliver_backend:
+      name: deliver_backend_staging
+      production: false
+      context:
+        - credentials-ehrenamtskarte
+      requires:
+        - pack_backend
```
ztefanie, line 31: no need to require pack_backend as this is already required by  `check_health_backend`

.circleci/src/workflows/deliver_production_administration.yml
```diff
@@ -0,0 +1,21 @@
+when: << pipeline.parameters.run_deliver_production_administration >>
+jobs:
+  - bump_version:
+      prepare_delivery: true
+      context:
+        - deliverino
+  - build_administration:
+      requires:
+        - check_administration
```
ztefanie, line 9: check_administration is missing

.circleci/src/workflows/deliver_production_backend.yml
```diff
@@ -0,0 +1,35 @@
...
+      requires:
+        - bump_version
+  - deliver_backend:
+      name: deliver_backend_production
+      production: true
+      context:
+        - credentials-ehrenamtskarte
+      requires:
+        - pack_backend
+        - check_health_backend
```
ztefanie, line 32: Same as above

ztefanie (changes requested):
You have not runned these worksflows yet, do you? Should be tested and working before merging into main.

.circleci/src/jobs/build_administration.yml
f1sh1918: because the post install would lead to an error as you send me the suggestion with ignore scripts? and we execute the post install later manually

.circleci/src/jobs/check_administration.yml
```diff
@@ -0,0 +1,31 @@
+docker:
+  - image: cimg/node:19.1.0
+environment:
+working_directory: ~/project/administration
+steps:
+  - checkout:
+      path: ~/project
```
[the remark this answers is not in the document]
f1sh1918: hm i mean we have different `working_directory` which is fine. but for me all checkout paths are `~/project` in all jobs. afaics

.circleci/src/jobs/build_administration.yml
ztefanie: yes, but you missed the `--ignore-scripts` tag here, when calling the command

f1sh1918: sorry i just didn't see the `not`

.circleci/src/jobs/check_backend.yml
f1sh1918: the working directory is already set to this for the whole job. so it should be sufficient to change the path to
```js
 ~/gradle/wrapper
~/gradle/caches
```

.circleci/src/jobs/deliver_backend.yml
f1sh1918: good point

.circleci/src/workflows/deliver_beta_administration.yml
f1sh1918: the problem here is @ztefanie if we set the this flag for the beta releases the git tags will be created.
If we then trigger the production release, we can not create the same tags again. 
we could add `-beta` to the tags or just create the tag on production

.circleci/src/jobs/deliver_administration.yml
f1sh1918: Hm i don't see the big advantage to write to the servers `known_hosts` tbh. This way is flexible and ensures that its temporary. I think i figured this out with @svenseeberg 
disable host key checking seems to be a red flag, because you can do host spoofing but i'm not an expert. Just read some blogs.

.circleci/src/jobs/check_administration.yml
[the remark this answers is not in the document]
ztefanie: unfortunately not.

.circleci/src/jobs/check_backend.yml
ztefanie: yes, but you need an dot before gradle, it should be `.gradle/wrapper`

.circleci/src/workflows/deliver_beta_administration.yml
ztefanie: Why not do it as we do it for Integreat:
Deploy beta -> Create github tags
Promote beta to production -> Create no tags
Deploy live directly (we rarly do this) -> Create tag

What would the problem be, if we have a github tag, but the version is only available on beta?

.circleci/src/jobs/check_backend.yml
f1sh1918: i adjusted it

.circleci/src/workflows/deliver_beta_administration.yml
f1sh1918: ok then i keep deploying directly live

ztefanie:
> i did and they work but the main issue is that we bump version for each of them. So if you run `deliver_administration_beta` tags gonna be created and if you want to run `deliver_administration_beta` it fails because tags already exist. I had to remove them manually before running second workflow.
> 

What is the problem? Running the same workflow (deliver_administraion_beta) twice?

f1sh1918:
> > i did and they work but the main issue is that we bump version for each of them. So if you run `deliver_administration_beta` tags gonna be created and if you want to run `deliver administration_beta` it fails because tags already exist. I had to remove them manually before running second workflow.
> > 
> 
> What is the problem? Running the same workflow (deliver_administraion_beta) twice?

Oh sry I meant deliver_backend_beta after the administration

ztefanie:
I see. We discussed if we want separate workflows for administration / backend / native or just on workflow to deliver all three at once. Not sure what we decided we want to do.
But of course we cannot trigger two workflows after each other, this is expected. If we run deliver backend and afterwards deliver administration they should have different version as we cannot ensure they are the same version, because it is possible there were commits done between these two workflows. To trigger all three at once we need a separate workflow for this.
In the bump_version job, we have the warning "Delivery was made on branch << pipeline.git.branch >>. Make sure to merge this branch before next delivery."

.circleci/src/jobs/deliver_to_server.yml
```diff
@@ -0,0 +1,47 @@
...
+  - add_ssh_keys:
+      fingerprints:
+        - a1:3f:a7:c3:ff:12:40:1d:85:de:a7:ab:12:3f:cc:05
+  - when:
+      condition: << parameters.production >>
+      steps:
+        - run:
+            name: SFTP upload package
+            command: |
+              echo "|1|iikuvSrIo3wkj+EqUgLRMsAq6yk=|r9bSjkawWFa94b45qE/se5Oio5k= ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIED3MwobbCs+ENMLXdyqlJb3bJ26TuIYt977TA3NrN66" >> known_hosts
```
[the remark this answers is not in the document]
f1sh1918: its a public key?!

f1sh1918:
ok i'm gonna check your issues @sarahsporck 
The administration delivery ran successfully finally :)

.circleci/src/jobs/deliver_to_server.yml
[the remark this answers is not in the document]
f1sh1918: don't think a public ssh fingerprint is a problem but will move it to circle ci context

.circleci/src/jobs/build_administration.yml
```diff
@@ -1,39 +1,30 @@
...
-        - v1-node-modules-{{ checksum "package.json" }}-{{checksum "package-lock.json" }}
+        - v2-node-modules-administration-{{ checksum "package.json" }}-{{checksum "package-lock.json" }}
   - install_dart_linux
   - install_protobuf_linux
+  - prepare_workspace
   - run:
-      name: Install node dependencies
-      command: "[ ! -d node_modules ] && npm ci --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache."
+      name: Install node dependencies on root level
+      command: "cd .. && [ ! -d node_modules ] && npm ci --ignore-scripts --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache."
```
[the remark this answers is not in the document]
f1sh1918: hm all command except of that need the `project/administration` directory so i decided to keep it or can i somehow set the `working_directory` on command level?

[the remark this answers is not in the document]
f1sh1918: i set the directory to the other commands, since the `restore_npm_cache` should work on root level

f1sh1918:
Ok both deliveries ran successful now.
The only thing we have to do is create an `deliver_beta` workflow where we execute frontend, backend, administration jobs to ensure we have everywhere the same version @ztefanie and @sarahsporck

sarahsporck (approved):
Any concrete testing instructions? Visually it seems fine.

f1sh1918:
> Any concrete testing instructions? Visually it seems fine.

@sarahsporck i mean you can just trigger the workflows f.e. `deliver_beta_administration` then a new version will be deployed to staging, but they already ran successfully

PR #1225, opened 2023-11-22
1224: Fix language switch

f1sh1918:
### Short description

App switches the language now completely and restores the chosen language on restart

### Proposed changes

- Save the language locale in settings. 
- Use translation from context
- Restore locale from settings on restart

### Resolved issues

Fixes: #1224 

Tested on android and ios. On android i built an apk and put it on the device to ensure that the app is properly closed and reopened

### DIFF
--- a/frontend/assets/l10n/app_de.json
+++ b/frontend/assets/l10n/app_de.json
@@ -5,10 +5,6 @@
     "disclaimer": "Haftung, Haftungsausschluss und Impressum",
     "infoTitle": "Info",
     "languageChange": "Sprache wechseln",
-    "languages(map)": {
-      "en": "Englisch",
-      "de": "Deutsch"
-    },
     "languageChangeSuccessful": "Ihre Sprache wurde erfolgreich geändert!",
     "licenses": {
       "one": "Lizenz",
--- a/frontend/assets/l10n/app_de.json
+++ b/frontend/assets/l10n/app_de.json
@@ -5,6 +5,10 @@
     "disclaimer": "Haftung, Haftungsausschluss und Impressum",
     "infoTitle": "Info",
     "languageChange": "Sprache wechseln",
+    "languages(map)": {
+      "en": "Englisch",
+      "de": "Deutsch"
+    },
     "languageChangeSuccessful": "Ihre Sprache wurde erfolgreich geändert!",
     "licenses": {
       "one": "Lizenz",
--- a/frontend/ios/Runner/Info.plist
+++ b/frontend/ios/Runner/Info.plist
@@ -28,14 +28,12 @@
 	<false/>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
-	<key>NSCameraUsageDescription</key>
-	<string>Einlesen von QR-Codes und Abfotografieren von Dokumenten für das Bewerbungsformular.</string>
-	<key>NSLocationAlwaysUsageDescription</key>
-	<string>Standort auf der Karte und Akzeptanzstellen in der Umgebung anzeigen.</string>
-	<key>NSLocationWhenInUseUsageDescription</key>
-	<string>Standort auf der Karte und Akzeptanzstellen in der Umgebung anzeigen.</string>
-	<key>NSPhotoLibraryUsageDescription</key>
-	<string>Anhängen von Bildern für das Bewerbungsformular.</string>
+	<key>CFBundleLocalizations</key>
+    <array>
+       <string>en</string>
+       <string>de</string>
+    </array>
+
 	<key>UIApplicationSupportsIndirectInputEvents</key>
 	<true/>
 	<key>UILaunchStoryboardName</key>
--- a/frontend/lib/about/language_change.dart
+++ b/frontend/lib/about/language_change.dart
@@ -4,12 +4,12 @@ import 'package:ehrenamtskarte/l10n/translations.g.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
+Map<String, String> nativeLanguageNames = {'en': 'English', 'de': 'Deutsch'};
+
 class LanguageChange extends StatelessWidget {
   const LanguageChange({super.key});
-
   @override
   Widget build(BuildContext context) {
-    final t = context.t;
     return Column(children: [
       ...buildConfig.appLocales.map((item) => DecoratedBox(
           decoration: BoxDecoration(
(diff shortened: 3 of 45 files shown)


f1sh1918:
@michael-markl it looks like the `TranslationProvider` does not affect on Pages that are loaded with the `BottomActionBar`. Not sure if there is a way to provide that information to all pages that will be opened with it and are not directly surrounded by the provider. Not sure if the best way is to just use this provider on every page where its needed (like i did on `SearchPage`)
But the other question is what happens with Pages that were already opened and are still in the `Tab Navigation`. They also have to be updated once the Locale changes. So i decided for this solution

michael-markl:
While I can't really say, it looks a bit like a hacky workaround for a shortcoming of the i10n lib to me.
Probably @steffenkleinle can review better here.

michael-markl:
To me, it seems it would be better to have the AppLocale in the context and have every translated widget subscribe to the context, so they are updated automatically if the locale changes?

f1sh1918:
> While I can't really say, it looks a bit like a hacky workaround for a shortcoming of the i10n lib to me. Probably @steffenkleinle can review better here.

It seems like there are additional l10n flutter libs that maybe can achieve that, but they don't seem to be very well supported and tbh i don't want to use some deprecated community lib (as we have lots of problems in react native)

michael-markl:
I think, slang also offers this solution with some more effort maybe, see "Getting Started" -> Step 4c.
https://github.com/slang-i18n/slang

f1sh1918:
> I think, slang also offers this solution with some more effort maybe, see "Getting Started" -> Step 4c. https://github.com/slang-i18n/slang

after 4b) comes 5?

f1sh1918:
> I think, slang also offers this solution with some more effort maybe, see "Getting Started" -> Step 4c. https://github.com/slang-i18n/slang

you mean dependency injection?

michael-markl:
Ah, sorry, I think slang does have the functionality built, see https://github.com/slang-i18n/slang#-changing-locale
They state:
> Widgets rebuild only if you use final t = Translations.of(context) or context.t.

Hence, I would propose to use one of the two in every occurence of `t` (Personally I find it weird, that they define a global `t`).

frontend/lib/about/language_change.dart
```diff
@@ -1,6 +1,9 @@
 import 'package:ehrenamtskarte/build_config/build_config.dart' show buildConfig;
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
 import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+
+import '../configuration/settings_model.dart';
 
 Map<String, String> languages = {'en': 'Englisch', 'de': 'Deutsch'};
```
steffenkleinle, line 8: Those should actually also go in the translations, the language names differ in german and english for example.

frontend/lib/app.dart
```diff
@@ -54,6 +54,13 @@ class App extends StatelessWidget {
 
     final String initialRoute = settings.firstStart ? introRouteName : homeRouteName;
 
+    // Load default language from settings
+    WidgetsBinding.instance.addPostFrameCallback((_) {
+      if (settings.language != null) {
+        LocaleSettings.setLocaleRaw(settings.language!);
+      }
+    });
+
```
steffenkleinle, line 63: Can we move this to `main.dart` to have everything language/locale related together and not have logic with possible side effects in two places?

Also, why do we need the `addPostFrameCallback`?

f1sh1918: no because we don't have the context of the `SettingsProvider` here
Because its a lifecycle hook that waits until the widget is rendered to not interfere

steffenkleinle: Thanks for explaining, guessed something like that.

frontend/lib/about/language_change.dart
f1sh1918: i used a  `Map` for this in translations

f1sh1918:
> Ah, sorry, I think slang does have the functionality builtin, see https://github.com/slang-i18n/slang#-changing-locale They state:
> 
> > Widgets rebuild only if you use final t = Translations.of(context) or context.t.
> 
> Hence, I would propose to use one of the two in every occurence of `t` (Personally I find it weird, that they define a global `t` in the first place).

ok i replace that and now everything should work.
One short question @steffenkleinle.
Since i added the missing translation strings for iOS permissions, is there also a way to provide custom permission strings for android? Didn't find anything about that

frontend/ios/Runner/Info.plist
```diff
@@ -28,14 +28,12 @@
 	<false/>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
-	<key>NSCameraUsageDescription</key>
```
f1sh1918, line 31: move to language specific. I think we don't use anything from photo library @michael-markl ? Not sure why this string exists

frontend/assets/l10n/app_de.json
```diff
@@ -5,6 +5,10 @@
     "disclaimer": "Haftung, Haftungsausschluss und Impressum",
     "infoTitle": "Info",
     "languageChange": "Sprache wechseln",
+    "languages(map)": {
+      "en": "Englisch",
+      "de": "Deutsch"
+    },
```
michael-markl, line 11: I think the languages in a language switch menu are usually not translated but only shown in their mother tongue.

frontend/ios/Runner/Info.plist
michael-markl: A few years ago ( :D ) we had some application form implemented in the native app. So, that's probably a left over of that.

michael-markl (commented):
LGTM. I did not test anything.

frontend/assets/l10n/app_de.json
f1sh1918: ok so you think we should use the `deviceLocale` ?

michael-markl: I propose to always show each language in its own language :D
So "Deutsch", "English", (... "Français", ...) etc, just like we do it in the integreat app.

f1sh1918: Indeed and therefore no translations are needed

steffenkleinle: My main issue with the previous solution was that it was `Englisch`. Completely fine with using own language instead of translations.

steffenkleinle (approved):
Tested on android, works like a charm! :tada:

Issue #1228, opened 2023-11-27
Wrong column headline activity log


f1sh1918:
**Describe the bug**
As a user i want to have the correct information about what informaton the column contains in the activity log

**To Reproduce**
1. Create a card for sozialpass
2. Go to Benutzereinstellungen
3. Open activity log
4. Check column names

**Expected behavior**
Pass-Id is used

**Screenshots**
<img width="910" alt="image" src="https://github.com/digitalfabrik/entitlementcard/assets/37902063/c31be813-efa1-4a7f-9c21-b89357d048dd">

PR #1231, opened 2023-12-04
1228: Rename nuernberg pass id in activity log

ztefanie:
### Short description

Renamed the nuernberg pass id in the activity log

### Resolved issues

Fixes: #1228


f1sh1918 (approved):
not test but looks legit to me

PR #1236, opened 2023-12-05
1235 Replace stores endpoint bavaria

f1sh1918:
### Short description

Bavaria now provides the store information xml to a new url

### Proposed changes

- switch url in backend config and docs

Fixes: #1235


michael-markl:
We should not forget to adjust the salt config too

sarahsporck (approved):
Tested locally.

dkehne:
they re-adjusted the url once more (final one): https://www.ehrenamt.bayern.de/xml-json/app-daten.xml

PR #1240, opened 2023-12-11
1235: Replace store endpoint bavaria

f1sh1918:
Update to new endpoint url for stores


ztefanie (approved):
lgtm, not tested the backend, but the link works in the browser

PR #1243, opened 2023-12-14
1242: DeepLink Web setup

f1sh1918:
### Short description

To use https deep links, we have to create a trust relation between our webserver and the apps. 

### Proposed changes
- create deepLink configs per project
- according to this configs create `apple -app-site-association` and `assetlinks.json` for each project
- since we have only one webserver instance for the druckerei, there have to be some redirects done f.e.:
`bayern.ehrenamtskarte.app/.well-known/assetslinks.json -> /bayern/assetlinks.json` to point to the particular files depending on the domain
- use `frontend/buildConfigs` as a file dependency since we need some app specific information

### Resolved issues

Fixes: #


frontend/build-configs/bayern/index.ts
```diff
@@ -3,6 +3,7 @@ import publisherText from "./publisherText"
 import disclaimerText from "./disclaimerText"
 
 export const bayernCommon: CommonBuildConfigType = {
+    activationPath: '/activation',
```
sarahsporck, line 6: What's this used for? When is it not '/activation'?

administration/config/webpack.config.ts
```diff
@@ -50,13 +52,52 @@ const imageInlineSizeLimit = parseInt(process.env.IMAGE_INLINE_SIZE_LIMIT || '10
 const cssRegex = /\.css$/
 const cssModuleRegex = /\.module\.css$/
 
+// https://developer.android.com/training/app-links/verify-site-associations#manual-verification
```
sarahsporck, line 55: this url does not work for me

frontend/build-configs/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "build-configs",
+  "version": "0.0.1",
+  "private": true,
+  "main": "dist/index.js",
+  "scripts": {
+    "preinstall": "tsc --build"
+  }
```
sarahsporck, line 8:
```suggestion
  "main": "index.ts",
```

frontend/build-configs/bayern/index.ts
f1sh1918: we can use this path for the router in the native frontend (modal will open to activate card). Furthermore it is used for the ios universal links to whitelist certain paths that open the app (apple-app-site-association)

frontend/build-configs/package.json
f1sh1918: this does not work for me @sarahsporck 
When i do an `npm i && npm run build` in `administration`, i'm facing the following error

```js
/Users/afischer/projects/entitlementcard/administration/node_modules/build-configs/index.ts:1
export {default as buildConfigNuernberg} from './nuernberg'
^^^^^^

SyntaxError: Unexpected token 'export'
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1149:20)
    at Module._compile (node:internal/modules/cjs/loader:1190:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1280:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/afischer/projects/entitlementcard/administration/node_modules/ts-node/src/index.ts:1608:43)
    at Module.load (node:internal/modules/cjs/loader:1089:32)
    at Function.Module._load (node:internal/modules/cjs/loader:930:12)
    at Module.require (node:internal/modules/cjs/loader:1113:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/Users/afischer/projects/entitlementcard/administration/src/project-configs/bayern/deepLinkConfig.ts:1:1)

```

sarahsporck: I'll review again tomorrow. But you might be right. Still I'm wondering why it is working for integreate. Well, we'll see tomorrow :)

f1sh1918: yes i also checked the integreat implementation and i tried for some time to figure out why this is working there and not here

sarahsporck: It seems the problem lies in cra not supporting [project references](https://github.com/facebook/create-react-app/issues/10695), and thus our build script (which originally was a cra script) does not support project references https://github.com/facebook/create-react-app/issues/10695. An easy workaround for now is to run tsc --build before run. I'll create an PR for you to visualize the changes. The current solution requires typescript to be installed globally and is therefore not sufficent in my opinion. However I think we should consider moving to a simpler webpack setup (as @michael-markl already [suggested in an older pr](https://github.com/digitalfabrik/entitlementcard/pull/1026#issue-1773363300)) at some point. Moving to ts-loader could also resolve this issue.

ztefanie (changes requested):
Why do we need the `apple-app-site-association`?
According to the docs Associated Domains Entitlement is needed for `shared web credentials, universal links, and App Clips.`. As we do not have a webpage that can display the app content, we do not support universal links. Custom URL Schemes should be enough to realize our goal, to link for the web/administration to the native app. I do not see the need for Associated Domains Entitlement, what is the use case?

For Android this makes sense, as this makes possible to redirect the user without requiring the user to select the app / confirm it should be opened by the app.

ztefanie (commented):
So we will have the link to the website and on the website we redirect to the app, if possible / installed?

The drawback i thought of is adding unnecessary complexity, buy using Associated Domains Entitlement, which is more complex than just a custom scheme.

ztefanie (changes requested):
For integreat, we just delete everything on the server and push the new dist/ folder to it. Aret here now files on the server that should not be overwritten? How can we prevent overwriting these files on the server when auto deliver or serve them every time we deliver a new verison? If there is anything we should take care of, please add a comment to the corresponding CICD issue

administration/tsconfig.json
```diff
@@ -26,7 +26,9 @@
     "src/**/*.d.ts",
     "src/graphql/customScalarTypes.d.ts"
   ],
-  "ts-node": {
```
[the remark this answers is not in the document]
f1sh1918: i think @sarahsporck removed it in her proposal. Maybe she can tell

administration/src/project-configs/bayern/deepLinkConfig.ts
```diff
@@ -0,0 +1,17 @@
+import { APP_STORE_TEAM_ID, buildConfigBayern } from 'build-configs'
```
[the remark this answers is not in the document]
f1sh1918: i think i didn't get your idea how to accomplish this.
How should i make two config files with different content generic?
I mean i could add the content of these configs to `frontend/build-configs` and then pass them.

administration/tsconfig.json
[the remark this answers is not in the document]
sarahsporck: This is neither required nor is it hurts to keep it. I removed it, while experimenting on the typescript setup. Currently we don't invoke ts-node anywhere, so we could even remove the dependency (maybe?). @f1sh1918 you can readd this :)

administration/config/webpack.config.ts
```diff
@@ -50,13 +52,52 @@ const imageInlineSizeLimit = parseInt(process.env.IMAGE_INLINE_SIZE_LIMIT || '10
...
+          sha256_cert_fingerprints: [config.android.sha256CertFingerprint],
+        },
+      },
+    ],
+    null,
+    2
+  )
+}
+
+const generateAppleAppSiteAssociation = (config: DeeplLinkingConfig) => {
```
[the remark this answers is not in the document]
f1sh1918: should be fine with last commit

administration/src/project-configs/bayern/deepLinkConfig.ts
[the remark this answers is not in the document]
f1sh1918: i'm not sure if i got the whole picture of how to accomplish what you want.

I mean we need a different config for each project because the content is different for every whitelabel. But I moved this config now to `fronted/buildConfigs`
Not sure if it makes sense to pass the particular `buildConfig` in webpack, since we only have one build not like integreat. So i execute the `getDeepLinkingConfigs()` where i have listed the particular build-configs in an array for all build-configs where we need the association files. Feel free to add a proposal or an example of what you would recommend if that is not fine for you, @ztefanie

frontend/build-configs/.npmignore
```diff
@@ -0,0 +1,3 @@
+# Files will be ignore while package installation
+/.prettierrc
+/.package-lock.json
```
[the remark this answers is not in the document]
ztefanie: I think you somehow missed to commit this? File is still here

ztefanie (commented):
As you just mentioned, there is only one deploy for both apps, i thought about this, and I think in the end there only should be only one `apple-app-site-association` file, which contains both appIDs. Same for assetlink.json.

Or will this be handled by subdomains of the server?

f1sh1918:
> As you just mentioned, there is only one deploy for both apps, i thought about this, and I think in the end there only should be only one `apple-app-site-association` file, which contains both appIDs. Same for assetlink.json.
> 
> Or will this be handled by subdomains of the server?

This will be handled by the server. We should not have this in one file to avoid people having issues when both apps are installed. Now the subdomains redirect on the particular project folder in the build folder

f1sh1918:
> For integreat, we just delete everything on the server and push the new dist/ folder to it. Aret here now files on the server that should not be overwritten? How can we prevent overwriting these files on the server when auto deliver or serve them every time we deliver a new verison? If there is anything we should take care of, please add a comment to the corresponding CICD issue

i don't think we have to do here anything @ztefanie the folder will be deleted and redeployed on update. All files that are needed will be recreated

frontend/build-configs/.npmignore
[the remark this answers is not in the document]
f1sh1918: hm i can not find the `npmignore` in the changed files anymore @ztefanie

ztefanie (approved):
Looks good :tada: 

I have not tested. If this can be tested locally and I should try, feel free to give me some short instructions and I will test it.

ztefanie:
oke yes i checked the creation of the artifacts previously, so should be fine :+1:

PR #1249, opened 2024-01-10
1242: Remove typechecking during build

sarahsporck:
Build app before start. Check out the webpack plugin [here](ForkTsCheckerWebpackPlugin)

### DIFF
--- a/administration/config/env.ts
+++ b/administration/config/env.ts
@@ -22,7 +22,7 @@ function getClientEnvironment(publicUrl: string) {
     NODE_ENV !== 'test' && `${paths.dotenv}.local`,
     `${paths.dotenv}.${NODE_ENV}`,
     paths.dotenv,
-  ].filter(Boolean)
+  ].filter(Boolean) as string[]
 
   // Load environment variables from .env* files. Suppress warnings using silent
   // if this file is missing. dotenv will never modify any environment variables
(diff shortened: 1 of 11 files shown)


administration/config/env.ts
```diff
@@ -22,7 +22,7 @@ function getClientEnvironment(publicUrl: string) {
     NODE_ENV !== 'test' && `${paths.dotenv}.local`,
     `${paths.dotenv}.${NODE_ENV}`,
     paths.dotenv,
-  ].filter(Boolean)
+  ].filter(Boolean) as string[]
```
sarahsporck, line 25: The typings were messed up here. Enforcing with as string[] fixes this.

administration/config/webpack.config.ts
```diff
@@ -518,6 +518,7 @@ function createWebpackConfig(webpackEnv: 'development' | 'production'): webpack.
             syntactic: true,
           },
           mode: 'write-references',
+          build: true,
```
michael-markl, line 521: Do you know what this does? I am not entirely sure :D

I am wondering:
Maybe we want to remove TS checking completely from the webpack build and directly transpile to JS to improve build times?
Type checking happens anyway in the language server in the IDE as well as on CI? Similarly, we don't check eslint / stylelint when building (one could see TS as a linter on steroids :D )

sarahsporck: > Do you know what this does? I am not entirely sure :D

This runs tsc --build before typechecking, which will generate the missing d.ts files.

But I think you're right. I added a typescript run to the CI, as it was missing and removed typechecking during build. @f1sh1918 it should work now even without the symlink. Would you mind giving it a try? :)

Issue #1210, opened 2023-11-15
Add hint to contact person of organization


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Currently some applicants add their own contact data to the contact person of the organization.

**Describe the solution you'd like**
- add a hint: "Bitte geben Sie hier die Daten der Person an, die ihr ehrenamtliches Engagement bestätigen kann"
- rename : "Kontaktperson der Organisation" to "Kontaktperson in der Organisation".
- rename the placeholder "Vor- und Nachname" to "Vor- und Nachname der Kontaktperson in der Organisation" (i don't think this is applicable) 

**Describe alternatives you've considered**
- Maybe we could also check if that's easily doable if the applicant name and surname are equal to the organization persons name and surname and show a hint (probably that needs additional effort)

PR #1252, opened 2024-01-17
1210: Add hint to contact person of organization

sarahsporck:
### Short description

Currently it is possible to set yourself as contact person in an application. The user should be informed that this will very likely not lead to a successful application.

### Proposed changes

- added hint about who this person is supposed to be
- added hint when applicant names are identical to the contact persons name

### Resolved issues

Fixes: #1210


ztefanie (approved):
Works as expectes :+1:

f1sh1918 (approved):
Nicely done.
Just an optional thing i personally prefer is to add some margin to the alert container
`margin: 8px 0` -->

image

PR #1253, opened 2024-01-17
WIP: add protobuf generation to the backend

sarahsporck:
### Short description

This adds protobuf generation to the backend as a preparation for the koblenz pass.
Sadly I couldn't get it work with the local protoc instance and had to install the protoc artifact. For now I'll keep this as a draft, so you can follow the progress.

### Proposed changes

- add protobuf plugin
- compile kotlin files during build


sarahsporck:
This is already done in #1257

PR #1255, opened 2024-01-18
Fix cache path for fvm

sarahsporck:
### Short description

The whole `check_frontend` job was running from working directory `frontend`. While the `build_android` and `build_ios` were not. This lead to the result that the cached file at `frontend/.fvm

### Proposed changes

- remove working directory from `check_frontend` job
- add working directory to the commands instead


ztefanie (approved):
Thanks for fixing this

Issue #1164, opened 2023-10-12
Handle increasing version number automatically


ztefanie:
Version numbers of frontend (`pubspec.yaml`) and administration (`package.json`) should be handled automatically when triggering releases.
Use the version calc command of the app-toolbelt

PR #1256, opened 2024-01-19
1164: Handle versioning

ztefanie:
### Short description
Versioning is automatically handled.

### Proposed changes
- Added entitlementcards to deliverino repos
- Adjusted toolbelt to allow version commits on all branches, see PR here: 
https://github.com/digitalfabrik/app-toolbelt/pull/2
- adjusted bump_verison step
- verison file should be used for frontend and administration

### Resolved issues

Fixes: #1164

### DIFF
--- a/.circleci/src/jobs/bump_version.yml
+++ b/.circleci/src/jobs/bump_version.yml
@@ -18,4 +18,21 @@ steps:
   - run:
       name: Calculate next version code
       command: echo "export NEW_VERSION_CODE=$(app-toolbelt v0 version calc | jq .versionCode)" >> ${BASH_ENV}
+  - when:
+      condition: << parameters.prepare_delivery >>
+      steps:
+        - run:
+            name: Bump git version
+            command: app-toolbelt v0 release bump-to ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --branch ${CIRCLE_BRANCH}
+# TODO 1199
+#  - when:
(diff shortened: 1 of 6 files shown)


version.json
```diff
@@ -1 +1 @@
-{ "versionName": "2024.1.3", "versionCode": 130 }
+{"versionName":"2024.1.4","versionCode":131}
```
sarahsporck, line 1: why did the formatting change here?

.circleci/src/jobs/bump_version.yml
```diff
@@ -18,4 +18,21 @@ steps:
...
+        - run:
+            name: Bump git version
+            command: app-toolbelt v0 release bump-to ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --branch ${CIRCLE_BRANCH}
+# TODO 1199
+#  - when:
+#      condition:
+#        and:
+#          - << parameters.prepare_delivery >>
+#          - not:
+#              equal: [main, << pipeline.git.branch >>]
```
sarahsporck, line 33: why not equal?

version.json
ztefanie: This was the auto commit of deliverino. Before was the manuall commit, where some whitespace were added by autoformat. We should stick with this new format deliverino uses.

.circleci/src/jobs/bump_version.yml
ztefanie: This is the condition for the step afterwards, which notifies us on mattermost, we need to merge this branch before a new release, because otherwise the next release will fail due to the same version numbers.

sarahsporck (approved):

PR #1257, opened 2024-01-19
Generate codes in backend

sarahsporck:
### Short description

This PR moves the creation of the Codes (encoding, hashing, storing) to the backend.  

### Open Questions

- PDFs are now created *after* the Code is generated. That means if there is an error during PDF generation the card is basically "dead" and can never be activated (because there is no QR-Code. Maybe we try delete the card in case of a PDF creation error?
- We cannot check the size (information size in bytes) of the generated code, which means that technically a Code can contain too much information to fit into our QR-Code scheme. I guess moving this to the backend and creating a custom error message for this case should be sufficient though, as it never came up until now anyway.
- Where do we want to set the expiration day? Client or Server?

### TODO:
- [x] tracking with Matomo
- [x] check for qr code size in backend at least

### Proposed changes

- Basically move generateActivationCodes and generateStaticVerificationCode from `CardBlueprint` to the backend
- Tests are mostly the same as in `administration`. 
- Doing the conversion Message -> CanonicalJson directly would be a lot simpler regarding logic and also has a better performance, but would be much harder to test. Therefore, I kept the original logic Message -> Object -> CanonicalJson


administration/src/cards/createCards.ts
```diff
@@ -22,37 +16,34 @@ export class CreateCardsError extends Error {
...
   }
+
+  return result.data.cards.reduce<CreateCardsResult>((acc, card) => {
+    const dynamicActivationCode = card.dynamicActivationCodeBase64 ? [DynamicActivationCode.fromBinary(base64ToUint8Array(card.dynamicActivationCodeBase64))] : []
+    const staticVerificationCode = card.staticVerificationCodeBase64 ? [StaticVerificationCode.fromBinary(base64ToUint8Array(card.staticVerificationCodeBase64))] : []
+    return {
+      dynamicActivationCodes: [...acc.dynamicActivationCodes, ...dynamicActivationCode],
+      staticVerificationCodes: [...acc.staticVerificationCodes, ...staticVerificationCode]
+    }
+  }, { dynamicActivationCodes: [], staticVerificationCodes: [] })
```
sarahsporck, line 46: We could define a much better type for this. But I think it makes sense to discuss this shortly in the team.

backend/build.gradle.kts
```diff
@@ -25,6 +25,8 @@ repositories {
 
 dependencies {
     implementation("com.google.protobuf:protobuf-kotlin:3.25.2")
+    // can probably be removed when upgrading protoc
+    // https://github.com/protocolbuffers/protobuf/issues/12056
```
sarahsporck, line 29: After updating protobuf I didn't need the additional dependencies. But as updating protoc is not in scope of this I'll readd the dependency again.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
```diff
@@ -0,0 +1,73 @@
...
+                    return o.joinToString(",", "[", "]") {
+                        serializeToString(it)
+                    }
+                }
+
+                else -> throw Error("Invalid argument of type ${o::class.simpleName} passed to serializeToCanonicalJson.")
+            }
+        }
+
+        fun serialize(message: GeneratedMessageV3): String {
```
sarahsporck, line 34: Note to myself: Remove this function again.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/types/CardCreationResultModel.kt
```diff
@@ -0,0 +1,6 @@
+package app.ehrenamtskarte.backend.verification.webservice.schema.types
+
+data class CardCreationResultModel constructor(
+    val dynamicActivationCodeBase64: String?,
+    val staticVerificationCodeBase64: String?,
+)
```
sarahsporck, line 6: This can be discussed again. In my implementation I return `CardCreatioResultModel[]`. But maybe sth like the following makes more sense
```kotlin
data class CardCreationResultModel constructor(
  val dynamicActivationCodesBase64: List<String>
  val staticVerificationCodeBase64: List<String>
)
```

sarahsporck (commented):
The pink fields pretty much show what has moved to the backend.
I removed hashing logic completely from the administration frontend, as it makes more sense to do everything in the backend (to avoid managing the "same" code twice).
The backend tests are basically the same as they were in administration.

create_cards_lifecycle drawio

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
```diff
@@ -0,0 +1,92 @@
+package app.ehrenamtskarte.backend.verification
+
+import com.google.protobuf.Descriptors
+import com.google.protobuf.Descriptors.FieldDescriptor.Type
+import com.google.protobuf.GeneratedMessageV3
+
+class CanonicalJson {
```
sarahsporck, line 7: Equivalent of implementation of `hashCardInfo.ts` in administration

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/QRCodeUtil.kt
```diff
@@ -0,0 +1,61 @@
+package app.ehrenamtskarte.backend.verification.webservice
+
+import Card
+import com.google.zxing.common.BitArray
+import com.google.zxing.qrcode.decoder.ErrorCorrectionLevel
+import com.google.zxing.qrcode.decoder.Mode
+import com.google.zxing.qrcode.decoder.Version
+
+class QRCodeUtil {
```
sarahsporck, line 9:
```suggestion
// implementation mostly taken from -> https://github.com/zxing-js/library/blob/master/src/core/qrcode/encoder/Encoder.ts
class QRCodeUtil {
```

backend/src/test/kotlin/app/ehrenamtskarte/backend/helper/ExampleCards.kt
sarahsporck, line 1: Helper file for tests to avoid copy pasting too much.
Usage: `ExampleCards.get(CardType.BavarianStandard)` to get a Bavarian example card.

michael-markl:
Nice stuff!
Probably you discussed the design already. My two cents: I am personally somewhat in favor of duplicating the hashing & activation-code-creation logic in the administration and the backend as I think it is quite a big selling point that customers can use the digitale Druckerei without ever sending personal information to us.
With this change, data privacy declarations might need change, so we probably cannot deploy this yet anyway(?).
(Also, I think the duplicated logic would be relatively well contained, well testable, and not subject to change (as all the customer-related stuff is inside the cardinfo, and we don't have to touch the hashing function when changing the cardinfo))

ztefanie:
Can you please add this svg to the docs? Or what would be a good place to put it?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
```diff
@@ -0,0 +1,92 @@
...
+    companion object {
+        /**
+         * Returns a canonical JSON object from the given protobuf message.
```
ztefanie, line 10: This comment just shows how this function, does break the single responsibility principle. If we want to change this to cleaner code (which i really prefer), it would be good to:
1) create a function `assertMessageIsValid` 
2) create three functions, which each of the asserts one of the three listed asserations and has a meaningful name, so the name of the functions tells us the same information as the content of the corresponding comment bullet point
3) Create a function, which adds the entries to the object
4) Only call the two functions described in 1) and 3) in the `messageToMap` function.
5) Get rid of the comment, but the last line, as our code is now self explaining.

With this canges, we have code that is self explaining, we do not have the duplication of code and comment, which can lead to inconsistency in the future, we follow the single responsibility principle and the statements in every function are at the same level of abstraction.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/QRCodeUtil.kt
```diff
@@ -0,0 +1,61 @@
...
+    companion object {
+        // Level 8 with EC of M gives 152 bytes
+        // Level 7 with EC of L gives 154 bytes
+        // Level 8 seems appropriate right now.
```
ztefanie, line 13: What are these levels?

ztefanie (commented):
I tested, when creating a qr code and try to activate it, i get an "no connection" error for bayern. I did run the generate protobuf and graphql commands, if they were needed.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
ztefanie: If it is easy to read, why dis there a comment explaining what is going on?
In my opinion the function solves two responsibilities: Asserting the input is valid and transforming it.
And there are different levels of abstraction in this function.
Function length should be no longer than 20 lines, according to uncle bob and his clean code book, which is quite the gold standard for clean code. This function is more than 40 lines.
For helping the user not to get lost in a lot of functions, we easily can use private functions.
(Just wanted to explain this is not a personal opinion thing on my site, but following the clean code principles of robert c. martin.)

I am a big fan of clean coding and highly recommend to follow these principles, but I am also fine if we as a team decide we do not want to follow them.

f1sh1918:
> Nice stuff! Probably you discussed the design already. My two cents: I am personally somewhat in favor of duplicating the hashing & activation-code-creation logic in the administration and the backend as I think it is quite a big selling point that customers can use the digitale Druckerei without ever sending personal information to us. With this change, data privacy declarations might need change, so we probably cannot deploy this yet anyway(?). (Also, I think the duplicated logic would be relatively well contained, well testable, and not subject to change (as all the customer-related stuff is inside the cardinfo, and we don't have to touch the hashing function when changing the cardinfo))

I got your point with data privacy, but it may be also difficult to maintain both ways in parallel if there are changes and keep them sync. Not sure if its worth to do that tbh

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
michael-markl: Clearly, code style is always opinionated. "clean code" principles are not the gold standard in my opinion.

My personal opinion is that dogmatic use of "clean code" principles do not serve the purpose of good readability and maintainability. Clearly it makes sense to not have >100 loc functions (unless in cases where it actually makes sense :D ).
In this very specific case, the actual number of lines that *do* something is maybe 25.

> If it is easy to read, why dis there a comment explaining what is going on?

As this function is central in the stable hashing procedure, it needs to be documented well. Therefore, adding a doc-string for this very specific function that explains the contract of the function (such that it can be replicated in other languages) in English does make sense in my opinion (at least it made sense to me at time of writing). Also adding comments where and how the procedure may be extended to keep the stability property seems reasonable to me. This cannot be documented in code.

> In my opinion the function solves two responsibilities: Asserting the input is valid and transforming it.

We could try checking the preconditions in a separate function. My feeling is that it won't increase readability.

> And there are different levels of abstraction in this function.

What do you mean by "levels of abstraction"?

sarahsporck:
> > but it may be also difficult to maintain both ways in parallel if there are changes and keep them sync.
> 
> I agree that there's some duplication and it would be great to not have it... Note, that we already duplicate the hashing logic in Dart, so that would a tripling of it :D Fortunately, these places are not subject to change in the near future (that might be a reason to allow it).
> 
> Unfortunately, code sharing between kotlin&JS&dart is quite hard I think...

So we could share the js code I think but at least for js_interop (dart) this support is still marked as experimental and relies on wasm, which can behave strange sometime. Therefore, not recommended.
> 
> (the decision is of course up to you as maintainers :) )

For Koblenz we'd have to use a different createCards endpoint in the backend anyway, right @f1sh1918. Therefore, I removed it. I see the selling point in keeping it, though.
Let's discuss this in our next weekly.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
michael-markl: Yup, definitely personal preference :D

Note, how the relevant code now is spread across ~75 lines and as a user of the public function I have to read private functions to understand the contract.

But let's close the discussion here^^ (unless you want to discuss it further :D )

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
```diff
@@ -0,0 +1,92 @@
...
+            if (o == null) {
+                return "null"
+            }
+            when (o) {
+                is String -> return "\"${o}\""
+                is Boolean, is Int -> return o.toString()
+                is Float -> return if (o.isFinite()) o.toString() else throw Error("Number with value $o cannot be passed to serializeToCanonicalJson.")
+                is Double -> return if (o.isFinite()) o.toString() else throw Error("Number with value $o cannot be passed to serializeToCanonicalJson.")
+                is Map<*, *> -> {
+                    return o.mapKeys { it.key.toString() }.toSortedMap(Comparator.naturalOrder()).entries.map {
```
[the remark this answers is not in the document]
michael-markl: No, at least in JS, object keys are always of type `String` or `Symbol` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#properties).
The JS implementation of `serializeToCanonicalJson` should probably throw when `Symbol`s occur, so we actually only have `String`s left.

For Dart, I am unsure what the line `https://github.com/digitalfabrik/entitlementcard/blob/231dcb336be2698165785a1205a3fb77d373a937/frontend/lib/util/json_canonicalizer.dart#L37C1-L37C55` does exactly. I'd hope it throws, if the type of the Key is not a string.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+
+        fun serializeToString(message: GeneratedMessageV3) = serializeToString(messageToMap(message))
+        private fun Int.isSafeInteger() = this > lowestSafeInt && this < highestSafeInt
+
+        fun serializeToString(o: Any?): String {
+            return when (o) {
+                null -> "null"
+                is String -> "\"${o}\""
```
michael-markl, line 95: This is not sufficient! You need to escape some characters. It would be better to use some existing JSON Encoder for Strings, numbers, etc.

michael-markl: Maybe also add a test case that includes a string containing `"` or containing a new line char.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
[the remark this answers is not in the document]
michael-markl: Turns out, Dart does in fact throw at runtime if the type is not a String:
https://dartpad.dev/?id=f86ddb228f3b1ba1cdbcc3702056b728

f1sh1918:
As we have discussed we should add a notification to the backend if pdf creation in frontend fails, to delete the (dead)cards.
I don't know if you want to implement this here @sarahsporck or we should create another task

sarahsporck (commented):
I've split removing the frontend code into a separate PR so it is easier to review. You can still test the new endpoint via e.g. postman (which has good graphqlsupport) or curl with a request similar to 
```curl
curl --location 'http://localhost:8000/' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer <auth-token-can-be-found-in-backend-logs>' \
--data '{"query":"mutation CreateCardsByCardInfos {\n    createCardsByCardInfos(\n        project: \"bayern.ehrenamtskarte.app\"\n        generateStaticCodes: true\n        encodedCardInfos: [\"Cg5NYXggTXVzdGVybWFubhCIchoICgIIASICCAA=\"]\n    ) {\n        staticVerificationCodeBase64\n        dynamicActivationCode {\n            cardInfoHashBase64\n            codeBase64\n        }\n    }\n}","variables":{}}'
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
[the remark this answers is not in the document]
sarahsporck: Ahh. So the number is converted to a symbol by the runtime? I'm wondering what's happening here then: https://github.com/digitalfabrik/entitlementcard/blob/ab799069c4e9082a90365b0eaee5d4064138a2b6/administration/src/util/serializeToCanonicalJson.test.ts#L120

administration/src/cards/hashCardInfo.ts
```diff
@@ -1,106 +0,0 @@
-import { AnyMessage, Message } from '@bufbuild/protobuf'
-
-import { CardInfo } from '../generated/card_pb'
-import serializeToCanonicalJson from '../util/serializeToCanonicalJson'
```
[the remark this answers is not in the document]
sarahsporck: Nope. I removed it in the other pr (#1284). 👍🏼

sarahsporck:
> As we have discussed we should add a notification to the backend if pdf creation in frontend fails, to delete the (dead)cards. I don't know if you want to implement this here @sarahsporck or we should create another task

This will be done in #1284

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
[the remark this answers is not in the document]
michael-markl: Yes, that's what JS is doing. the number is converted to a string at the time the engine parses `{ 42: ` :D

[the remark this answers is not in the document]
michael-markl: (It does not convert it to a `Symbol` (`Symbol`s are quite rarely used in application code), but it coerces the number to a `String`)

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -8,29 +9,157 @@ import app.ehrenamtskarte.backend.exception.service.ProjectNotFoundException
...
+                .setActivationSecret(ByteString.copyFrom(rawActivationSecret))
+                .build()
+
+            if (!QRCodeUtil.isContentLengthValid(dynamicActivationCode)) {
+                throw InvalidQrCodeSize(cardInfo.toByteArray().encodeBase64(), CodeType.DYNAMIC)
+            }
+
+            if (!cardInfo.extensions.hasExtensionRegion()) {
+                throw InvalidCardHashException()
+            }
```
michael-markl, line 65: Is it reasonable to move as much as possible outside of the DB transaction?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
```diff
@@ -0,0 +1,130 @@
...
+                        }
+                        Pair(it.key.number.toString(), messageToMap(subMessage))
+                    }
+
+                    else -> throw Error("Field ${it.key.name} has an unsupported type of ${it.key.type}")
+                }
+            }
+        }
+
+        fun serializeToString(message: GeneratedMessageV3) = serializeToString(messageToMap(message))
```
michael-markl, line 94:
```suggestion
        fun serializeToString(message: GeneratedMessageV3) = serializeToString(messageToMap(message))

```

michael-markl (approved):
Only minor stuff I think

ztefanie (approved):

f1sh1918 (approved):
I think dart has to be installed for fixing commit workflow
Tested for nuernberg and bavaria on android.
Works fine

PR #1258, opened 2024-01-22
Use npm workspaces for proper linking

sarahsporck:
### Short description

This adds npm workspaces to ensure that build-configs are linked properly to administration, as without this, symlinking did not reliably for everyone.

### Proposed changes

- added a root package.json which defines admisistration and frontend/build-configs as packages

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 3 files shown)


f1sh1918 (approved):
works fine. thank you

PR #1261, opened 2024-01-26
Fix android build

ztefanie:
New android versions could not read QR code due to some error in the CI. This PR should fix it. Somehow the flavor is not set properly with the gradle command. As a quick fix, i removed the gradle command entirely and the whole build is done with flutter comands and env variables. Not the best solution, but it works for now.

### DIFF
--- a/.circleci/src/workflows/delivery_beta_native.yml
+++ b/.circleci/src/workflows/delivery_beta_native.yml
@@ -15,14 +15,14 @@ jobs:
       context:
         - credentials-repo
         - credentials-ehrenamtskarte
-#  - deliver_android:
-#      name: deliver_android_bayern
-#      buildConfig: "bayern"
-#      context:
-#        - tuerantuer-google-play
-#      production_delivery: false
-#      requires:
-#        - build_android_bayern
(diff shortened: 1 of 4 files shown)


.circleci/src/workflows/delivery_beta_native.yml
```diff
@@ -69,11 +69,11 @@ jobs:
...
-        - build_ios_nuernberg
\ No newline at end of file
+#  - deliver_ios:
+#      name: deliver_nuernberg_ios
+#      buildConfig: nuernberg
+#      context:
+#        - tuerantuer-apple
+#      production_delivery: false
+#      requires:
+#        - build_ios_nuernberg
```
sarahsporck: I guess this will be uncommented at some point?

sarahsporck (approved):
Thank you for the fix!
Either way having two vars (flavor and build_config) is prone to having errornous results. We should think about merging the two or generating one value from the other (which would be a toPascalCase(build_config) or toKebapCase(flavor) and therefore quite simple to implement). But this can also be considered in #289

PR #1272, opened 2024-02-05
Update intellij config

f1sh1918:
There were some adjustments in the intejji config missing or incorrect:

- add font-glypth path for map styles
- adjust run config for administration test
- remove unneeded config file


.idea/vcs.xml
```diff
@@ -10,5 +10,6 @@
   <component name="VcsDirectoryMappings">
     <mapping directory="" vcs="Git" />
     <mapping directory="$PROJECT_DIR$/backend/ehrenamtskarte-maplibre-style" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/backend/ehrenamtskarte-maplibre-style/font-glyphs" vcs="Git" />
```
sarahsporck, line 13: You should add to the docs then that this submodule also needs to be initialized (as I did not do this yet ^.^)
Running `git submodule update --init --recursive` in the root repo `entitlementcard` will recursively initialize all submodules.

f1sh1918: ok so is this even needed then? Should be cloned auto?
```
8. Clone the submodule inside `backend/ehrenamtskarte-maplibre-style`
```

sarahsporck (approved):

.idea/vcs.xml
f1sh1918: you still didnt't tell if you checked if step 8 (clone submodule) is needed? I would like to update the docs

sarahsporck: It seems I had to update git, so that the submodules would be properly initialized on clone.

Issue #1199, opened 2023-11-09
Add mattermost notifications


ztefanie:
Mattermost notifications should be sent for:

- failing on main
- releases with realease notes and apk-link
- delivery on feature branch (see in bump_version)

PR #1277, opened 2024-02-07
1199: Add notify

ztefanie:
### Short description
Add notify command and end all jobs with it.

Notify only works if the toolbelt was installed successfully. 
The toolbelt is only notifying if it was a main branch or if it is a branch starting with "release-" and the "allow-all-branches" flag is set.

### Resolved issues

Fixes: #1199


.circleci/src/commands/notify.yml
```diff
@@ -0,0 +1,37 @@
...
+      command: |
+        if [ -n "<< parameters.success_message >>" ]; then
+            echo 'export MM_MESSAGE="<< parameters.success_message >>"' >> $BASH_ENV
+        fi
+      name: Mattermost - Prepare success message
+      when: on_success
+  - run:
+      command: |
+        echo 'export ALLOW_ALL_BRANCHES="<< parameters.allow-all-branches >>"' >> $BASH_ENV
+        app-toolbelt v0 notify mattermost --message "${MM_MESSAGE}" --channel << parameters.channel >> ${ALLOW_ALL_BRANCHES:+--allow-all-branches}
```
sarahsporck, line 34: What's this syntax/What does the '+' mean? `${ALLOW_ALL_BRANCHES:+--allow-all-branches}`

.circleci/src/jobs/check.yml
```diff
@@ -7,4 +7,7 @@ environment:
 shell: /bin/bash -eo pipefail
 steps:
   - checkout
-  - check_circleci_config
\ No newline at end of file
+  - install_app_toolbelt
+  - check_circleci_config
+  - notify:
+      allow-all-branches: true
```
sarahsporck, line 13: This will not send anything to mattermost if this fails, right?

.circleci/src/commands/notify.yml
ztefanie: It is the bash syntax for conditionally set the `allow-all-branches`-flag when the ALLOW_ALL_BRANCHES env var is set.

.circleci/src/jobs/check.yml
ztefanie: no this should be removed, thanks.

f1sh1918 (commented):
Just one question because i didn't find it in the code @ztefanie 
Is there also a notification for successful release? because it was mentioned in #1199 or will there be a separate pr

ztefanie:
> Just one question because i didn't find it in the code @ztefanie Is there also a notification for successful release? because it was mentioned in #1199 or will there be a separate pr

yes there will be a separate PR, i will open a ticket for this. We need the release-notes parsing first.

f1sh1918 (approved):
nicely done :)

sarahsporck (approved):

Issue #1268, opened 2024-01-31
Update maplibre to latest


f1sh1918:
**Is your feature request related to a problem? Please describe.**
We are still using an old version depending on a revision from march 23.

**Describe the solution you'd like**
We should update to the ref/tag of the latest version

**Additional context**
[Latest release](https://github.com/maplibre/flutter-maplibre-gl/releases/tag/0.18.0)

maxammann:
What issues do you experience with the current version? I heard something about rendering issues.

f1sh1918:
> What issues do you experience with the current version? I heard something about rendering issues.

I think you can find these issues in Sentry. There are no visible issues but lots of error reports. But we also have the for integreat. 
I mainly updated to get rid off the mapbox dep

PR #1281, opened 2024-02-07
1268: Update maplibre to latest

f1sh1918:
### Short description

Since we were using very old version of maplibre, i updated to the latest

### Proposed changes

- updated library to latest release
- remove unneeded pod specification for ios
- mapbox dep is now removed 📦 
- only notify listeners if value change to avoid unneeded notifications

### Side effects

- i had to add a constraint to gradle for resolving conflicting kotlin versions. Not sure if there is better fix. Check issue in the comment

### Resolved issues

Fixes: #1268


sarahsporck:
Current location is not working for me. When tapping the button on Android the screen becomes black and the map is not loaded.
<details>
<summary>Error & Stacktrace</summary>

```
E/flutter (20242): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Exception: A PagingController was used after being disposed.
E/flutter (20242): Once you have called dispose() on a PagingController, it can no longer be used.
E/flutter (20242): If you’re using a Future, it probably completed after the disposal of the owning widget.
E/flutter (20242): Make sure dispose() has not been called yet before using the PagingController.
E/flutter (20242): #0      PagingController._debugAssertNotDisposed.<anonymous closure> (package:infinite_scroll_pagination/src/core/paging_controller.dart:133:9)
E/flutter (20242): #1      PagingController._debugAssertNotDisposed (package:infinite_scroll_pagination/src/core/paging_controller.dart:142:6)
E/flutter (20242): #2      PagingController.notifyStatusListeners (package:infinite_scroll_pagination/src/core/paging_controller.dart:168:12)
E/flutter (20242): #3      PagingController.value= (package:infinite_scroll_pagination/src/core/paging_controller.dart:94:7)
E/flutter (20242): #4      PagingController.error= (package:infinite_scroll_pagination/src/core/paging_controller.dart:69:5)
E/flutter (20242): #5      ResultsLoaderState._fetchPage (package:ehrenamtskarte/search/results_loader.dart:106:25)
E/flutter (20242): <asynchronous suspension>
```
</details

f1sh1918:
> Current location is not working for me. When tapping the button on Android the screen becomes black and the map is not loaded.
> 
> Error & Stacktrace
> ```
> E/flutter (20242): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Exception: A PagingController was used after being disposed.
> E/flutter (20242): Once you have called dispose() on a PagingController, it can no longer be used.
> E/flutter (20242): If you’re using a Future, it probably completed after the disposal of the owning widget.
> E/flutter (20242): Make sure dispose() has not been called yet before using the PagingController.
> E/flutter (20242): #0      PagingController._debugAssertNotDisposed.<anonymous closure> (package:infinite_scroll_pagination/src/core/paging_controller.dart:133:9)
> E/flutter (20242): #1      PagingController._debugAssertNotDisposed (package:infinite_scroll_pagination/src/core/paging_controller.dart:142:6)
> E/flutter (20242): #2      PagingController.notifyStatusListeners (package:infinite_scroll_pagination/src/core/paging_controller.dart:168:12)
> E/flutter (20242): #3      PagingController.value= (package:infinite_scroll_pagination/src/core/paging_controller.dart:94:7)
> E/flutter (20242): #4      PagingController.error= (package:infinite_scroll_pagination/src/core/paging_controller.dart:69:5)
> E/flutter (20242): #5      ResultsLoaderState._fetchPage (package:ehrenamtskarte/search/results_loader.dart:106:25)
> E/flutter (20242): <asynchronous suspension>
> ```

Hm looks like its in a loop. The error is not really giving much information

f1sh1918:
@sarahsporck i finally figured out the reason for the rerendering cycle.
This has nothing to do with the error message but the settings provider triggers rerendering and determining position infinite.
I just found something about this disposal issue of a `PageController` you mentioned, but couldn't fix it.
https://stackoverflow.com/questions/73000965/how-to-dispose-pagingcontrollerinfinite-scroll-pagination-in-getxcontroller
I think we should just create a separate task for it

f1sh1918:
Okay i adjusted the `SettingsProvider` to only notify if an value has changed. Not sure if there is a better solution @sarahsporck and @michael-markl

michael-markl:
While I think it is okay to only notify when something actually changed, I do not quite get why this would fix a bug. There is probably something else that's triggering changes even if nothing would change?

frontend/lib/configuration/settings_model.dart
```diff
@@ -44,43 +44,60 @@ class SettingsModel extends ChangeNotifier {
...
+    notifyChange(currentLanguage, language);
   }
 
   @override
   String toString() {
     return 'SettingsModel{firstStart: $firstStart, hideVerificationInfo: $hideVerificationInfo, locationFeatureEnabled: $locationFeatureEnabled, enableStaging:$enableStaging, language:$language}';
   }
+
+  // only notify if value has changed
+  void notifyChange(dynamic oldValue, dynamic newValue) {
```
michael-markl, line 98: Not sure if this works, but if it does, it would be slightly better:
```suggestion
  void notifyChange<T>(T oldValue, T newValue) {
```

f1sh1918:
> While I think it is okay to only notify when something actually changed, I do not quite get why this would fix a bug. There is probably something else that's triggering changes even if nothing would change?

No it's the particular case with determineLocation that is running on widget build for location Button once. This Updates the settings and triggers a rerender which again updates the settings. I guess this function was added to listen to permission changes of the OS for location @michael-markl

michael-markl:
I think "rerenders" should not trigger setting updates blindly?

michael-markl:
Can you point me to where this settings update on rerender happens? I cannot seem to find it.

f1sh1918:
> Can you point me to where this settings update on rerender happens? I cannot seem to find it.

Sure. it also took me couple of time to find the issue.
https://github.com/digitalfabrik/entitlementcard/blob/587243ae00a5d83ce7e771693456277a5ebc504b/frontend/lib/search/location_button.dart#L29

And even on `userInteract= false` the settings will be updated by this function that was passed
https://github.com/digitalfabrik/entitlementcard/blob/587243ae00a5d83ce7e771693456277a5ebc504b/frontend/lib/search/location_button.dart#L86

frontend/lib/configuration/settings_model.dart
f1sh1918: changed it

frontend/lib/search/location_button.dart
```diff
@@ -79,14 +79,8 @@ class _LocationButtonState extends State<LocationButton> {
               await settings.setLocationFeatureEnabled(enabled: true);
             },
           )
-        : await determinePosition(
-            context,
-            requestIfNotGranted: false,
-            onDisableFeature: () async => settings.setLocationFeatureEnabled(enabled: false),
-            onEnableFeature: () async {
```
f1sh1918, line 86: this shouldn't be triggered on every widget rebuild "blindly"
Even if the user deactivates location permission in the OS, the location but does not change to disabled with this implementation.
The user has to click again on the button.
If we want to change this automatically we have to implement it on a proper way

michael-markl: Hmm... I think this would only get called once when the widget initializes, right? So that does not seem to be triggering a cycle (at first glance)?

f1sh1918: well it seem not to be case somehow because it actually caused the rerendering. When i commented out the function in the `WidgetsBinding.instance.addPostFrameCallback` the cycle was gone
Maybe something forces recreation?

michael-markl: >  Maybe something forces recreation?

Hmm... it seems like it... That seems to be the root cause, then. Maybe it has something to do with the changes to the routing / deep linking?

f1sh1918: Yes I also thought about that because it didn't happen before merging that issue. Even I have no idea how the GoRouter should cause an issue which seems not to be related with routing. But maybe sth triggers the app. dart to rerender

f1sh1918: i reverted the changes. thx for the pr

michael-markl (approved):
I did not test it

f1sh1918:
Would be nice to get another (tested) approval @sarahsporck or @ztefanie :)

sarahsporck (approved):
Tested on Android real device

PR #1284, opened 2024-02-08
Remove hashing from administratiom

sarahsporck:
### Short description

Removes the hashing logic from the administration frontend. As @michael-markl discussed this in #1257 we might want to keep the old endpoint for the admin frontend. Therefore, we can discuss this in our next weekly (this time for sure ;))

### Proposed changes

- removes hashCardInfo & serialze to json from frontend
- removes addCards graphql endpoint from backend

### Open TODOS (from other pr)

- [ ] delete created cards when pdf generation fails

### Side effects

After this change user data is technically sent to the backend (even though it is hashed) This might need to be discussed with our current clients. Would be nice if you could join this discussion in our weekly @dkhene


graphql.config.yml
sarahsporck, line 1: This is used by the IDE (both IntelliJ and VSCode work 🥳 )
For more insights in how to operate the graphql extension see https://github.com/JetBrains/js-graphql-intellij-plugin

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -25,29 +35,25 @@ const useCardGenerator = (region: Region) => {
...
 
+      cardBlueprints.forEach(cardBlueprint => new ActivityLog(cardBlueprint).saveToSessionStorage())
       downloadDataUri(pdfDataUri, 'berechtigungskarten.pdf')
       setState(CardActivationState.finished)
     } catch (e) {
+      if (codes !== undefined) {
+        // try rollback
+        try {
+          await deleteCards(client, projectConfig.projectId, region.id, extractCardInfoHashes(codes))
+        } catch {}
```
sarahsporck, line 55: maybe add a sentry report here? We'd then need to set up sentry for web, though.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
```diff
@@ -8,11 +8,20 @@ import app.ehrenamtskarte.backend.verification.database.CardEntity
...
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.isNull
 import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.deleteWhere
 import org.jetbrains.exposed.sql.select
 import java.time.Instant
 
 object CardRepository {
+    fun deleteInactiveCardsByHash(project: String, regionId: Int, cardInfoHashList: List<ByteArray>) {
+        Cards.deleteWhere {
+            firstActivationDate.isNull() and (cardInfoHash inList cardInfoHashList) and (Cards.regionId eq regionId)
```
sarahsporck, line 22: I think the region has to be passed here, as a user is only allowed to delete cards of one region, right?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
```diff
@@ -8,11 +8,20 @@ import app.ehrenamtskarte.backend.verification.database.CardEntity
...
+        }
```
michael-markl, line 23: Do you think we should return an error if some cards were not found or were already active or have the wrong regionId?

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
f1sh1918: you can create an issue and add a comment here. has not to be done is this task

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
sarahsporck: Not sure. For now this should be sufficient, as it's just the fallback for when pdf generation fails. But if we want to implement deletion by the user or an administrator this could look different. @f1sh1918 what do you think?

michael-markl: If we leave it like this, we should probably document the API endpoint that it silently fails / ignores these cases.

f1sh1918: i think users shouldn't be able to delete cards but revoke them if we gonna implement a "revoke" function one day.
Btw how can i test a pdf creation failure to initiate the deletion @sarahsporck ?

f1sh1918 (commented):
Looks fine to me
Tested card generation single and bulk for ehrenamt and sozialpass.
Any hint how to reproduce pdf generation error?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
sarahsporck: I just added a throw Error in the PdfFactory to test this 👍🏼

f1sh1918 (changes requested):
I just tested the deletion of cards which only seem to work for dynamic cards.
Is that the desired behavior @sarahsporck ?

sarahsporck:
> I just tested the deletion of cards which only seem to work for dynamic cards. Is that the desired behavior @sarahsporck

Nope it is not and I can already guess the culprit. Static cards don't have an activationDate by default. I'll fix this

administration/src/cards/CardBlueprint.ts
```diff
@@ -118,32 +115,15 @@ export class CardBlueprint {
...
-    return isContentLengthValid(dynamicCode) && isContentLengthValid(staticCode)
-  }
-
   generateCardInfo = (): CardInfo => {
     const extensionsMessage: PartialMessage<CardExtensions> = {}
 
     this.extensions.forEach(extension => {
-      if (extension.state === null || !extension.setProtobufData) {
-        // We allow to skip invalid extensions to enable computing the protobuf size.
+      if (!extension.setProtobufData) {
```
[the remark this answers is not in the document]
sarahsporck: It should not be removed (to allow a partial correct state). Although, now there are no errors because of removing this, as the protobuf computation has been removed.

f1sh1918:
I can still reproduce it @sarahsporck 
I created a card for nuernberg and the static card still remains in database

image

f1sh1918 (approved):
Nicely done :)
Now works as expected!

Issue #1271, opened 2024-02-01
Deliver administration & backend production


f1sh1918:
- Move beta artifacts to production server (promotion)
- Enable direct production deployment
- Unhold apk packages from installation on Production
- Deploy webhook config on production server
- Update documentation

PR #1290, opened 2024-02-13
1271: Promote backend administration

f1sh1918:
### Short description

This pr copies the beta artifacts to production and installs them. It also enables the direct production delivery. 

### Proposed changes

- [x] Move beta artifacts to production server and install them (promotion)
- [x] Enable direct production delivery (even we should avoid this)
- [x] Unhold apk packages from installation on production server
- [x] Deploy webhook config on production server
- [x] Merge workflows for administration and backend delivery, since they shouldn't be executed separately

### Note
- Please do not run production delivery :)

### Side effects

- the delivery jobs will take longer but ensures that if building or packaging fails for one of the artifacts, nothing will be deployed and installed on the server

### Resolved issues

Fixes: #1271


.circleci/config.yml
```diff
@@ -838,16 +872,10 @@ parameters:
...
         default: false
         type: boolean
-    run_deliver_beta_backend:
-        default: false
-        type: boolean
-    run_deliver_production_administration:
-        default: false
-        type: boolean
-    run_deliver_production_backend:
+    run_deliver_production_backend_administration:
```
f1sh1918, line 878: maybe we just call it backend?

ztefanie (approved):
Nice :rocket: Good job :+1: 

Just one thing: I think it would just be fine to delete the "deliver_production_backend_administration"-workflow, to enforce us, to only do betas, test then and then promote them. But as it is already there, I am also fine with keeping it.

.circleci/config.yml
ztefanie: I like this naming, as it is more clear what is happening. I prefer having a few extra chars instead of not understanding at the first look what is going on.

sarahsporck (approved):
Very nice! The deliver/promote commands are run via cron job, right? Maybe we could document this somewhere, as we are doing it for IG. No need to this in this PR.

f1sh1918:
> Nice 🚀 Good job 👍
> 
> Just one thing: I think it would just be fine to delete the "deliver_production_backend_administration"-workflow, to enforce us, to only do betas, test then and then promote them. But as it is already there, I am also fine with keeping it.

Ok then i think i just gonna add a comment to this workflow to use the beta and the promotion

f1sh1918:
> Very nice! The deliver/promote commands are run via cron job, right? Maybe we could document this somewhere, as we are doing it for IG. No need to this in this PR.

Yes this documentation will be done is a separate task when all delivery jobs are done

Issue #1265, opened 2024-01-29
Card storage does not survive app restart


f1sh1918:
**Describe the bug**

**To Reproduce**
1. Create a Pixel 7 with API 33 (Play Store) emulator
2. Activate a card
3. Close the app and reopen it
4. Card can't be loaded from storage

**Expected behavior**
Card should be loaded from storage

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy A34 5G SM-A346B/DSN
 - OS: Android 14 or higher

**Additional context**
Seems like an issue with `FlutterSecureStorage`
https://github.com/mogol/flutter_secure_storage/issues/653
Open pr that has to be tested
https://github.com/mogol/flutter_secure_storage/pull/664

ztefanie:
API level 33 is not working with android 14 or higher. Which versions are creating this bug?

vthejay:
Can not reproduce this. Browser Stack- Android 14, Pixel 8, The card is still there after closing the App.

vthejay:
Reported bug from user:
Device: Google Pixel 7
App version name: 2024.1.5 
Android version: Android 14 (SDK 34)

ztefanie:
Tried to reproduce this unsuccessfully. Browserstack Pixel 8 with Android 14. Force stopped the app and cleared cache. Card was still there. Also tried to reproduce locally on differend emulators with differend android version and on my real device Samsung Galaxy A33 with Android 13.

vthejay:
Reported per Mail in the last 7 days:
The same problem happens here:
1. Sony Experia i13113, Android-Version 10; App Version does not matter. In all Versions of the App.
2. Xiaomi Redmi Note 12, Androidversion 14 UKQ1.230917.001, App Version 2024.1.5

I could not reproduce this as well.

f1sh1918:
Maybe fixed with #1293. Will have to check with upcoming release

f1sh1918:
New release was deployed on production that includes a new `SecureLocalStorage` version. We should check the sentry logs and wait if it will be reported

f1sh1918:
There was a new review on google console that it still occurres.

- Gerät: [Samsung Galaxy S24 Ultra]
- Sprache des Geräts: Deutsch
- Code der App-Version: 168
- Name der App-Version: 2024.5.5
[...]

PR #1293, opened 2024-02-14
1265: Workaround empty secure storage

f1sh1918:
### Short description

We are facing some not reproducible issue on android devices that the user data (stored cards) can't be read from secure storage after app restart. This relies on an issue from `flutter secure storage`
https://github.com/mogol/flutter_secure_storage/issues/653
Some devs report that a fake read could solve this issue.

### Proposed changes
- add a fake read for android devices
- add a second read for android devices that may contain the user data
- report error to sentry if first read is null but second read contains data to collect more information

### Side effects

- the initialization of the cards may be a bit slower for android but it shouldn't be a big performance issue

### Resolved issues

related to #1265

**Note:** We have to contact users that facing the issue if it was solved by this workaround, since we can not really test it
Just test if a saved card will be restored/loaded (still) properly

### DIFF
--- a/frontend/lib/identification/user_code_store.dart
+++ b/frontend/lib/identification/user_code_store.dart
@@ -34,8 +34,10 @@ class UserCodeStore {
       String? userCodesBase64FirstRead = await storage.read(key: _userCodesBase64Key);
       userCodesBase64 = await storage.read(key: _userCodesBase64Key);
       // report error if first read is null but second read contains data to collect sentry data for reproduction
-      if (userCodesBase64FirstRead == null && userCodesBase64 != null) {
-        reportError('First read from secure storage is null but user data exists', null);
+      if (userCodesBase64FirstRead != userCodesBase64) {
+        bool firstIsNull = userCodesBase64FirstRead == null;
+        bool secondIsNull = userCodesBase64 == null;
+        reportError("First read from secure storage differs from second: firstNull=${firstIsNull}, secondNull=${secondIsNull}", null);
       }
     } else {
       userCodesBase64 = await storage.read(key: _userCodesBase64Key);
--- a/frontend/lib/identification/user_code_store.dart
+++ b/frontend/lib/identification/user_code_store.dart
@@ -37,7 +37,9 @@ class UserCodeStore {
       if (userCodesBase64FirstRead != userCodesBase64) {
         bool firstIsNull = userCodesBase64FirstRead == null;
         bool secondIsNull = userCodesBase64 == null;
-        reportError("First read from secure storage differs from second: firstNull=${firstIsNull}, secondNull=${secondIsNull}", null);
+        reportError(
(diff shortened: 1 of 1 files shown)


frontend/lib/identification/user_code_store.dart
```diff
@@ -24,7 +28,19 @@ class UserCodeStore {
...
-    final String? userCodesBase64 = await storage.read(key: _userCodesBase64Key);
+    String? userCodesBase64;
+    // make fake read on android - workaround for https://github.com/mogol/flutter_secure_storage/issues/653
+    if (Platform.isAndroid) {
+      String? userCodesBase64FirstRead = await storage.read(key: _userCodesBase64Key);
+      userCodesBase64 = await storage.read(key: _userCodesBase64Key);
+      // report error if first read is null but second read contains data to collect sentry data for reproduction
+      if (userCodesBase64FirstRead == null && userCodesBase64 != null) {
+        reportError('First read from secure storage is null but user data exists', null);
+      }
```
michael-markl, line 39: Maybe, we can report more potential misbehavior here to get a better understanding.
Maybe something like this:
```suggestion
      if (userCodesBase64FirstRead != userCodesBase64) {
        bool firstIsNull = userCodesBase64FirstRead == null;
        bool secondIsNull = userCodesBase64 == null;
        reportError("First read from secure storage differs from second: firstNull=${firstIsNull}, secondNull=${secondIsNull}", null);
      }
```

frontend/lib/identification/user_code_store.dart
```diff
@@ -24,7 +28,19 @@ class UserCodeStore {
...
+    } else {
+      userCodesBase64 = await storage.read(key: _userCodesBase64Key);
+    }
+
     if (userCodesBase64 == null) return [];
     return userCodesBase64
         .split(_storageDelimiter)
```
michael-markl, line 46: I have a question regarding the `importLegacyCard` function below. 

It constructs `UserCodeModel` without initializing it and then directly runs `insertCode` which in turn runs `UserCodeStore.store()`.
This seems quite bad.

I think it is not the cause of the problem, as `insertCode` is only called when a legacy code is found and thus should only be called once ever from importLegacyCard. But we should definitely refactor this.
`importLegacyCard` should return `Future<DynamicUserCode?>` and there should be a separate function `deleteLegacyCard` which should delete the key. The `initialize`  function of `UserCodeModel` should take care of calling these functions.

Also, we should add assertions to the functions of UserCodeModel which ensure they can only be called if UserCodeModel is initialized.

michael-markl (approved):

frontend/lib/identification/user_code_store.dart
f1sh1918: i'm gonna create a task for it, thx for the hint

f1sh1918:
Can anyone review AND test it :) @sarahsporck and @ztefanie

sarahsporck (approved):
Tested in Android emulator. Seems to work the same way as before 👍🏼

PR #1295, opened 2024-02-15
xxx-fix-ios-build-ci

f1sh1918:
### Short description

Looks like when updating associated domain the certificates have to be renewed
https://stackoverflow.com/questions/60788061/provisioning-profile-ios-team-provisioning-profile-doesnt-support-the-ass

Also the fvm_config is not available (seem to be not a new issue)

### Proposed changes

- [x] update the provisioning profiles to include associated domains
- fix xcode archive fails

### Side effects

-
-

### Resolved issues

Fixes: #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -886,45 +886,6 @@ workflows:
         jobs:
             - bump_version:
                 prepare_delivery: false
-            - check
-            - check_administration
-            - check_backend
-            - build_administration:
-                requires:
-                    - check_administration
-                    - bump_version
-            - pack_administration:
-                requires:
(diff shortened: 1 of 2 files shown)


f1sh1918:
ok i figured out that there was an issue with the provisioning profile name after i updated them to include `associated domains`. Somehow the existing profile wasn't updated but created a new one called `<bundleIdentifier> <some numbers>`
Probably there is a check if the profile name matches to the bundle identifier somewhere @ztefanie ? No idea tbh.
Really annoying not to get a proper error log from fastlane/xcode...but finally solved :)

Issue #850, opened 2023-03-06
Increase allowed email length for DB scheme


sarahsporck:
**Is your feature request related to a problem? Please describe.**
According to [stackoverflow](https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address) we should up our limit from 100 to at least 254.

**Describe the solution you'd like**
Increase the `varchar` in the DB to at least 254.

**Describe alternatives you've considered**

**Additional context**
Currently blocked by database migrations:
- #839

PR #1298, opened 2024-02-15
850: Increase allowed email length to 254

sarahsporck:
### Short description

Apparently emails can be longer than 100 characters. An email length of 254 characters can be considered the maximum length, therefore this PR increases the allowed length to 254.

### Proposed changes

- add migration to increase allowed email length

### Resolved issues

Fixes: #850


f1sh1918 (approved):
Tested migration locally. Works fine

PR #1305, opened 2024-02-21
Construct GoRouter only once to prevent tree recreation on settings change

michael-markl:
### Short description

After #179, the whole app tree was constantly recreated if the location feature was enabled.
This was because GoRouter recreated the whole app tree every time any the settingsmodel notified about changes. 

### Proposed changes

Do not recreate GoRouter after every Settings change. It seems this is the recommended way to use GoRouter.
(see this sample usage https://github.com/flutter/packages/blob/main/packages/go_router/example/lib/main.dart)

### Side effects

Hopefully none, except for fixing the bug.

### DIFF
--- a/frontend/lib/about/language_change.dart
+++ b/frontend/lib/about/language_change.dart
@@ -31,7 +31,7 @@ class LanguageChange extends StatelessWidget {
     final settings = Provider.of<SettingsModel>(context, listen: false);
     LocaleSettings.setLocaleRaw(language);
     settings.setLanguage(language: language);
-    Navigator.of(context).pop(true);
+    Navigator.of(context).pop();
     messengerState.showSnackBar(
       SnackBar(
         backgroundColor: Theme.of(context).colorScheme.primary,
--- a/frontend/lib/app.dart
+++ b/frontend/lib/app.dart
@@ -46,12 +46,6 @@ final GoRouter router = GoRouter(
       ],
     ),
     GoRoute(
-      path: homeRouteName,
-      builder: (BuildContext context, GoRouterState state) {
-        return HomePage(initialTabIndex: int.parse(state.pathParameters[homeRouteParamTabIndexName]!));
-      },
-    ),
-    GoRoute(
       path: '$homeRouteName/:$homeRouteParamTabIndexName',
       builder: (BuildContext context, GoRouterState state) {
         return HomePage(initialTabIndex: int.parse(state.pathParameters[homeRouteParamTabIndexName]!));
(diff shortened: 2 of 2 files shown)


f1sh1918 (commented):
Thx for the fix. Unfortunately some issues came up with that fix

I created an pr on your pr @michael-markl 
https://github.com/digitalfabrik/entitlementcard/pull/1307

f1sh1918 (approved):
now everything should be fine
Tested on android and ios

frontend/lib/app.dart
```diff
@@ -23,6 +23,52 @@ const homeRouteParamTabIndexName = 'tabIndex';
...
+          },
+        ),
+      ],
+    ),
+    GoRoute(
+      path: homeRouteName,
+      builder: (BuildContext context, GoRouterState state) {
+        return HomePage(initialTabIndex: int.parse(state.pathParameters[homeRouteParamTabIndexName]!));
+      },
+    ),
```
michael-markl, line 53: Is this actually necessary?

frontend/lib/about/language_change.dart
```diff
@@ -33,12 +31,7 @@ class LanguageChange extends StatelessWidget {
     final settings = Provider.of<SettingsModel>(context, listen: false);
     LocaleSettings.setLocaleRaw(language);
     settings.setLanguage(language: language);
-    Navigator.push(
-      context,
-      AppRoute(
-        builder: (context) => HomePage(),
-      ),
-    );
+    Navigator.of(context).pop(true);
```
michael-markl, line 34:
```suggestion
    Navigator.of(context).pop();
```

frontend/lib/app.dart
michael-markl: I removed it again @f1sh1918 as I think it is wrong 🤔

f1sh1918: Yes It is necassary. I described this in my proposal @michael.markl
Just passing an index is not needed my fault. Just forward to the Plain HomePage()

f1sh1918: If you confirm the intro slides there will be a no route found for home exception. please add 
```js
    GoRoute(
      path: homeRouteName,
      builder: (BuildContext context, GoRouterState state) {
        return HomePage();
      },
    )
```

michael-markl: Ah sorry, I didn't see your proposal PR. It's still a bit weird, that we need this route... Especially as HomePage now has three routes...
I added the route again.

f1sh1918: yes well the first is the inititalRoute which depends on settings and then we have a home route and a home route with index. Probably we could also remove that route and just redirect to to home/:0 for the map route `onFinishIntroCallback`
Maybe it also makes sense to use TabNames instead of index, because if we add a new tab between we will get redirected to the wrong tab with a given index.

f1sh1918 (approved):

Issue #1170, opened 2023-10-16
Push search and map routes to the root stack


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Reasoning can be found in #420 .

**Describe the solution you'd like**
We discussed this and decided it makes sense to push routes to the root navigation stack in the search and map tabs.

**Additional context**
#420

michael-markl:
Fixed in commit https://github.com/digitalfabrik/entitlementcard/commit/dcad66ef372ea4e9ba4b52193a25396b128d71e3

Issue #1288, opened 2024-02-13
Native back navigation moves app to background


sarahsporck:
**Describe the bug**
On Android when tapping the back arrow, the app moves to the background. Also the state is not persisted, when focussing the app again, resulting in the app loading the map screen.
This is very unpractical when having larger devices, as the in app back navigation is all the way on top. This was probably introduced by adding the `GoRouter`.
I didn't test native yet. Please do this when working on the ticket.

**To Reproduce**
1. Go to any route on android
2. Tap the native back arrow
3. App goes into the background

**Expected behavior**
App navigates back in the stack. If we are at the root stack the app moves to the background.

PR #1306, opened 2024-02-21
1170: Use root navigator when pushing subroutes

michael-markl:
### Short description

Subroutes are pushed to the root navigator.

### Proposed changes

Subroutes are pushed to the root navigator.

When navigating to map from the detail page of a search result, we have to first pop the detail page from the navigator stack and then navigate to the map tab.

### Side effects

None

### Resolved issues

Fixes: #1170 #1288

### DIFF
--- a/frontend/lib/about/language_change.dart
+++ b/frontend/lib/about/language_change.dart
@@ -1,8 +1,6 @@
 import 'package:ehrenamtskarte/build_config/build_config.dart' show buildConfig;
 import 'package:ehrenamtskarte/configuration/settings_model.dart';
-import 'package:ehrenamtskarte/home/home_page.dart';
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
-import 'package:ehrenamtskarte/routing.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 
@@ -28,17 +26,12 @@ class LanguageChange extends StatelessWidget {
     ]);
   }
 
-  switchLanguage(BuildContext context, String language) {
+  Future<void> switchLanguage(BuildContext context, String language) async {
     final messengerState = ScaffoldMessenger.of(context);
     final settings = Provider.of<SettingsModel>(context, listen: false);
     LocaleSettings.setLocaleRaw(language);
-    settings.setLanguage(language: language);
-    Navigator.push(
-      context,
-      AppRoute(
-        builder: (context) => HomePage(),
-      ),
-    );
+    await settings.setLanguage(language: language);
+    Navigator.pop(context);
     messengerState.showSnackBar(
       SnackBar(
         backgroundColor: Theme.of(context).colorScheme.primary,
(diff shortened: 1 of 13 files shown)


frontend/lib/search/results_loader.dart
```diff
@@ -8,6 +8,8 @@ import 'package:infinite_scroll_pagination/infinite_scroll_pagination.dart';
 
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
 
+import '../home/home_page.dart';
```
f1sh1918, line 11: please use package import

frontend/lib/store_widgets/accepting_store_summary.dart
```diff
@@ -11,18 +11,20 @@ import 'package:intl/intl.dart';
 
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
 
+import '../map/map_page.dart';
```
f1sh1918, line 14: please use correct import

frontend/lib/store_widgets/detail/detail_page.dart
```diff
@@ -11,11 +11,13 @@ import 'package:graphql_flutter/graphql_flutter.dart';
 
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
 
+import '../../map/map_page.dart';
```
f1sh1918, line 14: please adjust import

frontend/lib/about/language_change.dart
```diff
@@ -28,17 +26,12 @@ class LanguageChange extends StatelessWidget {
...
     LocaleSettings.setLocaleRaw(language);
-    settings.setLanguage(language: language);
-    Navigator.push(
-      context,
-      AppRoute(
-        builder: (context) => HomePage(),
-      ),
-    );
+    await settings.setLanguage(language: language);
+    Navigator.pop(context);
```
f1sh1918, line 34: this change will avoid showing the snackbar of the language change.
You could move the showSnackbar before the `settings.setLanguage`
Not sure if there is a better solution

f1sh1918 (commented):
There is an issue in the language change and some imports
Apart of the everything works as expected

frontend/lib/search/results_loader.dart
michael-markl: Is there no linter for that :o

frontend/lib/about/language_change.dart
michael-markl: Let's merge #1305 first and check again. I suspect, that the tree recreation removes the snackbar.

frontend/lib/search/results_loader.dart
michael-markl: I went ahead and enabled it and fixed a few more occurences

frontend/lib/about/language_change.dart
f1sh1918: updated the branch. just some conflicts to resolve

michael-markl: Snackbars are visible for me

f1sh1918 (approved):
Tested on android and ios. works as expected! Also language change is fine now

PR #1307, opened 2024-02-22
fix-recreating-home-proposal

f1sh1918:
### Short description

The fix of not recreating the router causes some issue for the language change -> duplicated bottomTabsMenu
When activating card with deep link and not confirming the intro slides and reopening the app, there is the home route missing when the user then accepts intro slides

### Proposed changes

- fix language change
- add missing home route

image


sarahsporck (approved):
Tested on android real device. Question: Why is the last tab bar item called "Mehr" in your screenshot? On Android there is only "About".

Issue #1313, opened 2024-02-26
Expose android ports


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a developer i have to expose the tcp ports manually for development to access the backend. It would be nice to expose them automatically on build

**Describe the solution you'd like**
- Execute a shell script when running the local flutter builds

PR #1314, opened 2024-02-26
1313: Expose android ports

f1sh1918:
### Short description

As a developer i have to expose the ports for local android builds manually.

### Proposed changes

- expose android ports for local builds if android device is connected

Fixes: #1313

### Testing:
- connect android device
- run local build
- stores should be loaded on map and search
- if no android device is connected the build should also start (ios build)


sarahsporck (approved):
Tested ✅ 
🙃 Personal preference: move port forwarding command into the root package.json, so that it can easily be invoked via cmd (so that vsc users also profit from this 🙈)

f1sh1918:
> Tested ✅ 🙃 Personal preference: move port forwarding command into the root package.json, so that it can easily be invoked via cmd (so that vsc users also profit from this 🙈)

Alright then please test it again :) @sarahsporck

sarahsporck (approved):
Nice ❤️ and tested ✅

PR #1324, opened 2024-02-27
1212: Print deeplink on pdf

f1sh1918:
### Short description

As a user wants to activate the card directly from a deep link, we have to print a proper link on the pdf 

### Proposed changes

- add PdfLinkArea property
- add function getDeepLinkFromQrCode
- add pdf deep link config to pdf config for bavaria
- add tests for deepLink generation
- disabled type checking for testing, since not needed (had some issues)
- update pdf layout
- add tsc build to `build-config` to fix import issues from `build-configs`

### Resolved issues

Tested on ios and android real device with several pdf viewers (recommend adobe acrobat, chrome or firefox pdf viewer)

Fixes: #1212

### Testing

- create a card (bavaria)
- copy the pdf on your device (drag and drop emulator and find it under files)
- (android dev environment only) go to AppSettings -> OpenByDefault -> Add supported links (required for unsigned apks)
- open the pdf and click on the qrCode

### Design (lastest)
[📍 Here](https://nextcloud.tuerantuer.org/index.php/apps/files/files/6032144?dir=/Projects/Digitale%20Berechtigungskarte/Sozialministerium%20Bayern/Releasedokumente&openfile=true)


administration/src/util/generateDeepLink.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { PdfQrCode } from '../cards/pdf/PdfQrCodeElement'
+import { QrCode } from '../generated/card_pb'
+import { uint8ArrayToBase64 } from './base64'
+
+const generateDeepLink = (qrCode: PdfQrCode): string => {
+  const qrCodeContent = new QrCode({
+    qrCode: qrCode,
+  }).toBinary()
+  return `${window.location.origin}/${ACTIVATION_PATH}/${ACTIVATION_FRAGMENT}${encodeURIComponent(
```
sarahsporck, line 11: the host and protocol should come from the project config, so that I can use this url during development.
-> `berechtigingskarte://bayern.ehrenamtskarte.app/...`

sarahsporck (commented):
The other link on the pdf doesn't work. Also 🤡 , there is a additional space in the text of **Variante 2**: 
"[...] auf dem Sie die App bereits installiert haben (**<--here**), [...]"

administration/src/project-configs/getProjectConfig.ts
```diff
@@ -60,11 +68,11 @@ export const setProjectConfigOverride = (hostname: string) => {
 
 const getProjectConfig = (hostname: string): ProjectConfig => {
   switch (window.localStorage.getItem('project-override') ?? hostname) {
-    case 'bayern.ehrenamtskarte.app':
-    case 'staging.bayern.ehrenamtskarte.app':
+    case BAYERN_PRODUCTION_ID:
```
f1sh1918, line 71: just replaced these magic strings here and also in the buildConfigs

administration/src/util/generateDeepLink.ts
f1sh1918: Now i'm facing compile errors @sarahsporck when i'm using the `getBuildConfig()`

f1sh1918: I only get it work with
https://github.com/digitalfabrik/entitlementcard/pull/1328

michael-markl:
I am a bit worried that a lot of people still get the document on paper. Then the whole paragraph is a bit confusing to people, because they don't know that this is even a link.
Would it be better to phrase it slightly different?

Maybe phrase it like:
```
Variante 2:
Sie lesen dieses Dokument bereits auf einem Gerät, auf dem die App installiert ist? Dann tippen Sie auf den QR-Code, um die Karte zu aktivieren.
```
And then add the link target on top of the qr code.

Another remark: The download link for the app at the top should be clickable, right?

f1sh1918:
> I am a bit worried that a lot of people still get the document on paper. Then the whole paragraph is a bit confusing to people, because they don't know that this is even a link.
> Would it be better to phrase it slightly different?
> 
> Maybe phrase it like:
> ```
> Variante 2:
> Sie lesen dieses Dokument bereits auf einem Gerät, auf dem die App installiert ist? Dann tippen Sie auf den QR-Code, um die Karte zu aktivieren.
> ```
> And then add the link target on top of the qr code.
> 
> Another remark: The download link for the app at the top should be clickable, right?

Good point. Gonna discuss this with @hauf-toni
Yes It should. 
Gonna check this. I had some issues with the first link but should be fixed with last commit

hauf-toni:
@michael-markl good point, particularly for non-technology-savvy users this will be accessible and understandable – I will implement your suggestion and deposit the pdf on nextcloud

michael-markl:
@hauf-toni Could you mark the download link for the app as a hyperlink from within InDesign? I think this is currently not the case. That's why I cannot tap on the link on my phone nor click on it on Firefox or Acrobat Reader. It seems, Chrome automatically detects the URL and makes it clickable. But we should mark it as a hyperlink for the other browsers / PDF viewers.

@f1sh1918 Other than that, it seems Firefox and possibly other PDF viewers do not support custom URL schemes like `berechtigungskarte://xyz` but I guess for production this should not be relevant (as we use https URLs), right?

f1sh1918:
> @hauf-toni Could you mark the download link for the app as a hyperlink from within InDesign? I think this is currently not the case. That's why I cannot tap on the link on my phone nor click on it on Firefox or Acrobat Reader. It seems, Chrome automatically detects the URL and makes it clickable. But we should mark it as a hyperlink for the other browsers / PDF viewers.
> 
> @f1sh1918 Other than that, it seems Firefox and possibly other PDF viewers do not support custom URL schemes like `berechtigungskarte://xyz` but I guess for production this should not be relevant (as we use https URLs), right? Edit: You already mentioned this in the PR description :D

@michael-markl
i updated the testing instruction. Sarah mentioned that she expects to have the custom scheme for dev environment.
I think we could just use the staging configuration and follow the instructions i added to `deep-linking.md`.
Otherwise we can not properly test via pdf.

f1sh1918:
Even the download link (app download) in the pdf is not working in some pdf viewers, this pr can be tested and reviewed.
The fixing of this link is not related to this pr and can be done in a separate issue

administration/src/util/getDeepLinkFromQrCode.ts
```diff
@@ -0,0 +1,25 @@
...
+  const qrCodeContent = new QrCode({
+    qrCode: qrCode,
+  }).toBinary()
+  const buildConfigProjectId = getBuildConfig(window.location.hostname).common.projectId
+  // custom link schemes don't work in browsers or pdf thats why we use the staging link scheme also for development
+  const host = isStagingMode() || isDevMode() ? buildConfigProjectId.staging : buildConfigProjectId.production
+  const deepLink = `${HTTPS_SCHEME}://${host}/${ACTIVATION_PATH}/${ACTIVATION_FRAGMENT}${encodeURIComponent(
+    uint8ArrayToBase64(qrCodeContent)
+  )}/`
+  // Log deepLink for development and testing purposes
```
ztefanie, line 19: please delete this comment

docs/deep-linking.md
```diff
@@ -40,24 +40,20 @@ We use a `fragment (code#)` for the activation code to avoid sending user data t
 Install: `npm i -g uri-scheme` if you want to use the simplified command
 
 ### a) Android
+
+#### Note
+For android trusted associations, which are required for https schemes, only work with signed apks out of the box.
+For local development you have to install the app and add these domains manually.
```
ztefanie, line 46: I did not have to do this, it was working out of the box for me.

ztefanie (approved):
Looks good :rocket: 

Tested on android emulator. Worked as expected.

docs/deep-linking.md
f1sh1918: hm were the trusted domains added there? Hm when i delete the apk on emulator and reinstall it they are not set by default. No idea why it worked but i would keep it in the docs even its working in some cases

f1sh1918:
Just created an issue for the not working app download link
https://github.com/digitalfabrik/entitlementcard/issues/1340

PR #1329, opened 2024-03-02
Update README.md

ztefanie:
added emoji code


f1sh1918 (commented):
:+1:: Nice i like the emojis :)

:wrench: i think it would increase the readability to add bulletpoints f.e.
- `:+1:` or `:+1::` : This is great! It always feels good when somebody likes your work. Show them!

f1sh1918 (approved):
👍 nice i like it

PR #1330, opened 2024-03-05
Fix cicd install bundler

ztefanie:
### Short description
delivery is failing due to wrong permissions while installing the gems, see here: https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/4572/workflows/75290bc2-3332-49b1-8762-74c3ea5ab9e9/jobs/35497

### Proposed changes

- fixed the duplication of messages sent
- added missing check step to beta delivery workflow
- fixed bundler update issue
- updated android images

### DIFF
--- a/.circleci/src/commands/notify.yml
+++ b/.circleci/src/commands/notify.yml
@@ -31,6 +31,10 @@ steps:
   - run:
       command: |
         echo 'export ALLOW_ALL_BRANCHES="<< parameters.allow-all-branches >>"' >> $BASH_ENV
+        name: Set ALLOW_ALL_BRANCHES flag
+        when: always
+  - run:
+      command: |
         app-toolbelt v0 notify mattermost --message "${MM_MESSAGE}" --channel << parameters.channel >> ${ALLOW_ALL_BRANCHES:+--allow-all-branches}
       name: Mattermost Status Alert
       when: always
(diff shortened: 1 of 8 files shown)


.circleci/src/commands/notify.yml
```diff
@@ -31,7 +31,10 @@ steps:
...
       command: |
         echo 'export ALLOW_ALL_BRANCHES="<< parameters.allow-all-branches >>"' >> $BASH_ENV
-        app-toolbelt v0 notify mattermost --message "${MM_MESSAGE}" --channel << parameters.channel >> ${ALLOW_ALL_BRANCHES:+--allow-all-branches}
-      name: Mattermost Status Alert
+      name: Set ALLOW_ALL_BRANCHES flag
       when: always
-
+  - run:
+      command: |
+        app-toolbelt v0 notify mattermost --message "<< parameters.success_message >>" --channel << parameters.channel >> <<#parameters.allow-all-branches>>--allow-all-branches<</parameters.allow-all-branches>>
```
f1sh1918, line 38: is this "#" correct here? `#parameters.allow-all-branches`

f1sh1918 (commented):
Looks fine to :)
Just one comment

.circleci/src/commands/notify.yml
ztefanie: yes it is correct, this is some circleci magic syntax to add the content between the both `<< >>` if the parameter is true.

f1sh1918: yes i know. but i'm talking about the additional `#` what does that mean?

f1sh1918 (approved):
👍 nice work :) thx for fixing

Issue #1333, opened 2024-03-05
Add activation route


f1sh1918:
**Is your feature request related to a problem? Please describe.**
When a user clicks on the activation link on the pdf or e-mail without having the particular app installed, a redirect to the login site will happen and probably confuse the user. Therefore we should add a hint that the app has to be downloaded

**Describe the solution you'd like**
- add an activation route
- show the download links for android and ios depending on the project (we may need this later for other projects that bavaria)
- add a "Activate Card" button opening the activation link, because the pdf is not opened anymore (links in pdf can not be targeted with new window)

**Describe alternatives you've considered**

**Additional context**

PR #1334, opened 2024-03-05
1333: Add activation route

f1sh1918:
### Short description

An activation page is needed to handle opening activation links without the app being installed on the smartphone.
Remind the user that the app has to be installed before the card can be activated.

### Proposed changes

- add an activation route
- add install instructions
- add a separate `ProjectSwitcher` component for dev issues (make it reusable)

Fixes: #1333 

#### a) dev environment
- create a card
- move the pdf on your device
- uninstall the app on the device (ehrenamt)
- for testing on dev environment you have to use the following linking scheme to test the route in your browser on mobile device .
`http://<networkip>:3000/activation/code#activationCode/` (use the activation code in the console)

There is no real good way to test it on dev environment unfortunately, because we need a `localhost/localNetworkIp` link to open the route locally. There also might be issues that `connection is lost` in the flutter app meanwhile you are in the browser and follow instructions. But this can only happen on dev environments (app in background).

#### b) staging
- create a card
- uninstall the app (ehrenamt)
- open the link in the pdf on smartphone


administration/src/mui-modules/activation/components/ActivationPageContent.tsx
```diff
@@ -0,0 +1,37 @@
...
+  if (!hash) {
+    return <InvalidActivationLink />
+  }
+
+  return (
+    <>
+      {activationText(config.name, downloadLink)}
+      <ActivationButton variant='outlined' color='primary' href={window.location.href}>
+        Aktiviere Karte
+      </ActivationButton>
```
michael-markl, line 32: Did you test whether this button actually works?
* Uninstall the app
* Load this activation page in the browser
* Install the app
* Go back to the browser
* Click on link

We should probably also test this in the three different browsers, i.e. Firefox, Chrome, Safari
I suspect that it might not work in every browser :/

michael-markl: Alternatively, we can also just remove this button and tell the user to go back and click the link again.

ztefanie (commented):
I tried to test this, but it did not work, or I did something wrong.

What I did:
1) Created a card
2) opened http://localhost:3000/activation/AtIEThFmwAN49h0rWeIwBcB1JoS7gTkn7zJ9UFp8pSU=
The string in the end, is the activation code that is shown below of the qr code on the pdf.
When the activation code does not contain a e.g. slash, i got this error message `Der von Ihnen geöffnete Link ist ungültig.
Der Link wurde fehlerhaft in den Browser übertragen. Versuchen Sie, den Link manuell aus der E-Mail oder dem PDF zu kopieren.`
when it does contain a slash, i was just redirected to the login page.

f1sh1918:
> I tried to test this, but it did not work, or I did something wrong.
> 
> What I did:
> 1) Created a card
> 2) opened http://localhost:3000/activation/AtIEThFmwAN49h0rWeIwBcB1JoS7gTkn7zJ9UFp8pSU=
> The string in the end, is the activation code that is shown below of the qr code on the pdf.
> When the activation code does not contain a e.g. slash, i got this error message `Der von Ihnen geöffnete Link ist ungültig.
> Der Link wurde fehlerhaft in den Browser übertragen. Versuchen Sie, den Link manuell aus der E-Mail oder dem PDF zu kopieren.`
> when it does contain a slash, i was just redirected to the login page.

Sorry I forgot to mention
This PR depends on #1324 the link Scheme has to be
`activation/code#activationCode and the activation Code has to be Uri encoded. Gonna add an example link in the description.

administration/src/project-configs/bayern/activationText.tsx
```diff
@@ -0,0 +1,19 @@
+import { Typography } from '@mui/material'
```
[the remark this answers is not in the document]
f1sh1918: Indeed

administration/src/mui-modules/activation/ActivationPage.tsx
```diff
@@ -0,0 +1,32 @@
+import { Card } from '@mui/material'
+import { styled } from '@mui/system'
+import React, { ReactElement, useContext } from 'react'
+
+import StandaloneCenter from '../../bp-modules/StandaloneCenter'
```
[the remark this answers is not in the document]
f1sh1918: BP modules are used for internal druckerei and are not mobile friendly. Mui modules are used for application process and are mobile friendly. I just use a template component here but could also create an additional one to avoid mixing. At some point we might refactor this to only have one component library but that's gonna be kind a huge effort.

administration/src/mui-modules/activation/components/ActivationPageContent.tsx
f1sh1918: Gonna test this. Why I added this link was because when you click on the qr code of the pdf it does not open new tab/window and the user has to go back after app installation. I mean we need this feature for Koblenz so it would be really bad if the 3 big browsers don't support it. I mean this is the main use case for DeepLinking

f1sh1918: Ah okay now I got the issue. This is a good point Koblenz. Gonna test this

f1sh1918: ok i tested this. In chrome in works out of the box but in other browsers you have to "allow open links with external apps" f.e. firefox. So i think you solution to reopen the pdf is the better way

administration/src/mui-modules/activation/ActivationPage.tsx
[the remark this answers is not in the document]
f1sh1918: to avoid mixing i just created a separate template component instead of using inheritance

administration/src/project-configs/bayern/activationText.tsx
```diff
@@ -0,0 +1,27 @@
...
+import React from 'react'
+
+export const ActivationText = (applicationName: string, downloadLink: string) => (
+  <div>
+    <Typography variant='h6' mb='8px'>
+      App nicht installiert
+    </Typography>
+    <ol>
+      <li>
+        Laden Sie sich die App <b>{applicationName}</b> im App- oder PlayStore auf Ihrem Smartphone herunter.
```
michael-markl, line 11:
```suggestion
        Laden Sie die App <b>{applicationName}</b> im App- oder PlayStore auf Ihrem Smartphone herunter.
```

f1sh1918:
@ztefanie i would appreciate a second review since michael did not test it. thx

ztefanie (commented):
I tested it on Android emulator.

I get this as the first screen wenn opening `http://localhost:3000/activation/code#ChQKBHRlc3QQoqMBGggKAggBIgIIABIQIuF4botoitZxNY86gioBKxoUJPaMJ8zLarOKBYaIDlfvjFLw0fU=`

on the emulator:
android emulator

After clicking on bayern, i see the screen i would have expected. But i think this first screen should not be displayed.

When opening in the browser (on my linux machine) it get:
First screen:
brwoser
second screen:
brwoser-2

f1sh1918:
This screen is dev only and we need it to decide which project config should be used because we have no proper URL. So this is fine @ztefanie

ztefanie (approved):

Issue #1321, opened 2024-02-27
Add function hexQRCodeFromProtobuf


sarahsporck:
**Is your feature request related to a problem? Please describe.**
For the csv export to datev we need to embedd the qrcode in a csv file. We found that a hex representation of the qrcode uses the least space in a string data format.

**Describe the solution you'd like**
A function that can convert a protobuf qrcode to a hex qrcode string. E.g.
1. generate protobuf qrcode 
2. convert to [qrcode matrix](https://github.com/digitalfabrik/entitlementcard/blob/03a0499a469610b16b215616d2cec7b5d93cb279/administration/src/util/qrcode.ts#L183)
3. convert "bitmap" to hexmap to avoid creating very large bitstrings

Step 2 -> 3 could look like this:
1110100100
1010011010
1110111100
0011110011
1011101001
0110001011
1100101011
1110110111
1011001101
1110010111

becomes 

3a4-29a-3bc-f3-2e9-18b-32b-3b7-2cd-397

using this conversion:
`matrix.map(row => parseInt(row, 2).toString(16)).join('-')`

Also make sure to write tests for this.

**Describe alternatives you've considered**

- do all of that in the backend

**Additional context**
We sent several proposals of encodings to the datev. This one was the most favored one from our side. The reason we don't send the "raw byte content" to the datev is because we use a custom qrcode format (see qrcode.ts) and we should not entrust an "outside" company to use this.
We could have also offered to provide a "package/module/script" that implements the qrcode conversion, but decided against that.

michael-markl:
Similarly important: QR codes can get as large as 177×177 (or larger?). This means a single row might have >64 bits, therefore **we need to use BigInt** for converting between binary and hex format of each row!

Thus, I suggest to use as encoding function:
```js
matrix.map(row => BigInt("0b" + row).toString(16)).join('-')
```

and as decoding:
```js
const rows = '3a4-29a-3bc-f3-2e9-18b-32b-3b7-2cd-397'.split('-') 
rows.map(hex => BigInt("0x" + hex).toString(2).padStart(rows.length, '0'))
```

ztefanie:
Thanks for adding this valuable hints. 

I still have some questions.

> We should make sure, that Datev uses the right inverse function of this encoding. The hex codes must be padded with zeros at the start when converting back to binary, like so:
> 
> ```js
> const rows = '3a4-29a-3bc-f3-2e9-18b-32b-3b7-2cd-397'.split('-') 
[...]

PR #1336, opened 2024-03-06
1321: Add convertProtobufToHexQRCode

ztefanie:
### Short description
Added convertProtobufToHexQRCode function

### Proposed changes
- get dynamicActivationCode
- get Matrix
- convert to HexMap

If i get this right, then we cannot create a DynamicActivationCode from the information in the database, but need cardInfo data, which we do not save. So I guess I did something wrong here, because we need to create the csv when requested from the data we have or will the csv be created, when a card is created and be saved somewhere? I am a bit lost with the application logic, so please let me know what the input of the function should be. Is `DynamicActivationCode` correct as the input?

### Resolved issues

Fixes: #1321


f1sh1918:
The dynamic activation code is the correct input

f1sh1918:
> ### Short description
> Added convertProtobufToHexQRCode function
> 
> ### Proposed changes
> * get dynamicActivationCode
> * get Matrix
> * convert to HexMap
> 
> If i get this right, then we cannot create a DynamicActivationCode from the information in the database, but need cardInfo data, which we do not save. So I guess I did something wrong here, because we need to create the csv when requested from the data we have or will the csv be created, when a card is created and be saved somewhere? I am a bit lost with the application logic, so please let me know what the input of the function should be. Is `DynamicActivationCode` correct as the input?
> 
> ### Resolved issues
> Fixes: #1321

The `DynamicActivationCode` is the correct input for the function. You can not recreate ActivationCodes from database. There is no user related data stored.

I think the process should be:
1. ImportCardsController -> add there a separate button for csv export 
2. Use `generateCards` fkt from `useCardGenerator` there should be maybe a flag passed to not `generatePdf` with the codes but `exportCSV` and there you can call your fkts inside.

administration/src/util/qrcode.ts
```diff
@@ -246,3 +248,20 @@ export const drawQRCode = (
     size
   )
 }
+
+export const convertProtobufToHexCode = (protobufQRCode: DynamicActivationCode): string => {
```
f1sh1918, line 252: we also need the `StaticVerificationCode` to be converted because it will also be printed on the pdf by DATEV

f1sh1918 (commented):
Added some small hints

administration/src/util/qrcode.ts
```diff
@@ -246,3 +248,20 @@ export const drawQRCode = (
...
+    },
+  }).toBinary()
+  const qrCodeMatrix = encodeQRCode(qrCode).getMatrix()
+  return convertBitmapToHexmap(qrCodeMatrix)
+}
+
+const convertBitmapToHexmap = (matrix: QRCodeByteMatrix): string =>
+  matrix
+    .getArray()
+    .map(row => parseInt(row.join(''), 2).toString(16))
```
michael-markl, line 266: See https://github.com/digitalfabrik/entitlementcard/issues/1321#issuecomment-1981174652

```suggestion
    .map(row => BigInt("0b" + row.join('')).toString(16))
```

administration/src/util/qrcode.ts
```diff
@@ -246,3 +248,21 @@ export const drawQRCode = (
...
+    .map(row => BigInt(`0b${row.join('')}`).toString(16))
+    .join('-')
+}
+
+export const convertHexmapToUInt8Array = (hexmap: string): Uint8Array[] => {
+  const zeroASCIICode = 48
+  const hexRows = hexmap.split('-')
+  const binaryStringMatrix = hexRows.map(hex => BigInt(`0x${hex}`).toString(2).padStart(hexRows.length, '0'))
+  const binaryUInt8Matrix = binaryStringMatrix.map(stringRow =>
+    Uint8Array.from(stringRow.split(''), x => x.charCodeAt(0) - zeroASCIICode)
```
f1sh1918, line 265: can you explain to me why we need this zeroAsciiCode stuff?
What is that better than just converting the string to an int?
`Uint8Array.from(stringRow.split(''), x => Number(x))`

f1sh1918 (commented):
👍 Nice implementation of this complex topic. Took some time for me to understand what and why we do this. Just added one thing i couldn't understand

administration/src/util/qrcode.ts
ztefanie: This also works and is a lot nicer. Thanks for the hint, i adjusted it.

PR #1342, opened 2024-03-11
Fix-ci-environment-variable

f1sh1918:
### Short description

The environment variable for production builds was not passed correctly to android and ios in the ci, so the environment fallback `showcase` was used instead production. That caused loading wrong stores

### Proposed changes

- remove quotes for base64 string for android
- pass the environment variable via xcargs for ios

### Testing
- check uploaded artifacts in browserstack (ipa testing)
- or get the artifacts here:
[Android](https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/4720/workflows/4788f1e6-f79b-4195-831b-967c36ec4123/jobs/35992/artifacts)
[iOS](https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/4738/workflows/c13863ab-e88b-4a34-9188-bd5b4fefa292/jobs/36053/artifacts)
- now the correct stores should be displayed in SearchList and Map

### DIFF
--- a/frontend/ios/fastlane/Fastfile
+++ b/frontend/ios/fastlane/Fastfile
@@ -85,7 +85,6 @@ platform :ios do
     # DO NOT OVERWRITE THE BUNDLE_CONFIG ENV VARIABLE! It is used by ruby bundle.
 
     sh("fvm flutter pub run build_runner build --define df_build_config=name=#{build_config_name}")
-    env_prod = sh("echo -n environment=production | base64")
 
     build_app(
        workspace: "Runner.xcworkspace",
@@ -93,7 +92,8 @@ platform :ios do
        output_directory: "#{ENV['HOME']}",
        output_name: "#{build_config_name}.ipa",
        export_method: "app-store",
-       xcargs: "DART_DEFINES='#{env_prod}'"
+       # pass base64 encoded env variable "environment=production" directly, because shell result does not work for gym as it does for gradle (android/Fastfile)
+       xcargs: "DART_DEFINES='ZW52aXJvbm1lbnQ9cHJvZHVjdGlvbg=='"
      )
   end
 
--- a/frontend/ios/fastlane/Fastfile
+++ b/frontend/ios/fastlane/Fastfile
@@ -85,6 +85,7 @@ platform :ios do
     # DO NOT OVERWRITE THE BUNDLE_CONFIG ENV VARIABLE! It is used by ruby bundle.
 
     sh("fvm flutter pub run build_runner build --define df_build_config=name=#{build_config_name}")
+    env_prod = sh("echo -n environment=production | base64")
 
     build_app(
        workspace: "Runner.xcworkspace",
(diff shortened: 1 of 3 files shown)


michael-markl (commented):
The apk seems to work, the IPA still has the same problem

michael-markl (commented):
The APK worked, but the IPA did not (tested on browser stack).

frontend/ios/fastlane/Fastfile
```diff
@@ -85,13 +85,15 @@ platform :ios do
...
     sh("fvm flutter pub run build_runner build --define df_build_config=name=#{build_config_name}")
+    env_prod = sh("echo -n environment=production | base64")
 
     build_app(
        workspace: "Runner.xcworkspace",
        scheme: flavor,
        output_directory: "#{ENV['HOME']}",
        output_name: "#{build_config_name}.ipa",
        export_method: "app-store",
+       xcargs: "DART_DEFINES='#{env_prod}'"
```
michael-markl, line 96: Maybe also remove the single-quotes? (Just guessing)
```suggestion
       xcargs: "DART_DEFINES=#{env_prod}"
```

f1sh1918: I tested this before. Didn't work. Thx for the hint with browserstack

f1sh1918: @michael-markl i had to pass the base64 string directly since somehow the result of this env_prod is wrong (LW4gZW52aXJvbm1lbnQ9cHJvZHVjdGlvbgo=)
Just decoded it and realized that it takes this for input `-n environment=production`

frontend/ios/fastlane/Fastfile
```diff
@@ -92,6 +92,8 @@ platform :ios do
        output_directory: "#{ENV['HOME']}",
        output_name: "#{build_config_name}.ipa",
        export_method: "app-store",
+       # pass base64 encoded env variable "environment=production" directly, because shell result does not work for gym as it does for gradle (android/Fastfile)
```
ztefanie, line 95: I think this is not correct, we did successful releases with passing vars like this.

frontend/ios/fastlane/Fastfile
```diff
@@ -92,6 +92,8 @@ platform :ios do
...
+       xcargs: "DART_DEFINES='ZW52aXJvbm1lbnQ9cHJvZHVjdGlvbg=='"
```
ztefanie, line 96: can you please put this base64 string into a named variable so we know what is happening here.

ztefanie (changes requested):
Can you please change the default, when the variable is not set properly to production, so if something is not working we have errors on dev not prod.

frontend/ios/fastlane/Fastfile
f1sh1918: You can test the ipa without and it won't work so no idea what happened meanwhile. I mean why we set this for gradle and not for gym?

frontend/ios/fastlane/Fastfile
ztefanie: I think it would be better to just adjust the command to handle the newline chars correctly. I think `echo environment=production | base64 | tr -d '\n'` should work.

ztefanie (approved):

Issue #1344, opened 2024-03-12
Add "Erklärung zur Barrierefreiheit" to Nuremberg-Pass and Ehrenamtskarte


dkehne:
**Is your feature request related to a problem? Please describe.**
Due to German law we need to add a declaration of accessibility to our apps. We can add the same declaration for both apps and it can be found here: https://berechtigungskarte.app/barrierefreiheit/

**Describe the solution you'd like**
Link it the same way as we link the data privacy declaration. Maybe we add it above software libraries. An icon has to be identified as well.

PR #1345, opened 2024-03-13
1344: Add accessibility policy

ztefanie:
### Short description
Added accessibility policy

### Resolved issues

Fixes: #1344

### DIFF
--- a/frontend/lib/about/about_page.dart
+++ b/frontend/lib/about/about_page.dart
@@ -123,7 +123,7 @@ class AboutPageState extends State<AboutPage> {
                 onTap: () => launchUrlString(buildConfig.dataPrivacyPolicyUrl, mode: LaunchMode.externalApplication),
               ),
               ListTile(
-                leading: const Icon(Icons.privacy_tip_outlined),
+                leading: const Icon(Icons.accessibility_new_outlined),
                 title: Text(t.about.accessibilityPolicy),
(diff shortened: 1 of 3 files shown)


f1sh1918:
I have no clue why the encodeUriComponent results in different strings sometimes for the DeepLink tests.
Debugged that and looks super random.
For now you can just deactivate the tests and create an issue.
I already spent lots of time on that before and the issue seemed to be instantly gone but looks like it isn't.
In one ci run the string is ok in the other one char changes.

frontend/lib/about/about_page.dart
```diff
@@ -117,9 +119,14 @@ class AboutPageState extends State<AboutPage> {
               ContentTile(icon: Icons.copyright, title: t.about.licenses(n: 1), children: getCopyrightText(context)),
               ListTile(
                 leading: const Icon(Icons.privacy_tip_outlined),
-                title: Text(t.about.privacyDeclaration),
+                title: Text(t.about.privacyPolicy),
                 onTap: () => launchUrlString(buildConfig.dataPrivacyPolicyUrl, mode: LaunchMode.externalApplication),
               ),
+              ListTile(
+                leading: const Icon(Icons.privacy_tip_outlined),
```
f1sh1918, line 126:
```suggestion
                leading: const Icon(Icons.accessibility_new_outlined),
```
i think its nicer to use a separate icon

f1sh1918 (approved):
👍 works nice, just think an appropriate icon would be good
Tested on ios

Issue #1346, opened 2024-03-13
DeepLink tests generate different activationCodes


f1sh1918:
**Describe the bug**
On different run times the deepLink tests generate different activationCodes

**To Reproduce**
1. Run test
2. Check activationCode
3. use different date
4. Get different activation

**Expected behavior**
The tests should check a particular string to particular timestamp to get same result

PR #1347, opened 2024-03-13
1346: Fix deep link test

f1sh1918:
### Short description

Activation codes in this test depend on timestamp, because the default expiry date will be calculated. So different runtime generate different activationCodes. This shouldn't be the case

### Proposed changes

- add fix timestamp for tests

Fixes: #1346


ztefanie (approved):
Thanks for fixing this :+1:

PR #1360, opened 2024-03-26
Update the installation guide

seluianova:
### Short description
Add small clarifications to the installation guide that will hopefully be helpful to "first-time-with-flutter" (and "first-time-with-protobuf") people

### Proposed changes
- add some detail
- use bullet points instead of nested numbered lists (1.1, 1.2, etc) because the latter are not supported in markdown

### Side effects
No (only docs affected)

### Resolved issues
Fixes: #

### DIFF
--- a/docs/development-setup.md
+++ b/docs/development-setup.md
@@ -1,21 +1,27 @@
 # Prerequisite
 
-Make sure you have installed [protoc](https://github.com/protocolbuffers/protobuf) on a current version.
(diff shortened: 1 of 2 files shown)


f1sh1918 (approved):
👍 thx for the detailed update of the documentation

ztefanie (approved):
:+1: Thanks for updating this :)

PR #1364, opened 2024-03-27
1211: Send DeepLinks via mail

f1sh1918:
### Short description

As a user i would like to receive a confirmation mail if my card was created by a service user to be informed about the current status. 

### Proposed changes

- add confirmation mail to user if card was created via application and a mail address was provided
- confirmation mail includes deep link
- pass mail address to `MailNotSendError`
- send mail as multipart to ensure better mobile mail client compatibility (created separate issue for adding html mail templates to other messages https://github.com/digitalfabrik/entitlementcard/issues/1374)

### Side effects

none

### Resolved issues

Fixes: #1211

### Testing
- go to applications and create a card (ensure the applicant has a mail address you can access) 
- check your mails
- open the mail on your smartphone and click activation link (works only on mail clients and browsers that support external app links)
- ensure config-local.yml exists with local mail configuration (dev only)


administration/src/cards/extensions/EMailExtension.ts
```diff
@@ -0,0 +1,24 @@
...
+  causesInfiniteLifetime() {
+    return false
+  }
+  fromString(state: string): void {
+    this.state = state
+  }
+  toString(): string {
+    return this.state ?? ''
+  }
+  isValid() {
```
f1sh1918, line 19: not sure if we need another check. But this field is optional, so if a card will be created without mail address (not created from application) it should not block the card creation

administration/src/cards/sendCardConfirmationMail.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+const sendCardConfirmationMail = async (
+  client: ApolloClient<object>,
+  project: string,
+  recipientAddress: string,
+  recipientName: string,
+  deepLink: string
+): Promise<void> => {
+  const result = await client.mutate<
```
michael-markl, line 23: This might also throw. Maybe, we should wrap the error in SendCardConfirmationMail?
Also we should probably adjust the error dialog when this happens (in the useCardGenerator hook?)?

f1sh1918: The error dialog could be more general like `Etwas ist schief gegangen beim erstellen der Karte` in we want it more precise we have to check if `MailExtension` is provided.

Can you specify what you mean by wrap the error? Enclose the `client.mutate` with try catch?

michael-markl: Hmm... maybe what you did is already enough. What happens when the backend throws `MailNotSentException`? The user may want to know which email address this exception belongs to 🤔

specs/backend-api.graphql
```diff
@@ -130,6 +130,8 @@ type Mutation {
   editAdministrator(adminId: Int!, newEmail: String!, newRegionId: Int, newRole: Role!, project: String!): Boolean!
   "Reset the administrator's password"
   resetPassword(email: String!, newPassword: String!, passwordResetKey: String!, project: String!): Boolean!
+  "Sends a confirmation mail to the user when the card creation was successful"
+  sendCardCreationConfirmationMail(deepLink: String!, project: String!, recipientAddress: String!, recipientName: String!): Boolean!
```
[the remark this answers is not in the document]
f1sh1918: this is auto generated

f1sh1918:
> What is the config-local.yml file? How do I create it?

this file is used to create local overwrites of the config which will be taken if you run the config on your local dev environment. This file will not be commited. There you can configure your mail credentials for sending mails locally
You can just copy the config and override the parts you want to override f.e.
```
    smtp:
      host: mail.gmx.net
      port: 587
      username: your mail address
      password: your password
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -226,4 +227,17 @@ class CardMutationService {
...
+    fun sendCardCreationConfirmationMail(dfe: DataFetchingEnvironment, project: String, recipientAddress: String, recipientName: String, deepLink: String): Boolean {
+        val context = dfe.getContext<GraphQLContext>()
+        val jwtPayload = context.enforceSignedIn()
+        transaction { AdministratorEntity.findById(jwtPayload.adminId) ?: throw UnauthorizedException() }
+        val backendConfig = dfe.getContext<GraphQLContext>().backendConfiguration
+        val projectConfig =
+            context.backendConfiguration.projects.find { it.id == project }
+                ?: throw ProjectNotFoundException(project)
+        Mailer.sendCardCreationConfirmationMail(backendConfig, projectConfig, deepLink, recipientAddress, recipientName)
+        return true
```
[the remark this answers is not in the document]
f1sh1918: as far as i read void return type goes against graphQL best practises and you would have to define a void scalar. 
If i remove boolean return type i can not update the graphql schema.
https://stackoverflow.com/a/61714123
Dunno if the true lleads to a status code tbh

administration/src/cards/sendCardConfirmationMail.ts
f1sh1918: hm i didn't know why i couldn't properly catch errors with this custom function but i adjusted it to use the standard mutation hook and now it is catched. added the error handling and toasts there. I set the error toast to timeout 0 and show the mail address, so the user has to close it manually but also has sufficient time to check the mail address. Not sure if thats fine for you

f1sh1918:
I still have some issue that the link is not properly recognized as a link in my mobile mail client :/
Just `https://staging.bayern` is marked as hyperlink the rest not :/

michael-markl:
@f1sh1918 Which mail client do you use?

f1sh1918:
> @f1sh1918 Which mail client do you use?

i use BlueMail on android but i'm sure there will be more mobile mail clients that are not capable of it

f1sh1918:
Hm just realized that some mail clients Display the HTML template quite small. I probably have to check how I can improve the configuration

michael-markl:
> Hm just realized that some mail clients Display the HTML template quite small. I probably have to check how I can improve the configuration

Which clients? It seems GitHub uses the same minimalistic approach in their email HTML texts, so we should probably be fine.

michael-markl (commented):
I guess, we should also print the email address somewhere on the Card Creation page.
Maybe it makes sense to opt out of sending these emails automatically.

Imo, it is not yet clear at all (from perspective of the administrator) that an email will be sent to the applicant (especially as this was not the case so far).

michael-markl:
However, if we build a feature flag and enable this only for dev / staging, this PR could be merged, and we move the remaining tasks to other issues to avoid blowing this PR up?

f1sh1918:
ok i just talked to daniel and we add a feature flag until he talked to bavaria in june

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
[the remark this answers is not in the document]
michael-markl: Yeah, we used to return a Boolean flag (which I believe was usually always `true` (never `false`) as nothing or `null` is returned in the error case).
Maybe that was not the best design decision from an API evolution standpoint; maybe it is better to always define an "object" type for every query, so that you can enrich the result of the API endpoint with other data without having to define a whole new query (and deprecating the old one). (that is why we have some graphql endpoints named `someQueryV2`).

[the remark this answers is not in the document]
f1sh1918: if we want to change this you can create a separate issue for this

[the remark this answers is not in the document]
michael-markl: Yep, but for the administration endpoints, it's not really relevant, as we can break backwards compatibility for those...

administration/src/errors/DefaultErrorMap.tsx
```diff
@@ -24,7 +24,7 @@ const defaultErrorMap = (extensions?: ErrorExtensions): GraphQLErrorMessage => {
   const defaultError = { title: 'Etwas ist schief gelaufen.' }
 
   if (!extensions || !extensions['code']) return defaultError
-
+  console.log(extensions['code'])
```
ztefanie, line 27: is this intended?

administration/src/project-configs/bayern/config.ts
```diff
@@ -64,6 +73,8 @@ const config: ProjectConfig = {
   },
   timezone: 'Europe/Berlin',
   pdf: pdfConfiguration,
+  // Wait for bavarian confirmation until we enable it on production
```
ztefanie, line 76: Please add a TODO at the beginning of this comment.

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -55,12 +56,12 @@ object Mailer {
                         .from(fromName, smtpConfig.username)
                         .withSubject(subject)
                         .withPlainText(message)
-                        .withHeader("Content-Type", "text/plain; charset=UTF-8")
+                        .withHTMLText(messageHTML)
```
ztefanie, line 59: I think this header should not be removed, is it moved somewhere or have you deleted it?

ztefanie (approved):
Tested on Android emulator, works as expected.

Will leave an approve, but highly recommend you readd the header with the encoding, because we will get responses about malformatted emails, if we do not set the encoding properly.

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
f1sh1918: the header will be calculated automatically depending on what you send and which charset your backend uses.

michael-markl: I believe a suitable header is already added automatically by the email library

michael-markl: But it doesn't hurt to double check :)

f1sh1918:
> Tested on Android emulator, works as expected.
> 
> Will leave an approve, but highly recommend you readd the header with the encoding, because we will get responses about malformatted emails, if we do not set the encoding properly.

When you check the raw mails you gonna see that it uses UTF-8 even without the header

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
f1sh1918: yes i checked it.

Issue #1206, opened 2023-11-14
Notes for applications


f1sh1918:
As a service user i want to be able to create and edit a note on an application to save additional information about the progress of the application

## Acceptance Criteria:
 - A service user i can create, edit and view note by clicking a button on the application form
 - the note will be deleted with the application

## Technical Tasks:
 - [ ] implement a CTA button to open a dialog where the user can create/update a note
 - [ ] implement the note form with texts and buttons.
 - [ ] maybe it makes sense to limit the amount of chars

hauf-toni:
Design Proposal can be found [📌 here](https://www.figma.com/file/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?type=design&node-id=290%3A47&mode=design&t=IBS2a6MVaGwSntJl-1)

hauf-toni:
@f1sh1918 V2 (show note when application is opened) can be found in the Figma-File linked above

PR #1370, opened 2024-04-02
1206 Notes for applications

f1sh1918:
### Short description

As a service user i might want to add notes to the current application, f.e. how often the applicant was contacted or sth similar.

### Proposed changes

- add `updateNote` fkt in the backend
- adjust database schema and ApplicationView for applications
- add database migration for applications table
- add reusable `TextAreaDialog` component
- add `NotesDialogController` for data handling
- fix loading spinner position to avoid container jumping while loading data
- limited notes to 1000 chars
- show quick note by hovering "Note anzeigen"
- add test

### Resolved issues

Fixes: #1206

### Testing
- run migrate db (dev only)
- open applications
- select an application
- click on 'Notiz anzeigen'
- add text and save -> text should be saved on open
- add text and close without saving (or using backdrop, click on overlay to close) -> text updates shouldn't be saved
- reach the character limit and check if note can be saved
- test if tooltip appears

Note: I slightly adjusted the design, since Button colors were wrong in the design and used the standard `DialogFooter` from blueprint


backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -128,4 +129,29 @@ class EakApplicationMutationService {
             }
         }
     }
+
+    @GraphQLDescription("Updates a note of an application")
+    fun updateApplicationNote(
+        applicationId: Int,
+        noteText: String,
```
michael-markl, line 136: I think we should put *some* restriction on `noteText`, e.g. at most 1000 characters or so.

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -130,79 +143,93 @@ const ApplicationCard = ({
...
+            loading={loading}
+            onCancel={() => setDeleteDialogOpen(false)}
+            onConfirm={() => deleteApplication({ variables: { applicationId: application.id } })}>
+            <p>Möchten Sie den Antrag unwiderruflich löschen?</p>
+          </Alert>
+        </SectionCard>
+      </ApplicationViewCard>
+      {openNoteDialog && (
+        <NoteDialogController application={application} isOpen={openNoteDialog} onOpenNoteDialog={setOpenNoteDialog} />
+      )}
```
michael-markl, line 231: Is it possible to move this inside the ApplicationViewCard? Then we don't need the fragment (which brings a pretty bad diff in this PR + more indents)

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -130,79 +143,93 @@ const ApplicationCard = ({
...
+        elevation={1}
+        icon={withdrawalDate ? <Icon icon='warning-sign' intent='warning' /> : undefined}
+        collapseProps={{ isOpen: isExpanded, onToggle: () => setIsExpanded(!isExpanded), keepChildrenMounted: true }}
+        collapsible
+        $hideInPrintMode={!isSelectedForPrint}>
+        <SectionCard>
+          <SectionCardHeader>
+            <NoteButton onClick={() => setOpenNoteDialog(true)} intent='none' icon='annotation'>
+              Notiz anzeigen
+            </NoteButton>
```
michael-markl, line 168: Would it make sense to show some excerpt of the note here too  so that the adminsitrators don't have to click a button before they can see the note? I imagine that could be useful.

administration/src/bp-modules/applications/NoteDialogController.tsx
```diff
@@ -0,0 +1,50 @@
...
+      const { title } = getMessageFromApolloError(error)
+      appToaster?.show({ intent: 'danger', message: title })
+    },
+    onCompleted: () => appToaster?.show({ intent: 'success', message: 'Notiz erfolgreich geändert.', timeout: 2000 }),
+  })
+  const applicationQuery = useGetApplicationByIdQuery({
+    variables: { applicationId: application.id },
+  })
+  const applicationQueryHandler = getQueryResult(applicationQuery)
+  if (!applicationQueryHandler.successful) return applicationQueryHandler.component
```
michael-markl, line 29: Isn't the note already on the application prop so we don't actually need to run another query?

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -128,4 +129,29 @@ class EakApplicationMutationService {
...
+
+        return transaction {
+            val application = ApplicationEntity.findById(applicationId) ?: throw UnauthorizedException()
+            // We throw an UnauthorizedException here, as we do not know whether there was an application with id
+            // `applicationId` and whether this application was contained in the user's project & region.
+
+            val user = AdministratorEntity.findById(jwtPayload.adminId)
+                ?: throw UnauthorizedException()
+
+            if (!mayViewApplicationsInRegion(user, application.regionId.value)) {
```
michael-markl, line 150: we should probably use/add a function called `mayUpdateApplicationsInRegion` or so

administration/src/bp-modules/applications/NoteDialogController.tsx
f1sh1918: yes but if its gonna be updated by the user, we would have to update all applications to keep it in sync. I thinks this should be avoided

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
f1sh1918: ok but the roles are the same for that so i thought its redundant to add a second function

administration/src/bp-modules/applications/ApplicationCard.tsx
f1sh1918: this make sense but was no requirement.

f1sh1918: check last commit

administration/src/bp-modules/applications/NoteDialogController.tsx
f1sh1918: But probably i should block the fetching somehow until its called by the `onSave` method. 
Have to check how i can do this to avoid a first unneeded call or do you have a better solution @michael-markl ?

f1sh1918: ok i realized there are lazy query hooks that seem to be suitable

administration/src/bp-modules/components/TextAreaDialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+      </Tooltip>
+    </ButtonContainer>
+  )
+
+  return (
+    <Dialog isOpen={isOpen} onClose={onClose}>
+      <TextArea
+        fill={true}
+        onChange={e => setText(e.target.value)}
+        value={text}
```
michael-markl, line 66:
```suggestion
        value={text}
        readOnly={loading}
```

administration/src/bp-modules/applications/NoteDialogController.tsx
```diff
@@ -0,0 +1,83 @@
...
+    onCompleted: () => {
+      appToaster?.show({ intent: 'success', message: 'Notiz erfolgreich geändert.', timeout: 2000 })
+    },
+  })
+
+  const onClose = () => onOpenNoteDialog(false)
+
+  const onSave = (text: string) => {
+    updateApplicationNote({ variables: { applicationId: application.id, text } })
+    onOpenNoteDialog(false)
```
michael-markl, line 51: Maybe this `onOpenNoteDialog(false)` should be moved to the `onCompleted` callback? Probably we don't want to close the dialog if an error occurs, so that the user can save his input?

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -128,4 +131,32 @@ class EakApplicationMutationService {
...
+            }
+            val user = AdministratorEntity.findById(jwtPayload.adminId)
+                ?: throw UnauthorizedException()
+
+            if (!mayUpdateApplicationsInRegion(user, application.regionId.value)) {
+                throw ForbiddenException()
+            }
+
+            ApplicationRepository.updateApplicationNote(applicationId, noteText)
+            return@transaction noteText
```
michael-markl, line 159: It seems quite hacky to me that we return the note again.

administration/src/bp-modules/applications/NoteDialogController.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  const getNoteExcerpt = (maxChars: number, text: string): string =>
+    text.length > maxChars ? `${text.slice(0, maxChars)} ...` : text
+
+  const note = applicationData ? applicationData.applicationNote : application.note
```
michael-markl, line 57: So this is used to show only the updated note if the note was updated, right? :D

Hmm.. can't we update the original `application` object similarly to how we remove an application object on delete?

administration/src/bp-modules/applications/NoteDialogController.tsx
f1sh1918: i moved it

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
f1sh1918: alright i reverted it to true (success)

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -76,6 +76,9 @@ const ApplicationsOverview = (props: { applications: Application[] }) => {
                 key={application.id}
                 application={application}
                 gotDeleted={() => setUpdatedApplications(sortedApplications.filter(a => a !== application))}
+                gotChanged={application =>
```
ztefanie, line 79: "gotChanged" seems to be the name of a boolean. As this is a function, i would expect it to be named something like "onChange".
(Maybe you can also adjust the "gotDeleted" naming)

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -199,4 +199,22 @@ object ApplicationRepository {
                 .sortByKeys({ it.id.value }, ids)
         }
     }
+
```
ztefanie, line 202: So these Repository.kt files are businesslogic mixed with database access?

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -128,4 +131,31 @@ class EakApplicationMutationService {
...
+    fun updateApplicationNote(
+        applicationId: Int,
+        noteText: String,
+        dfe: DataFetchingEnvironment
+    ): Boolean {
+        val context = dfe.getContext<GraphQLContext>()
+        val jwtPayload = context.enforceSignedIn()
+
+        return transaction {
+            val application = ApplicationEntity.findById(applicationId) ?: throw UnauthorizedException()
```
ztefanie, line 145: Unauthorized ist not the correct exception here, the message of UnauthorizedExeption ist ""JWT expired, invalid or missing".

Definition for 401 is: "JWT expired, invalid or missing".

You should either use 403 Forbidden or (what I think is more appropriate here) a 404 (Not found). I think 404 is the better choice here, as we do not provide information to someone unauthorized to know if this resource exists or not.

ztefanie (approved):
Tested on firefox, works as expected :tada:

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
michael-markl: I think mostly access to DB / persistence layer. There should not be a lot of businesslogic

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
f1sh1918: Alright makes sense
Just a question to @michael-markl 
You've added this comment
```
val application = ApplicationEntity.findById(applicationId) ?: throw UnauthorizedException()
            // We throw an UnauthorizedException here, as we do not know whether there was an application with id
            // `applicationId` and whether this application was contained in the user's project & region.

```
Does this really make sense? I mean you can not delete an application if you were not able to display it. Sure you could inject any id you want to but in `deleteApplication` a NotFoundExceptions makes also more sense?!

michael-markl: I am also partially confused with this comment :D I agree with @ztefanie that either 403 or 404 seems correct, maybe 404 is preferred as it's clearer what happend, 403 might be in some scenario potentially safer (to prevent scraping of active application ids crossproject/crossregion) but at the end it does not really matter probably :D

Issue #1184, opened 2023-10-20
Reactivate deleted card


f1sh1918:
**Is your feature request related to a problem? Please describe.**

Currently a card can only be activated and used on one device. If the activation code will be used on another device, the user gets a notification that the card will be deactivated on the other device. 
With the multi card implementation the user is able delete a card and reactivate.
In this case the user will also be notified that the card will be deactivated on the other device, which is not the case

**Describe the solution you'd like**
- set the `totpSecret` to null on card deletion
- The `firstActivationDate` shouldn't be overwritten since it may be used for statistics
- Show the already activated dialog only if the `totpSecret` is set

**Additional context**
Here is the discussion: https://github.com/digitalfabrik/entitlementcard/pull/1142#issuecomment-1772227549

f1sh1918:
@steffenkleinle maybe this can also labeled as a bug. 
Feel free to add suggestions for implementation

michael-markl:
I believe, instead of adding an `active` column to the cards table, we can also simply set the `totp-secret` to `NULL`. (It's also NULL before the first activation).
`firstActivationDate` should only be written if it has not been written yet, so this should not be a problem.

f1sh1918:
alright good point. i'm gonna adjust the description

michael-markl:
Its still unclear how to authorize the deletion. Probably using an otp code, but we should likely not use the same code as we use for verification.

f1sh1918:
#1266 might be good enough to not put more effort into this?

f1sh1918:
resolved with #1266

PR #1371, opened 2024-04-02
1266: Adjust reactivation message card active on another device

seluianova:
### Short description
Adjust reactivation message if the card has already been activated on another device.

Currently the user might be confused if the card was activated on another device and then deleted and activated on a new device that a message will be shown that the card will be deactivated on the other device.

### Proposed changes
I have adjusted the message suggested in the task to keep it a couple words shorter, but feel free to object.
image

### Side effects
- No?

### Resolved issues
Fixes: #1266


michael-markl (approved):
Nice clarification! Maybe we don't need #1184 with this change?

f1sh1918:
> Nice clarification! Maybe we don't need #1184 with this change?

I mean #1184 would be way cleaner but i don't see that the effort is really needed at this point so we can maybe put it on hold

PR #1372, opened 2024-04-02
Update migration docs

f1sh1918:
### Short description

If for any reason a db migration update does not run successfully after debian package installation (within post install), a devops user might want to run the database migration manually on the server.

### Proposed changes

- add section in migration docs with migration commands for server


ztefanie (approved):
:+1: Thanks for adding.

Issue #1340, opened 2024-03-11
PDF app download link not clickable


f1sh1918:
**Describe the bug**
In some pdf viewers f.e. firefox pdf plugin the app download link is not clickable, since not all pdf viewers recognize urls automatically. 

**To Reproduce**
1. Use firefox
2. https://staging.bayern.ehrenamtskarte.app/
3. Create a card
4. Open the pdf with the firefox build in pdf viewer
5. Try to click the app download link

**Expected behavior**
A click on the download link should open the download page

**Additional context**
Currently the app download link only works in chrome pdf viewer and adobe acrobat.
In the pdf template a link annotation has to be created to get it work in all pdf viewer

hauf-toni:
PDF with Hyperlink can be found [here](https://nextcloud.tuerantuer.org/index.php/apps/files/files/4496444?dir=/Projects/Digitale%20Berechtigungskarte/Sozialministerium%20Bayern/Releasedokumente).

seluianova:
Do we need this change for Nürnberg as well?

PR #1373, opened 2024-04-02
1340: Add a clickable link to app download to the PDF template

seluianova:
### Short description
In some pdf viewers f.e. firefox pdf plugin the app download link is not clickable, since not all pdf viewers recognize urls automatically.

### Proposed changes
- replace the pdf template with one with a clickable link

### Side effects
- no?

### Resolved issues
Fixes: #1340


f1sh1918 (approved):
👍 works nice

PR #1375, opened 2024-04-04
Upgrade fvm to 3.x

michael-markl:
### Short description

Upgrades the flutter version manager configuration files from 2.x to 3.x,

### Side effects

Likely, all developers (who installed fvm before Feb 09) need to upgrade [fvm](https://fvm.app) to version 3.x


ztefanie (approved):
Tested locally, app starts as expected.

f1sh1918 (approved):
tested on ios. works fine

Issue #1374, opened 2024-04-03
Provide html mail templates


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Some mobile mail clients can not properly resolve links inside mails. Therefore we should provide the mails additionally as html.

**Describe the solution you'd like**
- provide mail templates as html
- since the `Mailer.kt` is growing bigger it may make sense to create separate template files
- the solution has to be tested on several mail client

**Additional context**
Mattermost discussion: https://chat.tuerantuer.org/digitalfabrik/pl/qyj9p93fr788dksjizgw7945mh

PR #1378, opened 2024-04-06
1374: Add HTML templates for emails

michael-markl:
### Short description

I wrote a tiny templating wrapper around kotlinx.html.
This allows us to write only a single template for each email and then render a nice plain text message. from it

I added the css prop "word-break: break-word;" for links. @f1sh1918 let me know, if this fixed the issue on your mail client.

I'll have this PR as a Draft until #1211  is merged. Everything should work already, though.

### Side effects

Every email template was adjusted; I did not test each one, but at least most of them.

### Resolved issues

Fixes: #1374


f1sh1918:
@michael-markl the mail looks now fine. css fix works

f1sh1918 (approved):
Tested most of the mails. looked good in plain text and html
👍

Issue #1343, opened 2024-03-12
Add english store descriptions


dkehne:
We have received in new list for the Akzeptanzpartner: https://nextcloud.tuerantuer.org/index.php/f/6040252
Maybe the data need to be formatted before. For the first time we also received the english translations.

dkehne:
The englisch translations are only delivered/maintain for the description so we might just add the englisch text below the german text for now.

f1sh1918:
@dkehne should i add some filler text link?

// english version
<english description>

i really prefer some proper implementation (financed)

dkehne:
i will clarify that.

f1sh1918:
![image](https://github.com/digitalfabrik/entitlementcard/assets/37902063/c47e292c-7770-40aa-84ba-c9d88af8a9f8)

@dkehne

dkehne:
@f1sh1918 perfect

f1sh1918:
@dkehne we have to wait until next release with english description import until i can update the store data.
If we need them faster, i have to create a separate data import file that excludes the english translations but can be imported immediately

PR #1382, opened 2024-04-08
1343: Add english translations to store descriptions

f1sh1918:
### Short description

Since the proposal for store translations was not accepted by nuernberg, the english translations will be simply added with line breaks to the existing store description

### Proposed changes

- add additional column for csv import
- merge discounts for acceptance store
- split discounts for csv export
- adjusted documentation since we will get csv files instead excel sheets from nuernberg

### Side effects

- Not very happy with that implementation but since nuernberg doesn't want to pay for proper translation implementation, thats the simplest way to display them

### Resolved issues

Fixes: #1343 

### Notes

- open pr of the data repository can be found here:
https://github.com/digitalfabrik/sozialpass-nuernberg-data/pull/11

### Testing
- run "Import data" from backend runConfigs (dev only)
- check store descriptions in app

image


f1sh1918:
The release of this gonna be a little bit ugly. We have until this pr is released on staging and update the data repository with the new csv structure https://github.com/digitalfabrik/sozialpass-nuernberg-data/pull/11
Same for production

docs/sozialpass-stores-import.md
```diff
@@ -5,11 +5,11 @@ The source file will be provided in nextcloud by the city of nuernberg.
 This source file contains addresses which have to be converted to coordinates before they can be used for store import.
 The output file will be uploaded to the entitlement server.
 
-1. Enable `geocoding` and `csvWriter` in `/backend/src/resources/config`
-2. Save the recent Akzeptanzpartner excel file in csv format and put it into `/backend/src/resources/import`
+1. Enable `geocoding` and `csvWriter` in `/backend/src/resources/config-local.yml`
```
michael-markl, line 8:
```suggestion
1. Enable `geocoding` and `csvWriter` in `/backend/src/resources/config.local.yml`
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/nuernberg/utils/WriteCsv.kt
```diff
@@ -29,12 +29,14 @@ fun writeCsvWithGeoInformation(input: List<AcceptingStore>) {
...
             "Website",
-            "Rabatt",
+            "RabattDE",
+            "RabattEN",
             "Kategorie"
         )
     )
 
     input.forEach { record ->
+        val discounts = splitDiscount(record.discount)
```
michael-markl, line 39: Alternatively, I would remove this split function and only populate the "DE" version below for now and add a comment. The `splitDiscount` function seems to hacky for me :D

```kt
// For now, we do not properly support multi-language discount descriptions.
// Hence, the English and German translations of the discount were merged into the `discount` field
// of the `AcceptingStore` object in the `MapFromCsv` step.
// Therefore, we do not populate the CSV column for the english version, as splitting it back up is unreliable.
```

f1sh1918: Hm but we provide the updated csv (added geoinformation) to the customer on nextcloud

michael-markl: Hmm, I see... So every quartal Nürnberg takes the CSV files that we upload to nextcloud, adjusts it, and sends it back to us?

f1sh1918:
closed in favor of #1397

Issue #1323, opened 2024-02-27
Export generated cards to CSV


sarahsporck:
**Is your feature request related to a problem? Please describe.**
It should be possible to export generated cards to CSVs so that Nürnberg can send this csv to an external print service provider, that can generate the pdf from this CSV.
For that an additional buttons should be added to the button bar in the create cards view.

**Describe the solution you'd like**
- add a feature flag for the csv export
- add a new button to the button bar "Export to CSV"
- build and generate CSV (in the frontend) after creating a card; if the csv export fails make sure that the created cards are properly deleted again
- download the csv (csvs created in the "Einzelne Karte erstellen" tab should be called "Pass-ID.csv“ and csvs created in the "Mehrere Karten erstellen tab" should be called "Pass-ID[0]mass.csv"

**Describe alternatives you've considered**
Generating the CSV in the backend. As generating the QRCode in the backend is quite the effort I decided against that. 

Blocked by 
- #1321
- #1322

Issue #1322, opened 2024-02-27
Export CardResultModel + CardBlueprint to CSV line


sarahsporck:
**Is your feature request related to a problem? Please describe.**
Nürnberg wants to have a CSV export of our created cards, so that an external print service provider (datev) can take care af properly printing the pdf.
For that we need to add a new functionality to export created Cards as CSV.
This issue should allow to create a valid csv line for a Card Blueprint.

**Describe the solution you'd like**
- add function that generate a valid CSV line for a `CreateCardsResult` and a `CardBlueprint` (it's probably quite easy to extend the cardextension and reuse existing logic from the csv import)
- the line should consist of the same fields as in the csv-import + 'cardInfoHashBase64' + hexQRCode
- [Example CSV with valid rows](https://github.com/digitalfabrik/entitlementcard/files/14421904/nuernberg-datev-print.csv)

Also add tests for this.

**Describe alternatives you've considered**
Doing the CSV generation in the backend. This seems to be a lot of work as the qr logic must move for this. Therefore I'd implement the whole feature in the frontend for now

**Additional context**
Technically this is blocked by #1321, but this can also be done in parallel and the hexQRCode is added later.

PR #1383, opened 2024-04-09
1322: Csv export

ztefanie:
### Short description
Added the CSV export feature for nuernberg

### Proposed changes
- Renamed PDFerror to PdfError, to unify it with other occurences of "pdf"
- Added CSV fucntion
(Tests are still missing, will add them soon PR)

### Side effects
- Refactored PDF creation a bit, but should not affect the outcome.

### For testing
You can test the generated hex codes e.g. with this tool: https://bahamas10.github.io/binary-to-qrcode/
You need to remove the "-" from the hexcode, there should be no spaces or newlines.
After creating the csv you should be able to scan this code in the app successfully.

This feature is only available for nuernberg, the button for the csv export should only be visible for nuernberg, not for bayern. It should be possible to:
1) Export a single card
a) created via the form
b) created via the deep link (https://nuernberg.sozialpass.app/cards/add?Name=Thea+Test&Ablaufdatum=03.3.2026&Geburtsdatum=01.01.2000&Pass-ID=123&Adresszeile+1=Teststraße+3)
2) export multiple cards that were imported via csv.

(if you are a dev and lazy you can create a valid binary code with this snippet:
```
      const hexstrin = '... what you get from the csv'
      console.log(
        convertHexmapToUInt8Array(hexstring)
          .map(row => row.join(''))
          .join('\n')
      )
```

### Resolved issues

Fixes: #1322
Fixes: #1323


f1sh1918 (commented):
Looks quite nice. I left some comments for improvements.

i tried to scan the qr codes and even added a border and couldn't get them scanned properly. Probably because its the wrong qrCode type?!

What we use
Bildschirmfoto 2024-04-09 um 15 40 14

What i get with the generator

export_qr_code_test

Did you test scanning the qr codes in the app?
I probably could need some help to get correct qr codes

administration/src/cards/CsvFactory.ts
```diff
@@ -0,0 +1,67 @@
...
+import { convertProtobufToHexCode } from '../util/qrcode'
+import CardBlueprint from './CardBlueprint'
+import { CreateCardsResult } from './createCards'
+import AddressExtensions from './extensions/AddressFieldExtensions'
+import BirthdayExtension from './extensions/BirthdayExtension'
+import NuernbergPassIdExtension from './extensions/NuernbergPassIdExtension'
+import StartDayExtension from './extensions/StartDayExtension'
+import { findExtension } from './extensions/extensions'
+
+export class CsvError extends Error {
```
f1sh1918, line 11:
```suggestion
class CsvError extends Error {
```

f1sh1918, line 11: hm if i throw somewhere an csv error it should be handled it the `useCardGenerator` i think and show a proper toast. Currently if i click on csv export and an error occurs here i just get
`Etwas ist schiefgegangen beim erstellen der PDF.` even i didn't create the a pdf

administration/src/cards/CsvFactory.ts
```diff
@@ -0,0 +1,67 @@
...
+        new QrCode({
+          qrCode: {
+            case: 'staticVerificationCode',
+            value: createCardsResult.staticVerificationCode,
+          },
+        })
+      )
+    : ''
+
+  return `${cardBlueprint.fullName},${addressLine1},${addressLine2},${plz} ${location},${passId},${birthday},${startDay},${cardBlueprint.expirationDate},${createCardsResult.staticCardInfoHashBase64},${activationHex},${staticVerificationHex}\n`
```
michael-markl, line 58: I'd recommend using some library for generating the CSV contents to avoid doing all the nasty escaping etc. For example here a `,` (or even worse a newline) in any addressLine destroys the output.

michael-markl: Maybe this? https://csv.js.org/stringify/api/sync/

administration/src/cards/CsvFactory.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const getCSVFilename = (cardBlueprints: CardBlueprint[]): string => {
+  const filename =
+    cardBlueprints.length === 1
+      ? findExtension(cardBlueprints[0].extensions, NuernbergPassIdExtension)?.state?.passId
```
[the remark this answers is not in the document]
ztefanie: It is what Sarah defined in the issue, so I guess it was the explicit wish of nuernberg.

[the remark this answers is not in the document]
f1sh1918: Ok i understand. Sorry didn't ready this. But then I guess in both cases the Pass id should be used. So in the bulk The Pass id of the first array ítem. You took it as a hardcoded string

ztefanie:
> Looks quite nice. I left some comments for improvements.
> 
> i tried to scan the qr codes and even added a border and couldn't get them scanned properly. Probably because its the wrong qrCode type?!
> 
> What we use 
> 
> What i get with the generator
> 
> Did you test scanning the qr codes in the app? I probably could need some help to get correct qr codes


Yes, I tested this and it works as expected. Have you set the size correctly in the tool?

administration/src/cards/CsvFactory.ts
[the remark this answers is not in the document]
ztefanie: You think so? I am unsure what is really wanted here. I will ask Daniel, if he knows what is the desired output.

administration/src/cards/CsvFactory.ts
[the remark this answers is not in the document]
ztefanie: I do not get this comment. If I used the  nameColumnName, expiryColumnName, extensionColumnNames and add the extensions array, the order will not be the same, e.g. startDay will be in the wrong position. Also looping through the extensions will make the csv change if we add an extention, that should not change the csv, this will make the csv change without us to notice and cause errors for datev, who will expect always the same format.

f1sh1918:
> > Looks quite nice. I left some comments for improvements.
> > i tried to scan the qr codes and even added a border and couldn't get them scanned properly. Probably because its the wrong qrCode type?!
> > What we use
> > What i get with the generator
> > Did you test scanning the qr codes in the app? I probably could need some help to get correct qr codes
> 
> Yes, I tested this and it works as expected. Have you set the size correctly in the tool?

yes it put it to 49. Gonna try it again. The camera didn't recognized anything so i thought maybe borders are missing. Did you add sth to the qr code?

administration/src/cards/CsvFactory.ts
ztefanie: Good catch. Adjusted the error handling, should work now.

administration/src/cards/CsvFactory.ts
ztefanie: Good point, thanks for thinking of it :+1:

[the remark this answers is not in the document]
f1sh1918:
```js
extensionColumnNames: [
      'Startdatum',
      'Geburtsdatum',
      'Pass-ID',
      'Adresszeile 1',
      'Adresszeile 2',
      'PLZ',
      'Ort',
      null,
    ],
    defaultValidity: { years: 1 },
    extensions: [StartDayExtension, BirthdayExtension, NuernbergPassIdExtension, ...AddressExtensions, RegionExtension],
```
I'm just talking about the extensions. They should have the same orders.
The extension name should be the same as in the extensionColumnNames. If they shouldn't appear in the csv you can add `null` like for the `RegionExtension`

ztefanie:
> > > Looks quite nice. I left some comments for improvements.
> > > i tried to scan the qr codes and even added a border and couldn't get them scanned properly. Probably because its the wrong qrCode type?!
> > > What we use
> > > What i get with the generator
> > > Did you test scanning the qr codes in the app? I probably could need some help to get correct qr codes
> > 
> > 
> > Yes, I tested this and it works as expected. Have you set the size correctly in the tool?
> 
> yes it put it to 49. Gonna try it again. The camera didn't recognized anything so i thought maybe borders are missing. Did you add sth to the qr code?

No, this is the qr code I tested and that worked:
qr94

f1sh1918:
ok no idea how i get such a barcode with the description you wrote

image

ztefanie:
> ok no idea how i get such a qr code with the description you wrote
> 

Have you used the code snipped in the description. If you do not use this and do it on you own, you have to add zeros at the beginning. I build a jsfiddle for testing this: https://jsfiddle.net/ztefanie/934evpw8/2/

administration/src/cards/CsvFactory.ts
```diff
@@ -0,0 +1,67 @@
...
+  constructor(message: string) {
+    super(message)
+    this.name = 'CSVError'
+  }
+}
+
+export async function generateCsv(codes: CreateCardsResult[], cardBlueprints: CardBlueprint[]) {
+  try {
+    let csvContent =
+      'Name,AddressLine1,AddressLine2,AddressLocation,PassId,Birthday,StartDate,ExpirationDate,CardHash,ActivationCode,StaticUserCode\n'
```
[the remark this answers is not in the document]
f1sh1918: I have some additional suggestions. I think we should create a csvExportConfig in nuernberg/config and define there structure of the headline and a fkt that returns the content of a csv content line. Since this is very project specific and should be configurable differently for other upcoming projects if needed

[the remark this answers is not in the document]
ztefanie: Done

f1sh1918 (commented):
Looks good. Just checking the amount of columns for header and content would be good i think
Tested both qr codes and work fine 👍

administration/src/project-configs/nuernberg/__tests__/csvExport.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { DynamicActivationCode, StaticVerificationCode } from '../../../generated/card_pb'
+import { Region } from '../../../generated/graphql'
+import getProjectConfig from '../../getProjectConfig'
+import nuernbergConfig from '../config'
+import config from '../config'
+import { buildCsvLine } from '../csvExport'
+
+jest.mock('csv-stringify/browser/esm/sync', () => ({
+  stringify: (input: any[]) => input[0].join(',')
+}))
```
steffenkleinle, line 13: @ztefanie like this it works. I have no idea why wrapping it in `jest.fn()` does not work... apparently `jest.fn()` does not return what it should...

This should definitely work (but doesn't):
```suggestion
jest.mock('csv-stringify/browser/esm/sync', () => ({
  stringify: jest.fn((input: any[]) => input[0].join(','))
}))
```

If you need to change the implementation of the mock, you can also do:
```suggestion
jest.mock('csv-stringify/browser/esm/sync', () => ({
  stringify: jest.fn()
}))
```
and then further below:
```ts
mocked(stringify).mockImplementation((input: any[]) => input[0].join(',')) 
```

steffenkleinle: Why do we have to mock it in the first place? Because its an esm module?

administration/src/project-configs/nuernberg/config.ts
```diff
@@ -37,6 +38,12 @@ const config: ProjectConfig = {
   },
   pdf: pdfConfig,
   cardCreationConfirmationMailEnabled: false,
+  csvExport: {
+    enabled: true,
+    csvHeader:
```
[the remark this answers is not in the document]
ztefanie: Changed to an array and added a test for this.

administration/src/project-configs/nuernberg/__tests__/csvExport.test.tsx
ztefanie: Thanks @steffenkleinle 
Yes we need to mock it because it is an esm module.

administration/src/cards/CsvFactory.ts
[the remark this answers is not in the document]
f1sh1918: There is no update yet @dkehne ?

administration/src/cards/__tests__/CsvFactory.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import { CsvError, generateCsv, getCSVFilename } from '../CsvFactory'
+import { CreateCardsResult } from '../createCards'
+import NuernbergPassIdExtension from '../extensions/NuernbergPassIdExtension'
+import { findExtension } from '../extensions/extensions'
+
+jest.mock('csv-stringify/browser/esm/sync', () => ({
+  stringify: (input: any[]) => input[0].join(','),
+}))
+
+jest.mock('../../project-configs/showcase/config')
```
f1sh1918, line 14: why the showcase config is needed here?

f1sh1918 (approved):
👍 works nice :)

administration/src/cards/__tests__/CsvFactory.test.ts
ztefanie: It is loaded in the getProjectConfig.ts and i did not want to wrap the components with the ProjectConfigProvider, as this will fallback to showcase in tests and I needed to specify them.

PR #1385, opened 2024-04-09
Cleanup fkt backend

f1sh1918:
### Short description
Forgot to remove not needed function that i creating during #1206 
Also found a second function that is unused

### Proposed changes

- remove `findVerificationsByIds`
- remove `getApplicationById`


ztefanie (approved):
Not tested

PR #1390, opened 2024-04-11
1232: Create github release

### LINKED ISSUE #1165: Setup automatic release notes handling
Think of a good way to handle release notes. 

Options:
1) Automatic handling like in integreat can be done. We can add a field to which apps a release note belongs
2) We can not handle release notes automatically. So it needs to be done manually.

Other options?

### LINKED ISSUE #1232: Create github releases
When a release is done, a github release should be created.
The release should contain release notes and apk/ipa files

Adjust also, that for beta releases the "pre-release" tag for github release is set and unset when promotion is triggered.

### LINKED ISSUE #1319: Add success notify for releases
Add a success notify to mattermost for releases. The link to the github release should be there. 

blocked by: #1232 #1165

### LINKED ISSUE #1331: Add apks to github release
It should be easier for users to get our latest apk without using google play store. 

The apk should be added to the github release. Use octokit for adding it.

**Additional context**
We want a link that always links to the latest release for every app, e.g. `github.com/digitalfabrik/integreat-app/latest/integreat.apk`. See here: https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases
This link can be placed on e.g. the website.

Think about how to handle betas, they should not be latest, but get a "pre-release" tag or something, check if the linking would be still correct.

f1sh1918:
### Short description

As a dev i want that after successful delivery a github release with release notes will be created.

### Proposed changes

- add `notify_release` job to create a release, generate release notes, upload artifacts and send mattermost notification to release channel

### Resolved issues

https://github.com/digitalfabrik/app-toolbelt/pull/13 (merged)
resolves #1232, resolves #1319, resolves #1331, resolves #1165

### Testing (dev only)
1. Important deactivate delivery tasks in `deliver_beta_native`(!!!)
2. Push your commit
3. 
```sh
 curl -X POST https://circleci.com/api/v2/project/github/digitalfabrik/entitlementcard/pipeline \
  --header "Circle-Token: <"your token">" \
  --header "content-type: application/json" \
  --data '{"branch":"1232-create-github-release", "parameters":{"run_delivery_beta_native":true, "run_commit":false}}'

```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -818,7 +818,7 @@ jobs:
             - restore_environment_variables
             - install_app_toolbelt
             - run:
-                command: echo "export RELEASE_ID='$(app-toolbelt v0 release create all ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --production-release << parameters.production_delivery >> --should-use-predefined-release-notes false)'" >> ${BASH_ENV}
+                command: echo "export RELEASE_ID='$(app-toolbelt v0 release create all ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --production-release << parameters.production_delivery >>)'" >> ${BASH_ENV}
                 name: Create github release
             - run:
                 command: app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files "$(ls ~/attached_workspace/*.apk)" --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -805,6 +805,31 @@ jobs:
                         command: |
                             curl https://webhook.entitlementcard-test.tuerantuer.org/hooks/install-<< parameters.bundle >>-$EAK_WEBHOOK
                         name: Install package via webhook
+    notify_release:
+        docker:
+            - image: cimg/node:19.1.0
+        parameters:
+            production_delivery:
+                description: Whether builds are delivered to the production or beta lane of the play store.
+                type: boolean
+        resource_class: small
+        steps:
+            - prepare_workspace
+            - restore_environment_variables
+            - install_app_toolbelt
+            - run:
+                command: echo "export RELEASE_ID='$(app-toolbelt v0 release create all ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --production-release << parameters.production_delivery >> --should-use-predefined-release-notes false)'" >> ${BASH_ENV}
+                name: Create github release
+            - run:
+                command: app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files "$(ls ~/attached_workspace/*.apk)" --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
+                name: Upload android apks to github release
+            - run:
+                command: app-toolbelt v0 github-release-asset upload --releaseId ${RELEASE_ID} --files "$(ls ~/attached_workspace/*.ipa)" --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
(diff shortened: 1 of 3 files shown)


.circleci/src/jobs/notify_release.yml
```diff
@@ -0,0 +1,25 @@
...
+  - restore_environment_variables
+  - install_app_toolbelt
+  - run:
+      name: Create github release
+      command: echo "export RELEASE_ID='$(app-toolbelt v0 release create all ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --production-release << parameters.production_delivery >> --should-use-predefined-release-notes false)'" >> ${BASH_ENV}
+  - run:
+      name: Upload android apks to github release
+      command: app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files "$(ls ~/attached_workspace/*.apk)" --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
+  - run:
+      name: Upload ios ipas to github release
```
f1sh1918, line 20: not sure if we really need the ipas

ztefanie: probably not really needed. I think we can delete it, to save tokens, but not really relevant though

.circleci/src/jobs/notify_release.yml
```diff
@@ -0,0 +1,25 @@
...
+      command: app-toolbelt v0 github-release-asset upload --releaseId ${RELEASE_ID} --files "$(ls ~/attached_workspace/*.ipa)" --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
+  - notify:
+      success_message: <<^ parameters.production_delivery >>[Beta] <</ parameters.production_delivery >>Ehrenamtskarte Bayern und Sozialpass Nürnberg ${NEW_VERSION_NAME} have been released successfully! https://github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/releases/tag/${NEW_VERSION_NAME}-all
+      channel: releases
+      allow-all-branches: true
```
ztefanie, line 25: This should be set to false, as we only want to allow "release-*" and "main" branches.

.circleci/src/jobs/notify_release.yml
```diff
@@ -0,0 +1,25 @@
...
+docker:
+  - image: cimg/node:19.1.0
+resource_class: small
+steps:
+  - prepare_workspace
+  - restore_environment_variables
+  - install_app_toolbelt
+  - run:
+      name: Create github release
+      command: echo "export RELEASE_ID='$(app-toolbelt v0 release create all ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --production-release << parameters.production_delivery >> --should-use-predefined-release-notes false)'" >> ${BASH_ENV}
```
ztefanie, line 15: as this is a boolean, the convention is:
if it is set, it is true
if it is not set, it is false

But you would have to also adjust the toolbelt to fix this.

f1sh1918: okay i probably didn't get the problem. If we don't pass anything or false its fine for the beta use case.
If we pass true what will happen very unlikely that the `pre` will not set?
Can you explain me the issue?

ztefanie: The issue is to make a public interface more robust and following conventions.

f1sh1918: i switched to optional flags
https://github.com/digitalfabrik/app-toolbelt/pull/13
removed here the not needed flag `--should-use-predefined-release-notes false`

ztefanie (approved):
Nice PR. :+1:

.circleci/config.yml
```diff
@@ -805,6 +805,28 @@ jobs:
...
+                description: Whether builds are delivered to the production or beta lane of the play store.
+                type: boolean
+        resource_class: small
+        steps:
+            - prepare_workspace
+            - restore_environment_variables
+            - install_app_toolbelt
+            - run:
+                command: echo "export RELEASE_ID='$(app-toolbelt v0 release create all ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --production-release << parameters.production_delivery >>)'" >> ${BASH_ENV}
+                name: Create github release
```
ztefanie, line 822: You need to pass the "--production-release" flag properly as you adjusted it in the toolbelt

ztefanie (changes requested):

.circleci/config.yml
f1sh1918: hm i expected to be able to use it like this @ztefanie but looks like that its not valid in circle ci
```
<< parameters.production_delivery >>--production-release<</ parameters.production_delivery >>
```

ztefanie: You have to use the "#"-Syntax:

<<#parameters.production_delivery >>--production-release<</parameters.production_delivery >>

ztefanie (approved):

PR #1393, opened 2024-04-12
Unify issue template with Integreat and new labels

ztefanie:
### Short description
Just unified the templates a little with Integreat and adjusted to the new renamed labels.


f1sh1918 (approved):
👍 Looks good

Issue #1241, opened 2023-12-13
Mark juleica applications different


f1sh1918:
As a service user i may be confused about the displaying of organization status for juleica cards, that don't have organizations to verify. These applications are ready without any verifications and should be marked different

## Acceptance Criteria:
 - indicate the user that this application is ready to process

## Technical Tasks:
 - show the juleica chip and a green "ready" icon if the application has the type juleica

<img width="1056" alt="image" src="https://github.com/digitalfabrik/entitlementcard/assets/37902063/db34afe2-1427-4936-83e6-fd13c17aafd7">

PR #1394, opened 2024-04-14
1241: Mark juleica applications different

seluianova:
### Short description
As a service user i may be confused about the displaying of organization status for juleica cards, that don't have organizations to verify. These applications are ready without any verifications and should be marked different

### Proposed changes
- Show the juleica chip and a green "ready" icon if the application has the type juleica
image

### Side effects
No?

### Resolved issues
Fixes: #1241


administration/src/assets/juleica.png
ztefanie, line 1: Please change this to a ".svg". You can export this image from figma as a "svg".

administration/src/types/assets.d.ts
```diff
@@ -2,3 +2,8 @@ declare module '*.pdf' {
   const content: string
   export default content
 }
+
+declare module '*.png' {
```
ztefanie, line 6: please undo.

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -152,7 +158,12 @@ const ApplicationCard = ({
           )}
         </div>
       }
-      rightElement={<VerificationsQuickIndicator verifications={application.verifications} />}
+      rightElement={
```
ztefanie, line 161: I think it would make sense to introduce a new component
"JulicaVerificationQuickIndicator".
Then here you can render the correct component depending on "isJuleicaEntitlementType()"

administration/src/bp-modules/applications/VerificationsView.tsx
```diff
@@ -69,33 +76,64 @@ export const getStatus = (verification: Application['verifications'][number]) =>
...
-          />
-          <Indicator
-            status={VerificationStatus.Rejected}
-            text={verificationStati.filter(v => v === VerificationStatus.Rejected).length}
-          />
-        </UnFocusedDiv>
-      </Tooltip>
-    )
+    if (isJuleicaEntitlementType) {
+      return (
```
ztefanie, line 88: Please move this to the new "JulicaVerificationQuickIndicator" component.

ztefanie (changes requested):

ztefanie:
Also would be nice if you can add a test for this.

administration/src/bp-modules/applications/VerificationsView.tsx
```diff
@@ -16,6 +17,12 @@ const StyledIndicator = styled.span`
   padding: 4px;
 `
 
+const JuleicaIcon = styled.span`
```
f1sh1918, line 20: You can remove this span if you set the style on the parent.
Its better to set the height on the parent and to fit the children inside, so you could add any other assets without setting height to a particular value
Example:
```js
const UnFocusedDiv = styled.div`
  display: flex;
  align-items: center;
  flex: 1;
  :focus {
    outline: none;
  }
  height: 25px;
`
```

f1sh1918 (commented):
👍 nice implementation just left some comments to improve it

administration/src/types/assets.d.ts
seluianova: do you mean change it to svg, or there is a better way to import images?

administration/src/bp-modules/applications/VerificationsView.tsx
```diff
@@ -68,6 +73,24 @@ export const getStatus = (verification: Application['verifications'][number]) =>
   }
 }
 
+export const JulicaVerificationQuickIndicator = memo(() => {
```
f1sh1918, line 76: This should be put into a separate functional component and not exported here

administration/src/bp-modules/applications/VerificationsView.tsx
```diff
@@ -68,6 +73,24 @@ export const getStatus = (verification: Application['verifications'][number]) =>
...
+      }>
+      <UnFocusedDiv>
+        <Indicator status={VerificationStatus.Verified} />
+        <img src={JuleicaLogo} alt='juleica' height='100%' />
+      </UnFocusedDiv>
+    </Tooltip>
+  )
+})
+
 export const VerificationsQuickIndicator = memo(
```
f1sh1918, line 94: this can also be exported to a separate component. Sorry for all the refactoring issue!

administration/src/types/assets.d.ts
ztefanie: yes change to svg :+1:

administration/src/bp-modules/applications/VerificationsView.tsx
ztefanie: just to clarify: It should be put into a own file.

seluianova: I have pushed an update, please let me know if that's what you meant

administration/src/bp-modules/applications/VerificationsQuickIndicator.tsx
```diff
@@ -0,0 +1,47 @@
+import { Tooltip } from '@blueprintjs/core'
+import { memo } from 'react'
+import styled from 'styled-components'
+
+import { Application } from './ApplicationsOverview'
+import { Indicator, VerificationStatus, getStatus } from './VerificationsView'
+
+export const UnFocusedDiv = styled.div`
```
f1sh1918, line 8: (optional) this may be in a separate component file

ztefanie (approved):
Not tested again.
Thanks for addressing our comments :yellow_heart:

seluianova:
Tests creation moved into separate task https://github.com/digitalfabrik/entitlementcard/issues/1407

f1sh1918:
> Tests creation moved into separate task https://github.com/digitalfabrik/entitlementcard/issues/1407

@seluianova feel free to merge :)

PR #1397, opened 2024-04-15
Proposal: english store descriptions

michael-markl:
@f1sh1918  This is a slight overhaul of #1382 . Just a proposal with a slightly less hacky workaround (and some refactorings :grimacing:)

### Short description

Since the proposal for store translations was not accepted by nuernberg, the english translations will be simply added with line breaks to the existing store description

### Proposed changes

- add possibility to import "RabattEN" and "RabattDE" columns in csv import (if the old column "Rabatt" is present, then this column will be used for the german text, while the english text stays empty) 
- both translations are merged in the acceptance store pipeline
- For the csv export, the originally imported entries are used to write out the translations separately
- Refactored CSV import and export to use header names instead of hardcoded indices
- Remove unused ID column from written CSV

### Side effects

The backend should be deployed first, then the new data should be deployed (this PR is backward compatible).

### Resolved issues

Fixes: #1343 

### Notes

- open pr of the data repository can be found here:
https://github.com/digitalfabrik/sozialpass-nuernberg-data/pull/11

### Testing
- run "Import data" from backend runConfigs (dev only)
- check store descriptions in app


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/nuernberg/steps/DownloadCsv.kt
```diff
@@ -24,44 +23,53 @@ class DownloadCsv(config: ImportConfig, private val logger: Logger) :
...
+        parser.records.forEach { record ->
+            val name = record.get("Name")
+            if (name.isBlank()) {
+                throw IllegalArgumentException("CSV record " + record.recordNumber + " has an empty name.")
             }
+
+            val (discountDE, discountEN) =
+                if (parser.headerMap.containsKey("Rabatt")) {
+                    // Legacy format used until April 2024. Can be removed in future releases.
+                    Pair(record.get("Rabatt"), null)
```
f1sh1918, line 46: if i import it with the old csv format i get the `null` as a value in the database. It will be converted to a string

f1sh1918 (changes requested):
Nice refactoring 👍 
Unfortunately it breaks the old import format by adding `<Rabatt>null` to the database

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/nuernberg/steps/MapFromCsv.kt
```diff
@@ -44,7 +44,7 @@ class MapFromCsv(config: ImportConfig, private val logger: Logger) :
                 it.email.clean(),
                 it.telephone.clean(),
                 it.homepage.clean(),
-                it.discount.clean(false),
+                (it.discountDE + "\n\n" + it.discountEN).clean(false),
```
f1sh1918, line 47: here should be checked if `it.discountEN` is null before adding it

f1sh1918: sth like 
```
      val discount =
            if (it.discountEN == null) {
                it.discountDE
            } else {
                it.discountDE + "\n\n" + it.discountEN
            }
```

michael-markl: Thanks, a bit unfortunate that Kotlin does not give a compile time error :/

michael-markl:
@f1sh1918 Thanks for noticing! Fixed it.

f1sh1918 (approved):
Works fine thx

Issue #1396, opened 2024-04-15
On-screen keyboard opens above the map view


Docmale:
**Describe the bug**
After using the search function (and displaying the on-screen keyboard), when you subsequently open the location on the map, the on-screen keyboard opens again, even though there is no input field. (Nürnberg and Bavaria)

**To Reproduce**
1. Go to search page
2. Enter search string 'schw'
3. Tap a result
4. Details are shown and on-screen keyboard is closed
5. Tap 'Auf Karte zeigen'
6. Map is shown and on-screen keyboard is open

**Expected behavior**
On-screen keyboard should stay closed when displaying the map

**Screenshots**
![image](https://github.com/digitalfabrik/entitlementcard/assets/21236324/fbec864a-d42c-456f-9e3a-ca7273c85fb9)

PR #1398, opened 2024-04-16
1396: Hide keyboard when navigating to the map

seluianova:
### Short description
Bugfix:
After using the search function (and displaying the on-screen keyboard), when you subsequently open the location on the map, the on-screen keyboard opens again, even though there is no input field

### Proposed changes
- set the onTapOutside property for the search text box to hide the keyboard (i.e. to remove the focus from the field) when the user clicks outside the text box
- add the keyboardToggler variable and update _onSearchPressed() to preserve the current behavior of the search button - toggling the keyboard on click

### Side effects
- no?

### Resolved issues
Fixes: #1396

### DIFF
--- a/frontend/lib/widgets/app_bars.dart
+++ b/frontend/lib/widgets/app_bars.dart
@@ -84,17 +84,12 @@ class SearchSliverAppBarState extends State<SearchSliverAppBar> {
   final TextEditingController textEditingController = TextEditingController();
   final FocusNode focusNode = FocusNode();
 
-  var keyboardToggler = false;
-
   @override
   Widget build(BuildContext context) {
     final t = context.t;
     final foregroundColor = Theme.of(context).appBarTheme.foregroundColor;
     return SliverAppBar(
       title: TextField(
-        onTap: () {
-          keyboardToggler = true;
-        },
         onTapOutside: (PointerDownEvent event) {
           focusNode.nextFocus();
         },
@@ -115,7 +110,7 @@ class SearchSliverAppBarState extends State<SearchSliverAppBar> {
           IconButton(icon: const Icon(Icons.clear), onPressed: _clearInput, color: foregroundColor),
         IconButton(
           icon: const Icon(Icons.search),
-          onPressed: _onSearchPressed,
+          onPressed: () {},
           color: foregroundColor,
         )
       ],
@@ -127,15 +122,6 @@ class SearchSliverAppBarState extends State<SearchSliverAppBar> {
     super.dispose();
     focusNode.dispose();
     textEditingController.dispose();
-  }
-
-  _onSearchPressed() {
-    if (keyboardToggler && !focusNode.hasPrimaryFocus) {
-      keyboardToggler = false;
-    } else {
-      focusNode.requestFocus();
-      keyboardToggler = true;
-    }
   }
 
   _onSearchFieldTextChanged(String text) {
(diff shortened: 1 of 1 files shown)


frontend/lib/widgets/app_bars.dart
```diff
@@ -84,12 +84,20 @@ class SearchSliverAppBarState extends State<SearchSliverAppBar> {
...
   @override
   Widget build(BuildContext context) {
     final t = context.t;
     final foregroundColor = Theme.of(context).appBarTheme.foregroundColor;
     return SliverAppBar(
       title: TextField(
+        onTap: () {
+          keyboardToggler = true;
+        },
+        onTapOutside: (PointerDownEvent event) {
```
f1sh1918, line 98: if i do only
```js
        onTapOutside: (PointerDownEvent event) {
          FocusScope.of(context).unfocus();
        },
```
and revert all other changes it also seem to work as expected. Have i overseen anything that makes these flags required?

f1sh1918 (commented):
Tested and works 👍 
Maybe i found an easier implementation just check my comments

frontend/lib/widgets/app_bars.dart
seluianova: Yes, if you do so, the search button won't toggle (open-hide-open-hide) the keyboard anymore.

Current behavior (before my changes) is:
If you click search button once - it opens the keyboard.
If you click it again - it hides the keyboard.

I assumed that was the intent, and added a flag to preserve this behavior.

f1sh1918: on ios the open and closing by clicking on the icon doesn't work. As we discussed the icon is not really needed because the search happens on change of the text input, so we may change the button to an icon

seluianova: What do you think if we keep it as button, but it won't do anything, but splash?
Like, for those, who like to click on buttons after typing to search, even when the automatic search is working, just to make sure?

_Use case_:
You entered some value in the search field and didn't notice the search results updated.
Then you click on our magic "search" button, it splashes, and that gives you confidence that everything is okay, the search has been applied.

f1sh1918: i think i wouldn't just pass an empty function for the splash animation. I think its dirty to pattern :D
The icon in this case is even at the moment just an indicator for the search
I think we should basically use an icon with no function
```js
Padding(
          padding: EdgeInsets.only(right: 16),
          child: Icon(Icons.search),
        ),
```

seluianova: Pushed the icon with some opacity, to make it different from the real "clear" button

michael-markl: One more idea would maybe be to add the search icon in the `icon` property of the InputDecoration of the TextField. Maybe that would move the search icon to the left, but maybe that's actually better anyway? At least in many google apps the icon is to the left of the input field (if there is any).

michael-markl: Or maybe even replace the whole input with a flutter [SearchBar](https://api.flutter.dev/flutter/material/SearchBar-class.html) (which didn't exist back in the day this page was first developed - not sure if it fits here).

f1sh1918: i think a search bar could be discussed and implemented in a separate task if we want it.

f1sh1918 (approved):
Tested on android and ios

PR #1403, opened 2024-04-18
789: Avoid wrong production builds

f1sh1918:
### Proposed changes

-
-

### Side effects

-
-

### Resolved issues

Fixes: #


ztefanie (approved):
Thanks for this quickfix. Long term solution would be to only use one param.

Issue #1388, opened 2024-04-10
Add category 'Andere' (other)


Docmale:
**Is your feature request related to a problem? Please describe.**
There is no icon for category 'Andere' on filterpage therefore I cannot ask for : What else?

**Describe the solution you'd like**
- add a filter icon button to the search page

f1sh1918:
> @f1sh1918 is there a suitable icon for the category?

Currently we are using the normal marker icon as you can see on the map. Not sure if we want to use this. We also have to check if in the styles.json of the map repository this filter applies on the store symbol layer. I think currently this icon is used as a fallback if not category will be applied as filter

seluianova:
@f1sh1918 ah, right, I didn't notice that we have this category already. Thanks.
But I assume it should display not only the items, that fall into this category, but all the items, that do not fall into the categories available in the panel, right?

PR #1408, opened 2024-04-22
1388: Add the Others category to the filter bar in the app

seluianova:
### Short description
Add the "Others" category button to the filter bar.

### Proposed changes
- return the Others category in the filter bar

### Side effects
Actually this category was removed intentionally in this PR https://github.com/digitalfabrik/entitlementcard/pull/709/files
But I didn't find why?
Maybe it shouldn't be displayed in Nürnberg, but then we could remove it from the config probably?

### Resolved issues
Fixes: #1388

### DIFF
--- a/frontend/lib/search/filter_bar.dart
+++ b/frontend/lib/search/filter_bar.dart
@@ -14,13 +14,10 @@ class FilterBar extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final t = context.t;
-    final sortedCategories = [...categoryAssets(context)];
-    sortedCategories.sort((a, b) => a.shortName.length.compareTo(b.shortName.length));
 
-    // Move the "Others" category to the end of the list
-    final othersCategory = sortedCategories.where((category) => category.id == 9).single;
-    sortedCategories.remove(othersCategory);
-    sortedCategories.add(othersCategory);
+    final sortedCategories = [...categoryAssets(context).where((category) => category.id != 9)];
+    sortedCategories.sort((a, b) => a.shortName.length.compareTo(b.shortName.length));
+    sortedCategories.add(categoryAssets(context).where((category) => category.id == 9).single);
 
     final filteredCategories = sortedCategories.where((element) => buildConfig.categories.contains(element.id));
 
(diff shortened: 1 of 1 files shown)


frontend/lib/search/filter_bar.dart
```diff
@@ -15,8 +15,13 @@ class FilterBar extends StatelessWidget {
   Widget build(BuildContext context) {
     final t = context.t;
     final sortedCategories = [...categoryAssets(context)];
-    sortedCategories.removeWhere((category) => category.id == 9);
     sortedCategories.sort((a, b) => a.shortName.length.compareTo(b.shortName.length));
+
+    // Move the "Others" category to the end of the list
```
seluianova, line 20: I assumed we'd want to display it as the last one

f1sh1918:
The category was not removed by this pr it was only adjusted to be removed for entitlementcard and sozialpass. @seluianova 
I think its fine for nuernberg but probably should be discussed with @dkehne

seluianova:
@f1sh1918 right, I just meant it was "removed" from the widget, by calling removeWhere() function :)

michael-markl:
I think the decision (back at the time) was that we do not show `Others` as this is not a real category. But it's also fine to me if we show it :D
We should not encourage the Administrators to put too many stores into the  `others` category. instead they should improve their categorization scheme in this case.

frontend/lib/search/filter_bar.dart
f1sh1918: i prefer using here a named function like `moveOtherCategoryToEnd` or sth similar and do it there so there is no need for a comment. Not sure if adding and removing is the best way
In general we could think of only sorting all elements except of `otherCategoryId` and then add it
```js
 final sortedCategories = [...categoryAssets(context).where((category) => category.id != "otherCategoryId")];
 // then do the sorting and then add the category
 // then add the particular category
    sortedCategories.add(othersCategory);
 
```

f1sh1918 (commented):
Tested on ios and works fine. Left a comment

frontend/lib/search/filter_bar.dart
seluianova: Applied your suggestion of sorting all except 9 + adding 9

dkehne:
We can add the Others-category to all entitlement-cards without any flag etc. It's suitable for Nuremberg and Bavaria.

ztefanie (approved):
Looks good :+1: 
Tested on Android locally, works as expected :+1:

Issue #1361, opened 2024-03-27
Update used node version


ztefanie:
**Is your feature request related to a problem? Please describe.**
Currently some jobs (especially the ones using debian bullseye image) use node 12. Update so at least node 14 is used.
Check the install_toolbelt step finishes without warnings about a wrong node version.
Use a separate command for this, so future version updates only need to be done at one place.

michael-markl:
Probably updating debian to bookworm will be enough [(they package version 18)](https://packages.debian.org/bookworm/nodejs). Related to #1362

PR #1413, opened 2024-04-25
Upgrade debian images to bookworm

michael-markl:
### Short description

By now, we deploy the backend on debian 12 (bookworm).
This upgrades the debian CI images to v12.
This should fix problems related to extremely old Node versions on these builds.

### Side effects

- Likely none, except that CI warnings & errors related to old Node versions should be resolved.

### Resolved issues

Fixes: #1361

### DIFF
--- a/.circleci/src/workflows/commit_main.yml
+++ b/.circleci/src/workflows/commit_main.yml
@@ -1,7 +1,9 @@
 when:
-  and:
-    - << pipeline.parameters.run_commit_main >>
-    - equal: [main, << pipeline.git.branch >>]
+  or:
+    - equal: [<< pipeline.parameters.run_commit_main >>, "run"]
+    - and:
+      - equal: [<< pipeline.parameters.run_commit_main >>, "run_if_on_main"]
+      - equal: [main, << pipeline.git.branch >>]
 jobs:
   - bump_version:
         prepare_delivery: false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -966,8 +966,12 @@ parameters:
         default: true
         type: boolean
     run_commit_main:
-        default: true
-        type: boolean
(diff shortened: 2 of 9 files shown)


.circleci/src/workflows/commit_main.yml
```diff
@@ -1,7 +1,9 @@
 when:
-  and:
-    - << pipeline.parameters.run_commit_main >>
-    - equal: [main, << pipeline.git.branch >>]
+  or:
```
ztefanie, line 2: I would prefer if you undo this, as all our run_xx parameters are boolean and i would expect this to also be boolean, but I see it makes it easier to run this workflow on other branches, so I am also fine with you suggestion.

ztefanie (approved):
Thanks for fixing :rocket:

.circleci/src/workflows/commit_main.yml
michael-markl: Hmm.. currently this flag `run_commit_main` can only be used for skipping the `commit_main` workflow in case we are on branch `main`. As we probably never want to skip this workflow on main (?) it seems that this flag was somewhat useless before?

Is there a different possibility to run the workflow on other branches?

ztefanie: You are right, the before solution did not make sense. We could introduce an other boolean param, that is default false, and does run the commit_Main workflow without checking the branch. But this is not really better than your solution :thinking:

michael-markl: Hmm yeah, this seems to make it unclear which param to set. I'd say, we merge this until we come up with a better solution ? :D

ztefanie: Yes fine :+1:

PR #1438, opened 2024-05-09
Properly convert date string to instant using project timezone

michael-markl:
### Proposed changes

* Use Instant instead of TimeStamp (as we use Instant everywhere)
* Properly convert to Instant by using the project time zone
* slightly adjust SQL query

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
@@ -11,7 +11,6 @@ import app.ehrenamtskarte.backend.verification.webservice.schema.types.CardStati
 import org.jetbrains.exposed.dao.id.EntityID
 import org.jetbrains.exposed.sql.Coalesce
 import org.jetbrains.exposed.sql.JoinType
-import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.SortOrder
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.inList
@@ -21,9 +20,7 @@ import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.count
 import org.jetbrains.exposed.sql.deleteWhere
 import org.jetbrains.exposed.sql.intLiteral
-import org.jetbrains.exposed.sql.javatime.date
 import org.jetbrains.exposed.sql.select
-import java.sql.Timestamp
 import java.time.Instant
 
 object CardRepository {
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
```diff
@@ -75,36 +73,48 @@ object CardRepository {
...
             .orderBy(Regions.name, SortOrder.DESC)
             .orderBy(Regions.prefix, SortOrder.DESC)
-        val result = query
-            .map { CardStatisticsResultModel(region = it[Regions.prefix] + ' ' + it[Regions.name], cardsCreated = (it[allCards[numAlias]] ?: 0).toInt(), cardsActivated = (it[activeCards[numAlias]] ?: 0).toInt()) }
+        return query
+            .map {
+                CardStatisticsResultModel(
+                    region = it[Regions.prefix] + ' ' + it[Regions.name],
+                    cardsCreated = (it.getOrNull(cardsCreated[numAlias]) ?: 0).toInt(),
+                    cardsActivated = (it.getOrNull(activeCards[numAlias]) ?: 0).toInt()
```
michael-markl, line 115: @f1sh1918 We should use the `getOrNull` method here, as the subscript operator seems to return null although its function declaration says otherwise.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
```diff
@@ -75,36 +73,48 @@ object CardRepository {
         }
     }
 
-    fun getCardStatisticsByProjectAndRegion(project: String, dateStart: Timestamp, dateEnd: Timestamp, regionId: Int?): List<CardStatisticsResultModel> {
+    fun getCardStatisticsByProjectAndRegion(
```
ztefanie, line 76: Why the changing of the naming? I think "dateStart" and "dateEnd" are somehow clearer than "from" and "until".

ztefanie (approved):
Looks good :+1:

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/repos/CardRepository.kt
michael-markl: I think date sounds like a date (as in a local date or a a day of the calendar). However they are instants in time. We can also revert it though...

f1sh1918 (changes requested):
Unfortunately your change breaks the timeframe. Including the timezone i get the following values in the backend

image

using this filter
image

The Enddate should be included so  13.05.2024 22:00Z would be correct timestamp

michael-markl:
@f1sh1918 I left the decision open whether or not the "endzeitraum"-date should be exclusive or inclusive (that was an open question in your PR afaik). Currently it is exclusive. If you want I can also change it to inclusive in this PR.

michael-markl:
@f1sh1918 I rebase the branch and made the end date inclusive. (feel free to test it again though)

f1sh1918 (approved):

Issue #1439, opened 2024-05-10
Fix build_martin step in commit_main


michael-markl:
Currently, CI is failing on main in the step build_martin
https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/5221/workflows/a966b331-804c-4337-9fe4-32d656a3c6a6/jobs/38755?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

The issue seems to be a Rust update with breaking changes (a [previous successful build](https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/5207/workflows/188f3561-28c9-4dba-bb6b-af12ba32f5a0/jobs/38685) contained the message:)
```
warning: the following packages contain code that will be rejected by a future version of Rust: semver-parser v0.10.1
```

Probably, we have to upgrade martin to a more recent version (or downgrade rust for now).

michael-markl:
Upgrading martin seems to be quite involved, as they (annoyingly) made breaking changes with v0.7 (https://github.com/maplibre/martin/pull/456 , https://github.com/maplibre/martin/discussions/430 , https://github.com/maplibre/martin/releases/tag/v0.7.0)

PR #1444, opened 2024-05-13
Fix compilation of martin

michael-markl:
### Short description

This fixes the compilation of martin by upgrading the open ssl dependency while keeping the same version of martin. (I don't exactly know why this began to fail, though)

### Proposed changes

While this fixes the compilation error short term, we should still upgrade Martin middle-term :)

### Side effects

Likely none, except  that we can build martin again.

### Resolved issues

Fixes: #1439

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -458,10 +458,8 @@ jobs:
                     - v1-cargo-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}
             - prepare_workspace
             - run:
-                command: |
-                    apt update
-                    apt install -y openssl libssl-dev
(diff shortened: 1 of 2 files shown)


f1sh1918 (commented):
do we still need this after martin update?
If thats not the case, you should maybe add a comment

michael-markl:
No, we don't need this after the martin update.

However, we could merge this now until we know how we want to deploy martin in the near future (They distribute their own debian packages via GitHub, so we no longer have to build it here, see also https://git.tuerantuer.org/DF/salt/pulls/155#issuecomment-1364)

f1sh1918 (approved):

Issue #1215, opened 2023-11-15
Mark processed applications


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a service user i want to see a mark when a card was created from an application, to avoid multiple card creation for an application

**Describe the solution you'd like**
- add a column to the application table "cardCreated" (boolean, default= false)
- create a database migration to add this field
- when a card was created from an application set "cardCreated" field to true
- adjust the button text of "Karte erstellen" to "Karte erneut erstellen" if the "cardCreated" field is true
- add the filter checkbox in the AppStatus bar relates to #1216  ,  this can also be done in a separate tasks

f1sh1918:
@dkehne in the nextcloud file is written that we should use sth like "Karte wurde bereits 1x erstellt"
Is it really needed to count that? A flag with `true/false` should be sufficient?

dkehne:
@f1sh1918 a true/false flag is sufficient!

hauf-toni:
While working on #1216 , I discovered that these two new functions have great potential to be implemented together and in interaction with each other.
This would have the advantage that users would perceive the change as less big, as it would feel more like one new feature than two. they would also have a central point in the interface where they can organize applications. You can find my design proposal [📌 here](https://www.figma.com/file/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?type=design&node-id=392%3A6485&mode=design&t=HMI5f8Vb19qX1QD9-1).

hauf-toni:
as discussed with @f1sh1918 this issue will also be solved by implementing the design of #1216

f1sh1918:
@hauf-toni i think this was a misunderstanding. Even if the issue is included in the design you mentioned this task can be done separately. There is no "hard" relation between these two tasks. Even if the filter bar is not visible yet the button label can be changed and also the rest of the needed functionality implemented as mentioned in the description.

PR #1451, opened 2024-05-21
1215: Mark processed applications

ztefanie:
### Short description

Mark processed applications

### Proposed changes

- Added cardCreated boolean in the applications table
- send cardCreated flag to frontend
- display different button text in the applications view, depending on the flag
- set cardCreated flag, when a card for an application is created

### Side effects

- Maybe the card creation is somewhere messed up, as the creation process now takes an argument with applicationId. I tested it, and the applicationId is optional as it should be.

### Resolved issues

Fixes: #1215


administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -48,6 +48,7 @@ const useCardGenerator = (region: Region) => {
   })
   const [cardBlueprints, setCardBlueprints] = useState<CardBlueprint[]>([])
   const [state, setState] = useState(CardActivationState.input)
+  const [applicationId, setApplicationId] = useState<number>()
```
michael-markl, line 51: Shouldn't this be

```suggestion
  const [applicationId, setApplicationId] = useState<number | undefined>()
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -116,7 +117,8 @@ class CardMutationService {
         dfe: DataFetchingEnvironment,
         project: String,
         encodedCardInfos: List<String>,
-        generateStaticCodes: Boolean
+        generateStaticCodes: Boolean,
+        applicationId: Int? = null
```
michael-markl, line 121: It's not clear here, why there is (only) one applicationId for this endpoint (instead of one for every card info). Please add a comment in the GraphQLDescription.

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -213,7 +213,7 @@ const ApplicationCard = ({
               href={createCardQuery ? `./cards/add${createCardQuery}` : undefined}
               icon='id-number'
               intent='primary'>
-              Karte erstellen
+              {cardCreated ? 'Karte erneut erstellen' : 'Karte erstellen'}
```
michael-markl, line 216: I think we should make it more noticeable for the user when a card has already been created for this application.
Maybe we can make the user confirm a dialog when clicking on this button in this case? And/Or add a [success callout](https://blueprintjs.com/docs/#core/components/callout) before the button toolbar saying that a card has been created for this application?

michael-markl (approved):
Nice! (Not tested, only code review!)

administration/src/bp-modules/applications/ApplicationCard.tsx
f1sh1918: hm sounds reasonable but was not considered by ui/ux @hauf-toni

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
ztefanie: Good point, thanks you. I renamed the param to be more clear.

administration/src/bp-modules/applications/ApplicationCard.tsx
ztefanie: As the deadline is quite soon, I suggest we do this in a later step / separate issue.

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -49,6 +49,7 @@ const useCardGenerator = (region: Region) => {
   })
   const [cardBlueprints, setCardBlueprints] = useState<CardBlueprint[]>([])
   const [state, setState] = useState(CardActivationState.input)
+  const [applicationIdToMarkAsProcessed, setApplicationIdToMarkAsProcessed] = useState<number>()
```
f1sh1918, line 52:
```suggestion
  const [applicationIdToMarkAsProcessed, setApplicationIdToMarkAsProcessed] = useState<number | undefined>()
```
If it isn't initialized, shouldn't it also include `undefined`

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -119,7 +120,8 @@ class CardMutationService {
         dfe: DataFetchingEnvironment,
         project: String,
         encodedCardInfos: List<String>,
-        generateStaticCodes: Boolean
+        generateStaticCodes: Boolean,
+        applicationIdToMarkAsProcessed: Int? = null
```
f1sh1918, line 124: (optional) i wonder whether we need to send the `applicationId` to the backend if we don't do anything with it. Wouldn't be a boolean sufficient?

f1sh1918 (approved):
Works nice! Good work 👍 
Tested on chrome

Just one small thing. 
Due to "GroupByApplication" pr was not merged in the meantime, it was not possible to implement the last bullet point in the issue: https://github.com/digitalfabrik/entitlementcard/issues/1215

Could you please create a separate issue for that.
Even i think that was just an ui/ux improvement from toni and is not really mandatory for the milestone

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
f1sh1918: sorry i didn't see that same issue was already resolved. this can be ignored

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
ztefanie: Yes we do find the right application with it. Without sending the applicationId we only create the card and do not know which application it belongs to. If we don't do anything with it (e.g. for nuernberg or when creating a card that does not belong to a application) we can just pass null.

Issue #1391, opened 2024-04-11
Application note should only be queryable by administrators


michael-markl:
Currently, applicants and organizations that are linked to an application can query the application note.
As this note is private to the administrators, this should not be the case.

f1sh1918:
So whats your proposed solution @michael-markl ?
To create a separate `ApplicationView` for Administrators that include the note and remove it from `ApplicationView`

michael-markl:
Hmm not sure what the best approach is... As the field is nullable (if I remember correctly) we could just not populate it for the corresponding endpoints.

f1sh1918:
its indeed nullable. just created a pr. not sure if this goes in the right direction

PR #1452, opened 2024-05-21
1391: Notes only queryable by admins

f1sh1918:
### Short description

Currently a not authorized user such as applicant or verifier could add "note" into the graphl query and get private notes from the application.

### Proposed changes

- add a flag `includePrivateInformation` and only populates notes int the ApplicationView if its set to true

### Resolved issues

Fixes: #1391


ztefanie (approved):

Issue #1455, opened 2024-05-22
Improve debian package installation


f1sh1918:
**Is your feature request related to a problem? Please describe.**
The installation of the backend and administration fails quite often even if the debian packages are existing on the server. This may have different reasons like racing conditions or one webhook(administration) and second webhook(backend) blocks each other because they were triggered simultaneously

**Describe the solution you'd like**
- merge deploy administration and backend since they mostly depend on each other

**Describe alternatives you've considered**

**Additional context**
Relate to several salt prs: 
- https://git.tuerantuer.org/DF/salt/pulls/157
- https://git.tuerantuer.org/DF/salt/pulls/158

PR #1456, opened 2024-05-22
1455: Improve debian package installation

f1sh1918:
### Short description

The installation of the backend and administration fails quite often even if the debian packages are existing on the server. This may have different reasons like racing conditions or one webhook(administration) and second webhook(backend) blocks each other because they were triggered simultaneously

### Proposed changes

- merge package installation and trigger one webhook to avoid blocking installations
- separate package upload and package installation jobs

### Resolved issues

Fixes: #1455

Relate to several salt prs: 
- https://git.tuerantuer.org/DF/salt/pulls/157 (added a apt repo rebuild to ensure that packages will be recognized directly after uploading before installation was started
- https://git.tuerantuer.org/DF/salt/pulls/158 merge the webhook

### DIFF
--- a/.circleci/src/jobs/install_packages_to_server.yml
+++ b/.circleci/src/jobs/install_packages_to_server.yml
@@ -1,0 +1,31 @@
+docker:
+  - image: cimg/base:2022.09
+parameters:
+  production:
+    description: Whether builds are delivered to production or beta.
+    type: boolean
+steps:
+  - checkout:
+      path: ~/project
+  - run:
+      name: Install Curl, node and npm
+      command: |
+        sudo apt update
+        sudo apt install -y curl nodejs npm
+  - install_app_toolbelt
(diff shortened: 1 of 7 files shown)


.circleci/src/jobs/install_packages_to_server.yml
```diff
@@ -0,0 +1,31 @@
+docker:
+  - image: cimg/base:2022.09
```
ztefanie, line 2: Whouldn't it make more sense to use a debian package here too? Not really important, so fine for me to keep an ubuntu image.

ztefanie (approved):
Looks good :+1:

.circleci/src/jobs/install_packages_to_server.yml
f1sh1918: i think we can address this in the unified images issue

PR #1457, opened 2024-05-22
Fix check backend health

f1sh1918:
- add sudo for backend package postinstall
- fix tsc build: looks like due to merging issues some prop in a test was lost

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -659,7 +659,7 @@ jobs:
                 path: ~/project
             - attach_workspace:
                 at: /tmp/workspace
-            - run: apt update -y && apt install -y nodejs npm curl
+            - run: apt update -y && apt install -y nodejs npm curl sudo
             - install_app_toolbelt
             - run:
                 command: apt install ca-certificates-java -y
--- a/.circleci/src/jobs/check_health_backend.yml
+++ b/.circleci/src/jobs/check_health_backend.yml
@@ -10,7 +10,7 @@ steps:
       path: ~/project
   - attach_workspace:
       at: /tmp/workspace
-  - run: apt update -y && apt install -y nodejs npm curl
+  - run: apt update -y && apt install -y nodejs npm curl sudo
   - install_app_toolbelt
   - run:
       name: Install ca-certificates-java # workaround for current issue: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1030129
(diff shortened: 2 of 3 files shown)


michael-markl:
Btw: You can trigger the `commit_main` pipeline through circleci, if you click on  the "Trigger pipeline" button (https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard?branch=fix-check-backend-health) and add the string parameter "run_commit_main" with  value "run".

.circleci/config.yml
```diff
@@ -659,7 +659,7 @@ jobs:
                 path: ~/project
             - attach_workspace:
                 at: /tmp/workspace
-            - run: apt update -y && apt install -y nodejs npm curl
+            - run: apt update -y && apt install -y nodejs npm curl sudo
```
michael-markl, line 662: Possibly we should add `sudo` as a dependency to our debian package then?

michael-markl: https://github.com/digitalfabrik/entitlementcard/blob/release-mai-2024/.circleci/src/jobs/pack_backend.yml#L11C232-L11C255

f1sh1918:
> Btw: You can trigger the `commit_main` pipeline through circleci, if you click on  the "Trigger pipeline" button (https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard?branch=fix-check-backend-health) and add the string parameter "run_commit_main" with  value "run".

OK i normally use command line for that

.circleci/config.yml
f1sh1918: is that a feature circle-ci provides? i couldn't find any documentation how to do that in circleci @michael-markl 
Or are there specific debian image including sudo, which i also couldn't find in docker image documentation

ztefanie: Not really happy with this solution. Why do we want to add sudo, as everything runs on debian we can just remove the sudo's from the script. Also only adding sudo as a package seems not to be sufficent, as we need to add the current user to sudoers group and log in and out.

ztefanie (approved):

.circleci/config.yml
f1sh1918: currently when i try to switch user on staging server i need sudo. thats why the command is like that for migration in the post install script

michael-markl: > is that a feature circle-ci provides? i couldn't find any documentation how to do that in circleci @michael-markl 

The line I referred to uses the pack deb script which accepts dependencies for the deb package. Currently the only dependency is the Java runtime. We could add sudo there.

(Sorry for slow response, I'm on vacation)

Issue #1459, opened 2024-05-28
Applicant Application List view changes width


f1sh1918:
**Describe the bug**
When an applicant opens the application via link and opens the different application details the list containers gets bigger (Antragsdetails) or smaller (Persönliche Daten)

**To Reproduce**
1. Create an application
2. Open the application link in the received mail
3. Open the details for "Persönliche Informationen" and open details for "Antragsdetails"
4. See container jumping

**Expected behavior**
The container remains the original size

PR #1460, opened 2024-05-28
1459: Applicant application view width

f1sh1918:
### Short description

For bigger screens the list container jumps if "Antragsdetails" will be opened

### Proposed changes

- set a fixed width for non mobile devices to avoid container jumping

### Resolved issues

Fixes: #1459 

### Testing
- Create an application
- Open the link in the received mail
- Open the collapse "Antragsdetails"
- Container should not jump


seluianova (approved):
Tested on desktop and mobile Google Chrome, looks good to me

ztefanie (approved):
lgtm

Issue #1214, opened 2023-11-15
Add POIs to favorites


f1sh1918:
As a user I want to add pois to favorites to find them easier on the map and store list

## Acceptance Criteria:
 - a store can be added to favorites by clicking favorites cta button on the store detailpage (icon changes to filled)
 - the store will be marked with a favorite icon on the map instead of the category icon. The icon should have the same size as the other icons
 - In the ButtomNavigationBar a "Favorite" item should be added that navigates to the favorites page
 - the favorites page contains a list of acceptance stores the user added to favorites. It should behave like the of the search page. (click on the item -> store detail page)
 - a favorite store can be removed from favorites on the store detail page by clicking the favorites button (icon changes to outlined)
 - the favorites will be saved on the device, so they will maintain after the app was closed

## Technical Tasks:
 - add a favorite icon to the mapstyle repository
 - provide the functionality to add/remove a store to the favorites on the store detail page. save the information on the device
 - mark the favorites with the favorites icon on the map
 - add a "favorite" BottomNavigationBar item
 - add a favorite page that shows store favorites. style similar to the search list
 
 ## Design:
- provide an material icon for the favorites
- provide a cta/toggle for the detail page to add the store to favorites
- provide a design that includes a new favorite item in the BottomActionBar
- provide a design for the favorites pages that should be similar/equal to the SearchList

## Some additional info:
To simplify this issue #750 should be done before

IsabellaTG:
@f1sh1918 please review: https://www.figma.com/file/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?type=design&node-id=280%3A445&mode=design&t=wYIZWn6bcdOdFw8W-1

f1sh1918:
@IsabellaTG i addressed some comments to you

seluianova:
@dkehne could you please tell me if we need the ability to enable/disable this feature in the configuration per project?

dkehne:
@seluianova no, we do not need to. it can be published for all projects

PR #1461, opened 2024-05-28
1214: Add pois to favorites

seluianova:
### Short description
Add the ability to save POIs as favorites on device

### Proposed changes
- provide the functionality to add/remove a store to the favorites on the store detail page. save the information on the device
- display snack bar notification when a favorite has neen added/removed
- add a "favorite" BottomNavigationBar item
- add a favorite page that shows store favorites. style similar to the search list

**Not included**:
- mark the favorites with the favorites icon on the map

_Maybe we could separate this part?
It's taking a while, because I'm not quite sure how to do it yet, if we don't make “Favorites” as a category.
Or I might need some help with that._

**UPD**. As discussed in the channel, we will implement that separately.

### Side effects
Couldn't find any
Not tested on iOS 

### Resolved issues
Fixes: #1214


frontend/lib/app.dart
```diff
@@ -104,29 +105,31 @@ class App extends StatelessWidget {
...
-          child: MaterialApp.router(
-            theme: lightTheme,
-            darkTheme: darkTheme,
-            themeMode: ThemeMode.system,
-            debugShowCheckedModeBanner: false,
-            localizationsDelegates: [
-              GlobalMaterialLocalizations.delegate,
-              GlobalWidgetsLocalizations.delegate,
-              GlobalCupertinoLocalizations.delegate,
+        mapStyleUrl: mapStyleUrl,
```
ztefanie, line 108: This lines did not change, but are displayed here as "changed", due to formatting changes. Can you please undo. Not sure why the linter did not catch this.

frontend/lib/favorites/favorites_store.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  Future<List<int>> getFavorites() async {
+    List<int> favorites = [];
+    try {
+      String? stringOfIds = await _storage.read(key: _favoritesKey);
+      if (stringOfIds != null) {
+        favorites = (jsonDecode(stringOfIds) as List).cast<int>();
+      }
+    } catch (e) {
+      log('Failed to load favorites from secure storage', error: e);
```
ztefanie, line 20: We should give the user some hint, why it is not working, shouldn't we?

frontend/lib/favorites/favorites_store.dart
```diff
@@ -0,0 +1,47 @@
...
+    if (favorites.remove(storeId)) {
+      _writeFavorites(favorites);
+    }
+  }
+
+  Future<void> _writeFavorites(List<int> favorites) async {
+    try {
+      await _storage.write(key: _favoritesKey, value: jsonEncode(favorites));
+    } catch (e) {
+      log('Failed to persist favorites to secure storage', error: e);
```
ztefanie, line 44: Same as above.

ztefanie (approved):
Looks good. Nice job :rocket: 

Tested on Android, works as expected :+1:

f1sh1918 (changes requested):
Good work 👍 
There is one scenario we should catch. 
1. User adds a store to favorites.
2. Later the store will be deleted because it doesn't participate anymore.

Current behavior
The store is still in favorites. That ok but when i click on the store, the user is lost in this error screenshot. Retry doesn't help
Currently there is no proper error handling for this case. Maybe we need some support of ui/ux for that

frontend/assets/l10n/app_en.json
```diff
@@ -184,5 +184,11 @@
     "showOnMap": "Show on map",
     "unknownCategory": "Unknown category",
     "website": "Website"
+  },
+  "favorites": {
+    "title": "Favorites",
+    "noFavoritesFound": "You have not added any acceptance points to your favorites yet.",
+    "favoriteHasBeenAdded": "The acceptance point has been added to the favorites.",
+    "favoriteHasBeenRemoved": "The acceptance point has been removed from the favorites."
```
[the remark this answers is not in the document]
seluianova: I used "point" because we use it in other translations:
```
    "noAcceptingStoresFound": "No acceptance points found matching this search.",
    "acceptingStore": "Acceptance points",
    "acceptingStoreNotFound": "Acceptance point not found.",
```
Any objections if I update it everywhere then?

frontend/lib/favorites/favorites_store.dart
seluianova: I'm not sure we can say why, but I think we could be showing the snack bar notification with an error like 'Failed to save the store as a favorite'.
Does it sound good to you?

frontend/assets/l10n/app_en.json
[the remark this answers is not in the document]
f1sh1918: ah okay so. then my mistake. I was checking the database name and thought it sounds more intuitive to me. But you can also keep it as it is then.

frontend/lib/favorites/favorites_store.dart
seluianova: True, thanks.
Added error display for users when favorites failed to initialize and when favorites failed to update.

seluianova:
> There is one scenario we should catch

@f1sh1918 I think I would need your help to reproduce that.
I tried, but catched a different issue:
If favorites contain a store that is not in the database, AcceptingStoreByIdQuery returns it as Null, resulting in a casting error in _fetchPage().
I fixed that in the last commit, so the deleted store just won't show up in the favorites list.

> The store is still in favorites. That ok but when i click on the store, the user is lost in this error screenshot. Retry doesn't help
> We should show in this case probably another error message and provide a way to go back

Currently I'm not quite sure how we can achieve this result, because in order to display a store item in the favorites list we need data about this store (name, description, category). But from where we could get it, if not from the database?

seluianova:
- all comments have been addressed;
- more data about favorite stores is now saved on the device (id, name, category id);
- if a favorite store has been deleted from the database, it is still displayed in favorites with a note that the store is no longer available and with the option to remove it from favorites;
- favorites are now stored in shared preferences rather than secure storage because they are not sensitive data and we basically don't need encryption for them (hopefully this change doesn't break on iOS, as I haven't tested it);
- I added some tests, but will continue separately.

@ztefanie I withdrew your approval because the feature has been significantly reworked

seluianova:
So:
- all comments have been addressed;
- more data about favorite stores is now saved on the device (id, name, category id);
- if a favorite store has been deleted from the database, it is still displayed in favorites with a note that the store is no longer available and with the option to remove it from favorites;
- favorites are now stored in shared preferences rather than secure storage because they are not sensitive data and we basically don't need encryption for them (hopefully this change doesn't break on iOS, as I haven't tested it);
- I also started to add tests, but will continue within a separate task.

frontend/assets/l10n/app_de.json
```diff
@@ -176,13 +176,24 @@
   "store": {
     "acceptingStore": "Akzeptanzstelle",
     "acceptingStoreNotFound": "Akzeptanzstelle nicht gefunden.",
+    "acceptingStoreNotAvailable": "Akzeptanzstelle ist nicht mehr verfügbar",
```
f1sh1918, line 179:
```suggestion
    "acceptingStoreNotAvailable": "Diese Akzeptanzstelle ist nicht mehr verfügbar.",
```

frontend/assets/l10n/app_en.json
```diff
@@ -174,15 +174,26 @@
     "title": "Search"
   },
   "store": {
-    "acceptingStore": "Acceptance points",
+    "acceptingStore": "Acceptance point",
     "acceptingStoreNotFound": "Acceptance point not found.",
+    "acceptingStoreNotAvailable": "Acceptance point is no longer available",
```
f1sh1918, line 179:
```suggestion
    "acceptingStoreNotAvailable": "The acceptance point is no longer available.",
```

frontend/lib/store_widgets/removed_store_summary.dart
```diff
@@ -0,0 +1,149 @@
...
+              storeId: storeId,
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+}
+
+class RemovedStoreContent extends StatelessWidget {
```
f1sh1918, line 97: i think this widget can be moved to a separate component file

f1sh1918 (approved):
Super nice pr 👍 
Thx for adding tests :)
Tested on ios and works fine.
Just some small things
I think the icon size of the favorite icon should be increased a little bit.
I suggest `iconSize:36` to be more prominent
Added screenshot with increased icon size

image

Issue #1458, opened 2024-05-28
Reduce confusion of valid date for Goldkarte during creation


dkehne:
**Is your feature request related to a problem? Please describe.**
Goldkarten of the Bayerische Ehrenamtskarte are valid eternaly without a specific date to become invalid. In the current creation process we just grey-out the valid-date-suggestion for the Standardkarte. 

This lead in at least one case to confusion. The cards are independtly from that visualization created without a valid date (logic is working perfect). 

**Describe the solution you'd like**
Instead of just greying out the valid date, we should do further step to make clear that the valid date is not relevant for gold cards. Setting the greyed-out valid date to 31.12.9999 or undisplay the field would both work.

**Additional context**
It's not completely crucial because we only had that topic once in 2 years in our support tickets. So changing the greyed-out date might be enough

![Screenshot 2024-05-27 114046](https://github.com/digitalfabrik/entitlementcard/assets/4182469/4e0cfbad-2ff2-44a0-84a6-fb861a815d7c)

PR #1463, opened 2024-05-28
1458: Remove endDate gold card

f1sh1918:
### Short description

As a service user it might be confusing if the field "Ablaufdatum" shows a particular date if gold card is selected, since gold card has an infinite lifetime

### Proposed changes

- only render "Ablaufdatum" field if "Standard" is the selected for card type

### Side effects

- none

### Resolved issues

Fixes: #1458

### Testing
- Create a card and select "Goldkarte"
- Field Ablaufdatum should disappear
- Choose Standard and the field should appear again

### DIFF
--- a/administration/src/bp-modules/cards/AddCardForm.tsx
+++ b/administration/src/bp-modules/cards/AddCardForm.tsx
@@ -57,7 +57,6 @@ const CreateCardForm = ({ cardBlueprint, onRemove, onUpdate }: CreateCardsFormPr
         <FormGroup label='Ablaufdatum'>
           <TextField
             fullWidth
-            disabled={cardBlueprint.hasInfiniteLifetime()}
             type='date'
             required
             size='small'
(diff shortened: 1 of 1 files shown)


administration/src/bp-modules/cards/AddCardForm.tsx
```diff
@@ -53,33 +53,35 @@ const CreateCardForm = ({ cardBlueprint, onRemove, onUpdate }: CreateCardsFormPr
...
-              try {
-                cardBlueprint.expirationDate = PlainDate.from(e.target.value)
-                onUpdate()
-              } catch (error) {
-                console.error(`Could not parse date from string '${e.target.value}'.`, error)
+      {!cardBlueprint.hasInfiniteLifetime() && (
+        <FormGroup label='Ablaufdatum'>
+          <TextField
+            fullWidth
+            disabled={cardBlueprint.hasInfiniteLifetime()}
```
seluianova, line 60:
```suggestion
```
Maybe we don't need this attribute then

seluianova (approved):
Looks good to me

ztefanie (approved):
Looks good. not tested

Issue #1453, opened 2024-05-22
Fix failing notify


ztefanie:
See failing job here: https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/5365/workflows/22f3398a-e74e-4c39-896a-8af4c1364729/jobs/39712

Fix it.

PR #1466, opened 2024-05-29
release-2024.5.3

f1sh1918:
### Short description

Latest release with missing bavaria administration artifacts.
Fixed some ci issues to get the delivery work

### Proposed changes

- renamed some steps because of wrong naming
- commented out non working notify step https://github.com/digitalfabrik/entitlementcard/issues/1453

Notes: We have different version numbers for native artifacts (2024.5.5) and backend (2024.5.6) since backend job failed at first run. Couldn't delete native store version versions to redeploy all, since they are in review. But it think its okay. But versions are based on the same tag

### DIFF
--- a/.circleci/src/jobs/notify_release.yml
+++ b/.circleci/src/jobs/notify_release.yml
@@ -22,7 +22,8 @@ steps:
   - run:
       name: Upload backend debian packages to github release
       command: app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files "$(ls ~/attached_workspace/debs/backend/*.deb)" --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
-  - notify:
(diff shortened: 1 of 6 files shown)


.circleci/src/jobs/notify_release.yml
```diff
@@ -22,7 +22,8 @@ steps:
   - run:
       name: Upload backend debian packages to github release
       command: app-toolbelt v0 release upload --releaseId ${RELEASE_ID} --files "$(ls ~/attached_workspace/debs/backend/*.deb)" --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME}
-  - notify:
-      success_message: <<^ parameters.production_delivery >>[Beta] <</ parameters.production_delivery >>Ehrenamtskarte Bayern und Sozialpass Nürnberg ${NEW_VERSION_NAME} have been released successfully! https://github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/releases/tag/${NEW_VERSION_NAME}-all
-      channel: releases
-      allow-all-branches: false
+# https://github.com/digitalfabrik/entitlementcard/issues/1453
```
ztefanie, line 25: Do we really want to merge it like this?

f1sh1918: If you currently run the ci it fails. If the notify fails in other occurrences the whole pipeline fails f.e. the "delivery" and assets won't be uploaded or installed which is annoying.
Thats why i think we should spare it out until its fixed to enable delivery pipeline running successful

ztefanie (approved):
Sad approve :face_holding_back_tears:

Issue #1468, opened 2024-05-29
Bulk Csv card export fails


f1sh1918:
**Describe the bug**
When multiple csv input data is used for csv data export the export results in an unknown error

**To Reproduce**
1. Karten erstellen
2. Mehrere Karten erstellen and choose csv
3. Click CSV Export

**Expected behavior**
Csv card data will be exported

ztefanie:
Also the pdf export for multiple cards is broken, same error.

PR #1469, opened 2024-05-29
1468: Bulk csv card export fails

f1sh1918:
### Short description

Execution of export csv cards and print cards with bulk data fails

### Proposed changes

- adjust function executions

### Side effects

- none
-

### Resolved issues

Fixes: #1468

### Testing
1. Go to "Mehrere Karten erstellen"
2. use csv import file from `administration/resources/cards/sample-nbg.csv`
3. click csv export
4. repeat step 1. and 2. and click QR-Codes drucken


seluianova (approved):
lgtm, thanks for the quick fix

ztefanie (approved):
Tested, works as expected. This also fixes the broken pdf bulk export.

PR #1476, opened 2024-06-04
1453: Fix failing notify

f1sh1918:
### Short description

Notification fails due to node versions that have issues

### Proposed changes

- adjust node versions to lts
- use node img instead of base, since the nodejs version were not working
- add missing notifications to jobs

### Side effects

- the images of node may be little bit larger than base but don't require installation of additional packages and are fixed to a particular version

### Resolved issues

Fixes: #1453

### DIFF
--- a/.circleci/src/jobs/check_administration.yml
+++ b/.circleci/src/jobs/check_administration.yml
@@ -4,9 +4,9 @@ working_directory: ~/project
 steps:
   - checkout:
       path: ~/project
+  - install_app_toolbelt
   - install_dart_linux
   - install_protobuf_linux
-  - install_app_toolbelt
   - prepare_workspace
   - restore_npm_cache
   - run:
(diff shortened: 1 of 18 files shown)


.circleci/src/jobs/check_administration.yml
```diff
@@ -1,11 +1,12 @@
 docker:
-  - image: cimg/node:19.1.0
+  - image: cimg/node:20.13.1
 working_directory: ~/project
 steps:
   - checkout:
       path: ~/project
   - install_dart_linux
   - install_protobuf_linux
+  - install_app_toolbelt
```
ztefanie, line 9: It makes sense, to move the toolbelt up as the first step, as every step that is before install_app_toolbelt and fails will not notify. (same for other job below (check_backend))

ztefanie (commented):
Thanks for fixing this. Looks good :+1:

.circleci/src/jobs/check_administration.yml
f1sh1918: thx for the hint. i moved the step everywhere directly after checkout step (if no further apt tool installation is required)

ztefanie (approved):

Issue #1465, opened 2024-05-29
Replace link inside the digital request form of Bavarian Volunteer Card


dkehne:
Inside the form (page 2) there is a broken link which links to the requirements for the volunteer card. 
It should be replaced with the new/working link: https://www.ehrenamt.bayern.de/vorteile-wettbewerbe/ehrenamtskarte/index.php#sec3

PR #1477, opened 2024-06-04
1465: Replace requirements link

f1sh1918:
### Short description

The link for the requirements of the EC changed

### Proposed changes

- replace the link

### Resolved issues

Fixes: #1465

### Testing
1. http://localhost:3000/beantragen
2. type in personal data and go next
3. Check the link above "Blaue Ehrenamtskarte" -> "hier einsehen" for the requirements

### DIFF
--- a/administration/src/mui-modules/application/forms/StepCardTypeForm.tsx
+++ b/administration/src/mui-modules/application/forms/StepCardTypeForm.tsx
@@ -98,7 +98,7 @@ const StepCardTypeForm: Form<State, Options, ValidatedInput, AdditionalProps> = 
           Die Erfüllung der Voraussetzungen wird im nächsten Schritt des Antrags abgefragt. Weitere Informationen können
           Sie{' '}
           <a
-            href='https://www.ehrenamt.bayern.de/vorteile-wettbewerbe/ehrenamtskarte/'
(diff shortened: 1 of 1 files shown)


seluianova (commented):
I don't quite understand why, but if I click 'hier ansehen' in production form
(https://bayern.ehrenamtskarte.app/beantragen), the opened link is really broken:
https://www.ehrenamt.bayern.de/vorteile-wettbewerbe/index.php/ehrenamtskarte/voraussetzungen/index.php

So I'm not sure if adding “index.php#sec3” would fix that.

But this problem is not reproduced locally, the opened link works (with and without "index.php#sec3"). 

Do you know why the link is corrupted in the prod?

f1sh1918:
> I don't quite understand why, but if I click 'hier ansehen' in production form (https://bayern.ehrenamtskarte.app/beantragen), the opened link is really broken: https://www.ehrenamt.bayern.de/vorteile-wettbewerbe/index.php/ehrenamtskarte/voraussetzungen/index.php
> 
> So I'm not sure if adding “index.php#sec3” would fix that.
> 
> But this problem is not reproduced locally, the opened link works (with and without "index.php#sec3").
> 
> Do you know why the link is corrupted in the prod?

Maybe there is a wrong redirect in nginx because the link is actually working

seluianova (approved):
> we didn't release for a while on prod and in the meantime we adjusted the link once, and now we added an anchor

ah, okay. good to go then :)

Issue #1433, opened 2024-04-29
Create user data hash


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Since the user data is sensitive it should be hashed

**Describe the solution you'd like**
- choose a hashing algorithm (prefered argon2id)
- create a canonical json from the user data (name, birthday, aktenzeichen)
- hash the data using at least two iterations (tbd) and a secret salt that should be stored in a environment variable
- add unit tests
- provide the salt and an example hashing code snippet to koblenz

**Describe alternatives you've considered**
use the existing hashing algorithm

**Additional context**
It has discussed if we switch to the same hashing algorithm for everything (but this also may need additional resources which are not needed)
https://argon2.online/

michael-markl:
>  It has discussed if we switch to the same hashing algorithm for everything (but this also may need additional resources which are not needed)

There is no real need to use a "slow" hash algorithm for the card info hash (as we add random bytes to it - the pepper). Also we'd have to do it backward compatibly somehow.

f1sh1918:
> >  It has discussed if we switch to the same hashing algorithm for everything (but this also may need additional resources which are not needed)
> 
> There is no real need to use a "slow" hash algorithm for the card info hash (as we add random bytes to it - the pepper). Also we'd have to do it backward compatibly somehow.

Yes I was also thinking about that issues. "It has be discussed" Was what I wanted to write...
And yes password hashing would be difficult and we would have to support two ways of hashing maybe old passwords would never be changed if we don't force it

Issue #1421, opened 2024-04-29
Create createCard endpoint


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Create a new endpoint (or adjust existing one) that receive the user data from the self service portal

**Describe the solution you'd like**
- add a function that receives user data: Name, Aktenzeichen and Birthday
- Hash this data and check if the user is permitted or not
- if not permitted throw an error
- if permitted send the cardData (Name, Aktenzeichen, Birthday, StartDate, and the other general needed data)
- rate limiting: since this endpoint is not protected, it should be ensured that IPs with too many requests should be blocked, especially because the argon2id calculation is expensive.

**Describe alternatives you've considered**
We could extend the current createCard mutation and allow creating without jwt token if its a particular projectId (not sure if that is secure enough)

blocked by https://github.com/digitalfabrik/entitlementcard/pull/1499

seluianova:
> add a function that receives user data: Name, Aktenzeichen and Birthday

@f1sh1918 I guess the user would also need to select a region, wouldn't they?

seluianova:
> rate limiting: since this endpoint is not protected, it should be ensured that IPs with too many requests should be blocked, especially because the argon2id calculation is expensive.

just got a sudden idea.
what if instead of a web form to issue a card, we create such form in the flutter app?
I believe this would solve the issue above and may also be more convenient for users because there would be no need to download and scan pdfs

f1sh1918:
Its a good idea but we have lots of existing functionality in the "administration" like the card extensions, deeplinks which we can reuse. Further more we could also provide additionally a pdf if a service user supports a user that can not create the card by him/herself and just pass the pdf to him/her.

Normal use case is that the user doesn't have to scan or download the pdf but just create the card and then click a button with a `custom_scheme` deeplink which opens the app and activates the card.

seluianova:
@f1sh1918 is a BUT that doesn't make sense to discuss or doch? :)

> which we can reuse
but is it bad if we don't have to?

> the card extensions
is it applicable to Koblenz? is it something that an unauthorized user does? (I'm not familiar with this process)

> Normal use case is that the user doesn't have to scan or download the pdf
right, but still an extra action and extra browser interaction.

seluianova:
@f1sh1918 is a no-way but or to be discussed?

> which we can reuse

is it bad if we won't have to? I assumed it could mean for us to do less, not more

> the card extensions

[...]

PR #1499, opened 2024-06-19
1433: Create user hashing for Koblenz

ztefanie:
### Short description
Add argon2 hashing for card info

### Proposed changes
- Added argon2id hashing for user data
- Added test for argon2id hashing
- Added Koblenz region and refactored region db setup a little
- Added tests for koblenz canonical json creation
- Added docs for Koblenz how to create valid hashes
- Added pepper to passbold
- Added pepper to env var in Saltstack pr: https://git.tuerantuer.org/DF/salt/pulls/174

### Resolved issues

Fixes: #1433


backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        fun hashUserData(cardInfo: Card.CardInfo): String? {
+            val canonicalJson = CanonicalJson.messageToMap(cardInfo)
+            val hashLength = 32
+            if (!isCanonicalJsonValid(canonicalJson)) {
+                throw Exception("Invalid Json input for hashing")
+            }
+
+            val pepper = Environment.getVariable(KOBLENZ_PEPPER_SYS_ENV) //TODO handle if Null
```
ztefanie, line 54: What error handling do we want, if the env var cannot be accesed / is not set?

docs/CreateKoblenzHash.md
```diff
@@ -0,0 +1,69 @@
...
+    regionId: 95
+  }
+  extension_birthday {
+    birthday: 12213
+  }
+  extension_koblenz_pass_id {
+    pass_id: "123K"
+  }
+}
+```
```
michael-markl, line 29: (Note that I might not have the full context)

I'd prefer if this new hash procedure here would not be based on CardInfo; instead just a new object format à la `{ fullName: "Karla Koblenz", birthday: 123123, passId: "123K" }` (also, `passId` might not make sense; it should be an identifier for the person and not the pass, I think?).
Otherwise, we cannot add other card specific fields (such as expiration dates, or similar) to the card info.

f1sh1918: Yes i agree it should  be sth like PersonInfoHash and not connected to Cardainfo

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -163,6 +166,25 @@ class CardMutationService {
         return activationCodes
     }
 
+    @GraphQLDescription("Creates a new digital koblenz card and returns it")
```
f1sh1918, line 169: i think this endpoint shouldn't be part of this pr.
This task is about hashing the user data.

Some explanations from my side:
I think this function will be used later if the user sends data from the self service portal.
Then we should receive a proper user data object which we again hash and check if the hash is in the database.
Then we create from the userData the cardData and send them back to the frontend

f1sh1918 (commented):
I'm a little bit confused here about the scope of this task and probably there was a misunderstanding.

I expected this task to:
- create a UserInfoModel (name, birthday, referenceNr)
- hash this model with `argon2id`
- create tests for the hashing
- provide sample input and output data to koblenz service provider.

For other tasks like CardCreation we have different tasks, f.e.
https://github.com/digitalfabrik/entitlementcard/issues/1421

docs/CreateKoblenzHash.md
ztefanie: Thanks for this hint, i was unsure how to handle this and change it now tot the new object format you suggested.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
```diff
@@ -89,6 +91,10 @@ class CanonicalJson {
             }
         }
 
+        fun koblenzUserToString(koblenzUser: KoblenzUser): String {
+            return ObjectMapper().writeValueAsString(koblenzUser)
```
michael-markl, line 95: We need a canonical JSON string (RFC 8785) -- I don't think Jackson guarantees this ([yet?](https://github.com/FasterXML/jackson-core/issues/1037)).
I recommend converting `koblenzUser` to a `Map<String, Any?>` manually here, and calling `serializeToString` with the map.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
```diff
@@ -0,0 +1,76 @@
...
+                    .withSalt(pepperByteArray)
+                    .withParallelism(1)
+                    .withMemoryAsKB(16)
+                    .build()
+
+            val generator = Argon2BytesGenerator()
+            generator.init(params)
+            val result = ByteArray(hashLength)
+            generator.generateBytes(canonicalJson.toCharArray(), result)
+            return encode(result, params)
```
michael-markl, line 73: I think we should not use the encoding that you suggest (instead, we can simply use the base64 encoding of `result`).
There's simply no reason to add these parameters to each hash as we will use the hash as an index and not as a password (So we will have to stick to these parameters unless we can recompute all hashes -- which would need to be done by Koblenz). Also I think, we use the salt as a global credential. Then, it does not make sense to append it to every hash.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
```diff
@@ -0,0 +1,76 @@
...
+        fun hashKoblenzUserData(userData: KoblenzUser): String? {
+            val canonicalJson = CanonicalJson.koblenzUserToString(userData)
+            val hashLength = 32
+
+            val pepper = Environment.getVariable(KOBLENZ_PEPPER_SYS_ENV) // TODO handle if Null
+            val pepperByteArray = pepper?.toByteArray(StandardCharsets.UTF_8)
+            val params =
+                Argon2Parameters
+                    .Builder(Argon2Parameters.ARGON2_id)
+                    .withVersion(19)
```
f1sh1918, line 62: what is this version referring?

f1sh1918 (commented):
Nice implementation!
Added some minor comments
Unfortunately i can not start the backend receiving this error

```js
Exception in thread "main" java.lang.Error: Required project 'koblenz.pass.app' not found!
	at app.ehrenamtskarte.backend.regions.database.SetupKt.insertOrUpdateRegions$createOrUpdateRegion(Setup.kt:21)
	at app.ehrenamtskarte.backend.regions.database.SetupKt.access$insertOrUpdateRegions$createOrUpdateRegion(Setup.kt:1)
	at app.ehrenamtskarte.backend.regions.database.SetupKt$insertOrUpdateRegions$1.invoke(Setup.kt:61)
	at app.ehrenamtskarte.backend.regions.database.SetupKt$insertOrUpdateRegions$1.invoke(Setup.kt:38)
	```

File: backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt (hunk 5 above)
Comment (f1sh1918) [reply backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt:73]: ah i think we talked about some kind of "minimal" validation. But i agree that the that the encoded form is not very handy.
What do you mean with "append". We use here the salt as a pepper (global) to even make it more difficult to brute force it

Comment (michael-markl) [reply backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt:73]: The `encode` function adds the parameters as well as the salt to the "encoded" hash. I suggest to not use the `encode` function and instead simply return `base64Encode(result)`.

**Comment by f1sh1918** [comment]
> Nice implementation! Added some minor comments Unfortunately i can not start the backend receiving this error
> 
> ```js
> Exception in thread "main" java.lang.Error: Required project 'koblenz.pass.app' not found!
> 	at app.ehrenamtskarte.backend.regions.database.SetupKt.insertOrUpdateRegions$createOrUpdateRegion(Setup.kt:21)
> 	at app.ehrenamtskarte.backend.regions.database.SetupKt.access$insertOrUpdateRegions$createOrUpdateRegion(Setup.kt:1)
> 	at app.ehrenamtskarte.backend.regions.database.SetupKt$insertOrUpdateRegions$1.invoke(Setup.kt:61)
> 	at app.ehrenamtskarte.backend.regions.database.SetupKt$insertOrUpdateRegions$1.invoke(Setup.kt:38)
> 	```
> 	
> ```
@ztefanie 
You probably have to create an entry for koblenz in the backend configuration. (config.yml)
Additionally we need for deployment on staging and production configs in salt or the backend service will not start

You probably should talk to daniel and sven that we need some domain for koblenz. Just to prepare that

File: backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt (hunk 4 above)
Comment (f1sh1918) [reply backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt:95]: Do we also have to make ensure lower or uppercase of the strings?

File: backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt (hunk 5 above)
Comment (ztefanie) [reply backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt:73]: Hmh, I think we should keep the encoding (it also is quite a standart, not something I invented), as we can do some minimal form of validation and testing, when something is not working as expected. At least we can check for parameters being set and we will not get confused if we change something in the future, e.g. the salt or the version. It can be really hard (or impossible) without this encoding to check what is going wrong, so I would really prefer to keep this encoding, as the only downside of keeping it, is slightly longer entry in the database, but I think this is not really important for the 15-20k entries.

File: backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt (hunk 6 above)
Comment (ztefanie) [reply backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt:62]: version of the argon2 algorithm

File: backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt [hunk 7]
```diff
@@ -0,0 +1,76 @@
...
+                    .withIterations(2)
+                    .withSalt(pepperByteArray)
+                    .withParallelism(1)
+                    .withMemoryAsKB(16)
+                    .build()
+
+            val generator = Argon2BytesGenerator()
+            generator.init(params)
+            val result = ByteArray(hashLength)
+            generator.generateBytes(canonicalJson.toCharArray(), result)
```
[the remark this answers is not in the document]
ztefanie: Changed to `toByteArray()`, which uses UTF_8 per default

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/CanonicalJson.kt
ztefanie: I changed it to using a Map and the serializeToString function, result stays the same.

ztefanie:
> > Nice implementation! Added some minor comments Unfortunately i can not start the backend receiving this error
> > ```js
> > Exception in thread "main" java.lang.Error: Required project 'koblenz.pass.app' not found!
> > 	at app.ehrenamtskarte.backend.regions.database.SetupKt.insertOrUpdateRegions$createOrUpdateRegion(Setup.kt:21)
> > 	at app.ehrenamtskarte.backend.regions.database.SetupKt.access$insertOrUpdateRegions$createOrUpdateRegion(Setup.kt:1)
> > 	at app.ehrenamtskarte.backend.regions.database.SetupKt$insertOrUpdateRegions$1.invoke(Setup.kt:61)
> > 	at app.ehrenamtskarte.backend.regions.database.SetupKt$insertOrUpdateRegions$1.invoke(Setup.kt:38)
> > ```

I can start the backend with no problems, maybe some problem with your local setup. We can talk about this, when your vacation is over.

> 
> @ztefanie You probably have to create an entry for koblenz in the backend configuration. (config.yml) Additionally we need for deployment on staging and production configs in salt or the backend service will not start @svenseeberg can do that

Not part of this issue.

> 
> You probably should talk to daniel and sven that we need some domain for koblenz. Just to prepare that

Getting the domains for Koblenz is not part of this issue and not even related

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
michael-markl: We should probably at least strip the secret salt from the hashes when writing the hashes to the DB. That's the only problem I see with the encoding.

docs/CreateKoblenzHash.md
```diff
@@ -0,0 +1,62 @@
...
+    full_name: "Karla Koblenz"
+    birthday: 12213
+    referenceNumber: "123K"
+}
+```
+
+- The full_name must be `FirstnameSpaceLastname`. Every char must exaclty match the user input, as otherwise there is not possiblity to match the data Koblenz transfers with the input the user makes.
+e.g. `Karla Koblenz` will match neither with `Karla Lisa Koblenz` nor with `Karlá Koblenz`.
+- The birthday is defined in our protobuf [card.proto](../frontend/card.proto) file: It counts the days since the birthday (calculated from 1970-01-01).
+  All values of this field are valid, including the 0, which indicates that the birthday is on 1970-01-01. Birthdays before 1970-01-01 have negative values.
```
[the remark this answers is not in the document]
ztefanie: I see your point, would keep it like this, so at least our code is consistent?

docs/CreateKoblenzHash.md
```diff
@@ -0,0 +1,62 @@
...
+| Version      | 19                                                                             | 
+| Iterations   | 2                                                                              | 
+| Parallellism | 1                                                                              | 
+| Memory       | 16                                                                             | 
+| HashLength   | 32                                                                             | 
+| Salt         | Secret Salt will be shared with Koblenz<br/>for the example use `123456789ABC` | 
+
+
+### 4. The result...
+...for the example data and example salt must be:
```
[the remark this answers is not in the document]
ztefanie: I would prefer to keep it for the same reason, i wanted to keep it in the database: To make it easier (or possible at all) to debug. Is there any harm to keep it here?

docs/CreateKoblenzHash.md
```diff
@@ -0,0 +1,62 @@
...
+Hash with Argon2id with the following parameters:
+
+| Parameter    | Value                                                                          |
+|--------------|--------------------------------------------------------------------------------|
+| Version      | 19                                                                             | 
+| Iterations   | 2                                                                              | 
+| Parallellism | 1                                                                              | 
+| Memory       | 16                                                                             | 
+| HashLength   | 32                                                                             | 
+| Salt         | Secret Salt will be shared with Koblenz<br/>for the example use `123456789ABC` |
```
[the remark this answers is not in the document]
ztefanie: I am not an expert here too, but can read into some details. I though to use "smaller/faster" parameters as this is does not really need to be save for bruteforcing or anything if the database gets exposed.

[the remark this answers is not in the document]
michael-markl: I think the protection from bruteforcing is the main reason we use argon2id.

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
michael-markl: Well, it's a secret. We should regard it as a clear text password. We wouldn't sprinkle all data & interfaces with clear text password usually :)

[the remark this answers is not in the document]
ztefanie: It is not saved as clear text but also hashed

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
ztefanie: OWASP recommends to use: Use Argon2id with a minimum configuration of 19 MiB of memory, an iteration count of 2, and 1 degree of parallelism. Or:

> We recommend the following configuration settings:
m=47104 (46 MiB), t=1, p=1 (Do not use with Argon2i)
m=19456 (19 MiB), t=2, p=1 (Do not use with Argon2i)
m=12288 (12 MiB), t=3, p=1
m=9216 (9 MiB), t=4, p=1
m=7168 (7 MiB), t=5, p=1

Allo of the above options are equally "safe", only differ in CPU vs RAM usage.

See: [here](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#:~:text=Use%20Argon2id%20with%20a%20minimum,a%20parallelization%20parameter%20of%201.)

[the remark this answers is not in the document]
ztefanie: So would you be fine if we set it to: m=12288 (12 MiB), t=3, p=1

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
michael-markl: In your example below
```
$argon2id$v=19$m=16,t=2,p=1$MTIzNDU2Nzg5QUJD$UIOJZIsSL8vXcuCB82xZ5E8tpH6sQd3d4U0uC02DP40
                           |---Base64Salt---|-----------actual-hash---------------------|
```
So it's clear text (modulo base64 encoding): `Base64Decode("MTIzNDU2Nzg5QUJD")="123456789ABC"`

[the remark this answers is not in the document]
michael-markl: Maybe I was not clear enough in the first question:
Should Koblenz send the an "encoded" argon2id hash that includes the salt in its "encoding"?
If they do, we should remove the salt from the "encoding" before storing the hashes to the DB.
If they don't , then we don't need to do it.

From the perspective that the salt is a secret, I would tend to "They should send include the salt

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
michael-markl: @maxammann / @svenseeberg might have better insights than me.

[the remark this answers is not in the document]
michael-markl: (@ max for context: For Koblenz's self service portal, the user enters their Name+Birthday+ReferenceNumber into the web portal. We hash their data, and if we find the corresponding hash in our dataset, we create a card for the user (with the same data of course). Now, to protect against rainbow table attacks -- if the DB is leaked we don't want attackers to find out who is receiving social benefits -- the current plan is to use argon2id together with a secret but global salt (the same salt for all hashes).)

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
ztefanie: You are absolutely right. I will adjust it.

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
f1sh1918: we use it this standard (dd.MM.yyyy) for query params (used by 3rd party applications from nuernberg) and csv import. I'm with steffi in this case

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -1,18 +1,45 @@
...
+        regionName: String,
+        regionPrefix: String,
+        regionWebsite: String
+    ) {
+        val project =
+            projects.firstOrNull { it.project == regionProjectId }
+                ?: throw Error("Required project '$regionProjectId' not found!")
+        val region = dbRegions.singleOrNull { it.projectId == project.id }
+        if (region == null) {
+            RegionEntity.new {
```
maxammann, line 24: Is this happening in a transaction? I think this might be race-able when creating two regions in with parallel requests.

Likely not a huge issue though as only admins can create or update regions I guess.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
```diff
@@ -0,0 +1,73 @@
...
+                    .withIterations(2)
+                    .withSalt(pepperByteArray)
+                    .withParallelism(1)
+                    .withMemoryAsKB(16)
+                    .build()
+
+            val generator = Argon2BytesGenerator()
+            generator.init(params)
+            val result = ByteArray(hashLength)
+            generator.generateBytes(canonicalJson.toByteArray(), result)
```
maxammann, line 69: This comment is more related to CanonicalJson. I just briefly checked its implementation and it seems like it is not enforing an order of the fields. It seems like the protobuf implementation determines how things are sorted: https://github.com/digitalfabrik/entitlementcard/blob/aa83fa44fede13446b1f1b48f3886f4f525dd754/backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/CanonicalJson.kt#L61-L65


I could be wrong here though. If the above is true we should probably test the CanonicalJson implementation.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
```diff
@@ -0,0 +1,73 @@
...
+                .append(parameters.lanes)
+            stringBuilder.append("$").append(b64encoder.encodeToString(hash))
+            return stringBuilder.toString()
+        }
+
+        fun hashKoblenzUserData(userData: KoblenzUser): String? {
+            val canonicalJson = CanonicalJson.koblenzUserToString(userData)
+            val hashLength = 32
+
+            val pepper = Environment.getVariable(KOBLENZ_PEPPER_SYS_ENV) // TODO handle if Null
```
maxammann, line 54: This is used as argon2 salt, so we should probably name this a salt :D

docs/CreateKoblenzHash.md
```diff
@@ -0,0 +1,62 @@
...
+//Example:
+{
+    full_name: "Karla Koblenz"
+    birthday: 12213
+    referenceNumber: "123K"
+}
+```
+
+- The full_name must be `FirstnameSpaceLastname`. Every char must exactly match the user input, as otherwise there is not possibility to match the data Koblenz transfers with the input the user makes.
+e.g. `Karla Koblenz` will match neither with `Karla Lisa Koblenz` nor with `Karlá Koblenz`.
```
maxammann, line 25: So improve matching we could create multiple hashes that e.g. hash the initials. I don't know the usecase but this might be an option if issues arise.

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
maxammann: ISO 8601 might be problematic because there are maybe multiple encodings of the same date. Not sure though. I'm just thinking about the "+Z" and "+00:00"

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
maxammann: Some remarks:

* Argon2id is difficult to deploy in a browser. So if we intend to calculate the hash in the frontend we might go with a different one like pbkdf2, which is more legacy but still oke to use
* The minimum parameters (which I think we went with) are good enough. We are paying for the processing if hashing happens in the backend, we better start with the minimal req. https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html
* The salt shoudl be unique for each user. So in the DB we would store a random salt next to the hash.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
```diff
@@ -0,0 +1,73 @@
...
+            val pepperByteArray = pepper?.toByteArray(StandardCharsets.UTF_8)
+            val params =
+                Argon2Parameters
+                    .Builder(Argon2Parameters.ARGON2_id)
+                    .withVersion(19)
+                    .withIterations(2)
+                    .withSalt(pepperByteArray)
```
maxammann, line 61: This salt should be unique for each user and randomly generated.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
maxammann: How does the canonical json react to inputs that need to be escaped? E.g. names that include a quote?

maxammann:
Another approach would have been to hash on the client and just the data to the server for the lookup, right?

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
michael-markl: The order is canonicalized here: https://github.com/digitalfabrik/entitlementcard/pull/1499/files/eb2d7bbd19f7ee123a4725e74ba8bb473252022f#diff-c688a2cede2d0e62a8d75f41502f13cf619cb1d3d5c126b7ccc75329c3da8fc1R135

The escaping is (probably/hopefully) done by https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-primitive/ (which we construct here https://github.com/digitalfabrik/entitlementcard/pull/1499/files/eb2d7bbd19f7ee123a4725e74ba8bb473252022f#diff-c688a2cede2d0e62a8d75f41502f13cf619cb1d3d5c126b7ccc75329c3da8fc1R125 )

maxammann:
I believe this endpoint should have some rate limiting as it might be a target for DoS (argon2id is costly). The alternative approach would be to generate the hash on the client.

michael-markl:
> Another approach would have been to hash on the client and just the data to the server for the lookup, right?

Not really possible, as we don't want that the client can decide what to write to the card.

maxammann:
> > Another approach would have been to hash on the client and just the data to the server for the lookup, right?
> 
> Not really possible, as we don't want that the client can decide what to write to the card.

I believe we could still send the user data to the server and do the additional check only if we find the matching hash. Then the expensive argon2id/pbkdf verification only runs if the hash can be found. The user would still send the name, birthdate etc to the server.

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
ztefanie: We decided to not use a salt, but a pepper (same for all users). It was defined like this in the issue, i am not really sure how this decision was made.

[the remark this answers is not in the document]
michael-markl: We cannot have a unique salt for each user, as we use the hash as a lookup.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
michael-markl: Yes it's in a transaction (see below) and it is also only run once when the backend starts.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
ztefanie: A salt is a unique and not secret element added for every entry.
A pepper is a non-unique but secret element added for every entry.

We are using a pepper, so I named it like this.

maxammann: Interesting was not aware of that definition :) so yeah then it makes sense

ztefanie:
> I believe this endpoint should have some rate limiting as it might be a target for DoS (argon2id is costly). The alternative approach would be to generate the hash on the client.

Will be done here: https://github.com/digitalfabrik/entitlementcard/issues/1421 
and is already mentioned in the issue, that it needs rate limiting.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/utils/Environment.kt
```diff
@@ -0,0 +1,8 @@
+package app.ehrenamtskarte.backend.common.utils
+
+// This helper class was created to enable mocking getenv in Tests
```
seluianova, line 3: alternatively, we can set env variables for tests in gradle configuration:

```
tasks.withType<Test> {
    environment("KOBLENZ_PEPPER", "123456789ABC")
}
```

ztefanie: Thanks for sharing this idea, i tried different solutions and result of my research was that this mockable helper was the most robust solution.

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -30,22 +57,7 @@ fun insertOrUpdateRegions() {
...
-                regionIdentifier = null
-                website = "https://nuernberg.de"
-            }
-        } else {
-            nuernbergRegion.name = "Nürnberg"
-            nuernbergRegion.prefix = "Stadt"
-            nuernbergRegion.website = "https://nuernberg.de"
-        }
+        createOrUpdateRegion(NUERNBERG_PASS_PROJECT, "Nürnberg", "Stadt", "https://nuernberg.de")
+        createOrUpdateRegion(KOBLENZ_PASS_PROJECT, "Koblenz", "Stadt", "https://koblenz.de/")
```
f1sh1918, line 61: this will only work when the project was created before in the database, which requires a `config.yml` entry

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
f1sh1918: yes but we use it here as a pepper to make it more difficult to generate the hash but we have to ensure that we can regenerate this hash. With a random salt we couldn't

docs/CreateKoblenzHash.md
```diff
@@ -0,0 +1,62 @@
...
+| Parallellism | 1                                                                              | 
+| Memory       | 19                                                                             | 
+| HashLength   | 32                                                                             | 
+| Salt         | Secret Salt will be shared with Koblenz<br/>for the example use `123456789ABC` | 
+
+
+### 4. The result...
+...for the example data and example salt must be (output in encoded form with salt removed):
+
+`$argon2id$v=19$m=16,t=2,p=1$UIOJZIsSL8vXcuCB82xZ5E8tpH6sQd3d4U0uC02DP40`
```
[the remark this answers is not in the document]
seluianova: Or this value should be escaped using the quotation marks in csv

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
f1sh1918: i added the config in last commit

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
f1sh1918: ok i didn't see quotation marks in your example here: https://github.com/digitalfabrik/entitlementcard/pull/1528
We should provide a proper example csv for koblenz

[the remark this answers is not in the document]
seluianova: btw do you know why there is a $ delimiter for some params but comma for others? Having $ everywhere would also solve the problem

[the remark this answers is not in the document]
f1sh1918:
```
userHash, startDate, endDate, revoked, regionKey
"$argon2id$v=19$m=19,t=2,p=1$78SNlzz8llqPwetRlp8Lnag/WZGPc5XbimYbShpwVMQ", 12.05.2024, 12.10.2028, false, 09761
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
f1sh1918: just added a basic exception. Do we need a separate exception for that?

f1sh1918:
Just updated the pr by adding a project config, aligning some values from the tests and the docs which were different to result in the correct hash.
Is there anything else that is missing?

backend/src/test/kotlin/app/ehrenamtskarte/backend/cards/Argon2IdHasherTest.kt
```diff
@@ -17,7 +17,7 @@ internal class Argon2IdHasherTest {
         assertEquals(Environment.getVariable("KOBLENZ_PEPPER"), "123456789ABC")
 
         val hash = Argon2IdHasher.hashKoblenzUserData(KoblenzUser("Karla Koblenz", 12213, "123K"))
-        val expectedHash = "\$argon2id\$v=19\$m=19,t=2,p=1\$78SNlzz8llqPwetRlp8Lnag/WZGPc5XbimYbShpwVMQ" // This expected output was created with https://argon2.online/
+        val expectedHash = "\$argon2id\$v=19\$m=16,t=2,p=1\$UIOJZIsSL8vXcuCB82xZ5E8tpH6sQd3d4U0uC02DP40" // This expected output was created with https://argon2.online/
```
f1sh1918, line 20: was there any reason to increase that standard value?

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
michael-markl: The memory still needs to be increased (and the docs adjusted).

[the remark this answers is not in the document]
michael-markl: (ping @f1sh1918 : not sure if you saw my comment) I probably won't have time for a full review unfortunately :(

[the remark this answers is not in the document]
f1sh1918: i addressed it in the last commit (now). :)

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/Argon2IdHasher.kt
```diff
@@ -0,0 +1,73 @@
...
+            val hashLength = 32
+
+            val pepper = Environment.getVariable(KOBLENZ_PEPPER_SYS_ENV) ?: throw Exception("No koblenz pepper found")
+            val pepperByteArray = pepper.toByteArray(StandardCharsets.UTF_8)
+            val params =
+                Argon2Parameters
+                    .Builder(Argon2Parameters.ARGON2_id)
+                    .withVersion(19)
+                    .withIterations(2)
+                    .withSalt(pepperByteArray)
```
seluianova, line 61: Nit-picky one:
If we use pepper instead of salt, shouldn't it be set through the withSecret() method?
(I feel like it won't make a difference, just maybe reduce the confusion a bit?)

Otherwise we pass withSalt() params into encodeWithoutSalt() function

backend/src/main/kotlin/app/ehrenamtskarte/backend/user/KoblenzUser.kt
```diff
@@ -0,0 +1,7 @@
+package app.ehrenamtskarte.backend.user
```
seluianova, line 1: maybe should be moved to userdata package, created for the user import endpoint?
or vice versa.
I think keeping both packages can be confusing

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
seluianova: maybe we should set it through the withSecret() function then?
I feel like it's a bit confusing otherwise. we set not-a-salt value as a salt and then pass it into encodeWithoutSalt() function 🙈

docs/CreateKoblenzHash.md
```diff
@@ -0,0 +1,61 @@
...
+Hash with Argon2id with the following parameters:
+
+| Parameter    | Value                                                                          |
+|--------------|--------------------------------------------------------------------------------|
+| Version      | 19                                                                             | 
+| Iterations   | 2                                                                              | 
+| Parallellism | 1                                                                              | 
+| Memory       | 19456 (19MiB)                                                                  | 
+| HashLength   | 32                                                                             | 
+| Salt         | Secret Salt will be shared with Koblenz<br/>for the example use `123456789ABC` |
```
seluianova, line 48: Is it finally salt or not 🙈

f1sh1918: We use here a secret pepper shared with Koblenz as a additional security layer.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/Argon2IdHasher.kt
f1sh1918: the problem is that we can not reproduce the same hash as we have in https://argon2.online/ if we use with secret. i wouldn't change this

docs/CreateKoblenzHash.md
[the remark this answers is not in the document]
f1sh1918: i'm gonna keep the calculated birthday here if thats okay for everyone

backend/src/main/kotlin/app/ehrenamtskarte/backend/user/KoblenzUser.kt
f1sh1918: moved that

docs/CreateKoblenzHash.md
seluianova: I just meant that the terms are a bit mixed up, but maybe it doesn't matter

f1sh1918: i think the same terms should be used like in the online hash creator from argon

seluianova (approved):

michael-markl:
> I believe we could still send the user data to the server and do the additional check only if we find the matching hash. Then the expensive argon2id/pbkdf verification only runs if the hash can be found. 

I just noticed that this is also not possible, as the user would know the "secret salt/pepper" to generate the hash. So I think we can only do rate limiting.

Issue #1489, opened 2024-06-12
Display note icon application header


dkehne:
As a user I want to see with one view if a note is maintained for an appliaction.

## Acceptance Criteria:
- I don't want to collapse the application to see if it's maintained.

PR #1500, opened 2024-06-20
1489: Note icon application header

f1sh1918:
### Short description

As a user i want to get a quick indicator if a note was set

### Proposed changes

- add note icon to the application header if a note was set

### Side effects

-none

### Resolved issues

Fixes: #1489

### Testing
1. Go to applications and create a note. Icon should be displayed

### DIFF
--- a/administration/src/bp-modules/applications/ApplicationCard.tsx
+++ b/administration/src/bp-modules/applications/ApplicationCard.tsx
@@ -10,7 +10,7 @@ import {
   SectionCard,
   Tooltip,
 } from '@blueprintjs/core'
-import { memo, useContext, useMemo, useState } from 'react'
+import { ReactElement, memo, useContext, useMemo, useState } from 'react'
 import styled, { css } from 'styled-components'
 
 import getMessageFromApolloError from '../../errors/getMessageFromApolloError'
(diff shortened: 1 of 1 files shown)


administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -148,6 +155,19 @@ const ApplicationCard = ({
     return !!blueCardJuleicaEntitlement
   }
 
+  const RightElement = (): ReactElement => {
+    return (
+      <RightElementContainer>
+        {application.note && <Icon icon='annotation' intent='none' />}
```
seluianova, line 161: might be out of scope, but I feel like it could be nice to display the content of the note in the tooltip maybe?

seluianova, line 161:
```suggestion
        {application.note && application.note.trim() && <Icon icon='annotation' intent='none' />}
```
as a nit-picky corner case, we could also check that the note is not just a space.
the use-case - if the user wanted to remove the note but accidentally left a space for example.

seluianova (approved):
added couple minor suggestions, but lgtm

administration/src/bp-modules/applications/ApplicationCard.tsx
f1sh1918: i thought about that. But it's out of scope. They can request that feature in the future

Issue #1462, opened 2024-05-28
Map vanishes when restart fom app-list


Docmale:
When  I open the list of  background apps choosing another app and come back to 'Ehrenamt' the same way the map disappeared (black screen of one physikal device [Samsung A53; V14] , white one on the other [Samsung Note 8;V9]). The android emulator dosen't show this behaviour but _sometimes_ crashes at this point. 
I will ad ios behaviour later on.

**To Reproduce**

1. Start Ehrenamt
2. open background list
3. Choose whatever other app
4. come back to Ehrenamt the same way
5. open Map when not already open
6. find black/white screen
7. for heeling remove Ehrenamt from background list and start it again

Related to:~https://github.com/maplibre/flutter-maplibre-gl/issues/327~
~Should be fixed by this pr: https://github.com/maplibre/flutter-maplibre-gl/pull/406~

**Update to latest maplibre release**
~It should be fixed by updating to https://github.com/maplibre/flutter-maplibre-gl/releases/tag/v0.20.0~
Looks like kotlin plugin for android has also to be updated

**Update:** its not fixed by maplibre v0.20 -> Blocked, waiting for upstream fix

This bug happens on android 9 and android 14. You can upload your apk to browserstack for testing

Docmale:
not reproducable on ios

f1sh1918:
Might be fixed with maplibre 0.26.0. Actually we could upgrade to 0.27.0, since this was merged and released
https://github.com/maplibre/flutter-maplibre-gl/pull/364

PR #1506, opened 2024-06-25
1462: Map vanishes app list

f1sh1918:
### Short description

On some device likes Samsung S24Ultra (Android 14) the map vanishes if you switch to app list and back.
Unfortunately this is not fixed by maplibre update and the issue is open for some time already.
So i implemented a little workaround that reloads the map on `resume`
The current state is not super nice, since selected stores on the map and the position will be lost.
I could also pass an initial store selection or anything but maybe anyone knows a better solution to avoid this.

### Proposed changes

- update maplibre and gradle
- add Observer to reload Map widget on AppLifeCycleState.resume

### Side effects

- Current state of the map f.e. selected store is lost

### Resolved issues

Fixes: #1462

### DIFF
--- a/frontend/android/app/build.gradle
+++ b/frontend/android/app/build.gradle
@@ -64,6 +64,10 @@ ext.setupVariant = { variant, buildConfigName ->
 android {
     compileSdkVersion 34
 
+    kotlinOptions {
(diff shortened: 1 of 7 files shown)


f1sh1918:
will maybe close this draft and hope that there will be an upstream fix for maplibre

Issue #1514, opened 2024-06-27
Use predefined platform for git tag


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Since we don't want to create git tags for all available platforms in a `bump-up` step in some scenario, we have to pass a particular tag f.e. `native | web | all` depending on what release workflow we trigger.

**Describe the solution you'd like**
- add a parameter for platform that will be used for creating tag, default 'all' since its the common case
- if no parameter is provided, tags for all platforms will be created

PR #1515, opened 2024-06-27
1514: Use predefined platform git tag

f1sh1918:
### Short description

Since we don't want to create git tags for all available platforms in a `bump-up step` in some scenarios, we have to pass a particular tags f.e.` native | web | all` depending on what release workflow we trigger.

### Proposed changes

- pass parameters for platform specific deployments
- you can pass string separated by "/" for platforms f.e. `web/native` if you need

### Dependency
depends on this pr to be merged
https://github.com/digitalfabrik/app-toolbelt/pull/14

### Side effects

- none

### Resolved issues

Fixes: #1514

### Testing (dev only)
1. (Before depending pr is merged) You can test it by replacing the following the repo url in `install_app_toolbelt`
with: https://github.com/digitalfabrik/app-toolbelt/archive/refs/heads/bump-up-add-platform-option.tar.gz
2. Choose a workflow that uses `bump_version` with `prepare_delivery: true` like `deliver_beta_backend_administration` but comment out all steps except `bump_version`
3. trigger workflow
4. Check for new git tag
https://github.com/digitalfabrik/entitlementcard/tags


ztefanie (approved):
Looks good, then toolbelt pr needs to be merged first.

Issue #1517, opened 2024-07-01
Activation of golden cards throws error


dkehne:
**Describe the bug**
Try to activate a golden card inside the Bavarian Volunteer App. It throws the attached error.

**Expected behavior**
The card should be activated and stored in the App.

**Screenshots**
![signal-2024-07-01-081242](https://github.com/digitalfabrik/entitlementcard/assets/4182469/ef6e9162-3f55-4a71-bcc8-0f2946c7143c)

PR #1518, opened 2024-07-01
1517: Failed gold card activation

f1sh1918:
### Short description

Gold cards can not be activated since `expirationDate = 0` isBefore today

### Proposed changes

- return false if `expirationDate = 0` for `isExpired()`

### Side effects

- none

### Resolved issues

Fixes: #1517

### Testing
- Create a gold card


f1sh1918:
closed in favor of #1519

michael-markl:
Expiration day should be null for golden cards. There must be something else wrong.

PR #1519, opened 2024-07-01
release-2024.5.3-hotfix

f1sh1918:
### Short description

Gold cards can not be activated since `expirationDate = 0` isBefore today

### Proposed changes

- return false if `expirationDate = 0` for `isExpired()`

### Side effects

- none

### Resolved issues

Fixes: #1517

### Testing
- Create a gold card


f1sh1918:
wrong release branch was used

michael-markl:
As I mentioned in the other pr, this is not the root of the cause.
ExpirationDay should be null for the golden card.

Please fix it in the card mutation service, and then add a migration script that sets those entries to null that have the value 0.

PR #1520, opened 2024-07-01
release-2024.6.1-hotfix:

f1sh1918:
### Short description

Gold cards can not be activated since the `expirationDate `  was wrongly set to `0`

### Proposed changes

- check if card has expirationDay and only set it then, else null
- add migration to fix cards that have `expirationDay = 0`
- add missing null check for startDay for staticCards

### Side effects

- none

### Resolved issues

Fixes: #1517

### Testing
- Create a gold card


michael-markl:
As I mentioned in the other pr, this is not the root of the cause.
ExpirationDay should be null for the golden card.

Please fix it in the card mutation service, and then add a migration script that sets those entries to null that have the value 0.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
```diff
@@ -70,7 +70,7 @@ class CardMutationService {
             CardRepository.insert(
                 hashedCardInfo,
                 activationSecretHash,
-                cardInfo.expirationDay.toLong(),
+                expirationDay = if (cardInfo.hasExpirationDay()) cardInfo.expirationDay.toLong() else null,
```
michael-markl, line 73: To make it consistent with the other optional fields:
```suggestion
                expirationDay = cardInfo.expirationDayOrNull?.toLong(),
```

Also, the same must be done for `createStaticVerificationCode`

f1sh1918: updated pr

ztefanie (approved):
Thanks for fixing this. Tested on Android emulator, works as expected.

backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/webservice/schema/CardMutationService.kt
michael-markl: @f1sh1918 I don't think so 🤔 at least I cannot see it yet.

f1sh1918: sorry ssh timeout :/ gonna push it soon

f1sh1918: @michael-markl now its pushed

michael-markl (approved):
If you have double checked that the DB migration does not update any (other) existing cards, this looks good to me :)

Issue #1507, opened 2024-06-26
Mobile scanner does not recognize barcodes


f1sh1918:
**Describe the bug**
On some Android 14 devices like Samsung Galaxy A23 5g the mobile scanner does not recognize barcodes

**To Reproduce**
1. Use Samsung Galaxy A23 5G
2. Activate Card
3. Scan the activation barcode
4. Nothing happens

**Expected behavior**
The barcode will be recognized and the card can be activated

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy A23 5G
 - OS: Android 14

**Additional context**
Related issue: https://github.com/juliansteenbakker/mobile_scanner/issues/1077

Update: This issue comes from upstream android camera issue and is device related
https://github.com/juliansteenbakker/mobile_scanner/issues/698

ztefanie:
Issue was again reported with these details:
Samsung Handy A 23 5 G.
Die Android Version ist 14.
Die Version der App ist 2024, 3.16 (die aktuellste über Play Store verfügbar)

ztefanie:
Happend again:
goldene Ehrenamtskarte.
Kaufbeuren 
App-Version 2024.5.5

vthejay:
The problem exists with the Version 2024.5.5. as well (the latest Version)
Device Samsung A23 5g
Android 14

vthejay:
with the test device Samsung A23 5G, I can reproduce the problem. The scanner does not recognize the QR code at all.
I have already activated the Developer Options on the Samsung A23.
I can get the logfiles through ADB later today.  :grinning: 
So, finally, we can reproduce this problem....
BTW, interesting: The test device has Android 13

f1sh1918:
There is a patch mentioned to get more logs but it looks like we first have to ugrade to recent version v5.1.1 which has some breaking changes

f1sh1918:
fixed by #1507

PR #1524, opened 2024-07-01
1507: Mobile scanner does not recognize barcodes

f1sh1918:
### Short description

Some android cameras don't recognize barcodes due to some upstream android camera issues.
This seem to be related to the defaultCamera resolution each device uses
https://github.com/juliansteenbakker/mobile_scanner/issues/698
Its not that easy to detect which cameras may be issued to not apply this on all android devices.

### Proposed changes

- provide a fixed camera resolution for specific android models

### Side effects

- aspect ratio is not correct. so we can discuss if this is good enough, because it looks ugly
- i played around with the values but didn't find a working solution with correct aspect ratio so far 

### Resolved issues

Related: #1507 

### Note
I'm not sure if we should set the cameraResolution to all android devices because it looks ugly. But its also difficult to maintain particular models, so let me know what you think. We could keep the issue open if new devices will be reported.

### Testing (dev only)
This bugfix is only testable with the particular device, but you can just check if the controller applies, if you add your model to `devicesWithoutQRCodeDetection` list and check if the resolution of the camera changes

### DIFF
--- a/frontend/lib/identification/qr_code_scanner/qr_code_scanner.dart
+++ b/frontend/lib/identification/qr_code_scanner/qr_code_scanner.dart
@@ -3,6 +3,7 @@ import 'dart:typed_data';
 
 import 'package:ehrenamtskarte/identification/qr_code_scanner/qr_code_scanner_controls.dart';
 import 'package:ehrenamtskarte/identification/qr_code_scanner/qr_overlay_shape.dart';
+import 'package:ehrenamtskarte/widgets/error_message.dart';
 import 'package:flutter/material.dart';
 import 'package:mobile_scanner/mobile_scanner.dart';
 
@@ -22,7 +23,7 @@ class QrCodeScanner extends StatefulWidget {
 }
 
 class _QRViewState extends State<QrCodeScanner> {
-  bool _hasCameraIssues = false;
+  late Future<bool> _hasCameraIssues;
 
   @override
   void initState() {
@@ -30,11 +31,7 @@ class _QRViewState extends State<QrCodeScanner> {
     // Workaround for https://github.com/juliansteenbakker/mobile_scanner/issues/698
     // Check once the qr code scanner was initialized if the device has camera issues
     // Depending on that set a controller with predefined camera solution to fix that qr code reading issues
-    WidgetsBinding.instance.addPostFrameCallback((_) async {
-      setState(() async {
-        _hasCameraIssues = await isDeviceWithCameraIssues();
-      });
-    });
+    _hasCameraIssues = isDeviceWithCameraIssues();
   }
 
   final MobileScannerController _controller = MobileScannerController(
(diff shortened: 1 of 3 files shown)


frontend/lib/identification/qr_code_scanner/qr_code_scanner.dart
```diff
@@ -21,11 +24,14 @@ class QrCodeScanner extends StatefulWidget {
 
 class _QRViewState extends State<QrCodeScanner> {
   final MobileScannerController _controller = MobileScannerController(
-    torchEnabled: false,
-    detectionSpeed: DetectionSpeed.normal,
-    formats: [BarcodeFormat.qrCode],
-    returnImage: false,
-  );
+      torchEnabled: false, detectionSpeed: DetectionSpeed.normal, formats: [BarcodeFormat.qrCode], returnImage: false);
+  final MobileScannerController _controllerCameraResolution = MobileScannerController(
```
f1sh1918, line 28: not sure how i can avoid defining different controllers twice since `cameraResolution` has no setter to change it

frontend/lib/util/android_utils.dart
```diff
@@ -12,3 +12,13 @@ Future<bool> certificateIsRequired() async {
   AndroidDeviceInfo androidInfo = await deviceInfo.androidInfo;
   return androidInfo.version.sdkInt < 25;
 }
+
+Future<bool> isDeviceWithCameraIssues() async {
+  if (!Platform.isAndroid) {
+    return false;
+  }
+  List<String> devicesWithoutQRCodeDetection = ['SM-A236B'];
```
michael-markl, line 20: There are more model IDs for the Samsung Galaxy A23. We should probably take all listed here:
https://storage.googleapis.com/play_public/supported_devices.html

I think as a workaround it's fine if it does not look great on these devices, but at least works well enough.

Have you checked whether a higher resolution would work too?

f1sh1918: Well it doesn't matter what resolution you take it's just important to set a default resolution. It will not really be used. I mean you can check it on your device by adding your model and test it. It's just changing the aspect ratio which looks weird but for some devices its ok-

I can add these devices, even we don't really know it they also cause this issue. Were only reported for A23 5G.
It would be really better to maintain that list somewhere where we could add it without triggering new release.

michael-markl: There are ~10 model IDs for the A23 5G.

Not sure if it's worth the effort to do more engineering around this workaround. We should rather try to fix the upstream issue if we want to invest more time :D

frontend/lib/identification/qr_code_scanner/qr_code_scanner.dart
```diff
@@ -20,12 +22,29 @@ class QrCodeScanner extends StatefulWidget {
...
-    formats: [BarcodeFormat.qrCode],
-    returnImage: false,
-  );
+      torchEnabled: false, detectionSpeed: DetectionSpeed.normal, formats: [BarcodeFormat.qrCode], returnImage: false);
+  final MobileScannerController _controllerPredefinedCameraResolution = MobileScannerController(
+      torchEnabled: false,
+      detectionSpeed: DetectionSpeed.normal,
+      formats: [BarcodeFormat.qrCode],
+      returnImage: false,
+      cameraResolution: const Size(640, 480));
```
michael-markl, line 47: I am not entirely sure if it works, but you could make the _controller non-final and simply overwrite if/once we detect it is a problematic device (instead of setting the `_hasCameraIssues` flag).

michael-markl (approved):
Looks good (enough) to me :D Let's see if this fixes it for the reports that we got.

If you have tested on a separate device and this should be mergable I think :D

frontend/lib/identification/qr_code_scanner/qr_code_scanner.dart
f1sh1918: That does not work

f1sh1918:
> Looks good (enough) to me :D Let's see if this fixes it for the reports that we got.
> 
> If you have tested on a separate device and this should be mergable I think :D

Yes since i adjusted some minor things, i will test it today again with the real device

f1sh1918:
@michael-markl i just tested it again and realized that this implementation needed a future builder to handle async functions in the `initState`
Not very nice to need so much additional code just for async functions in initState but i think its not avoidable
Some good infos initState and StateManagement in flutter
https://www.dhiwise.com/post/the-role-of-flutter-linitstate-in-state-management

Issue #1416, opened 2024-04-29
Create endpoint for user data


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Create an http put endpoint that receives user data.
**Describe the solution you'd like**
- create new table: userHash, startDate, endDate, invalid, lastUpdated (timestamp)
- create a endpoint and function that receives data with a csv body
- check data validity: hash length, valid date format

seluianova:
Input data example:
```
userHash, startDate, endDate, valid
dashj21sasd32, 12.05.2024, 12.10.2028, true
```

seluianova:
Do we want to have 'koblenz' in the name of the table? Like, `koblenzusers` ?

seluianova:
Might be nice to know how much data we expect in the CSV.
The processing time for 1000 lines takes about 2.82 seconds on my machine locally.

michael-markl:
Regarding hash validity checks: We could also require that the first entry is a dummy entry with a hash deduced from dummy data. I think for other hashes we can only check their length. (as I've written in #1499 I think we should not add these parameters to the hashes).

michael-markl:
> Re-tested for 20000 entries:
29.13 sec before warming up
17.78 sec after warming up
>
> We might want to think about the performance optimization then?

I don't think 30s would be a problem. It should only run once a week and we may require that it only runs at night for example.

seluianova:
**Another open question:**
Do we want to receive userHash as a full argon2id-encoded value?
Example: `$argon2id$v=19$m=16,t=2,p=1$MTIzNDU2Nzg5QUJD$UIOJZIsSL8vXcuCB82xZ5E8tpH6sQd3d4U0uC02DP40`
Or maybe it would make more sence to receive only the hash-part?
Considering that the params are static (are they?)

Issue #1417, opened 2024-04-29
Implement authentication for user data endpoint


seluianova:
**Is your feature request related to a problem? Please describe.**
Client authentication must be implemented for the user data endpoint. 
It should be only valid for one ProjektID.

**Describe the solution you'd like**
Authentication method is to be defined.
Possible options:
- TLS Client Certificate (preferred)
- HTTP Basic Authentication

**Describe alternatives you've considered**
TBD

**Additional context**

#blocked by #1491

f1sh1918:
@seluianova i added blocked since we can not create tokens yet, but it could also be done with a hardcoded token for the moment if we don't enable it for production

Issue #1491, opened 2024-06-13
Add a possibility to create a token for api requests


ztefanie:
**Is your feature request related to a problem? Please describe.**
A user that is logged in should be able to create a token which than can be used to make api requests to the protected "user data uploading" endpoint

**Describe the solution you'd like**
- Think about / discuss which role should have the rights to create such a token, either the project admin or a special role we introduce for this
- Add a function to create this token, decide which algorithm is good for this (secure)
- this token must be saved to the database hashed
- Add a section to the settings in the digitale druckerei, where an authenticated user can create this token and delete it, but not few it, as it is hashed and this is not possible, add a hint about this to the ui

Edit: For Koblenz it is ok if the existing admin role can create the api tokens.

f1sh1918:
@ztefanie are you working in this or can i take it since you are soon on vacation and so other issues depend on it

PR #1528, opened 2024-07-03
1416: Create endpoint for user data

seluianova:
### Short description
Create an http entpoint for Koblenz project to import user data

### Proposed changes
- create new table to store user data with the columns: userHash, startDate, endDate, revoked, projectId, lastUpdated
- create an endpoint that receives data in the csv file, validates it, and stores it in the database
- if userHash already exists in the database, the record is updated, otherwise it is created

**Questionable items**:

- I decided to create a POST endpoint instead of PUT because:
As I see it, PUT is typically used to update or create the target resource indicated in the URI. Like, PUT /user/{id}, for example.
But in our case it's more like a procedure call and in such case IMO it's more common to use POST.
But if it's already agreed upon, let me know, and I'll just change it to PUT.

- I named the table UserEntitlements because I think it will be more accurate than just Users or UserData.
But feel free to object.

- I named the column 'revoked' instead of 'invalid', because 'valid' sounds like data validity to me.
But feel free to object (2)

**Not included**:
- auto-tests (come next)
- userHash validation (to be added after #1433, with the potential to re-use some code from there maybe)
- authentication (to be done in #1417)

### Resolved issues
Fixes: #1416

### Testing
CSV data example:
```
userHash, startDate, endDate, revoked, regionId
dashj21sasd32, 12.05.2024, 12.10.2028, false, 1
```


backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0013_CreateUserEntitlementsTable.kt
```diff
@@ -0,0 +1,39 @@
+package app.ehrenamtskarte.backend.migration.migrations
```
f1sh1918, line 1: please merge recent main, since there already is a migration 13

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    private fun importData(csvParser: CSVParser) {
+        val headers = csvParser.headerMap.keys
+        val requiredColumns = setOf("userHash", "startDate", "endDate", "revoked")
+        if (!headers.containsAll(requiredColumns)) {
+            throw UserImportException("Missing required columns: ${requiredColumns - headers}")
+        }
+
+        transaction {
+            // TODO replace with koblenz.pass.app ?
```
f1sh1918, line 65: Please add in comment also that we have to send an apiToken (this validation is also missing for sure) and then we can validate which project sends data and set a projectId

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    private fun importData(csvParser: CSVParser) {
+        val headers = csvParser.headerMap.keys
+        val requiredColumns = setOf("userHash", "startDate", "endDate", "revoked")
+        if (!headers.containsAll(requiredColumns)) {
+            throw UserImportException("Missing required columns: ${requiredColumns - headers}")
+        }
+
+        transaction {
+            // TODO replace with koblenz.pass.app after #1428
```
f1sh1918, line 66: We have to receive also an valid api token https://github.com/digitalfabrik/entitlementcard/issues/1491 and then we can check which project it is. Please add this to the todo comment

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    private fun isValidUserHash(userHash: ByteArray): Boolean {
+        // TODO implement userHash validation after #1433
+        return true
+    }
+
+    private fun parseDate(dateString: String): Instant {
+        // TODO use timezone from Koblenz config after #1428
+        try {
+            return dateStringToStartOfDayInstant(dateString, "dd.MM.yyyy", ZoneId.systemDefault())
```
f1sh1918, line 106: Hm do we really need a timezone. i think the system default UTC should be fine?

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,111 @@
...
+import java.time.ZoneId
+import java.time.format.DateTimeParseException
+
+class UserImportHandler {
+
+    private val logger: Logger = LoggerFactory.getLogger(UserImportHandler::class.java)
+
+    fun handle(ctx: Context) {
+        try {
+            val files = ctx.uploadedFiles("file")
```
f1sh1918, line 26: We have to receive also a valid api token (connected to administrators) here for auth https://github.com/digitalfabrik/entitlementcard/issues/1491 and then we can check which project it is, check if this project is allowed (should be added as a flag in config.yml) and pass the projectID, else throw an error. Please add this todo. I think a separate issue to do the api token check would be good to create

seluianova: > I think a separate issue to do the api token check would be good to create

I think it supposed to be this task: https://github.com/digitalfabrik/entitlementcard/issues/1417

f1sh1918: ah perfect. then just refer to this issue

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
seluianova: true. no idea why I thought that

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/database/Schema.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.jetbrains.exposed.dao.IntEntityClass
+import org.jetbrains.exposed.dao.id.EntityID
+import org.jetbrains.exposed.dao.id.IntIdTable
+import org.jetbrains.exposed.sql.javatime.CurrentTimestamp
+import org.jetbrains.exposed.sql.javatime.timestamp
+
+object UserEntitlements : IntIdTable() {
+    val userHash = binary("userHash")
+    val startDate = timestamp("startDate")
+    val endDate = timestamp("endDate")
```
michael-markl, line 14: Please make these align with the [Schema of the Cards table](https://github.com/digitalfabrik/entitlementcard/blob/main/backend/src/main/kotlin/app/ehrenamtskarte/backend/verification/database/Schema.kt) We use integer numbers (`long`) for the number of days since 1970. There are helper functions to convert dates to and from this format.

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/database/Schema.kt
```diff
@@ -0,0 +1,33 @@
...
+    val revoked = bool("revoked")
+    val projectId = reference("projectId", Projects)
```
michael-markl, line 16: We should probably have a `regionId` rather than a project id here as cards are always assigned to a region rather than a project

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,113 @@
...
+import java.io.InputStreamReader
+import java.time.Instant
+import java.time.ZoneId
+import java.time.format.DateTimeParseException
+
+class UserImportHandler {
+
+    private val logger: Logger = LoggerFactory.getLogger(UserImportHandler::class.java)
+
+    fun handle(ctx: Context) {
```
michael-markl, line 24: Probably we should throw an error if we're on production for now.

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/database/Schema.kt
seluianova: Could you please tell me why?
For example, it will be less readable if we need to check the dates in the database for some support purposes. 
But what's the idea behind it?

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/database/Schema.kt
seluianova: I think this should be disscussed then.
Because the current idea was that we extract the projectId based on the authorization token of the user who performs the import.
Do we want to have a region-level token?
Or should be query regionId by projectId?
What if the project has more than 1 region?

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
seluianova: Done

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,113 @@
...
+        transaction {
+            // TODO as part of #1417: use projectId defined from the auth token
+            val project = ProjectEntity.find { Projects.project eq "showcase.entitlementcard.app" }.single()
+
+            for (entry in csvParser) {
+                if (entry.toMap().size != csvParser.headerMap.size) {
+                    throw UserImportException("The entry [${entry.joinToString(",")}] has missing data")
+                }
+
+                val userHash = entry.get("userHash").toByteArray()
```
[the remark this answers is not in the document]
seluianova: > assuming userHash will be simply be the Base64 encoded hash -- not using the argon2id encoding that contains the parameters and salt

I thought we will get userHash in argon2id encoding actually.
@ztefanie do you know that by chance?

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/database/Schema.kt
seluianova: > We might as well use the [date type of postgres](https://www.postgresql.org/docs/current/datatype-datetime.html), but I don't know whether exposed supports that.

Yes, it does! I have changed dates to date type

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private fun isValidUserHash(userHash: ByteArray): Boolean {
+        // TODO implement userHash validation after #1433
+        return true
+    }
+
+    private fun parseDate(dateString: String): LocalDate {
+        try {
+            return LocalDate.parse(dateString, DateTimeFormatter.ofPattern("dd.MM.yyyy"))
```
michael-markl, line 107: If the format is not yet set in stone, I'd prefer if we'd use an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date (as in "2024-12-24") (aka the most standard date format in all software)

seluianova: I would wait for @f1sh1918's reply as he gave me a sample data

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,113 @@
...
+                if (!isValidUserHash(userHash)) {
+                    throw UserImportException("Invalid userHash format: ${userHash.joinToString(",")}")
+                }
+
+                val startDate = parseDate(entry.get("startDate"))
+                val endDate = parseDate(entry.get("endDate"))
+                if (startDate.isAfter(endDate)) {
+                    throw UserImportException("Start date [$startDate] cannot be after end date [$endDate]")
+                }
+
```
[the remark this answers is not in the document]
seluianova: Nice idea, added

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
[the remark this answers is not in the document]
ztefanie: We came to the conlusion in my PR (#1433), that we will add all parameters, but the salt.

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/database/Schema.kt
f1sh1918: We have a user based token and the user belongs to a region and to a project. so thats fine.
If the project has more region the user that imports for what region the cards can be created.

michael-markl: So the idea is to retrieve the region from the "user" making the "PUT API request" and then storing this region in the table `UserEntitlement`?

f1sh1918: Now i understand the issue from victoria.
Lets say we use the import for a project like bavaria and we have one project manager or a similar role that doesn't apply to a region because it wants to import users/cards from different regions. Then in the import csv should be a field `regionId` and we just check to which `regionId` belongs the project and check if the api token of the user is valid for this project.
The api token is user based (https://github.com/digitalfabrik/entitlementcard/issues/1491) so it belongs to a project

Nevertheless we need a regionId to create a card and we should use it as a reference here.
To keep this as generic as possible we should probably send a `regionId/regionIdentifier` in the csv and if its valid set the reference

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
f1sh1918: The sample data was not my decision but provided by the customer and can be found in nextcloud.
You can just ping @dkehne that he should inform the customer about that we want to use this date format here

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/database/Schema.kt
seluianova: So, I make regionId column instead of projectId and fill it from csv.
And also add a validation, that the passed regionId belongs to the project, associated with the user.
Right?

seluianova: So, I make regionId column instead of projectId and fill it from csv.
And also add a validation, that the passed regionId belongs to the project, associated with the user.

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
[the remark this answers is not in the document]
seluianova: So, I leave the current implementation as is

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
seluianova: Leaving as is

seluianova:
All the comments have been addressed

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,120 @@
...
+                    throw UserImportException(entry.recordNumber, "Failed to validate userHash")
+                }
+
+                val startDate = parseDate(entry.get("startDate"), entry.recordNumber)
+                val endDate = parseDate(entry.get("endDate"), entry.recordNumber)
+                if (startDate.isAfter(endDate)) {
+                    throw UserImportException(entry.recordNumber, "Start date cannot be after end date")
+                }
+
+                val revoked = entry.get("revoked").toBoolean()
```
f1sh1918, line 95:
```suggestion
                val revoked = entry.get("revoked").toBooleanStrictOrNull()?: throw UserImportException(entry.recordNumber, "Revoked must be a boolean value")
```

f1sh1918 (approved):
Very good implementation. Tested with postman. 👍 
There is just one error to catch i think for revoked.
I'm not happy with the naming of the table tbh.
Added a proposal

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/TablesRegistry.kt
```diff
@@ -26,6 +27,7 @@ object TablesRegistry {
         AcceptingStores,
         Addresses,
         Contacts,
-        Categories
+        Categories,
+        UserEntitlements
```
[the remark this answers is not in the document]
seluianova: EntitledUsers - sound to me like it should contain user data.
While the table contains entitlements data.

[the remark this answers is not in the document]
f1sh1918: ok i'm also fine with your naming. no worries

backend/src/main/kotlin/app/ehrenamtskarte/backend/userdata/webservice/UserImportHandler.kt
```diff
@@ -0,0 +1,120 @@
...
+            // TODO as part of #1417: define the project by auth token
+            val project = ProjectEntity.find { Projects.project eq "bayern.ehrenamtskarte.app" }.single()
+            val regionsByProject = Regions.select { Regions.projectId eq project.id }.map { it[Regions.id].value }
+
+            for (entry in csvParser) {
+                if (entry.toMap().size != csvParser.headerMap.size) {
+                    throw UserImportException(entry.recordNumber, "Missing data")
+                }
+
+                val regionId = entry.get("regionId").toInt()
```
f1sh1918, line 79: i'm just not pretty sure if we should use sth not project based and use a lookup instead. Maybe at some point f.e. database migration from baselane or sth with less or more other regions from the bavaria project is needed and then we don't have the same regionId for koblenz.
Maybe we could use the `regionIdentifier` to get the `regionId` since it won't change @seluianova

seluianova: Just to make sure:
If for some reason we get the same userHash for different regions - 2 records will be created. Because there is a unique constraint - userHash + regionKey.
Do you think it's okay?

seluianova: Just to make sure:
If for some reason we get the same userHash for different regions - a record for each region will be created. Because I added a unique constraint: userHash + regionKey.
Do you think it's okay?

seluianova: Changed regionId to regionKey in the UserEntitlements table

seluianova: As discussed, I changed regionId to regionKey in the UserEntitlements table

And just to make sure:
If for some reason we get the same userHash for different regions - a record for each region will be created. Because I added a unique constraint: userHash + regionKey.
Do you see any issues with that?

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0014_CreateUserEntitlementsTable.kt
```diff
@@ -17,7 +17,7 @@ internal class V0014_CreateUserEntitlementsTable : Migration() {
                 "startDate" date NOT NULL,
                 "endDate" date NOT NULL,
                 "revoked" boolean NOT NULL,
-                "regionId" integer NOT NULL,
```
f1sh1918: oh looks like we had a misunderstanding. You can keep here the `regionId` we just use the `regionKey` to identify the correct region (id) from the regions table @seluianova

seluianova: Fixed 🙈

f1sh1918:
Tested the import again. everything fine from my side @seluianova

Issue #1508, opened 2024-06-26
Leading comma in the 1st line


Docmale:
**Describe the bug**
Missing Region befor aleading comma in the headline

**To Reproduce**
1. export statistic
2. open CSV-file
3. see no Region

**Expected behavior**
Title 'Region' is set befor leading comma

**Screenshots**
![image](https://github.com/digitalfabrik/entitlementcard/assets/21236324/e2a113f0-d7b8-40fe-9b8e-4dfdb4ce036e)

michael-markl:
The comma is fine but there should be a name for the column like "Stadt" or so in front of it :D

PR #1533, opened 2024-07-10
1508: Add region header to the CSV export of cards statistics

seluianova:
### Short description
Region header is missing in the CSV export of cards statictics

### Proposed changes
Add 'region' to the mapping of statistics keys

### Side effects
no?

### Resolved issues
Fixes: #1508


f1sh1918 (approved):
Tested on chrome. works fine 👍

PR #1534, opened 2024-07-10
1530: Change warning text in the application form

seluianova:
### Short description
Change warning text in the application form when the applicant and the approver are identical

### Side effects
no?

### Resolved issues
Fixes: #1530
Fixes: #1490


seluianova:
Added commit for #1490, as this is also a minor adjustment to the application form

ztefanie (approved):
Looks good, thanks you :+1:

Issue #1294, opened 2024-02-15
Rename `verification` package to `cards` in backend


sarahsporck:
**Is your feature request related to a problem? Please describe.**
In the backend project we have the verification package. This is because it contains logic regarding verification. However, this name is misleading as it also handles e.g. card creation. Therefore, rename it to `cards`.

**Describe the solution you'd like**
Rename verification package to cards.

**Describe alternatives you've considered**
Don't rename it.

sarahsporck:
Also check the graphql directories e.g. frontend/graphql-queries & administration/src/graphql

PR #1535, opened 2024-07-10
1294: Rename verification package to cards

seluianova:
### Short description
Rename verification package to cards because it better describes the purpose of the package

### Side effects
I tested the flow from creating the card in administration to activating the card in the app, it seems to work as before.
But not sure I haven't missed anything.

### Resolved issues
Fixes: #1294


ztefanie (approved):
Looks good :+1: 

Tested a quick workflow of creating a card and activating it, works as expected.

Issue #1318, opened 2024-02-27
Setup backend test database


sarahsporck:
To write tests for the endpoints we generally need a Database Connection.
That's why we need a test DB.
There are two approaches:
1. Use the local DB And fill it with test data (this is the approach the cms uses)
➕  db can be evaluated even after running the tests to simplify debugging
➕  can be used to load (fixed) test data for development testing
➖  local db could become corrupted if test data and other local data clashes. Especially if we use fixed unique keys in the test data this option can become unstable
➖ probably more effort to set up in the CI

2. Use an In-Memory DB with only the necessary data. This is often the quick and ea
➕ often easy setup
➕ static test data ensures that you always get the same result
➖ test performance can be bad and does not scale well with more tests

deen13:
Hey @sarahsporck, have you come across [testcontainers](https://testcontainers.com/)? It bridges the gap between local and in-memory database and is easy to integrate.

deen13:
[Testcontainers](https://testcontainers.com/) is also a good solution for bridging the gap between in-memory and local databases. Perhaps it could be helpful :)

Issue #1532, opened 2024-07-10
Add integration tests for backend


seluianova:
**Is your feature request related to a problem? Please describe.**
I think it would be nice to have integration tests for backend.
Javalin allows to implement that:
https://javalin.io/tutorials/testing

The question is, what database we would like to use for integration tests?

**Option-1**:
Database from our local environment.
In this case, the tests will change the database state. Not sure if that's desirable.
If we use this option, we can either accept these implications or revert the changes after each test.
Another disadvantage would be - migrations, if the database version is larger than the application version, we will have to resolve this problem manually.

**Option-2** (_sounds the best to me_):
Run a separate database container for tests.
This can be implemented using the Testcontainers framework: https://testcontainers.com
In this case we will have a clean database for every test run.
Which also means, we can easily apply migrations.

**Option-3**:
Mock Exposed framework.
But as far as I've tried, it doesn't seem like it can be easily mocked.

seluianova:
We already have an issue for the database setup: #1318

PR #1536, opened 2024-07-15
1318: Setup backend test database

seluianova:
### Short description
To add integration tests for the backend (#1532), we need to setup a database.
In my opinion, using test containers library is the most convenient way to implement that.

### Proposed changes
Add IntegrationTest base class, where database setup is performed. 
To add new tests that require a database, new test classes must extend the IntegrationTest class.

### Side effects
Would be nice to check if everything works on different OS.
I have only tested it on Ubuntu 22.04.4 LTS.

### Resolved issues
Fixes: #1318


f1sh1918 (commented):
Looks good to me.
Just one questions. How can i trigger all tests at once?
Just found out how to trigger every single one.
The test execution should be added to circle ci `check_backend` even these are only temporary tests.

Nicely done :) 👍

seluianova:
@f1sh1918 
All backend tests are run with `./gradlew test`, so the new tests are also included in CI already

f1sh1918 (approved):
Ok i think everything should be fine.

`gradle test` does not work for because of some jvm issue but that should rely on my local configuration
Have to fix that

seluianova:
@f1sh1918 try gradlew instead of gradle, it means “gradle wrapper”

seluianova:
@ztefanie could you please also check that you can run the new tests locally?

backend/src/test/kotlin/app/ehrenamtskarte/backend/IntegrationTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        assertEquals(2L, projectCount)
+    }
+
+    @Test
+    fun verifyRegionCount() {
+        var regionCount = 0L
+        transaction {
+            regionCount = RegionEntity.count()
+        }
+        assertEquals(94L, regionCount)
```
deen13, line 63: You might consider using the `@Transactional` annotation to wrap each of the annotated methods, which would render the transaction block unnecessary. This change would enhance the test code by allowing you to use immutable `val` instead of mutable `var`.

seluianova: We use transaction blocks throughout the project.
Not sure if it makes sense to use a different approach for tests.

seluianova: It also seems to me that we should begin the transaction exactly where it is needed.
We can have a long test scenario in which various actions are performed (sending http requests, etc.) and only at the end can be the database interaction.

f1sh1918:
maybe @steffenkleinle can test it, since steffi will be on vacation

f1sh1918:
> @f1sh1918 try gradlew instead of gradle, it means “gradle wrapper”

doesn't help. its my local jvm config. nevermind

backend/src/test/kotlin/app/ehrenamtskarte/backend/IntegrationTest.kt
deen13: That makes sense! You can still eliminate the mutable variable declaration by writing a transaction function that returns the results. This approach aligns better with the [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#immutability). :)

steffenkleinle (approved):
Tested, works for me!

backend/src/test/kotlin/app/ehrenamtskarte/backend/IntegrationTest.kt
seluianova: Thanks I will consider that! But let me leave it as is here, because I remove these particular tests right in the next PR :)

Issue #1538, opened 2024-07-17
Disable stack traces in JSON responses


svenseeberg:
**Describe the bug**
Errors on the server lead to stack traces in responses. This should not happen. Stack traces should be printed into log files (which in contrast are relatively empty).

**To Reproduce**
1. Go to https://staging.bayern.ehrenamtskarte.app/ after applying https://git.tuerantuer.org/DF/salt/pulls/181/files
2. Try to login
3. See response body for login request

Example: https://github.com/digitalfabrik/entitlementcard/issues/1537

**Expected behavior**
The stack trace should not be returned to the user, only printed into the log file.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: all
 - Browser all
 - Version all

**Smartphone (please complete the following information):**
 - Device: all
 - OS: all
 - Browser all
 - Version all

**Additional context**

PR #1539, opened 2024-07-19
Do not send stack traces in GraphQL responses

michael-markl:
### Short description

To remove the stack traces from the Graphql errors, I had to remove the usage of  `GraphqlErrorException` as  it always serializes the whole exception including stacktrace.

### Proposed changes
Make `GraphQLBaseException` no longer inherit from the above mentioned `GraphqlErrorException` and implement a custom DataFetcherExceptionHandler that only wraps unknown exceptions into `ExceptionWhileDataFetching` objects and converts the known ones (ie. the subclasses of GraphQLBaseException) into GraphQLErrors without wrapping it.

If an ExceptionWhileDataFetching occurs, the wrapped exception is rethrown and either handled by `handle` if it's a 40x error, or by Javalin (if it's a completely unexpected exception) and then a 500 is returned and the exception is logged by Javalin.

### Side effects

Hopefully none. 

### Resolved issues

Fixes: #1538


f1sh1918 (approved):
Looks fine to me.
Responses do not contain stack traces anymore.
Nicely done 👍

Issue #1492, opened 2024-06-17
Add option to import from freinet


ztefanie:
See this csv created from freinet, it should be possible to add this in the digitale druckerei.

To test this you the csv files named bulk-import-freinet or bulk-import-freinet2 here: https://github.com/digitalfabrik/entitlementcard/tree/main/administration/resources/cards

PR #1541, opened 2024-07-23
1492: Add bulk import from freinet

ztefanie:
### Short description
Adding bulk import csv option from freinet csv format

### Proposed changes
- Added new route "import/?freinet=true"
- Added project config setting to enable freinet import
- Add function that adjusted freinet csvs to expected csv input format
- Add button to import from freinet
- Tests are missing, I will create a ticket for them if i will not manage to add some this week before my vacation

### Side effects
- Maybe standart csv import is somehow affected

### Resolved issues

Fixes: #1492


administration/src/bp-modules/cards/ImportCardsFromFreinetController.tsx
```diff
@@ -0,0 +1,67 @@
...
+const mergeFirstAndLastnameIntoNewColumn = (line: string[], csvHeader: string[], nameCoulmnName: string) => {
+  const columnNameDoesNotExist = csvHeader.indexOf(nameCoulmnName) == -1
+  if (columnNameDoesNotExist) {
+    csvHeader[csvHeader.length] = nameCoulmnName
+  }
+
+  const indexVorname = csvHeader.indexOf(FREINET_FIRSTNAME_COLUMN_HEADER)
+  const indexNachname = csvHeader.indexOf(FREINET_LASTNAME_COLUMN_HEADER)
+  const fullname = `${line[indexVorname]} ${line[indexNachname]}`
+  line[csvHeader.indexOf('Name')] = fullname.trim()
```
f1sh1918, line 18: shouldn't this also be `nameColumnName` instead of hardcoded string

administration/src/bp-modules/cards/ImportCardsFromFreinetController.tsx
```diff
@@ -0,0 +1,67 @@
...
+const FREINET_LASTNAME_COLUMN_HEADER = 'nachname'
+const FREINET_CARDTYPE_COLUMN_HEADER = 'inhaber_ehrenamtskarte'
+
+const mergeFirstAndLastnameIntoNewColumn = (line: string[], csvHeader: string[], nameCoulmnName: string) => {
+  const columnNameDoesNotExist = csvHeader.indexOf(nameCoulmnName) == -1
+  if (columnNameDoesNotExist) {
+    csvHeader[csvHeader.length] = nameCoulmnName
+  }
+
+  const indexVorname = csvHeader.indexOf(FREINET_FIRSTNAME_COLUMN_HEADER)
```
f1sh1918, line 15: maybe we should add some error handling if columns do not exists

administration/src/bp-modules/cards/ImportCardsFromFreinetController.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const renameExpirationDateHeader = (csvHeader: string[], expiryColumnName: string) => {
+  const indexExpirationDate = csvHeader.indexOf(FREINET_EXPIRATION_DATE_COLUMN_HEADER)
+  if (indexExpirationDate != -1) {
+    csvHeader[indexExpirationDate] = expiryColumnName
+  }
+}
+
+const getCardTypeByFreinetValue = (line: string[], indexCardTypeFreinet: number): string =>
+  line[indexCardTypeFreinet].includes('Gold') || line[indexCardTypeFreinet].includes('gold')
```
f1sh1918, line 29:
```suggestion
line[indexCardTypeFreinet].toLowerCase().includes('gold')
```

f1sh1918 (commented):
Added some comments.
In general it think we should have an error handling if the column names do not exist
If the column names in the header (f.e. vorname, nachname) do not match our format, i just get undefined name but can create cards.

image

f1sh1918:
@ztefanie 
i used the import files in the mentioned issue. But i get all these empty lines, caused by `;`

image

ztefanie:
> @ztefanie i used the import files in the mentioned issue. But i get all these empty lines, caused by `;`
> 
> A second (optional) thing could be to set a `null` for the columnNames mail extension (my issue) `extensionColumnNames: ['Kartentyp', null, null],` to not show the mail notification column We could think about providing this as a feature that they could export the mail address of the user and send the mail notification also for bulk imports. But i think we didn't make a decision on this
> 
> image

Yes both is as it should be. 
In the example input are empty lines and they should be displayed like this.
About the email notification, we can do this if they request it at a later point.

f1sh1918 (approved):
Alright works as expected 👍 

For later testing i think it could help to also provide a working csv example.
Sorry i was a bit confused while testing it ^^

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/***
+ * Converts Freinet CSV export data into valid input for CSV import
+ * @param line The line of the Freinet CSV export
+ * @param csvHeader the header of the Freinet CSV Export
+ *
+ * Format of the input must be:
+ *      columns with the header "vorname" and "nachname" must exist
+ *      column with the header expiration date must exist and be called "eak_datum"
+ *      column with the name "inhaber_ehrenamtskarte" can exist, if so it is used, if not the expiration date is used
```
steffenkleinle, line 58: :question: why is the `inhaber` and the `expiration date` interchangable?

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
```diff
@@ -0,0 +1,66 @@
+import { BAVARIA_CARD_TYPE_GOLD, BAVARIA_CARD_TYPE_STANDARD } from '../../cards/extensions/BavariaCardTypeExtension'
+import { ProjectConfig } from '../../project-configs/getProjectConfig'
+
+const FREINET_EXPIRATION_DATE_COLUMN_HEADER = 'eak_datum'
+const FREINET_FIRSTNAME_COLUMN_HEADER = 'vorname'
+const FREINET_LASTNAME_COLUMN_HEADER = 'nachname'
+const FREINET_CARDTYPE_COLUMN_HEADER = 'inhaber_ehrenamtskarte'
+
+const mergeFirstAndLastnameIntoNewColumn = (line: string[], csvHeader: string[], nameColumnName: string) => {
```
steffenkleinle, line 9: :upside_down_face: Is all of this done in place as no copies or anything are returned? Wouldn't it be a lot safer to work with copies? Looks kindof error prone to me.
E.g. if the header is adjusted in an early iteration, doesn't that have implications for later iterations?

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
```diff
@@ -0,0 +1,66 @@
...
+  const columnNameDoesNotExist = csvHeader.indexOf(nameColumnName) == -1
+  if (columnNameDoesNotExist) {
+    csvHeader[csvHeader.length] = nameColumnName
+  }
+
+  const indexVorname = csvHeader.indexOf(FREINET_FIRSTNAME_COLUMN_HEADER)
+  const indexNachname = csvHeader.indexOf(FREINET_LASTNAME_COLUMN_HEADER)
+  const fullname = `${line[indexVorname] ?? ''} ${line[indexNachname] ?? ''}`
```
steffenkleinle, line 17: :question: So it is okay if there is only a first or last name?

steffenkleinle, line 17:
```suggestion
  ```suggestion
  const indexFirstName = csvHeader.indexOf(FREINET_FIRSTNAME_COLUMN_HEADER)
  const indexLastName = csvHeader.indexOf(FREINET_LASTNAME_COLUMN_HEADER)
  const fullname = `${line[indexFirstName] ?? ''} ${line[indexLastName] ?? ''}`
```

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
```diff
@@ -0,0 +1,66 @@
...
+  line[csvHeader.indexOf(nameColumnName)] = fullname.trim()
+}
+
+const renameExpirationDateHeader = (csvHeader: string[], expiryColumnName: string) => {
+  const indexExpirationDate = csvHeader.indexOf(FREINET_EXPIRATION_DATE_COLUMN_HEADER)
+  if (indexExpirationDate != -1) {
```
steffenkleinle, line 23: Is the linting somehow broken? Or is it just configured way less strict?
```suggestion
  if (indexExpirationDate !== -1) {
```
A few more below.

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const setCardType = (line: string[], csvHeader: string[], cardTypeColumnName: string, expiryColumnName: string) => {
+  const indexCardTypeFreinet = csvHeader.indexOf(FREINET_CARDTYPE_COLUMN_HEADER)
+
+  const hasValidCardTypeColumn = csvHeader.indexOf(cardTypeColumnName) != -1
+  if (!hasValidCardTypeColumn) {
+    csvHeader[csvHeader.length] = cardTypeColumnName
+  }
+  const indexCardType = csvHeader.indexOf(cardTypeColumnName)
```
steffenkleinle, line 42:
```suggestion
  const indexCardType = csvHeader.indexOf(cardTypeColumnName)
  if (indexCardType === -1) {
    csvHeader[csvHeader.length] = cardTypeColumnName
  }

```

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
```diff
@@ -0,0 +1,66 @@
...
+  const hasFreinetCardType = indexCardTypeFreinet != -1
+
+  line[indexCardType] = hasFreinetCardType
+    ? getCardTypeByFreinetValue(line, indexCardTypeFreinet)
+    : (line[indexCardTypeFreinet] = getCardTypeByExpirationDate(line, csvHeader.indexOf(expiryColumnName)))
```
steffenkleinle, line 47: :wrench: An assignment as else value in a ternary looks very weird. I think this would be better:
```suggestion
if (hasFreinetCardType) {
  line[indexCardType] = getCardTypeByFreinetValue(line, indexCardTypeFreinet)
} else {
  line[indexCardTypeFreinet] = getCardTypeByExpirationDate(line, csvHeader.indexOf(expiryColumnName))
}
```
Or am I missing something here?

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
ztefanie: It is due to the format Freinet uses. The "inhaber_ehrenamtskarte" is set to either "Blau" or "Gold". Its just the what the export of freinet looks like. If this value is not set,w e use the expiration date. If an expiration date exists, it is a blue card. If no expiration date is set, it is a gold card.

administration/src/Router.tsx
```diff
@@ -37,6 +37,8 @@ const Main = styled.div`
   }
 `
 
+export const FREINET_PARAM = 'freinet'
```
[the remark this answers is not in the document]
ztefanie: Well it is the name of the URL-param so i think the Router is the best location to define it. Where do you think this should be defined?

administration/src/bp-modules/cards/CreateCardsController.tsx
```diff
@@ -29,6 +33,13 @@ const CreateCardsController = () => {
       <Buttons vertical>
         <CardFormButton text='Einzelne Karten erstellen' icon='add' onClick={() => navigate('./add')} />
         <CardFormButton text='Mehrere Karten importieren' icon='upload' onClick={() => navigate('./import')} />
+        {freinetCSVImportEnabled && (
+          <CardFormButton
+            text='Karten aus Freinet Export importieren'
```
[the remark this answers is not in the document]
ztefanie: I see your point, but I decided to make this name really self explaining.

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
ztefanie: Yes it does impact later iterations, but this is handled and works as expected. You are right it may be not the best implementation, but we are not planning to have this feature for a ever or a long time, it is just a workaround, we will remove this as soon as Freinet and entitlemtcard are connected.

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
ztefanie: yes it is, the users will see the cards and need to check the manually if they are really correct.

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
steffenkleinle: Thanks for explaining, makes totally sense :+1:

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
ztefanie: but you are right, i will add a check

administration/src/bp-modules/cards/ImportCardsFromFreinetController.ts
ztefanie: Thanks you are right. I fixed this.

administration/src/Router.tsx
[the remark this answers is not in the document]
steffenkleinle: I don't know either :see_no_evil: Probably its fine and can be adjusted later if needed.

steffenkleinle (commented):
Tested on firefox. I can open the CSV you linked in the issue (https://github.com/digitalfabrik/entitlementcard/issues/1492#issuecomment-2199477046), but it looks like this:
image

So it checked that there are invalid values and therefore blocks creating the QR codes, but I am not able to change the names or anything? Am I doing something wrong?

If this is supposed to happen in another issue/PR, it works as expected!


> Tests are missing, I will create a ticket for them if i will not manage to add some this week before my vacation

Did you manage to do this?

steffenkleinle:
And perhaps you could add the csv here too? https://github.com/digitalfabrik/entitlementcard/tree/main/administration/resources/cards

f1sh1918:
> Tested on firefox. I can open the CSV you linked in the issue ([#1492 (comment)](https://github.com/digitalfabrik/entitlementcard/issues/1492#issuecomment-2199477046)), but it looks like this: image
> 
> So it checked that there are invalid values and therefore blocks creating the QR codes, but I am not able to change the names or anything? Am I doing something wrong?
> 
> If this is supposed to happen in another issue/PR, it works as expected!
> 
> > Tests are missing, I will create a ticket for them if i will not manage to add some this week before my vacation
> 
> Did you manage to do this?

this is expected. just remove the ";" from the csv and create a card :) @steffenkleinle

steffenkleinle (approved):

f1sh1918:
resolved merge conflicts and merge @ztefanie to put it in the next release

PR #1542, opened 2024-07-23
Add react lint check to pre commit

ztefanie:
### Short description
Added react linting to pre commit hook.

### DIFF
--- a/.github/hooks/pre-commit
+++ b/.github/hooks/pre-commit
@@ -5,7 +5,7 @@
 # ln -s ../../.github/hooks/pre-commit .git/hooks/pre-commit
 #
 
-REACT_FILES=$(git diff --cached --name-only administration/)
+REACT_FILES=$(git diff --cached --name-only administration/ | sed 's| |\\ |g')
 
 if [ ! -z "$REACT_FILES" ]; then
   cd administration && yarn lint && cd ..
--- a/.github/hooks/pre-commit
+++ b/.github/hooks/pre-commit
@@ -5,7 +5,7 @@
 # ln -s ../../.github/hooks/pre-commit .git/hooks/pre-commit
 #
 
-REACT_FILES=$(git diff --cached --name-only administration/ | sed 's| |\\ |g')
+REACT_FILES=$(git diff --cached --name-only -- administration/ | sed 's| |\\ |g')
 
 if [ ! -z "$REACT_FILES" ]; then
   cd administration && yarn lint && cd ..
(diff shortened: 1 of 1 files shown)


.github/hooks/pre-commit
```diff
@@ -5,6 +5,12 @@
 # ln -s ../../.github/hooks/pre-commit .git/hooks/pre-commit
 #
 
+REACT_FILES=$(git diff --cached --name-only administration/ | sed 's| |\\ |g')
+
+if [ ! -z "$REACT_FILES" ]; then
+  cd administration && yarn lint && cd ..
```
seluianova, line 11: maybe I'm doing something wrong, but if I change any ts files in administration and run this script, if-condition is false and yarn lint command is not executed

.github/hooks/pre-commit
```diff
@@ -5,6 +5,12 @@
 # ln -s ../../.github/hooks/pre-commit .git/hooks/pre-commit
 #
 
+REACT_FILES=$(git diff --cached --name-only administration/ | sed 's| |\\ |g')
```
seluianova, line 8:
```suggestion
REACT_FILES=$(git diff --cached --name-only -- administration/ | sed 's| |\\ |g')
```

I guess we also need -- here:

```
fatal: ambiguous argument 'administration/': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

f1sh1918: hm i expected this should work `bash .github/hooks/pre-commit` but doesn't

ztefanie: yes it does work, with this command: bash .github/hooks/pre-commit
but it will do nothing if there are not staged files and skip all steps and therefore create no output

f1sh1918: i think this proposed adjustment has to be done. Rest looks fine to me then

seluianova: I actually meant, how to make it work automatically, as a hook.
For example, I make some changes in administration and click 'commit' in Idea.
As I understand, there must be some output from this hook, but I don't get where to look for it.

ztefanie: You need to set you hook path in the git config to the .github/hooks directory. You can do this by
`git config --local core.hooksPath .github/hooks`

f1sh1918 (approved):

seluianova (approved):

PR #1546, opened 2024-07-26
Fix thread exhaustion/leak

### LINKED ISSUE #1544: Memory leaks backend
**Describe the bug**
Since the release 2024.5.1 we are facing memory leaks on production.
https://github.com/digitalfabrik/entitlementcard/releases/tag/2024.5.1-all
Rollout of the update on production: 2024-06-07 - 11am
It may be related to the javalin update, but its not clear what causes this issue
Currently the backend service will be restarted (scheduled) every 12h as a workaround.

**To Reproduce**
1. Go to grafana and check increasing memory usage
2. Also check the server logs if you can find something

**Expected behavior**
Memory usage is not constantly increasing

**Screenshots**
image

**Additional context**
We already tried to fix it by a possible issue with jetty (used by javalin) but it seems like it didn't fix the issue
https://github.com/digitalfabrik/entitlementcard/pull/1510

Check directly grafana here:
https://grafana.tuerantuer.org/d/l3Gr9b5Vk/server-vitals-with-host-selection?orgId=1&refresh=5m&var-HOST=entitlementcard.tuerantuer.org&viewPanel=8&from=now-24h&to=now

maxammann:
### Short description

We create many MatomoTrackers which require manual cleanup/closing. Since some version the tracker cleans up after itself automatically when created in a try-catch resource construct: https://github.com/matomo-org/matomo-java-tracker/issues/215#issuecomment-1986844774

### Proposed changes

- Upgrade matomo dependency
- make sure we only have a single tracker

### Resolved issues

Fixes #1544

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
@@ -22,8 +22,8 @@ import java.net.URI
 import java.util.concurrent.ExecutionException
 
 object Matomo {
-    val logger = LoggerFactory.getLogger(Matomo::class.java)
-    var tracker: MatomoTracker? = null
+    private val logger = LoggerFactory.getLogger(Matomo::class.java)
+    private var tracker: MatomoTracker? = null
 
     private fun getTracker(config: BackendConfiguration): MatomoTracker {
         val tracker = tracker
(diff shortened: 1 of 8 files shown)


maxammann:
This PR requires a config update

f1sh1918:
> This PR requires a config update

Can you create a salt issue or even better pr for that? :)

maxammann:
done :)

michael-markl (approved):
LGTM (except we have to be careful with deployment of config & backend order)

backend/src/main/kotlin/app/ehrenamtskarte/backend/matomo/Matomo.kt
```diff
@@ -22,12 +23,19 @@ import java.util.concurrent.ExecutionException
 
 object Matomo {
     val logger = LoggerFactory.getLogger(Matomo::class.java)
+    var tracker: MatomoTracker? = null
```
f1sh1918, line 26: (optional) tracker could be private var

f1sh1918 (approved):
Looks good to me
also approved salt pr. 
You can merge both

Issue #1223, opened 2023-11-22
Error log if there are no applications


f1sh1918:
**Describe the bug**
Receive an error log when fetching data for `/application` and `/verification`

**To Reproduce**
1. create an application
2. click on the mail link to view your application
3. remove some chars from the magic link
4. Error can be found in the backend log

**Expected behavior**

**LOG**
```js
Nov 22 13:46:39 entitlementcard.tuerantuer.org backend[120716]: [JettyServerThreadPool-620] ERROR app.ehrenamtskarte.backend.common.webservice.GraphQLExceptionHandler - Exception while fetching data (/application) : null
```

michael-markl:
> @sarahsporck the only place where we use /application is for the applicant view attachment

I am pretty sure, this log does not come from the application attachment endpoint (because this would not generate a graphql error afaics).
The `/application` in the graphql error message says, that the graphql endpoint could not retrieve the property `application` of the specified query/mutation.
Queries that have this property:
* https://github.com/digitalfabrik/entitlementcard/blob/main/administration/src/graphql/applications/getApplication.graphql#L2
* https://github.com/digitalfabrik/entitlementcard/blob/main/administration/src/graphql/applications/getApplicationByApplicant.graphql#L2

The solution would be to find where the exception is thrown and instead do a proper error handling, I think?

f1sh1918:
it is thrown when the passed `accessKey` can't be found in the database and that happens if the a broken link was opened

f1sh1918:
i just updated the steps to reproduce.
This error is probably connected with broken links the user may receive via mail.
I gonna block this issue due to this pr https://github.com/digitalfabrik/entitlementcard/pull/1378 which maybe solves the issue or at least let it less occur.

f1sh1918:
we still have link 60-80 logs within 24hours. Probably due to broken link. We could check if the `accessKey` length is valid

f1sh1918:
are there any opinions to that @michael-markl , @ztefanie or @seluianova

michael-markl:
Is this the only thing that is logged? 
```
Nov 22 13:46:39 entitlementcard.tuerantuer.org backend[120716]: [JettyServerThreadPool-620] ERROR app.ehrenamtskarte.backend.common.webservice.GraphQLExceptionHandler - Exception while fetching data (/application) : null
```

f1sh1918:
yes we could maybe try to add a log for the `accessKey` that is used

[...]

PR #1553, opened 2024-07-30
1223: Log no applications

f1sh1918:
### Short description

We are facing some issues that we get a lot of server logs that no applications where found.
To be able to better investigate why this happens we log the `accessKeys` of the applicants and verifiers that cannot get applications.

### Proposed changes

- add logger if `application == null` that logs ip address and accessKey
- some small refactoring to clarify which accessKeys will be passed

### Side effects

- none

### Resolved issues

relates: #1223 

### Testing (dev only)
1. Create an application that needs an organization to verify
2. Open the verification and the application link and check that everything is still working
3. Remove some characters of the `accessKey` in the links and check the backend log

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
@@ -7,7 +7,6 @@ import app.ehrenamtskarte.backend.application.database.Applications
 import app.ehrenamtskarte.backend.application.webservice.schema.view.ApplicationView
 import app.ehrenamtskarte.backend.application.webservice.schema.view.JsonField
 import app.ehrenamtskarte.backend.application.webservice.utils.ExtractedApplicationVerification
-import app.ehrenamtskarte.backend.auth.webservice.schema.ResetPasswordMutationService
 import app.ehrenamtskarte.backend.common.database.sortByKeys
 import app.ehrenamtskarte.backend.common.webservice.GraphQLContext
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidLinkException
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -97,29 +100,43 @@ object ApplicationRepository {
         }
     }
 
-    fun getApplicationByApplicant(accessKey: String): ApplicationView {
+    fun getApplicationByApplicant(accessKey: String, dfe: DataFetchingEnvironment): ApplicationView {
+        val logger = LoggerFactory.getLogger(ResetPasswordMutationService::class.java)
```
seluianova, line 104:
```suggestion
        val logger = LoggerFactory.getLogger(ApplicationRepository::class.java)
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -97,29 +100,43 @@ object ApplicationRepository {
...
+                throw InvalidLinkException()
+            } else {
+                application.let { ApplicationView.fromDbEntity(it) }
+            }
         }
     }
 
-    fun getApplicationByApplicationVerificationAccessKey(applicationVerificationAccessKey: String): ApplicationView {
+    fun getApplicationByApplicationVerificationAccessKey(applicationVerificationAccessKey: String, dfe: DataFetchingEnvironment): ApplicationView {
+        val logger = LoggerFactory.getLogger(ResetPasswordMutationService::class.java)
```
seluianova, line 119:
```suggestion
        val logger = LoggerFactory.getLogger(ApplicationRepository::class.java)
```

f1sh1918: :D copy paste. thx

seluianova (approved):
I still think moving the logger declaration outside of the functions would be better, but lgtm anyway :)

Issue #1547, opened 2024-07-29
Add project-config administration


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Create a basic project config for koblenz in administration.

**Describe the solution you'd like**
- copy the project-config from nuernberg
- rename `name` `projectId` by replacing nuernberg by Koblenz
- Extensions: StartDayExtension, BirthdayExtension, RegionExtension (also adjust the extension column names), new extension aktenzeichen will be added here #1419
- `csvExport` can be disabled and has not to be configured
- `cardStatistics` can be disabled

**Describe alternatives you've considered**

**Additional context**

PR #1554, opened 2024-07-30
1547: Add project config for koblenz

f1sh1918:
### Short description

To be able to start the druckerei with koblenz config we need a basic project configuration

### Proposed changes

- added basic project config for koblenz
- added favicon icon
- added runConfigs for the project roles (not sure if we need anything except project admin, since cards will be created via self service portal)
- remove ability for card creation in the administration, since they will be created in the user portal

### Side effects

- none

### Resolved issues

Fixes: #1547 

### Testing
- Create users and login


.idea/runConfigurations/Create_project_admin__env_local_buildConfig_koblenz_.xml
seluianova, line 1: just wondering, for nurnberg we only have a region admin.
won't it be the same for koblenz?

administration/src/project-configs/bayern/config.ts
```diff
@@ -83,6 +83,7 @@ const config: ProjectConfig = {
       primaryColorLight: '#c6c0d8',
     },
   },
+  cardCreation: true,
```
seluianova, line 86: nit-picky one:
maybe to follow the style we already have
```suggestion
cardCreation: {
  enabled: false,
},
```

f1sh1918: i think we just need this object style if we add some additional parameters/configs if it is enabled. Which we don't do here, so i think thats not needed here

steffenkleinle (approved):
LGTM.
Tested creating a user via CLI and UI for koblenz in firefox. Login works as expected.
I just had to switch between the projects, otherwise I get `Server nicht erreichbar`.

administration/src/project-configs/koblenz/pdf.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const renderCardHash = ({ cardInfoHash }: InfoParams) => {
+  return cardInfoHash
+}
+
+// TODO 1422 Create PDF for koblenz
+
+const pdfConfiguration: PdfConfig = {
+  title: 'Koblenz-Pässe',
```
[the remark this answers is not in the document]
f1sh1918: Indeed we have a static and a dynamic card for koblenz and this is used as a doc name of the pdf. we have the same for nuernberg.

f1sh1918:
> LGTM. Tested creating a user via CLI and UI for koblenz in firefox. Login works as expected. I just had to switch between the projects, otherwise I get `Server nicht erreichbar`.

Dunno why the switch didn't work directly. I checked that again by removing the local storage variable and start administration again. for me it works directly if i click on switch to "Koblenz"

PR #1556, opened 2024-07-31
Fix protobuf version and adjust docs

steffenkleinle:
### Short description

I had to upgrade the protobuf version to get it to work on my machine. Let me know if that is alright or whether I should revert it.

### Proposed changes

- Upgrade protobuf
- Adjust docs

### Side effects

None.

### Resolved issues

Fixes: None.


backend/build.gradle.kts
```diff
@@ -24,7 +24,7 @@ repositories {
 }
 
 dependencies {
-    implementation("com.google.protobuf:protobuf-kotlin:3.25.2")
+    implementation("com.google.protobuf:protobuf-kotlin:4.27.2")
```
seluianova, line 27: Remembering how I struggled with the protobuff setup, I'm scared to test this change 😄

f1sh1918 (approved):
Tested:
- Running all protobuf creation scripts
- Created a card
- Activated a card

Looks like it works fine. Maybe someone else should doublecheck but i couldn't see any other breaking changes that may cause issues

seluianova (approved):
Seems to work for me

backend/build.gradle.kts
steffenkleinle: @seluianova I get that, would be super cool if you could still test this though please :pray:

seluianova: maybe would be helpful to specify the exact version of proto compiler in the dev instruction?

Issue #1381, opened 2024-04-08
strip spaces around names before comparison


Docmale:
**Is your feature request related to a problem? Please describe.**
The comparison of the applicant's name with the name of the contact person in the organization fails if a name is accompanied by a space.

**Describe the solution you'd like**
Leading and trailing spaces should be removed before comparison

**Describe alternatives you've considered**
Leave as is.

Consider removing all non word character(hyphens, dots) and convert to lowercase letters before comparison:
Dr. Hans-Peter Müller Lüdenscheid -> drhanspeterüllerlüdenscheid 
Dr Hans Peter müller Lüdenscheid ->  drhanspeterüllerlüdenscheid

michael-markl:
Where does this comparison take place?

f1sh1918:
> Where does this comparison take place?

https://github.com/digitalfabrik/entitlementcard/pull/1252/files#

PR #1559, opened 2024-08-01
1381: Make name comparison more robust

steffenkleinle:
### Short description

Make name comparison more robust.

### Proposed changes

- Use normalize-strings
- Trim whitespaces before comparing
- Remove special characters before comparing

### Side effects

None.

### Resolved issues

Fixes: #1381.


f1sh1918 (approved):
Looks good 👍

Issue #1487, opened 2024-06-06
Improve sorting store list


Docmale:
**Is your feature request related to a problem? Please describe.**
As a user i want to sort the store item either alphabetically or by distance. In the current implementation the sorting is not clear and can not be changed

**Describe the solution you'd like**
- Indicate in the headline of the store list (currently "Suchergebnisse" ) whether its sorted alphabetically or by distance by adding the particular label
- Show the FAB button (that only will be displayed once after permitting location) permanently.
- Sort the store list by default alphabetically if location permission is not given
- Sort the store list by store distance if location permission is given
- Depending on which sorting is applied, give the user the ability with the FAB button to change the sorting

**Additional context**

f1sh1918:
For me the button only disapears if it was clicked once.
I agree that it might be confusing that there is no "indicator" which sorting was used.
I think we should sort them automatically nearby if user location is available and alphabetically if user location is disabled and indicate the sorting somewhere

f1sh1918:
i think if we implement option 2 an option could be.

1. Default label for search results "Suchresultate (alphabetisch sortiert)
2. Click on in meiner Nähe suchen 
3. The label "Suchresultate" will change to "Suchresultate (in der Nähe)
4. FAB Button changes to "Alphabetisch suchen"
5. Click on it sorting changes to default and label changes to 1.

What do you think about that @Docmale , @ztefanie , @michael-markl  and @seluianova 
I could also ask toni for help

michael-markl:
Yeah it's a tough one, because (as far as I can see) there's really no reason to sort alphabetically except if there is no location available. so having a huge button just for sorting Alphabetically seems bad to me.

michael-markl:
I removed the `bug` flag because it's not a bug (the button only displays if the location permission was not given at the time of app startup), but I admit it might be confusing to the user

f1sh1918:
Maybe we ask @hauf-toni for her opinion

f1sh1918:
@hauf-toni thx for your input. I updated the task description and the task is ready to go

PR #1560, opened 2024-08-02
1487: Allow switching of sorting mode for store list

steffenkleinle:
### Short description

Improve sorting for store list by allowing to switch the sorting mod and giving additional information.

### Proposed changes

- Add information about sorting mode to search results
- Always show FAB to allow switching between sorting mode
- Display snackbar if permission is rejected

### Side effects

None.

### Resolved issues

Fixes: #1487.


frontend/lib/home/home_page.dart
```diff
@@ -91,15 +91,13 @@ class HomePageState extends State<HomePage> {
...
-      floatingActionButton: _currentTabIndex == mapTabIndex
-          ? FloatingActionMapBar(
-              bringCameraToUser: (position) async {
-                await mapPageController?.bringCameraToUser(position);
-              },
-              selectedAcceptingStoreId: selectedAcceptingStoreId,
-            )
-          // Returning a Container() instead of null avoids animations
-          : Container(),
+      floatingActionButton: FloatingActionMapBar(
```
steffenkleinle, line 94: Moving this check inside of `FloatingActionMapBar` was necessary to have enough space for the snackbar. Let me know if there are better ways to achieve this.

frontend/lib/search/location_button.dart
```diff
@@ -64,28 +64,65 @@ class _LocationButtonState extends State<LocationButton> {
...
+        ),
+      );
     }
   }
 
+  Future<void> _showFeatureDisabled() async {
+    final messengerState = ScaffoldMessenger.of(context);
+    final t = context.t;
+    messengerState.showSnackBar(
+      SnackBar(
```
steffenkleinle, line 88: The snackbar is currently shown on top of the FAB, let me know if I should adjust this.

steffenkleinle (commented):
Please be critical with this PR, wasn't really sure about some decisions.

frontend/lib/home/home_page.dart
f1sh1918: hm i can not find an issue with the old implementation?
If i click change language and switch to map.
Snackbar appears. Button is moved a bit up.  Even i think its not a big issue if it overlaps for some seconds tbh
[Screen_recording_20240805_101121.webm](https://github.com/user-attachments/assets/1dd3c5aa-2d04-4d22-b5d9-f6a737d932bb)

frontend/lib/search/location_button.dart
f1sh1918: thats okay i think. Can you give me a hint how i can test this snackbar message

frontend/lib/search/location_button.dart
```diff
@@ -32,21 +36,17 @@ class _LocationButtonState extends State<LocationButton> {
...
   @override
   Widget build(BuildContext context) {
-    if (_locationStatus == LocationRequestStatus.requestSuccessful) {
-      return Container();
-    }
     final settings = Provider.of<SettingsModel>(context);
     final t = context.t;
-    if (settings.locationFeatureEnabled) {
+    if (widget.sortingMode == SortingMode.alphabetically) {
       return Container(
```
f1sh1918, line 42: since there is lots of duplicated code, i think its nicer to create ab FAB widget and use there all the static parameters and pass the only text, icon and onPress fkt.
I think you should also add an icon for alphabetically sorting to keep the look and feel equal
https://api.flutter.dev/flutter/material/Icons-class.html

frontend/lib/search/search_page.dart
```diff
@@ -18,11 +22,25 @@ class _SearchPageState extends State<SearchPage> {
...
+          ? SortingMode.byDistance
+          : SortingMode.alphabetically);
+    });
+  }
 
   @override
   Widget build(BuildContext context) {
     final t = context.t;
-    final currentCoordinatesInput = _coordinates;
+    if (_sortingMode == null) {
```
f1sh1918, line 41: I would rather prefer to set  a default value for the sorting mode and make it mandatory instead of checking here for null value. Even you could remove this "!" for `sortingMode`
`SortingMode _sortingMode = SortingMode.alphabetically;`

frontend/lib/search/search_page.dart
```diff
@@ -50,14 +68,22 @@ class _SearchPageState extends State<SearchPage> {
...
+          setCoordinates: (position) => setState(() {
+            _coordinates = CoordinatesInput(lat: position.latitude, lng: position.longitude);
+            _sortingMode = SortingMode.byDistance;
+          }),
+          setSortingMode: (sortingMode) {
+            if (sortingMode == SortingMode.alphabetically || _coordinates != null) {
+              setState(() => _sortingMode = sortingMode);
+            }
+          },
+          sortingMode: _sortingMode!,
```
f1sh1918, line 85:
```suggestion
          sortingMode: _sortingMode,
```

frontend/lib/search/search_page.dart
```diff
@@ -50,14 +68,22 @@ class _SearchPageState extends State<SearchPage> {
             ResultsLoader(
               searchText: searchFieldText,
               categoryIds: _selectedCategories.map((e) => e.id).toList(),
-              coordinates: currentCoordinatesInput,
+              coordinates: _sortingMode == SortingMode.byDistance ? _coordinates : null,
             )
           ],
         ),
         LocationButton(
```
f1sh1918, line 75: We have two `LocationButtons` (map/search) i think since we now switch different sortings with this button we should rename it to "SortingButton" or sth. similar.

f1sh1918 (commented):
Very nice pr. Works fine 👍 
Just added some comments which are minor

frontend/lib/search/location_button.dart
steffenkleinle: Okay. Sure, block the location permission and press the FAB :)

frontend/lib/search/search_page.dart
steffenkleinle: The reason why I did not opt for a default value is that this might lead to flickering, as the sorting mode is automatically updated/changed at a later point if the location feature is enabled and user coordinates are available.

frontend/lib/search/search_page.dart
steffenkleinle: Somehow needed, otherwise I get an error. Not sure why, I thought dart would handle stuff like this automatically.

frontend/lib/search/search_page.dart
f1sh1918: i couldn't recognize a flickering. but i'm also fine to keep your solution

frontend/lib/search/search_page.dart
f1sh1918: yes its just changeable by setting default sortingMode. so i you keep your solution. you can also keep this

frontend/lib/widgets/extended_floating_action_button.dart
```diff
@@ -0,0 +1,39 @@
...
+import 'package:flutter/material.dart';
+
+class ExtendedFloatingActionButton extends StatelessWidget {
+  final String label;
+  final bool loading;
+  final IconData icon;
+  final void Function() onPressed;
+
+  const ExtendedFloatingActionButton(
+      {super.key,
```
f1sh1918, line 11: please run linter

frontend/lib/search/location_button.dart
f1sh1918: not its okay. its just temporary

f1sh1918 (approved):
Nicely done 👍 
Just run formatter and its good to go

frontend/lib/home/home_page.dart
steffenkleinle: Without this, I just get the following error and no snackbar is shown:
image

frontend/lib/search/location_button.dart
steffenkleinle: @f1sh1918 what do you mean by that?

f1sh1918: the snackbar is just showing up temporary for some seconds. so its okay for me

ztefanie (approved):
Tested. Works as expected.

Issue #1424, opened 2024-04-29
Create a mapping-database table for linking koblenz-users-table and cards-table


ztefanie:
**Is your feature request related to a problem? Please describe.**
we need a new table, which links the koblenz-users-table with the card that is connected to the entry in the koblenz users table.

**Describe the solution you'd like**
- Create a new table with two fields:
- id of the cards table
- id of koblenz-users-table
- extend createCard function->fill this table with the particular ids 

**Describe alternatives you've considered**
Add a field "cardID" to the koblenz users table

seluianova:
**Questions**:
Do we want to allow multiple cards to be created for the same user?
When an additional card is created, should we keep the old cards active or we want to deactivate them?

seluianova:
@ztefanie @f1sh1918 
Could you please clarify if there is a specific reason why we want to store the relation in a separate mapping table?
I believe this approach is usually used for many-to-many relationships. While we have a one-to-many relation in our case.

> Describe alternatives you've considered:
> Add a field "cardID" to the koblenz users table

I would create an entitlements_id column in cards table.
Because for one user we can have many cards (one active, others revoked).
Any objections to that?

f1sh1918:
no i think thats fine. 
i think the reason was that we didn't want to adjust the card table with another optional id
the only thing we have to ensure later in the "extend card" function is that the old card will be revoked after new card was activated but that should work with this solution

ztefanie:
yes thats fine.

PR #1563, opened 2024-08-06
1421: Add createCardFromSelfService mutation (Koblenz)

seluianova:
### Short description
Add createCardFromSelfService mutation to create cards from the Self-Service Portal

### Proposed changes
- add new mutation (receive user data --> make hash --> check user entitlements --> create static and dynamic card)
- make issuerId column in the cards table nullable for cards created wth the new mutation

**Open question**:
Do we want to add smthg like a userentitlements_id column to the cards table?

**Not included:**
- tests (come next)
- rate limiting (should be discussed)

### Side effects
Hopefully none?

### Resolved issues
Fixes: #1421


backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -115,6 +124,88 @@ class CardMutationService {
...
+            regionId,
+            numberOfDynamicCards = 1,
+            numberOfStaticCards = if (generateStaticCode) 1 else 0
+        )
+
+        return activationCode
+    }
+
+    private fun enrichCardInfoWithExpDayAndStartDay(cardInfo: Card.CardInfo, expirationDay: Int, startDay: Int): Card.CardInfo {
+        return cardInfo.toBuilder()
```
seluianova, line 186: Is there a less ugly way to update card info?

seluianova, line 186: Maybe there is a less ugly way to update card info to set dates from the user entitlements table?

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
```diff
@@ -27,11 +27,11 @@ object RegionsRepository {
...
 
     fun findByIdInProject(project: String, id: Int): RegionEntity? {
-        val query = (Projects innerJoin Regions)
+        return (Projects innerJoin Regions)
             .slice(Regions.columns)
             .select { Projects.project eq project and (Regions.id eq id) }
-            .single()
-        return RegionEntity.wrapRow(query)
+            .singleOrNull()
+            ?.let { RegionEntity.wrapRow(it) }
```
seluianova, line 34: The return type was nullable in the declaration, but the function never returned null

seluianova, line 34: The return type was nullable in the declaration, but the function never returned null

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/Argon2IdHasher.kt
```diff
@@ -23,7 +23,7 @@ class Argon2IdHasher {
         private fun encodeWithoutSalt(
             hash: ByteArray?,
             parameters: Argon2Parameters
-        ): String? {
+        ): String {
```
seluianova, line 26: It’s never null, or I missed something?

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -115,6 +124,88 @@ class CardMutationService {
         }
     }
 
+    @GraphQLDescription("Creates a new digital entitlementcard from self-service portal")
+    fun createCardFromSelfService(
```
seluianova, line 128: Open for suggestions regarding the name

f1sh1918: i like it

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -115,6 +124,88 @@ class CardMutationService {
...
+        if (context.backendConfiguration.production) {
+            throw ForbiddenException()
+        }
+        if (project != KOBLENZ_PASS_PROJECT) {
+            throw NotKoblenzProjectException()
+        }
+
+        val cardInfo = parseEncodedCardInfo(encodedCardInfo)
+        val regionId = cardInfo.extensions.extensionRegion.regionId
+        transaction { RegionsRepository.findByIdInProject(project, regionId) } ?: throw RegionNotFoundException()
```
seluianova, line 145: Maybe it's a redundant check, since we check entitlements per region anyway

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/repos/CardRepository.kt
```diff
@@ -57,7 +57,7 @@ object CardRepository {
             this.expirationDay = expirationDay
             this.issueDate = Instant.now()
             this.regionId = EntityID(regionId, Regions)
-            this.issuerId = EntityID(issuerId, Administrators)
+            this.issuerId = if (issuerId != null) EntityID(issuerId, Administrators) else null
```
steffenkleinle, line 60: :question: would it perhaps make sense to set the `issuerId` to something representing the self-service portal instead of setting it to null, i.e. to introduce a hard-coded "administrator" to identify the self-service portal.

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -115,6 +124,88 @@ class CardMutationService {
...
     }
 
+    @GraphQLDescription("Creates a new digital entitlementcard from self-service portal")
+    fun createCardFromSelfService(
+        dfe: DataFetchingEnvironment,
+        project: String,
+        encodedCardInfo: String,
+        generateStaticCode: Boolean
+    ): CardCreationResultModel {
+        // TODO disable for production until the Koblenz project is ready to go live
```
steffenkleinle, line 134: Should this be disabled in this PR?

seluianova: Why not?

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/repos/CardRepository.kt
seluianova: I considered this option as an alternative, but was not sure about the benefit

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
seluianova: Ah, I’ve got your question. It is disabled in this PR in the next lines.
Probably the statement should be like “TODO remove the condition once the Koblenz project is ready to go live”

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/repos/CardRepository.kt
steffenkleinle: Yes, not sure either. I just thought it might prevent accidentally missing the issuerId. But I guess this is only a small benefit.

f1sh1918: hm the problem is we might need a new role for that because we need a "protected" user that will be created once in this case if we set an issuer id. (i think we should avoid this) i thought about it could make sense to set a flag in the card table like a source (SelfService) (InternalService) or sth meaningful to indicate that the issuerId is not missing by mistake.

seluianova: @f1sh1918 or as I noted in "open questions", maybe we can add a reference column to user entitlements table.
it might also indicate, that it's a self-service.
what do you think?

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -115,6 +124,88 @@ class CardMutationService {
...
+        val context = dfe.getContext<GraphQLContext>()
+        if (context.backendConfiguration.production) {
+            throw ForbiddenException()
+        }
+        if (project != KOBLENZ_PASS_PROJECT) {
+            throw NotKoblenzProjectException()
+        }
+
+        val cardInfo = parseEncodedCardInfo(encodedCardInfo)
+        val regionId = cardInfo.extensions.extensionRegion.regionId
```
f1sh1918, line 144: i don't think we can get the `regionId` from the `cardInfos` from the self service form since we have no user that is logged in and creates the card (There we normally get the `regionId`, because the regionManager or regionAdmin has a particular `regionId`)
You have to get it from `userentitlements` and enrich it.

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
f1sh1918: According to my comment before we shouldn't need it, since we get it from userEntitlements and if the user was not found you already throw an exception

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -115,6 +124,88 @@ class CardMutationService {
...
+        )
+
+        val userHash = Argon2IdHasher.hashKoblenzUserData(user)
+
+        val userEntitlements = transaction { UserEntitlementsRepository.findUserEntitlements(userHash.toByteArray(), regionId) }
+        if (userEntitlements == null || userEntitlements.revoked || userEntitlements.endDate.isBefore(LocalDate.now())) {
+            throw InvalidUserEntitlementsException()
+        }
+
+        val cardInfoWithExpDayAndStartDay = enrichCardInfoWithExpDayAndStartDay(
```
f1sh1918, line 160: please also enrich regionId
(optional) i don't like shortening function names since its not that easy to understand so i would prefer
`enrichCardInfoWithRegionIdExpirationDayAndStartDay`
Or just name it `enrichMissingCardInfo` if it gets too long

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/MigrationsRegistry.kt
```diff
@@ -15,6 +15,7 @@ object MigrationsRegistry {
         V0011_AddRegionCardConfirmationMailActivation(),
         V0012_AddApplicationCardCreated(),
         V0013_FixExpirationDateGoldCard(),
-        V0014_CreateUserEntitlementsTable()
+        V0014_CreateUserEntitlementsTable(),
+        V0015_MakeCardIssuerIdColumnNullable()
```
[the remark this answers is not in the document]
f1sh1918: maybe you can check my pr and we can merge it if its okay since it came first :)

f1sh1918 (commented):
Added some comment but just some small things
Good work :)

Just use the regionKey `07111` for the import its connected to koblenz

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/repos/CardRepository.kt
f1sh1918: there is a separate task for that. I think we agreed on to keep the card table as it is but create a separate table for that relationship
https://github.com/digitalfabrik/entitlementcard/issues/1424

seluianova: True, I forgot about that, thanks

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
seluianova: Ah, then I also need to update the user import a bit, currently it allows to upload the same user hash for different regions.
I thought we will pass the region through the form somehow.

f1sh1918:
Hm for me the encoded card info doesn't fit with the entitlement card import

image

seluianova:
@f1sh1918 ops, sorry, I copy-pasted the wrong hash. I have just fixed the PR description

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
f1sh1918: yes but you can also do it in a separate task

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
seluianova: Renamed to enrichCardInfo(), as basically from the implementation perspective it doesn't matter if the data is missing or not, it will be added anyway

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
seluianova: Already done here

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -115,6 +124,88 @@ class CardMutationService {
...
+        project: String,
+        encodedCardInfo: String,
+        generateStaticCode: Boolean
+    ): CardCreationResultModel {
+        // TODO disable for production until the Koblenz project is ready to go live
+        val context = dfe.getContext<GraphQLContext>()
+        if (context.backendConfiguration.production) {
+            throw ForbiddenException()
+        }
+        if (project != KOBLENZ_PASS_PROJECT) {
```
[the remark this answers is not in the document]
seluianova: like boolean selfServiceEnabled ?

f1sh1918:
@seluianova 
Could you adjust the project in the `UserImportHandler` (LN71) to koblenz. Then we could directly use the correct region in the import csv

seluianova:
> Could you adjust the project in the UserImportHandler (LN71) to koblenz

@f1sh1918 set to Koblenz, updated test data in the PR description

f1sh1918 (approved):
Tested. Works fine 👍

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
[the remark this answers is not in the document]
f1sh1918: yes. i overseen this. adding to local config true. everything works fine

Issue #1471, opened 2024-06-03
Update Android SDK 7.4.0


f1sh1918:
**Describe the bug**
A crash issue and a potential fix was provided in the google play console
https://play.google.com/console/u/0/developers/5269030016285360668/app/4973007456755967197/vitals/crashes/6baa03584c1adfd4b7d3aff7574e804d/details?days=28. which seems to be connected to `Android SDK 7.3.1`

I'm not sure how to reproduce it but it should be fixed with updated `Android SDK 7.4.0`

**Expected behavior**
App does not crash

PR #1376, opened 2024-04-04
1354: Migrate to declarative gradle plugins

michael-markl:
### Changes

* Applied migration according to [Flutter's migration guide](https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply) 
* Removed some dependencies workaround
* Removed a path lookup workaround

### Side effects

Hopefully none.

### Resolved issues

Fixes: #1354

PR #1577, opened 2024-08-19
1471: Update android sdk

f1sh1918:
### Short description

This issue was reported from the play store even i couldn't reproduce any crash.

### Proposed changes

- update android sdk version

### Side effects

- none

### Resolved issues

Fixes: #1471 

### Testing
- build and start the app on android

### DIFF
--- a/frontend/android/settings.gradle
+++ b/frontend/android/settings.gradle
@@ -18,7 +18,7 @@ pluginManagement {
 
 plugins {
     id "dev.flutter.flutter-plugin-loader" version "1.0.0"
-    id "com.android.application" version "7.3.1" apply false
+    id "com.android.application" version "7.4.0" apply false
     id "org.jetbrains.kotlin.android" version "1.7.10" apply false
 }
 


ztefanie (changes requested):
...

f1sh1918:
> ...

what does this mean @ztefanie ?

ztefanie:
Yeah sorry, i wrote that the android plugin version must match the dependency version, which I am not sure if this is currently the case. Then i wanted to check, but I am lost in the rabbit hole of getting protoc plugin working again. So not sure it your PR is valid as i can not check. I wanted to edit the message as soon as i get my setup fixed, but it takes forever.

ztefanie (approved):
I am a little bit confused as i cannot find the dependencies class import and this needs to match the plugin version you updated. Where is our buildscript gradle settings block?

But I could start the app without issues on the emulator so i will approve

steffenkleinle (approved):
Not sure what exactly to test/look for here but I built nürnberg in the production version and navigated around a bit without anything crashing on my real android device.

f1sh1918:
@ztefanie not sure if this issue is related to this and if this dependency class import is still needed with declarative plugins
https://github.com/digitalfabrik/entitlementcard/pull/1376/files

PR #1579, opened 2024-08-21
696: Add migration to disable applications for nuernberg project regions

ztefanie:
### Short description
Changed regions of the nuernberg project to be appication-disabled.

### Proposed changes
- Add a migration to update the activatedForApplication cloumn for nuernberg region of nuernberg project to be false

### Side effects
- hopefully non, maybe anything that is related to creating applications in nurnberg, which should be nothing, maybe also things that are related to create cards in nuernberg.

### Resolved issues

Fixes: #696


f1sh1918:
@ztefanie
i don't think this is the solution for this ticket.
The flag you set disables the possibility to choose the region in the application process (on the frontend).
This is currently also not possible because the nuernberg project does not use the application process at all.
To disable this in the database is not wrong but i think what should be done here is to add a check to the application related mutation. In the issue max mentioned to add a flag to the backend configuration.
This flag should be used in the application mutation.

ztefanie:
> @ztefanie i don't think this is the solution for this ticket. The flag you set disables the possibility to choose the region in the application process (on the frontend side). This is currently (even without this value) not possible because the nuernberg project does not use the application process at all. To disable this in the database is not needed here. I think what should be done instead is to add a check to the application related mutation. In the issue max mentioned to add a flag to the backend configuration. This flag should be used in the application mutation and throw a forbidden error response for projects that are not enabled.

This database field is checked on the backend side. I tried with graphiql that you can add an application currently, but not any longer with my changes. This is a backend check.

I also thought about the solution to add this to the project config but i think is not necessary and adds unneeded complexity, as 
we only have one region for nuernberg, koblenz (and frankfurt) and therefore adding it to the project just duplicates the logic/complexity and we need the have some hierarchy of the two statements that forbit the applications. With my changes we avoid having the case, that it is forbidden project wise, but allowed region wise.

This is a backend side fix.

f1sh1918:
> > @ztefanie i don't think this is the solution for this ticket. The flag you set disables the possibility to choose the region in the application process (on the frontend side). This is currently (even without this value) not possible because the nuernberg project does not use the application process at all. To disable this in the database is not needed here. I think what should be done instead is to add a check to the application related mutation. In the issue max mentioned to add a flag to the backend configuration. This flag should be used in the application mutation and throw a forbidden error response for projects that are not enabled.
> 
> This database field is checked on the backend side. I tried with graphiql that you can add an application currently, but not any longer with my changes. This is a backend check.
> 
> I also thought about the solution to add this to the project config but i think is not necessary and adds unneeded complexity, as we only have one region for nuernberg, koblenz (and frankfurt) and therefore adding it to the project just duplicates the logic/complexity and we need the have some hierarchy of the two statements that forbit the applications. With my changes we avoid having the case, that it is forbidden project wise, but allowed region wise.
> 
> This is a backend side fix.

Alright i missed some pieces, you are right.
The only issue i see here then is that we set `activatedForApplication` default true. 
Maybe we can set this project base by a backend config variable?

ztefanie:
> > > @ztefanie i don't think this is the solution for this ticket. The flag you set disables the possibility to choose the region in the application process (on the frontend side). This is currently (even without this value) not possible because the nuernberg project does not use the application process at all. To disable this in the database is not needed here. I think what should be done instead is to add a check to the application related mutation. In the issue max mentioned to add a flag to the backend configuration. This flag should be used in the application mutation and throw a forbidden error response for projects that are not enabled.
> > 
> > 
> > This database field is checked on the backend side. I tried with graphiql that you can add an application currently, but not any longer with my changes. This is a backend check.
> > I also thought about the solution to add this to the project config but i think is not necessary and adds unneeded complexity, as we only have one region for nuernberg, koblenz (and frankfurt) and therefore adding it to the project just duplicates the logic/complexity and we need the have some hierarchy of the two statements that forbit the applications. With my changes we avoid having the case, that it is forbidden project wise, but allowed region wise.
> > This is a backend side fix.
> 
> Alright i missed some pieces, you are right. The only issue i see here then is that we set `activatedForApplication` default true. Maybe we can set this project based by a backend config variable? What do you think? If not we would have to make a database migration for each region we add that has no application feature in the frontend

Or we set the default to false, as for mulit-region projects the users than can activate there region if wanted in the settings?

f1sh1918:
> > > > @ztefanie i don't think this is the solution for this ticket. The flag you set disables the possibility to choose the region in the application process (on the frontend side). This is currently (even without this value) not possible because the nuernberg project does not use the application process at all. To disable this in the database is not needed here. I think what should be done instead is to add a check to the application related mutation. In the issue max mentioned to add a flag to the backend configuration. This flag should be used in the application mutation and throw a forbidden error response for projects that are not enabled.
> > > 
> > > 
> > > This database field is checked on the backend side. I tried with graphiql that you can add an application currently, but not any longer with my changes. This is a backend check.
> > > I also thought about the solution to add this to the project config but i think is not necessary and adds unneeded complexity, as we only have one region for nuernberg, koblenz (and frankfurt) and therefore adding it to the project just duplicates the logic/complexity and we need the have some hierarchy of the two statements that forbit the applications. With my changes we avoid having the case, that it is forbidden project wise, but allowed region wise.
> > > This is a backend side fix.
> > 
> > 
> > Alright i missed some pieces, you are right. The only issue i see here then is that we set `activatedForApplication` default true. Maybe we can set this project based by a backend config variable? What do you think? If not we would have to make a database migration for each region we add that has no application feature in the frontend
> 
> Or we set the default to false, as for mulit-region projects the users than can activate there region if wanted in the settings?

Hm the issue was to set that default true, that new regions for bavaria should use the new process instead of having to `opt-in` it because they may never do :D

ztefanie:
@f1sh1918 Do I now have to remove the migration?

Locally, if I change the setup and restart, the values are as expected, but I am not really sure if this is also what happens on the server.

f1sh1918:
> @f1sh1918 Do I now have to remove the migration?
> 
> Locally, if I change the setup and restart, the values are as expected, but I am not really sure if this is also what happens on the server.

Yes we don't need it.

The createOrUpdate function will run on start/restart
So when we apply next release package

f1sh1918:
> > @f1sh1918 Do I now have to remove the migration?
> > Locally, if I change the setup and restart, the values are as expected, but I am not really sure if this is also what happens on the server.
> 
> Yes we don't need it.
> 
> The createOrUpdate function will run on start/restart So when we apply next release package or manually trigger service restart

Hm well i just set the value to true in the database and restarted backend job. The function will be definitly executed but doesn't override the value in the database @ztefanie have to check why

f1sh1918 (approved):
Works fine :)
Just set activatedForApplication to true in the database and restarted the backend service. This deactivated the `activatedForApplication`

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -59,7 +62,7 @@ fun insertOrUpdateRegions() {
             }
         }
         // TODO #1551: Adjust regionidentifier_unique constraint
-        createOrUpdateRegion(NUERNBERG_PASS_PROJECT, "Nürnberg", "Stadt", null, "https://nuernberg.de")
-        createOrUpdateRegion(KOBLENZ_PASS_PROJECT, "Koblenz", "Stadt", "07111", "https://koblenz.de/")
+        createOrUpdateRegion(NUERNBERG_PASS_PROJECT, "Nürnberg", "Stadt", null, "https://nuernberg.de", false)
```
steffenkleinle, line 65: :question: Wouldn't it make sense to read these values from the build config? Otherwise the config is a little spread between different places, isn't it?

steffenkleinle (commented):
Sadly does not work for me, but maybe I am doing something wrong. The UI doesn't change with both `true` or `false` in the DB. Manually setting it to `true` and restarting did not change it back to `false` for me :/ What am I missing?

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
ztefanie: good point. But where should it be put? The project configs are in the react / administration part. 
We only could add it to the config.yaml.
@f1sh1918 Any preferences?

steffenkleinle: Oh sorry, I missed that there are no build configs for the backend. Then nvm.

steffenkleinle: Oh sorry, I missed that there are no build configs for the backend. Then nvm.

f1sh1918: i think the `config.yaml` is fine. yes it would be good to have one buildConfig for all project parts. The problem is that we use different source types in the backend which we may have to translate

ztefanie: ok, what do you mean by "The problem is that we use different source types in the backend which we may have to translate". I dont understand this.

f1sh1918: this is just a side note that when we created a big `projectConfig` in f.e. typescript we can not simply import it for the backend but have to convert it to valid `yaml` syntax

f1sh1918:
> Sadly does not work for me, but maybe I am doing something wrong. The UI doesn't change with both `true` or `false` in the DB. Manually setting it to `true` and restarting did not change it back to `false` for me :/ What am I missing?

I'm not really sure what your expectation is here. Only the database value should change and if you manually trigger a graphql mutation with application data you will get an error response. The frontend is not connected to the sendApplication mutation for sozialpass

steffenkleinle:
> > Sadly does not work for me, but maybe I am doing something wrong. The UI doesn't change with both `true` or `false` in the DB. Manually setting it to `true` and restarting did not change it back to `false` for me :/ What am I missing?
> 
> I'm not really sure what your expectation is here.

Me neither, that is why I am asking. 

>Only the database value should change

It did not for me, but maybe I was testing wrong.

> and if you manually trigger a graphql mutation with application data you will get an error response.

Okay, how do I do that?

> The frontend is not connected to the sendApplication mutation for sozialpass

I did not test/expect the frontend to do anything, I meant administration frontend.

> I think some testing instructions could help in this case

Yes, testing instructions would be nice.

f1sh1918:
1. Change the `activatedForApplication` in the database for region nuernberg to `true`
2. Rerun backend service. Value should change to false

ztefanie:
> Sadly does not work for me, but maybe I am doing something wrong. The UI doesn't change with both `true` or `false` in the DB. Manually setting it to `true` and restarting did not change it back to `false` for me :/ What am I missing?

I assume you have not restarted docker.

ztefanie:
Sorry for the late response, I am really busy :face_exhaling: 

Here a video how I do it and the upload is applied:
https://github.com/user-attachments/assets/391dec06-1213-4e12-8fd1-daeae6041e83

ztefanie:
> > and if you manually trigger a graphql mutation with application data you will get an error response.
> 
> Okay, how do I do that?

- Start the backend
- go to: [this link](http://0.0.0.0:8000/graphiql/?query=mutation%20addEakApplication(%24regionId%3A%20Int!%2C%20%24application%3A%20ApplicationInput!%2C%20%24project%3A%20String!)%20%7B%0A%20%20result%3A%20addEakApplication(regionId%3A%20%24regionId%2C%20application%3A%20%24application%2C%20project%3A%20%24project)%0A%7D%0A&variables=%7B%0A%20%20%22regionId%22%3A%2094%2C%0A%20%20%22application%22%3A%20%7B%0A%20%20%20%20%22applicationDetails%22%3A%20%7B%0A%20%20%20%20%20%20%22applicationType%22%3A%20%22FIRST_APPLICATION%22%2C%0A%20%20%20%20%20%20%22cardType%22%3A%20%22BLUE%22%2C%0A%20%20%20%20%20%20%22givenInformationIsCorrectAndComplete%22%3A%20true%2C%0A%20%20%20%20%20%20%22hasAcceptedEmailUsage%22%3A%20true%2C%0A%20%20%20%20%20%20%22hasAcceptedPrivacyPolicy%22%3A%20true%2C%0A%20%20%20%20%20%20%22wantsDigitalCard%22%3A%20true%2C%0A%20%20%20%20%20%20%22wantsPhysicalCard%22%3A%20false%2C%0A%20%20%20%20%20%20%22blueCardEntitlement%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%22entitlementType%22%3A%20%22VOLUNTEER_SERVICE%22%2C%0A%20%20%20%20%20%20%20%20%22volunteerServiceEntitlement%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%22certificate%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%22data%22%3A%201%0A%20%20%20%20%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%20%20%20%20%22programName%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22Volunteer%20Program%22%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%2C%0A%20%20%20%20%22personalData%22%3A%20%7B%0A%20%20%20%20%20%20%22address%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%22country%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22Germany%22%0A%20%20%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%20%20%22houseNumber%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22123%22%0A%20%20%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%20%20%22location%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22Munich%22%0A%20%20%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%20%20%22postalCode%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%22shortText%22%3A%20%2280331%22%0A%20%20%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%20%20%22street%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22Example%20Street%22%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%22dateOfBirth%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%22date%22%3A%20%221990-01-01%22%0A%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%22emailAddress%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%22email%22%3A%20%22johndoe%40example.com%22%0A%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%22forenames%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22John%22%0A%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%22surname%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22Doe%22%0A%20%20%20%20%20%20%7D%2C%0A%20%20%20%20%20%20%22telephone%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%22shortText%22%3A%20%22123456789%22%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%2C%0A%20%20%22project%22%3A%20%22nuernberg.sozialpass.app%22%0A%7D)
- run it and get expected error: REGION_NOT_ACTIVATED_FOR_APPLICATION
- change project to bayern in the request and region id to a region of bayern project, e.g. 1 and see the success message

steffenkleinle (approved):
Hmm, I thought I did exactly that and it did not work. Anyway, dumped the whole database, reinstalled everything and now works as expected.

Thanks for the detailed answers! :heart:

Issue #1395, opened 2024-04-15
Add code splitting to administration


f1sh1918:
**Is your feature request related to a problem? Please describe.**
To reduce the bundle size of each js file and reduce loading times, we should implement code splitting.

**Describe the solution you'd like**
- add suspense for using lazy import
- use lazy import for each route

**Additional context**
We have some proper working implementation in integreat, so maybe it can be reused

f1sh1918:
code splitting does not reduce the bundle size that much, so its not worth to implement that

PR #1582, opened 2024-08-21
1395: Code splitting

f1sh1918:
### Short description

As most users only use "beantragen" we might want to use code splitting to reduce to bundle

### Proposed changes

- code splitting and lazyImport for all routes
- used the implementation from integreat app

### Side effects

- hopefully none

### Resolved issues

Fixes: #1395

### Testing
- `npm run build`
- `serve -s build`

### Additional Information
Loaded bundle size for "beantragen" **with** code splitting
Bildschirmfoto 2024-08-21 um 16 14 55

Loaded bundle size for "beantragen" **without** code splitting
Bildschirmfoto 2024-08-21 um 16 19 36

### Note
I'm even not sure of this small win is it worth to use code splitting tbh

### DIFF
--- a/administration/src/Router.tsx
+++ b/administration/src/Router.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useMemo } from 'react'
+import React, { Suspense, useContext, useMemo } from 'react'
 import { Outlet, RouteObject, RouterProvider, createBrowserRouter } from 'react-router-dom'
 import styled from 'styled-components'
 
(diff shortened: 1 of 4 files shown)


f1sh1918:
I'm not pretty sure if this small enhancement really justify using code splitting in this case.
But the bundle size was reduced by 500kb at least

I created a draft to discuss if we should really implement this

ztefanie (commented):
Hmh, I do not think this is worth it.

I think this whole retry function gets a bit cluttered and unclear through the direct use of low level functions like "console.log" and "setItem", I think using helper functions like

```
const wait = (interval: number) => new Promise(resolve => setTimeout(resolve, interval))

const getPageForceRefreshed = (): boolean => JSON.parse(localStorage.getItem(PAGE_FORCE_REFRESHED_KEY) || 'false')
const setPageForceRefreshed = (value: boolean) => localStorage.setItem(PAGE_FORCE_REFRESHED_KEY, JSON.stringify(value))

const logError = (message: string, error: unknown) => {
  console.warn(message)
  console.warn(error instanceof Error ? error.message : 'Unknown error')
}

```

Would make it a lot easier to follow the high level of abstraction logic.

f1sh1918:
> Hmh, I do not think this is worth it.
> 
> I think this whole retry function gets a bit cluttered and unclear through the direct use of low level functions like "console.log" and "setItem", I think using helper functions like
> 
> ```
> const wait = (interval: number) => new Promise(resolve => setTimeout(resolve, interval))
> 
> const getPageForceRefreshed = (): boolean => JSON.parse(localStorage.getItem(PAGE_FORCE_REFRESHED_KEY) || 'false')
> const setPageForceRefreshed = (value: boolean) => localStorage.setItem(PAGE_FORCE_REFRESHED_KEY, JSON.stringify(value))
> 
> const logError = (message: string, error: unknown) => {
>   console.warn(message)
>   console.warn(error instanceof Error ? error.message : 'Unknown error')
> }
> ```
> 
> Would make it a lot easier to follow the high level of abstraction logic.

Alright thx for the input. I think we don't need a deep dive review in this case. I even would tend to close this pr since it doesn't improve the package size that much and code splitting can cause issues in rare cases.

f1sh1918:
won't be included

steffenkleinle:
@f1sh1918 do you want to keep the branch?

Issue #1551, opened 2024-07-29
Adjust regionidentifier_unique constraint


f1sh1918:
**Is your feature request related to a problem? Please describe.**
At the moment we can not add a `regionIdentifier` for nuernberg because we already have this region with the same identifier in the bavarian project and there is a unique constraint on `regionIdentifier`. 
For future projects like koblenz we need to have an `regionidentifier` for every region. So this could be problematic if another future project will also use the region koblenz.

**Describe the solution you'd like**
- remove current unique constraint for region identifier
- set a unique constraint for the columns (projectId, regionIdentifier)
- add the missing `regionIdentifier` for sozialpass region nuernberg.

PR #1588, opened 2024-08-26
1551: Adjust regionidentifier_unique constraint

seluianova:
### Short description
Adjust regionidentifier_unique constraint to allow regions with the same identifier across different projects

### Proposed changes
- re-create constraint
- add the missing regionIdentifier for sozialpass region nuernberg

### Side effects
- no?

### Resolved issues
Fixes: #1551

### Testing
- migrate DB, run backend
- check regions table: Nürnberg with regionIdentifier 09564 should be represented in both EAK and Nürnberg Pass projects


f1sh1918 (changes requested):
I think the migration won't work like this.
You have to replace the null value also in the migration, because after installation of a new release package the migration will run and then the service will be restarted.
Your adjustment in the `Setup.kt` will be triggert after migration not before

seluianova:
@f1sh1918 indeed. thanks. updated.

_I added `SET "regionIdentifier" = '00000' WHERE "regionIdentifier" IS NULL` instruction, assuming it's more universal and could be applied to different data sets. however not sure if it can really be useful._

anyway it works for our case as well, during migration we set this dummy value for Nürnberg and then in Setup it is updated to the real one.

f1sh1918 (approved):
Its fine for me to set it temporarly to `0000`

seluianova:
We probably also need to update:
- stores import for nurnberg (currently we don't set acceptingstores.regionId there)
- add not null constraint for acceptingstores.regionId

steffenkleinle (approved):
Tested, works like a charm!

PR #1589, opened 2024-08-26
Add update circleci config script

steffenkleinle:
### Short description

#
Add script to update circleci config.

### Proposed changes

Add command to package.json.

### Side effects

None.

### Resolved issues

Fixes: None.


f1sh1918 (approved):

f1sh1918:
@steffenkleinle please add not testable label the the pr if there is no issue with this label

steffenkleinle:
> Sorry if I am missing something, but we also have this script in ./scripts/circleci-update-config.sh What is the difference to run it here or there? Or it's triggered automatically from package.json somewhere?

Ahh, good to know, I did not now about that one. I personally like to run it like this because I use the command in Integreat too, but we can also just close and reject this if you feel like this is too much of a duplicate.

Issue #1596, opened 2024-08-26
Improve linting for administration


steffenkleinle:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Issue #1296, opened 2024-02-15
Enable eslint rule "@typescript-eslint/no-unnecessary-condition"


sarahsporck:
**Is your feature request related to a problem? Please describe.**
I realized we sometimes have used optional chains or conditions where it was not necessary. There is a linting rule for this. Enable it.

**Describe the solution you'd like**
Linting shows me an error when i accidentally check a type even though it is known.
For this you have to set "parserOptions.project" to "tsconfig.json".

steffenkleinle:
Perhaps it makes sense to use the eslint config from integreat. Perhpas we are too strict there though.

steffenkleinle:
Already done in #1604.

PR #1604, opened 2024-08-28
1596: Add @typescript-eslint rules and fix issues

steffenkleinle:
### Short description

Add @typescript-eslint rules and fix issues (and add a few eslint-disables).

### Proposed changes

- Rewrite .eslintrc to :eslintrc.js
- Use eslint config from integreat-app (and disable a few rules, might be enabled later)
- Enable @tyescript-eslint rules
- Fix most issues

### Side effects

None.

### Testing

That everything still works?

### Resolved issues

Partly fixes: #1596.


administration/src/Router.tsx
```diff
@@ -80,7 +80,7 @@ const Router = () => {
                 { path: 'region', element: <RegionsController /> },
               ]
             : []),
-          ...(projectConfig.cardStatistics ? [{ path: 'statistics', element: <StatisticsController /> }] : []),
+          ...(projectConfig.cardStatistics.enabled ? [{ path: 'statistics', element: <StatisticsController /> }] : []),
```
steffenkleinle, line 83: Is this correct?

seluianova:
It also fixes this one, right? https://github.com/digitalfabrik/entitlementcard/issues/1296

administration/src/Router.tsx
f1sh1918: yes

administration/src/Router.tsx
```diff
@@ -100,14 +100,7 @@ const Router = () => {
...
     return createBrowserRouter(routes.filter((element): element is RouteObject => element !== null))
-  }, [
-    authData,
-    projectConfig.applicationFeature,
-    signIn,
-    signOut,
-    projectConfig.cardStatistics,
-    projectConfig.cardCreation,
-  ])
+  }, [authData, signIn, signOut, projectConfig])
```
f1sh1918, line 103: is this really working meanwhile with stringify or anything to use objects as deps?

administration/src/bp-modules/NavigationBar.tsx
```diff
@@ -14,11 +14,11 @@ const PrintAwareNavbar = styled(Navbar)`
   }
 `
 
-interface Props {
+type Props = {
```
f1sh1918, line 17: I think we should define a proper naming within the project. It not has to be fixed here but i think we should name it <component-name>Props or sth similar

f1sh1918 (changes requested):
Cannot start webpack anymore

```
Failed to compile.

Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration.resolve.modules[2] should be a non-empty string.
   -> Folder name or directory path where to find modules.

npm ERR! Lifecycle script `start` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: administration 
npm ERR!   at location: /Users/afischer/projects/entitlementcard/administration 

Process finished with exit code 1
```

administration/src/bp-modules/NavigationBar.tsx
steffenkleinle: You mean this should e.g. be `NavigationBarProps`? Agreed, but I don't think there is a rule for this, we just have to take care of this manually/during reviewing. Do you want me to create an issue for this? Or should we do this on the fly?

administration/src/Router.tsx
steffenkleinle: This should work (and is as far as I can see). But it depends. If the object is recreated every render, e.g. as in [this example](https://react.dev/reference/react/useEffect#removing-unnecessary-object-dependencies) of the react docs, the hook is executed again on every render.
However, if the object is not redefined/does not change on every render, this works just fine. This is e.g. the case if it is a value from `useState`, `useMemo`, `useContext` or defined outside of a react component, which is the case here.

steffenkleinle:
> Cannot start webpack anymore
> 
> ```
> Failed to compile.
> 
> Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
>  - configuration.resolve.modules[2] should be a non-empty string.
>    -> Folder name or directory path where to find modules.
> 
> npm ERR! Lifecycle script `start` failed with error: 
> npm ERR! Error: command failed 
> npm ERR!   in workspace: administration 
> npm ERR!   at location: /Users/afischer/projects/entitlementcard/administration 
> 
> Process finished with exit code 1
> ```

Ohhh... funny, this was caused by this little change:
``` ts
      modules: ['node_modules', paths.appNodeModules].concat(modules.additionalModulePaths || []),
```
to
``` ts
      modules: ['node_modules', paths.appNodeModules].concat(modules.additionalModulePaths ?? []),
```

Should be fixed now @f1sh1918, can you please test this again? Thanks.

.idea/dataSources.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="DataSourceManagerImpl" format="xml" multifile-model="true">
-    <data-source source="LOCAL" name="ehrenamtskarte@localhost" uuid="d83188d7-bad5-4b46-b4d7-13cabcdaf7c7">
+    <data-source source="LOCAL" name="ehrenamtskarte@db-postgis" uuid="d83188d7-bad5-4b46-b4d7-13cabcdaf7c7">
```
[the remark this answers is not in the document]
steffenkleinle: Because it keeps changing for me. Reverted it neverthelesss

ztefanie (commented):
Looks good :+1:
Thanks for the adjustments. Not tested again.

ztefanie (approved):

f1sh1918 (approved):
Looks good 👍

PR #1607, opened 2024-08-29
1491: Add api token feature

ztefanie:
### Short description
Add possibility to add and delete tokens and to view metadata of the tokens. Tests will be done in https://github.com/digitalfabrik/entitlementcard/issues/1618

### Proposed changes
- Added new database table for tokens
- added three graphql endpoints
- added api token section in administration settings for koblenz and project admins

### Testing

login as a koblenz project admin. go to settings, create and delete tokens.

### Resolved issues

Fixes: #1491


f1sh1918:
Just tested the implementation and it works fine 👍 
Before reviewing it i want to discuss some general things (we also might improve in the future):
1. It might make sense to add a token name like in git so you can distinguish between the tokens.
2. Then we could display this instead "Email des Erstellers" because its always the same and doesn't bring any value to show it.
3. I think an creation date would be good and also display sth like "Added on <date created>
4. Before deleting the token i would expect an Alert like "Do you really want to delete this". We already have a component for that can be easily used.
5. A copy paste field where the generated token is displayed and can be copied via one click on it without missing sth by selecting.

ztefanie:
> Just tested the implementation and it works fine 👍 Before reviewing it i want to discuss some general things (we also might improve in the future):
> 
> 1. It might make sense to add a token name like in git so you can distinguish between the tokens.
> 2. Then we could display this instead "Email des Erstellers" because its always the same and doesn't bring any value to show it.
> 3. I think an creation date would be good and also display sth like "Added on
> 4. Before deleting the token i would expect an Alert like "Do you really want to delete this". We already have a component for that can be easily used.
> 5. A copy paste field where the generated token is displayed and can be copied via one click on it without missing sth by selecting.
> 
> And could you please resolve the migration conflicts :)

Yes these are all valid improvments, but I though we somehow agreed an a minimum usable feature and not a very good one, as it will only be used by very few people very rarly. So i think the solution I suggested is good enough and a lot more is not covered by our offer.

f1sh1918:
> > Just tested the implementation and it works fine 👍 Before reviewing it i want to discuss some general things (we also might improve in the future):
> > 
> > 1. It might make sense to add a token name like in git so you can distinguish between the tokens.
> > 2. Then we could display this instead "Email des Erstellers" because its always the same and doesn't bring any value to show it.
> > 3. I think an creation date would be good and also display sth like "Added on
> > 4. Before deleting the token i would expect an Alert like "Do you really want to delete this". We already have a component for that can be easily used.
> > 5. A copy paste field where the generated token is displayed and can be copied via one click on it without missing sth by selecting.
> > 
> > And could you please resolve the migration conflicts :)
> 
> Yes these are all valid improvments, but I though we somehow agreed an a minimum usable feature and not a very good one, as it will only be used by very few people very rarly. So i think the solution I suggested is good enough and a lot more is not covered by our offer.

ok then just two small things.
The colors of the font and background of the token are in a really bad contrast. I suggest just use some gray as background. And lets keep the token and the label `New Token` separate so copying is easier. Thats no big issue to adjust

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ApiTokenService.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.jetbrains.exposed.sql.transactions.transaction
+import java.security.SecureRandom
+import java.time.LocalDate
+import java.util.Base64
+
+fun getByteArrayLength() = 3 * TOKEN_LENGTH / 4 // 4*(n/3) chars are needed to represent n bytes
+
+@Suppress("unused")
+class ApiTokenService {
+    @GraphQLDescription("Creates a new api token for user upload endpoint")
```
f1sh1918, line 27:
```suggestion
    @GraphQLDescription("Creates a new api token for user import endpoint")
```

endpoint is /users/import

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ApiTokenService.kt
```diff
@@ -0,0 +1,93 @@
...
+        val jwtPayload = context.enforceSignedIn()
+
+        val random = SecureRandom()
+        val bytes = ByteArray(getByteArrayLength())
+        random.nextBytes(bytes)
+        val token = Base64.getEncoder()
+            .encodeToString(bytes)
+
+        val encryptedToken = BCrypt.withDefaults().hash(11, token.toCharArray())
+        val expirationDate = LocalDate.now().plusMonths(expiresIn.toLong())
```
f1sh1918, line 39: For better consistence and security it would make sense to extend the `Authorizer` by mayAdd-/mayCreate-/mayDeleteApiTokensInProject and throw a `ForbiddenException` for all new mutations

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ApiTokenService.kt
```diff
@@ -0,0 +1,93 @@
...
+        val context = dfe.getContext<GraphQLContext>()
+        val jwtPayload = context.enforceSignedIn()
+        return transaction {
+            val admin =
+                AdministratorEntity.findById(jwtPayload.adminId)
+                    ?: throw UnauthorizedException()
+
+            (ApiTokens leftJoin Administrators)
+                .select { ApiTokens.projectId eq admin.projectId }
+                .map { row ->
```
f1sh1918, line 82: this should also do the trick
```
                 .map {
                    ApiTokenMetaData(
                        it[ApiTokens.id].value,
                        it[Administrators.email],
                        it[ApiTokens.expirationDate].toString()
                    )
                }
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/MigrationsRegistry.kt
```diff
@@ -18,6 +18,7 @@ object MigrationsRegistry {
         V0014_CreateUserEntitlementsTable(),
         V0015_UpdateRoleRegionCombinationConstraint(),
         V0016_MakeCardIssuerIdColumnNullable(),
-        V0017_AlterUserEntitlementsUserHashConstraint()
+        V0017_AlterUserEntitlementsUserHashConstraint(),
+        V0018_AddApiTokenTable()
```
f1sh1918, line 22: Please merge recent main and update migrations to avoid issues

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ApiTokenService.kt
```diff
@@ -0,0 +1,93 @@
...
+                (ApiTokens.id eq id) and (creatorId eq admin.id)
+            }
+        }
+        return id
+    }
+}
+
+@Suppress("unused")
+class ApiTokenQueryService {
+    @GraphQLDescription("Gets metadata of all api tokes for a project")
```
f1sh1918, line 71:
```suggestion
    @GraphQLDescription("Gets metadata of all api token for a project")
```

f1sh1918 (commented):
Good work! 👍 
Tested and works fine
Added some comments but nothing crucial

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -0,0 +1,142 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: 15px;
+`
+
+const NewTokenText = styled.p`
+  font-size: 18px;
+  color: #007bff;
+  background: #e9f7ff;
```
f1sh1918, line 36: please check my design suggestion or use sth with better contrast between text and background

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -0,0 +1,142 @@
...
+  word-break: break-all;
+`
+
+const ApiTokenSetting = () => {
+  const metaDataQuery = useGetApiTokenMetaDataQuery({})
+
+  const appToaster = useAppToaster()
+
+  const [tokenMetaData, setTokenMetadata] = useState<Array<ApiTokenMetaData>>([])
+  const [createdToken, setCreatedToken] = useState<string>('')
```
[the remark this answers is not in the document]
ztefanie: if i do not init it, i need to adjust the type to "string|undefined" and i did not want to do it.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ApiTokenService.kt
```diff
@@ -0,0 +1,89 @@
...
+
+fun getByteArrayLength() = 3 * TOKEN_LENGTH / 4 // 4*(n/3) chars are needed to represent n bytes
+
+@Suppress("unused")
+class ApiTokenService {
+    @GraphQLDescription("Creates a new api token for user import endpoint")
+    fun createApiToken(expiresIn: Int, dfe: DataFetchingEnvironment): String {
+        val context = dfe.getContext<GraphQLContext>()
+        val jwtPayload = context.enforceSignedIn()
+        val admin = AdministratorEntity.findById(jwtPayload.adminId)
```
seluianova, line 32: transaction is missing

seluianova, line 32:
```suggestion
        val admin = transaction { AdministratorEntity.findById(jwtPayload.adminId) }
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ApiTokenService.kt
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        return token
+    }
+
+    @GraphQLDescription("Deletes a selected API token")
+    fun deleteApiToken(id: Int, dfe: DataFetchingEnvironment): Int {
+        val context = dfe.getContext<GraphQLContext>()
+        val jwtPayload = context.enforceSignedIn()
+        val admin = AdministratorEntity.findById(jwtPayload.adminId)
```
seluianova, line 55:
```suggestion
        val admin = transaction { AdministratorEntity.findById(jwtPayload.adminId) }
```

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -0,0 +1,147 @@
...
+    onError: error => {
+      const { title } = getMessageFromApolloError(error)
+      appToaster?.show({
+        intent: 'danger',
+        message: title,
+      })
+    },
+  })
+
+  const [deleteToken] = useDeleteApiTokenMutation({
```
seluianova, line 74: optional: maybe would be nice to add a confirmation dialog? to avoid accidental deletion

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -0,0 +1,147 @@
...
+              <th>Ablaufdatum</th>
+              <th></th>
+            </tr>
+          </thead>
+          <tbody>
+            {tokenMetaData.map(item => (
+              <tr key={item.id}>
+                <td>{item.creatorEmail}</td>
+                <td>{formatDate(item.expirationDate)}</td>
+                <td>
```
seluianova, line 135: nit-picky:
maybe we can add vertical alignment in the row to center because currently the delete button and the text are a bit at different levels
image

administration/src/project-configs/getProjectConfig.ts
```diff
@@ -114,6 +114,7 @@ export interface ProjectConfig {
   freinetCSVImportEnabled: boolean
   cardCreation: boolean
   storeManagement: StoresManagement
+  userUploadApiEnabled: boolean
```
seluianova, line 117:
```suggestion
  userImportApiEnabled: boolean
```
just to stick to the same naming

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -138,4 +138,13 @@ object Authorizer {
     fun mayUpdateStoresInProject(user: AdministratorEntity, projectId: Int): Boolean {
         return user.projectId.value == projectId && user.role == Role.PROJECT_STORE_MANAGER.db_value
     }
+
+    fun mayAddApiTokensInProject(user: AdministratorEntity): Boolean =
+        user.role == Role.PROJECT_ADMIN.db_value
+
+    fun mayViewApiMetadataInpProject(user: AdministratorEntity): Boolean =
```
seluianova, line 145:
```suggestion
    fun mayViewApiMetadataInProject(user: AdministratorEntity): Boolean =
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
```diff
@@ -54,3 +55,21 @@ class AdministratorEntity(id: EntityID<Int>) : IntEntity(id) {
     var notificationOnVerification by Administrators.notificationOnVerification
     var deleted by Administrators.deleted
 }
+
+val TOKEN_LENGTH = 60
```
seluianova, line 59:
```suggestion
const val TOKEN_LENGTH = 60
```

seluianova (commented):
Looks good to me in general! Added a few suggestions.

Also, I agree with Andy regarding the confirmation dialog before deletion and I think it might be important if Koblenz stores a token in some app config for example. Then if they accidentally delete it, they might have to redeploy their app.
Not sure if it's a plausible case though.
What do you think?

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/ApiTokenService.kt
ztefanie: Why? transaction only makes sense if there are multiple statements inside the transaction block.

seluianova (approved):
Thanks! Looks nice to me

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -0,0 +1,175 @@
...
+  margin-top: 15px;
+  word-break: break-all;
+`
+
+const TableData = styled.td`
+  vertical-align: middle !important;
+`
+
+const DeleteIcon = styled(Delete)`
+  display: block !important;
```
f1sh1918, line 47: please add here
`cursor: pointer;`

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -0,0 +1,175 @@
...
+const ApiTokenSetting = () => {
+  const metaDataQuery = useGetApiTokenMetaDataQuery({})
+
+  const appToaster = useAppToaster()
+
+  const [tokenMetaData, setTokenMetadata] = useState<Array<ApiTokenMetaData>>([])
+  const [createdToken, setCreatedToken] = useState<string>('')
+  const [expiresIn, setExpiresIn] = useState<number>(1)
+
+  //const [confirmationDialogIsOpen, setConfirmationDialogIsOpen] = useState<boolean>(false)
```
f1sh1918, line 59: please remove commented line

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -86,61 +97,78 @@ const ApiTokenSetting = () => {
...
+        }}>
+        <p>Möchten Sie das Token unwiderruflich löschen?</p>
+      </Alert>
+      <SettingsCard>
+        <H2>Api Token</H2>
+        <Container>
+          <H4>Neues Token erstellen</H4>
+          <p>
+            Ein neu erzeugtes Token wir nur einmalig angezeigt und kann danach nicht wieder abgerufen werden. Bitte
+            speichern sie dieses Token an einem sicheren Ort.
```
f1sh1918, line 122: Bitte speichern "Sie" . Not sure why i cannot make a suggestion here

administration/src/bp-modules/user-settings/ApiTokenSettings.tsx
```diff
@@ -86,61 +97,78 @@ const ApiTokenSetting = () => {
...
+    <>
+      <Alert
+        cancelButtonText='Abbrechen'
+        confirmButtonText='Token löschen'
+        icon='trash'
+        intent='danger'
+        isOpen={tokenToDelete !== null}
+        onCancel={() => setTokenToDelete(null)}
+        onConfirm={() => {
+          if (tokenToDelete != null) {
```
f1sh1918, line 109: `!== null`

f1sh1918 (approved):
Works great! Nice 👍 
Just some small things

PR #1615, opened 2024-09-03
Tests for createCardsByCardInfos mutation + authorization setup

seluianova:
### Short description
- Add TestAdministrators enum to create admins and auth tokens in tests
- Add tests for createCardsByCardInfos mutation where authorization token is used as an example

### Side effects
no, just tests

### Testing
n/a

### Resolved issues
n/a


f1sh1918 (approved):
Tested works as expected :)

ztefanie (approved):

PR #1619, opened 2024-09-09
Update docs

f1sh1918:
- added important note about backend config
- added koblenz project id in docs

### DIFF
--- a/docs/production-setup.md
+++ b/docs/production-setup.md
@@ -88,6 +88,12 @@ fvm flutter build ipa --flavor Bayern --release --dart-define=environment=produc
 
 
 ### Backend and Administration
+
+#### Important: 
+- any updates in the backend `config.yml` have to added/adjusted in salt before deployment!
+- Example: https://git.tuerantuer.org/DF/salt/src/branch/master/pillars/entitlementcard/production/bayern.sls
+
+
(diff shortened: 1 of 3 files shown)


docs/production-setup.md
```diff
@@ -88,6 +88,12 @@ fvm flutter build ipa --flavor Bayern --release --dart-define=environment=produc
 
 
 ### Backend and Administration
+
+#### Important:
```
f1sh1918, line 92: please mind this @ztefanie , @seluianova, @steffenkleinle if you add any values to the backend `config.yml`

seluianova: I don't have an access to the link below, can you grant it to me?

seluianova (approved):

docs/production-setup.md
ztefanie: ok, that is ugyl :cry: 

I think this hint should be in the config file in a comment, because then, when changing the file you get the reminder.

I created a ticket to solve this issue: https://github.com/digitalfabrik/entitlementcard/issues/1627

ztefanie (changes requested):
Please add this hint in the config file, so we get a reminder, when changing this.

docs/production-setup.md
f1sh1918: oh sure i moved the comment to the `config.yml`

ztefanie (approved):

PR #1629, opened 2024-09-10
1628: Exclude fullName field from user hashing

seluianova:
### Short description
Exclude fullName field from user hashing

> Koblenz-Pass News: "Aktenzeichen in der Kombination mit Geburtsdatum sind datenschutzrechtlich vereinbar. Die Übertragung des Aktenzeichen und des Geburtsdatums sind möglich."

### Proposed changes
- Adjust Argon2IdHasher, docs, tests

### Side effects
Card can be issued under any name

### Testing
1. Import user entitlements via POST http://localhost:8000/users/import
Include CSV file as a form-data:
```
```
The above hash **does not contain** a name of the user

2. Send graphql request to create card:
```
mutation CreateCardFromSelfService {
    createCardFromSelfService(
        project: "koblenz.sozialpass.app"
        encodedCardInfo: "Cg1LYXJsYSBLb2JsZW56GhIKAghfEgQI6r4BMgYKBDEyM0s="
        generateStaticCode: true
    ) {
        dynamicActivationCode {
            cardInfoHashBase64
            codeBase64
        }
        staticVerificationCode {
            cardInfoHashBase64
            codeBase64
        }
    }
}
```
The encodedCardInfo above **contains** a name of the user

### Resolved issues
Fixes: #1628

### DIFF
--- a/docs/CreateKoblenzHash.md
+++ b/docs/CreateKoblenzHash.md
@@ -6,7 +6,6 @@ The example data is
 
 | Parameter    | Value         |
 |--------------|---------------|
-| Name         | Karla Koblenz |
 | Geburtstag   | 10.06.2003    |
 | Aktenzeichen | 123K          |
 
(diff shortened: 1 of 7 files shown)


docs/CreateKoblenzHash.md
```diff
@@ -15,14 +14,11 @@ The example data is
...
     birthday: 12213
     referenceNumber: "123K"
 }
 ```
 
-- The full_name must be `FirstnameSpaceLastname`. Every char must exactly match the user input, as otherwise there is not possibility to match the data Koblenz transfers with the input the user makes.
-e.g. `Karla Koblenz` will match neither with `Karla Lisa Koblenz` nor with `Karlá Koblenz`.
 - The birthday is defined in our protobuf [card.proto](../frontend/card.proto) file: It counts the days since the birthday (calculated from 1970-01-01).
   All values of this field are valid, including the 0, which indicates that the birthday is on 1970-01-01. Birthdays before 1970-01-01 have negative values.
 - referenceNumber is set to the "Aktenzeichen"
```
seluianova, line 24: to be added:
> "The file numbers (Aktenzeichen) currently have a minimum length of five characters and a maximum length of 13 characters, but we ask that the solution be technically prepared for 15 characters." - written by Koblenz today

f1sh1918: i think that should be fine. We have to limit this in the frontend. We had some kind of signature buffer which will not survive :(
But even we had the same issue with the extension of the card holder name to 40 chars

seluianova:
@steffenkleinle sorry 🙈 have you managed that already or shall I extend the testing steps now?

PR #1631, opened 2024-09-11
1424: Add entitlementsId column in cards table

seluianova:
### Short description
Create an entitlements_id column in the cards table to add a reference between cards and user entitlements

### Proposed changes
- add new entitlements_id column in the cards table
- add a constraint - either entitlements_id or issuer_id must be not null in the cards table
- revoke old cards by entitlements_id when issuing a new card with the same entitlements
- adjust tests 

### Side effects
- no?

### Testing
1. Import user entitlements via POST http://localhost:8000/users/import
Include CSV file as a form-data:
```
07111,"$argon2id$v=19$m=19456,t=2,p=1$57YPIKvU/XE9h7/JA0tZFT2TzpwBQfYAW6K+ojXBh5w",01.01.2024,01.01.2025,false
```
2. Send a graphql request to create cards, check that the cards have been created and the entitlements_id column is populated
```
mutation CreateCardFromSelfService {
    createCardFromSelfService(
        project: "koblenz.sozialpass.app"
        encodedCardInfo: "Cg1LYXJsYSBLb2JsZW56GhIKAghfEgQI6r4BMgYKBDEyM0s="
        generateStaticCode: true
    ) {
        dynamicActivationCode {
            cardInfoHashBase64
            codeBase64
        }
        staticVerificationCode {
            cardInfoHashBase64
            codeBase64
        }
    }
}
```
3. Send graphql request to create cards again, check that the new cards are created and old cards are updated to revoked=true

### Resolved issues
Fixes: #1424


backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -74,25 +77,25 @@ class CardMutationService {
         if (!cardInfo.extensions.hasExtensionRegion()) {
             throw InvalidCardHashException()
         }
-        return transaction {
```
seluianova, line 77: moved the transaction up a level to keep all actions in a single transaction:
- revoke old cards
- create dynamic card
- create static card

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -156,10 +159,16 @@ class CardMutationService {
...
         )
 
-        val activationCode = CardCreationResultModel(
-            createDynamicActivationCode(updatedCardInfo, userId = null),
-            if (generateStaticCode) createStaticVerificationCode(updatedCardInfo, userId = null) else null
-        )
+        val activationCode = transaction {
+            val revokedCount = CardRepository.revokeByEntitlementsId(userEntitlements.id.value)
+            if (revokedCount > 0) {
+                logger.info("Revoked {} cards associated with the user entitlements {}", revokedCount, userEntitlements.id.value)
```
seluianova, line 165: I assume we want to revoke old cards and not delete, or?

f1sh1918: we can think about some cleanup functionality at a later point. I'm not sure what happens at the moment if a user has a activated on the phone and the card entry gets deleted

f1sh1918 (approved):
Looks good! 👍 
Tested successfully

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/Schema.kt
```diff
@@ -36,6 +37,7 @@ object Cards : IntIdTable() {
     val cardInfoHash = binary("cardInfoHash", CARD_INFO_HASH_LENGTH).uniqueIndex()
     val codeType = enumeration("codeType", CodeType::class)
     val firstActivationDate = timestamp("firstActivationDate").nullable()
+    val entitlementsId = reference("entitlementsId", UserEntitlements).nullable()
```
steffenkleinle, line 40: :question: Is the plural `s` correct here?

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/Schema.kt
```diff
@@ -51,6 +53,12 @@ object Cards : IntIdTable() {
                 ) or
                 ((activationSecretHash neq null) and (codeType eq CodeType.DYNAMIC))
         }
+        check("issuerid_or_entitlementsid_not_null") {
```
steffenkleinle, line 56: :wrench: Above check uses camel case, this uses snake case. I'd prefer the naming to be consistent.

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/Schema.kt
```diff
@@ -51,6 +53,12 @@ object Cards : IntIdTable() {
...
+            (
+                ((issuerId neq null) and (entitlementsId eq null))
+                    or ((issuerId eq null) and (entitlementsId neq null))
```
steffenkleinle, line 59: :question: Am I correct in the assumption that for bavaria/nuernberg `issuerId` must not be null and for koblenz `entitlementsId` must not be null? Shouldn't we rather make the check take that into account (if that is even possible)?

steffenkleinle (approved):
Tested in firefox, works as expected!

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/Schema.kt
seluianova: in my head it was, but I think you are right

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/Schema.kt
seluianova: I generally agree, but the above check uses camel case with the first letter capitalized. 
In other tables we also have camel case with the first letter not capitalized.
And in some tables we also have snake case.

I personally like the snake case the most, because it matches the naming in the database (all the letters are in lower case there).

I think it would be nice to make the same style everywhere, but that would require applying additional migrations, so I'm not sure if it's worth it, to be honest.

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/database/Schema.kt
seluianova: changed to singular

Issue #1297, opened 2024-02-15
Refactor importLegacyCard


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Related to this hint the function to importLegacyCard not implicity ensures `initialization`
[Here](https://github.com/digitalfabrik/entitlementcard/pull/1293#discussion_r1490959928) is some additional info

**Describe the solution you'd like**
- ensure that secure storage was initialized before inserting the legacy code

PR #1651, opened 2024-09-26
Report error to sentry if initialization of userCodeModel fails

michael-markl:
### Short description

Report errors to sentry if initialization of UserCodeModel fails, instead of silently swallowing potentially thrown exceptions.

Also, if initialization fails, do not implicitly use empty array of user codes, but show error.

Also throw Errors when using UserCodeModel when it is not properly initialized.

### Side effects

If the initialization actually failed, the "Ausweisen" tab cannot be opened.

### Testing

Activating card via Scanner and via Deeplink still works.
Activated cards survive app restart.

### Resolved issues

Fixes: #1297 

Related: #1265


f1sh1918 (approved):
Nice refactoring, tested on android and ios!

f1sh1918 (approved):
Thx for the fix! Works fine on android and ios

Could you please create an issue for the error handling improvement and add the particular issue id to the todo. thx

michael-markl:
> Could you please create an issue for the error handling improvement and add the particular issue id to the todo. thx

Done! https://github.com/digitalfabrik/entitlementcard/issues/1656

frontend/lib/identification/user_code_model.dart
```diff
@@ -1,38 +1,60 @@
...
+      _initState = _InitializationState.failed;
     } finally {
-      _isInitialized = true;
       notifyListeners();
     }
   }
 
   void insertCode(DynamicUserCode code) {
+    if (!isInitialized) {
+      throw StateError('UserCodeModel not initialized!');
```
ztefanie, line 56: Maybe you could move this error text to a constant that is reused, to avoid duplication

ztefanie (approved):
Thanks for this PR, only one small suggestion. Not tested

frontend/lib/identification/user_code_model.dart
michael-markl: Done

Issue #1509, opened 2024-06-26
Memory leaks in jetty.http


f1sh1918:
**Describe the bug**
We are facing some memory leaks on the backend service.
They may be related to the new javalin version, using a jetty dependency that causes memory leaks.
https://github.com/jetty/jetty.project/pull/11780

**To Reproduce**
1. Check the server logs
2. Check grafana

**Expected behavior**
No memory leaks occur

**Screenshots**

<img width="2231" alt="Bildschirmfoto 2024-06-26 um 11 21 45" src="https://github.com/digitalfabrik/entitlementcard/assets/37902063/4162aec4-06d0-469b-a380-f11484f05489">

**Additional context**
Link to grafana: https://grafana.tuerantuer.org/d/l3Gr9b5Vk/server-vitals-with-host-selection?orgId=1&refresh=5m&var-HOST=entitlementcard.tuerantuer.org&viewPanel=8&from=now-30d&to=now

PR #1652, opened 2024-09-26
Update potentially vulnerable backend dependencies

michael-markl:
### Short description

Update some backend dependencies that might have vulnerabilities.

Also revert the wrong fix for #1509, #1544


f1sh1918 (approved):
Tested the major queries and mutations. Looks good

ztefanie (approved):
Thanks for updating :)

PR #1658, opened 2024-10-01
Remove duplicate graphql provider

michael-markl:
### Short description

The graphql provider was twice in the widget tree: The `App` and `HomePage` widgets both "rendered" the graphql provider.
I removed the one at `HomePage` as the one in `App` was further up the tree. 

### Side effects

Likely none. Maybe caching works better now.... Not sure what other consequences there are when there are two graphql providers in place.

### Testing

General smoke test ? :D

### DIFF
--- a/frontend/lib/graphql/configured_graphql_provider.dart
+++ b/frontend/lib/graphql/configured_graphql_provider.dart
@@ -73,12 +73,7 @@ class ConfiguredGraphQlProviderState extends State<ConfiguredGraphQlProvider> {
         ]),
       ),
     );
-    return GraphQLProvider(
-      client: client,
-      child: CacheProvider(
-        child: widget.child,
-      ),
-    );
+    return GraphQLProvider(client: client, child: widget.child);
   }
 
   void _printDebugMessage(BuildContext context, String message) {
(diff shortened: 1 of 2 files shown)


frontend/lib/graphql/configured_graphql_provider.dart
```diff
@@ -73,12 +73,7 @@ class ConfiguredGraphQlProviderState extends State<ConfiguredGraphQlProvider> {
         ]),
       ),
     );
-    return GraphQLProvider(
-      client: client,
-      child: CacheProvider(
```
f1sh1918, line 78: why do you remove the cache provider? How improves that caching? :)

f1sh1918 (approved):
Works fine! Just added some questions for the context

frontend/lib/graphql/configured_graphql_provider.dart
michael-markl: I guess the CacheProvider once had a meaning, but through upgrades of graphql_flutter, it is a seemingly empty widget. I asked for some clarification on this widget here: https://github.com/zino-hofmann/graphql-flutter/issues/1459. For now I just assumed, that is a leftover class that should be removed in their lib.
Also, I checked whether caching still worked after removing, and it seems to work.

ztefanie (approved):
Good find, thanks for adjusting :+1:

Issue #1666, opened 2024-10-05
Adjust react form state to be immutable


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
At the moment, both CardBlueprint and Extension are class instances and used as react state.
We modify properties frequently by just assigning new values, for example:
``` ts
card.fullName = event.target.value
onUpdate()
```
Since react state SHOULD be immutable/read-only and react only checks whether the instance changed, it is necessary to make sure to do some hacky workaround in `onUpdate` to "notify" react of the state update.

This is not best practice and increases the risk of future bugs.

**Describe the solution you'd like**
We should refactor our code to avoid changing immutable state and using hacky workarounds.
Instead, the state should only contain the actual "state" as a plain object.

In general, assigning new properties to objects/in-place changes should be avoided as these side-effects are sometimes hidden/missed and make the code harder to understand and easier to introduce bugs.
Instead of doing something like the following should therefore be avoided. 
``` ts
    csvHeader[indexExpirationDate] = expiryColumnName
```

**Describe alternatives you've considered**
Keep the hacky workarounds.

**Additional context**
https://react.dev/reference/react/useState#ive-updated-the-state-but-the-screen-doesnt-update
https://react.dev/learn/updating-objects-in-state#treat-state-as-read-only

PR #1667, opened 2024-10-05
1666: Immutable state for cards and extensions

steffenkleinle:
### Short description

As discussed in #1666, mutating react state is bad practice and commonly leads to bugs as react is not aware of any changes.
Furthermore, mutating objects in place is generally not advisable as it introduces side effects and is also a source of bugs and makes the code less understandable.

This PR aims to address this issues in the case of cards and extensions by splitting the pure state (to be used in useStates/hooks) and helper methods around it to construct/mutate/check this state.

### Proposed changes

- Use simple objects for extensions
- Use correct react components for extension forms
- Use simple objects for cards (blueprints)
- Refactoring helper functions to reflect these changes

### Resulting Improvements

- No more `notifyUpdate` necessary
- Simple and easy adjustment of state (typesafe)
- Easy accessing of extension values

### Side effects

Hopefully none?

### Testing
1. Single card creation for bavaria, nuernberg and koblenz (/erstellen)
2. Create cards from link with queryParams find examples here: https://github.com/digitalfabrik/entitlementcard/pull/1089 Please also check correct addressData on pdf.
3. Generate cards via csv for nuernberg and bavaria. Find bulk import files here: (administration/resources/cards)
4. Import csv from freinet (bavaria)
5. Test csv export for nuernberg
6. Please also activate these cards to check if everything works with the qrCodes and deeplinks

### Resolved issues

Fixes: #1666.


administration/src/bp-modules/cards/ImportCardsController.tsx
```diff
@@ -59,23 +59,15 @@ const InnerImportCardsController = ({ region }: { region: Region }): ReactElemen
...
-      headers.forEach(header => {
-        const idx = csvHeader.indexOf(header)
-        if (idx === -1) {
-          // column is missing in csv
-          return
-        }
-        cardBlueprint.setValue(header, line[idx])
-      })
-      return cardBlueprint
+      return initializeCardFromCSV(projectConfig.card, line, csvHeader, region)
```
steffenkleinle, line 68: This is still really confusing, why are haders and csvHeader mixed here? Which one do I have to use?

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.ts
```diff
@@ -27,26 +25,25 @@ type UseCardGeneratorSelfServiceReturn = {
...
   downloadPdf: (code: CreateCardsResult, fileName: string) => Promise<void>
 }
-const initializeCardBlueprintForProject = (projectConfig: ProjectConfig): SelfServiceCard => {
-  const headers = getHeaders(projectConfig)
-  const selfServiceCard = new SelfServiceCard('', projectConfig.card)
-  headers.forEach(header => {
-    selfServiceCard.setValue(header, '')
-  })
-  return selfServiceCard
-}
```
steffenkleinle, line 42: Why was this necessary? It is just setting empty values anyway?

administration/src/cards/CsvFactory.ts
```diff
@@ -37,8 +36,6 @@ export function generateCsv(
 
 export const getCSVFilename = (cardBlueprints: CardBlueprint[]): string => {
   const filename =
-    cardBlueprints.length === 1
-      ? findExtension(cardBlueprints[0].extensions, NuernbergPassIdExtension)?.state?.passId
-      : 'Pass-ID[0]mass'
```
steffenkleinle, line 42: This only works for nuernberg, does it? What is the file name for other projects?

administration/src/cards/SelfServiceCard.ts
```diff
@@ -1,10 +0,0 @@
-import { CardBlueprint } from './CardBlueprint'
-
-class SelfServiceCard extends CardBlueprint {
-  // Override isValid() to remove expiration check
-  isValid(): boolean {
-    return this.isFullNameValid() && this.extensions.every(ext => ext.isValid())
-  }
```
steffenkleinle, line 7: Why?

administration/src/cards/CardBlueprint.ts
```diff
@@ -5,144 +5,205 @@ import { CardExtensions, CardInfo } from '../generated/card_pb'
...
+
+const createRandomId = () => Math.floor(Math.random() * 1000000)
+
+export const initializeCardBlueprint = (
+  cardConfig: CardConfig,
+  region: Region | undefined = undefined,
+  { id, fullName, expirationDate, extensions }: Partial<CardBlueprint> = {}
+): CardBlueprint => {
+  // TODO Why only if region is defined?
+  // const defaultExpirationDate = region ? PlainDate.fromLocalDate(new Date()).add(cardConfig.defaultValidity).toDaysSinceEpoch() : null
```
steffenkleinle, line 35: Why are we setting a expiration date only if a region is passed?

administration/src/cards/__tests__/SelfServiceCard.test.ts
```diff
@@ -1,30 +1,45 @@
...
   })
 
   it('should generate CardInfo', () => {
-    const card = new SelfServiceCard('Karla Koblenz', cardConfig)
-    expect(card.generateCardInfo().toJson({ enumAsInteger: true }) as object).toEqual({
+    const card = initializeCardBlueprint(cardConfig, undefined, { fullName: 'Karla Koblenz' })
+    expect(generateCardInfo(card).toJson()).toEqual({
       fullName: 'Karla Koblenz',
-      extensions: {},
+      // TODO why should this be empty/null?
```
steffenkleinle, line 33: Why should a newly initialized card have no values set yet?

administration/src/cards/extensions/AddressFieldExtensions.ts
```diff
@@ -1,45 +1,47 @@
...
-  public readonly name = AddressLine2Extension.name
-}
+const getAddressFieldExtension = <T extends AddressFieldExtension>(
+  name: T
+): Extension<AddressFieldExtensionState<T>> => ({
+  name,
+  Component: () => null,
+  getInitialState: () => ({ [name]: '' } as AddressFieldExtensionState<T>),
+  causesInfiniteLifetime: () => false,
+  // TODO Why no protobuf data?
```
steffenkleinle, line 23: Is it correct that address fields (and other extensions) are not included in protobuf data?

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -1,99 +1,73 @@
...
+export type BirthdayExtensionState = { [BIRTHDAY_EXTENSION_NAME]: number }
 
 const minBirthday = new PlainDate(1900, 1, 1)
+const getInitialState = (): BirthdayExtensionState => ({ birthday: minBirthday.toDaysSinceEpoch() })
 
-class BirthdayExtension extends Extension<BirthdayState, null> {
-  public readonly name = BirthdayExtension.name
+const BirthdayForm = ({ value, setValue, isValid }: ExtensionComponentProps<BirthdayExtensionState>): ReactElement => {
+  const { viewportSmall } = useWindowDimensions()
+  const inputColor = value === getInitialState() ? Colors.GRAY1 : Colors.BLACK
```
steffenkleinle, line 17: Shouldn't we rather set `null` as initial state and render `01.01.1900` as placeholder?

administration/src/cards/extensions/EMailNotificationExtension.ts
```diff
@@ -1,24 +1,19 @@
...
-  isValid(): boolean {
-    return true
-  }
+const EMailNotificationExtension: Extension<EmailNotificationExtensionState> = {
+  name: EMAIL_NOTIFICATION_EXTENSION_NAME,
+  Component: () => null,
+  // TODO Why initial null?
+  getInitialState: () => ({ [EMAIL_NOTIFICATION_EXTENSION_NAME]: '' }),
+  causesInfiniteLifetime: () => false,
+  // TODO Why no protobuf data?
```
steffenkleinle, line 12: Why is this initially set to null?

administration/src/cards/CardBlueprint.ts
```diff
@@ -5,144 +5,205 @@ import { CardExtensions, CardInfo } from '../generated/card_pb'
...
-        break
-      default:
-        const extensionIdx = this.cardConfig.extensionColumnNames.indexOf(key)
-        if (extensionIdx === -1) {
-          return
-        }
-        this.extensions[extensionIdx].fromString(value)
-    }
-  }
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
steffenkleinle, line 56: I was not able to construct the correct type, feel free to try :/

administration/src/cards/CardBlueprint.ts
```diff
@@ -5,144 +5,205 @@ import { CardExtensions, CardInfo } from '../generated/card_pb'
...
-    )
+export const isExpirationDateValid = (cardBlueprint: CardBlueprint, nullable = false): boolean => {
+  const today = PlainDate.fromLocalDate(new Date())
+  const expirationDate =
+    cardBlueprint.expirationDate !== null ? PlainDate.fromDaysSinceEpoch(cardBlueprint.expirationDate) : null
+  const startDay = cardBlueprint.extensions.startDay
+  const startDayDate = startDay !== undefined ? PlainDate.fromDaysSinceEpoch(startDay) : null
+
+  // TODO Why should this not be checked for self service? Why is this allowed to be nullable for csv/self service?
+  // TODO Adjust isValid
```
steffenkleinle, line 82: See other question, why is this allowed to be nullable in some cases? Why shouldn't this be checked for self service cards?

steffenkleinle:
We are mixing the names `Card` and `CardBlueprint` quite a lot. What is the meaning of the `blueprint` in this case? Can't we just use `card` consistently?

administration/src/cards/extensions/BirthdayExtension.tsx
f1sh1918: we still have design issues for that

administration/src/cards/SelfServiceCard.ts
f1sh1918: because it will be added in the backend. We don't have information in the frontend for that

administration/src/cards/CardBlueprint.ts
f1sh1918: tbh i have no idea. i was also really confused about this initialization stuff.
We have to pass this region in general because it will not be provided by form input nor csv file. And we get it from login information

administration/src/cards/extensions/AddressFieldExtensions.ts
f1sh1918: it depends on the extension if its included
```js
  setProtobufData(message: PartialMessage<CardExtensions>): void {
    message.extensionBirthday = {
      birthday: this.state?.birthday,
    }
  }
```

administration/src/cards/extensions/BirthdayExtension.tsx
f1sh1918: the problem is that then the current date is used if you open the datepicker. we can do that but when i remember right the more annoying thing was to the scroll hell if not setting a default value to sth closer to most common birthdays

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.ts
f1sh1918: because of the `isValid()` and `isFullnameValid()` fkts. Currently they will be executed directly because we have no untouched state

administration/src/cards/CsvFactory.ts
f1sh1918: we don't have csv export for other projects, so no filename needed or what do you mean?

administration/src/cards/extensions/EMailNotificationExtension.ts
f1sh1918: if it will not be passed as a queryParam or from the applications view it will be null since we have no information about email in some cases when cards will be created directly without application (nuernberg project)

administration/src/cards/CardBlueprint.ts
f1sh1918: hm i think the name shouldn't be nullable and the extension depend on which extension i think.
For selfServiceCards we have no expiration day in frontend

f1sh1918:
> We are mixing the names `Card` and `CardBlueprint` quite a lot. What is the meaning of the `blueprint` in this case? Can't we just use `card` consistently?

for me card would be fine

administration/src/cards/extensions/AddressFieldExtensions.ts
steffenkleinle: Yes, sure. But why is the address not included for example?

administration/src/cards/CsvFactory.ts
steffenkleinle: Ahh thanks for the explanation :+1:

f1sh1918: but actually this fallback is ugly i would use sth like `SozialpassMassExport`

administration/src/cards/extensions/AddressFieldExtensions.ts
f1sh1918: because its not needed in the qr code. this is just for the pdf letterhead and will not be transferred to the app.

administration/src/util/helper.ts
```diff
@@ -1,2 +1,11 @@
 export const isDevMode = (): boolean => window.location.hostname === 'localhost'
 export const isStagingMode = (): boolean => !!window.location.hostname.match(/staging./)
+
+export const updateArrayItem = <T>(array: T[], updatedItem: T, index: number): T[] => [
+  ...(index > 0 ? array.slice(0, index) : []),
+  updatedItem,
+  ...(index < array.length - 1 ? array.slice(index + 1) : []),
+]
```
michael-markl, line 8: I think it would make sense here to throw an Error if index is not in range.

administration/src/cards/CardBlueprint.ts
steffenkleinle: Okay, I'll change this to always initialize it.

administration/src/cards/CardBlueprint.ts
```diff
@@ -5,144 +5,203 @@ import { CardExtensions, CardInfo } from '../generated/card_pb'
...
 const MAX_ENCODED_NAME_LENGTH = 50
 
-export type JSONCardBlueprint<E = ExtensionInstance> = {
+export type CardBlueprint = {
   id: number
   fullName: string
-  expirationDate: string | null
-  extensions: (E extends Extension<infer T, unknown> ? JSONExtension<T> : never)[]
+  // TODO null necessary? csv?
+  expirationDate: number | null
```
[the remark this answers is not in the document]
steffenkleinle: Yes, and apparently apart from this it is necessary for self service cards. Not sure if it wouldn't make more sense to refactor it as an extension then too, though.

steffenkleinle:
@michael-markl sorry for rebasing, I did not realize you are currently reviewing.

administration/src/bp-modules/cards/ImportCardsController.tsx
f1sh1918: we have headers defined in the config  (headers) and the headers that were read from the csv (csvHeaders).

f1sh1918:
Could test that much @steffenkleinle when i click on create card i get this error

image

f1sh1918:
i added the most important testing scenarios @steffenkleinle 
If i forgot sth @ztefanie please feel free to add

administration/src/bp-modules/cards/ImportCardsController.tsx
steffenkleinle: I will create a refactoring issue as this is really confusing...

administration/src/cards/CardBlueprint.ts
```diff
@@ -5,144 +5,195 @@ import { CardExtensions, CardInfo } from '../generated/card_pb'
...
-    this.expirationDate = initParams ? PlainDate.fromLocalDate(new Date()).add(cardConfig.defaultValidity) : null
-    this.extensions = cardConfig.extensions.map(Extension => new Extension())
-    this.id = Math.floor(Math.random() * 1000000) // Assign some random ID
-    if (initParams) {
-      this.initialize(...initParams)
-    }
+type CardBlueprintKey = keyof Omit<Omit<CardBlueprint, 'extensions'>, 'id'>
+type KeyType = CardBlueprintKey | ExtensionKey | null
+
+const createRandomId = () => Math.floor(Math.random() * 1000000)
```
michael-markl, line 26: It's quite likely that there will be collisions. if we generate IDs with only 1,000,000 buckets. With 1000 cards, the probability that there is a collision is roughly 40% (see [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem))

What are these ids used for ? Is it possible to just use global counter?

steffenkleinle: Actually not sure, I just kept this from the previous implementation without any logical changes. I was wondering the same though... Not sure what they are used for.

michael-markl: Alright, maybe we should create an issue then. Seems quite suspicious.

f1sh1918:
merged recent main to get rid of db migration problems

f1sh1918 (changes requested):
One thing i forgot in the testing instructions @steffenkleinle 
Nuernberg also exports csv data to send it to a printer service from datev.
The export leads to wrong dates
Second table entry shows the correct output

Bildschirmfoto 2024-10-22 um 10 59 07

Another issue i faced that i can't create cards for koblenz.
I get "Ein unbekannter Fehler ist aufgetreten" even i type i correct data.
And i face some weird issued in the backend log for value encoding

```sh
SQL: INSERT INTO cards ("activationSecretHash", "cardInfoHash", "codeType", "entitlementId", "expirationDay", "issueDate", "regionId", revoked, "startDay") VALUES ($2a$11$IjK19SdDJNL3tevl9MkNZ.NDSJFAeS0kD7e3XlHUzmpzAooX0v1eu, �$���|v���gR�9��nU�Ɩ���7�, 1, 27, 20089, '2024-10-22T09:14:13.499388', 95, FALSE, 19723)
SQL: INSERT INTO cards ("activationSecretHash", "cardInfoHash", "codeType", "entitlementId", "expirationDay", "issueDate", "regionId", revoked, "startDay") VALUES (NULL, ���M��h0>���̵>�C�}4D6�Pp1�Ǥ\, 0, 27, 20089, '2024-10-22T09:14:13.529543', 95, FALSE, 19723)
```
I was facing this also when i switched from an array to a single card blueprint object when i tried to implement this first and then switched back to that array implementation. But no idea why this happens.
I think sth doesn't work correctly while creating hash

steffenkleinle:
> One thing i forgot in the testing instructions @steffenkleinle Nuernberg also exports csv data to send it to a printer service from datev. The export leads to wrong dates Second table entry shows the correct output

Should be fixed.

> Another issue i faced that i can't create cards for koblenz. I get "Ein unbekannter Fehler ist aufgetreten" even i type i correct data. And i face some weird issued in the backend log for value encoding

In the self service portal? I can't create/import cards with any user as `add/cards` is not available for me. For the self service portal, everything works for me.

> 
> ```shell
> SQL: INSERT INTO cards ("activationSecretHash", "cardInfoHash", "codeType", "entitlementId", "expirationDay", "issueDate", "regionId", revoked, "startDay") VALUES ($2a$11$IjK19SdDJNL3tevl9MkNZ.NDSJFAeS0kD7e3XlHUzmpzAooX0v1eu, �$����|v���gR��9����nU�Ɩ���7�, 1, 27, 20089, '2024-10-22T09:14:13.499388', 95, FALSE, 19723)
> SQL: INSERT INTO cards ("activationSecretHash", "cardInfoHash", "codeType", "entitlementId", "expirationDay", "issueDate", "regionId", revoked, "startDay") VALUES (NULL, ���M��h0>����̵>�C�}4D6�Pp1�Ǥ�\, 0, 27, 20089, '2024-10-22T09:14:13.529543', 95, FALSE, 19723)
> ```
>
> I was facing this also when i switched from an array to a single card blueprint object when i tried to implement this first and then switched back to that array implementation. But no idea why this happens. I think sth doesn't work correctly while creating hash

What exactly is weird/wrong about that? I just tested with main and I don't see a difference here...

administration/src/bp-modules/cards/ImportCardsController.tsx
steffenkleinle: https://github.com/digitalfabrik/entitlementcard/issues/1700

f1sh1918:
@steffenkleinle 
Looks like bulk import is now broken.
If i make a csv import for nuernberg or bavaria and try to export csv or create cards i get "Es ist ein unbekannter Fehler aufgetreten"

administration/src/cards/CardBlueprint.ts
f1sh1918: hm i removed the comment and don't get any error

f1sh1918 (approved):
Tested all use cases. works great. 👍

administration/src/cards/CardBlueprint.ts
ztefanie: i think this is the correct type:
`Array<{
  extension: Extension<InferExtensionStateType<typeof Extensions[number]>>;
  state: InferExtensionStateType<typeof Extensions[number]>;
}>`

administration/src/util/PlainDate.ts
```diff
@@ -56,11 +58,26 @@ class PlainDate {
...
 
+  static safeFromCustomFormat(value: string | null, format: string = 'dd.MM.yyyy'): PlainDate | null {
+    if (value === null) {
+      return null
+    }
+    try {
+      return PlainDate.fromCustomFormat(value, format)
+    } catch {
+      return null
+    }
```
ztefanie, line 74:
```suggestion
return value && PlainDate.fromCustomFormat(value, format).catch(() => null);
```

administration/src/util/helper.ts
```diff
@@ -1,2 +1,16 @@
...
 export const isStagingMode = (): boolean => !!window.location.hostname.match(/staging./)
+
+export const updateArrayItem = <T>(array: T[], updatedItem: T, index: number): T[] => {
+  if (index >= array.length || index < 0) {
+    throw new RangeError(`Index ${index} is out of range`)
+  }
+  return [
+    ...(index > 0 ? array.slice(0, index) : []),
+    updatedItem,
+    ...(index < array.length - 1 ? array.slice(index + 1) : []),
```
ztefanie, line 11:
```suggestion
    ...array.slice(0, index),
    updatedItem,
    ...array.slice(index + 1),
```

ztefanie (commented):
I did review this, but to be honest I did not really fully understand what you did here :-(
But everything looks fine to me and like a good improvement.
I added some small adjustment suggestions.

Tested works as expected

steffenkleinle:
> I did review this, but to be honest I did not really fully understand what you did here :-( But everything looks fine to me and like a good improvement. I added some small adjustment suggestions.
> 
> Tested works as expected

Please have a look at the issue, I think its mostly explained there. But basically, updated state is only recognized by react if the value does not shallow equal (i.e. it is the same object reference) as before. This prevents things like `cardBlueprint.expirationDate = newExpirationDate` as we did before. Therefore I replaced the usages of the classes `CardBlueprint` and `Extension` with corresponding plain objects.
If you have any other questions, feel free to ask and I can explain more in detail or in a call.

administration/src/cards/CardBlueprint.ts
steffenkleinle: Thank you @ztefanie, nice! :tada:

administration/src/util/PlainDate.ts
steffenkleinle: Sadly this does not work as
a) `value &&` might return empty strings `""`
b) `.catch()` does not exist on non-promise values

michael-markl:
I get a validation error on this MailNotification column (Bayern). However, clicking on "QR Codes drucken" still works.

Also, it's weird that the last line does not give a validation error as it's a Goldkarte and according to the table has a expiration date, although Goldkarten do not have expiration dates. Also 01.01.0025 looks like a very invalid date :D

grafik

f1sh1918:
> I get a validation error on this MailNotification column (Bayern). However, clicking on "QR Codes drucken" still works.
> 
> Also, it's weird that the last line does not give a validation error as it's a Goldkarte and according to the table has a expiration date, although Goldkarten do not have expiration dates. Also 01.01.0025 looks like a very invalid date :D
> 
> grafik
> 

Thx for testing. These issues existed before. I think we should remove the Mail notification field for csv. I will create an issue for that even it could also be used as a feature in the future so send mails with deeplinks. That should be discussed.

I think since it's a goldcard it will be ignored. Is not really good ux though. You can create a ticket

ztefanie (approved):

Issue #1659, opened 2024-10-02
NPM workspaces installing dependencies incorrectly


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
We have two npm workspaces: `administration` and `frontend/buildConfigs`.
Since the `frontend/buildConfigs` workspace has only a dependency to `typescript`, only this package and its subdependencies should be installed in the root.
However, a lot of other dependencies are installed in the root as well, seemingly arbitrarily.

**Describe the solution you'd like**
Clean up the dependencies and/or figure out why this happens. The dependencies should be installed correctly.

For example, the eslint plugin for jest fails to detect the version of jest since one is installed in the root and one in administration. This should be fixed (remove setting the version manually in `.eslintrc.js`).

**Describe alternatives you've considered**
Ignore it.

**Additional context**
Mattermost discussion: https://chat.tuerantuer.org/digitalfabrik/pl/8zmj3gaombrgtb1jqpspbpctzy

PR #1668, opened 2024-10-05
1596: Add various linting rules for react and typescript

steffenkleinle:
### Short description

Add various linting rules for react and typescript.
This brings the linting to nearly the exact level of integreat.

### Proposed changes

- Add various linting rules for typescript
- Add various linting rules for react/styling/...

### Side effects

None.

### Testing

Smoke test the app.

### Resolved issues

Fixes: #1596.


administration/scripts/build.ts
```diff
@@ -48,15 +56,92 @@ const writeStatsJson = argv.indexOf('--stats') !== -1
 // Generate configuration
 const config = configFactory('production')
 
+// Create the production build and print the deployment instructions.
+const build = (
```
steffenkleinle, line 60: These big changes are from moving code to come before usages.

administration/src/bp-modules/applications/JsonFieldArray.tsx
```diff
@@ -53,11 +53,11 @@ const JsonFieldArray = ({
   expandedRoot,
 }: JsonFieldViewProps<JsonField<'Array'>>) => {
   const [isExpanded, setIsExpanded] = useState(hierarchyIndex !== 1 || expandedRoot)
-  const children = jsonField.value.map((jsonField, index: number) => (
+  const children = jsonField.value.map(jsonField => (
     <JsonFieldView
       jsonField={jsonField}
       baseUrl={baseUrl}
-      key={index}
+      key={jsonField.name}
```
michael-markl, line 60: I think index is a better key here. I believe `name`  is not unique

administration/src/mui-modules/application/forms/AddressForm.tsx
```diff
@@ -20,14 +22,12 @@ const SubForms = {
...
-type Options = {}
-type AdditionalProps = {}
 
-const AddressForm: Form<State, Options, ValidatedInput, AdditionalProps> = {
+const AddressForm: Form<State, ValidatedInput> = {
   initialState: { ...createCompoundInitialState(SubForms), country: { shortText: 'Deutschland' } },
   getArrayBufferKeys: createCompoundGetArrayBufferKeys(SubForms),
   validate: createCompoundValidate(SubForms, {}),
-  Component: ({ state, setState }) => (
+  Component: ({ state, setState }: FormComponentProps<State>) => (
```
michael-markl, line 30: Can `FormComponentProps<State>` no longer be inferred?

administration/src/mui-modules/application/util/FormType.ts
```diff
@@ -14,15 +19,64 @@ export type ValidationError = { type: 'error'; message?: string }
...
+  Options extends Record<string, unknown> = Record<string, unknown>
+> = AdditionalProps &
   OptionsProps<Options> & {
     state: State
     setState: SetState<State>
   }
+
+// import { ReactElement } from 'react'
+//
+// import { SetState } from '../hooks/useUpdateStateCallback'
```
michael-markl, line 43: Probably this commented code should be removed?

administration/src/mui-modules/application/util/FormType.ts
```diff
@@ -2,7 +2,12 @@ import { ReactElement } from 'react'
```
michael-markl, line 2: Can you explain these changes?

administration/src/mui-modules/application/util/compoundFormUtils.ts
```diff
@@ -1,27 +1,30 @@
...
     // (see https://github.com/microsoft/TypeScript/issues/12936)
     const validKey = key as keyof VPre
     const validValue = value as VPre[typeof validKey]
-    result[validKey] = map(validValue, validKey)
-  }
+    return {
+      ...result,
+      [validKey]: map(validValue, validKey),
+    }
+  }, {})
```
michael-markl, line 19: Please revert

michael-markl (commented):
Just some drive-by comments :D

steffenkleinle:
> Just some drive-by comments :D

Long time since one of my PRs was reviewed by @michael-markl :tada:

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -22,14 +22,13 @@ export enum CardActivationState {
...
 }
 
-const extractCardInfoHashes = (codes: CreateCardsResult[]) => {
-  return codes.flatMap(code => {
+const extractCardInfoHashes = (codes: CreateCardsResult[]) =>
+  codes.flatMap(code => {
     if (code.staticCardInfoHashBase64) {
       return [code.dynamicCardInfoHashBase64, code.staticCardInfoHashBase64]
     }
     return code.dynamicCardInfoHashBase64
```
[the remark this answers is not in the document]
steffenkleinle: Perhaps, I am not familiar with this code yet. Perhaps @f1sh1918?

administration/src/mui-modules/application/forms/AddressForm.tsx
steffenkleinle: It can still be inferred, but we have a linting rule that requires all props of components to be explicitly defined. Personally, I think this rule makes sense to
a) make it obvious which props are needed
b) avoid any excess/wrong props

administration/src/mui-modules/application/util/FormType.ts
steffenkleinle: > The any type in TypeScript is a dangerous "escape hatch" from the type system. Using any disables many type checking rules and is generally best used only as a last resort or when prototyping code. This rule reports on explicit uses of the any keyword as a type annotation.

See https://typescript-eslint.io/rules/no-empty-object-type. Basically, using an empty object `{}` as type represents any non-nullish value, including literals like 0 and "". Since in this case we only want actual objects (or more specifically, records), I constrained the type to only records. Since we don't know the types of the values, th  type is `Record<string, unknown>`. Does that make sense?

administration/config/env.ts
```diff
@@ -64,10 +66,10 @@ function getClientEnvironment(publicUrl: string): {
...
     .filter(key => REACT_APP.test(key))
     .reduce(
-      (env: Record<string, string | boolean | undefined>, key: keyof typeof process.env) => {
-        env[key] = process.env[key]
-        return env
-      },
+      (env: Record<string, string | boolean | undefined>, key: keyof typeof process.env) => ({
+        ...env,
+        [key]: process.env[key],
+      }),
```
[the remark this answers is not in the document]
steffenkleinle: Can you explain why this should be reverted? In general it is not good practice to reassign params/objects, I think code has less side effects if you only work with new copies.
The linting rule is this one:
https://eslint.org/docs/latest/rules/no-param-reassign

administration/config/modules.ts
```diff
@@ -49,23 +49,20 @@ function getAdditionalModulePaths(paths: ReturnType<typeof getPaths>, options: {
...
 
-  const baseUrlResolved = path.resolve(paths.appPath, baseUrl)
-
-  if (path.relative(paths.appPath, baseUrlResolved) === '') {
+  if (baseUrlResolved && path.relative(paths.appPath, baseUrlResolved) === '') {
     return {
       src: paths.appSrc,
     }
   }
+  return {}
```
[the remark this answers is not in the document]
steffenkleinle: odd as in code style wise or odd as in wrong? :D I tried to avoid another return but I guess you are right, this is not very self explanatory. Reverted it and just added another return.

administration/config/env.ts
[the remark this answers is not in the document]
michael-markl: It creates many unnecessary copies of the object and is just unnecessarily slower (likely quadratic-ish runtime instead of linear-ish) and generates higher load for the garbage collector.

Also, in this kind of reduce functions, it is quite clear that mutations are totally fine.

I think I am, in general, not a huge fan of this linting rule :D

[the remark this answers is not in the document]
michael-markl: Also, the motivation of the rule only references the [function arguments object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) which apparently does not even exist for arrow functions.

administration/config/modules.ts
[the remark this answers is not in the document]
michael-markl: Yeah, I thought the previous version was a bit easier to read... But that's just taste I guess :D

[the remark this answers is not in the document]
steffenkleinle: Upon looking at the code I wrote again I agree :D

administration/config/env.ts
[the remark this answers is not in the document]
steffenkleinle: Works for me, don't really mind. I personally like the destructuring syntax more and think it's less error prone, but that is definitely a matter of personal preferences. I also see your point in terms of runtime and garbage collection, not sure if that has any noticeable effect in our case though.

administration/config/webpack.config.ts
```diff
@@ -230,7 +227,7 @@ function createWebpackConfig(webpackEnv: 'development' | 'production'): webpack.
               'scheduler/tracing': 'scheduler/tracing-profiling',
             }
           : {}),
-        ...(modules.webpackAliases && modules.webpackAliases.src ? modules.webpackAliases : {}),
+        ...(modules.webpackAliases.src ? modules.webpackAliases : {}),
```
ztefanie, line 230: Is it ensured that webpackAliases cannot be null?

So, if modules.webpackAliases is undefined or null, the old expression will return {}, whereas the new will throw an error because it tries to access .src on an undefined or null value.

administration/config/webpackDevServer.config.ts
```diff
@@ -32,6 +35,7 @@ function createWebpackDevServerConfig(proxy: Configuration['proxy'], allowedHost
     // really know what you're doing with a special environment variable.
     // Note: ["localhost", ".localhost"] will support subdomains - but we might
     // want to allow setting the allowedHosts manually for more complex setups
+    // eslint-disable-next-line no-nested-ternary
```
ztefanie, line 38: Can't the nested ternary be avoided? Would be nicer

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -29,10 +29,14 @@ const ApplicationList = styled(FlipMove)`
...
 // Necessary for FlipMove, as it cannot handle functional components
-export class ApplicationViewComponent extends React.Component<React.ComponentProps<typeof ApplicationCard>> {
+// eslint-disable-next-line react/prefer-stateless-function
+export class ApplicationViewComponent extends React.Component<ApplicationCardProps> {
   render(): ReactElement {
+    // eslint-disable-next-line react/prop-types
+    const { application } = this.props
     return (
-      <ApplicationListCard key={this.props.application.id}>
+      // eslint-disable-next-line react/prop-types
```
ztefanie, line 38: Can't we fullfill the rules?

ztefanie (approved):
Tested, works as expected :rocket:

administration/src/bp-modules/applications/ApplicationsOverview.tsx
steffenkleinle: I think not possible for class components without installing deprecated prop-types again :disappointed: I did not spend too much time investigating though.

administration/config/webpackDevServer.config.ts
steffenkleinle: Since this is just a configuration file and copied from another project, I did not feel like spending too much time on this. But yes, they could be avoided. I think its okay here though.

administration/src/errors/DefaultErrorMap.tsx
```diff
@@ -23,10 +23,16 @@ type ErrorExtensions = {
...
-  if (!extensions || !extensions['code']) {
+  if (!extensions || extensions.code === undefined) {
     return defaultError
   }
-  switch (extensions['code']) {
+
+  const cardInfo = extensions.encodedCardInfoBase64
+    ? CardInfo.fromBinary(base64ToUint8Array(extensions.encodedCardInfoBase64))
+    : null
+  const codeTypeText = extensions.codeType === CodeType.Dynamic ? 'Aktivierungscode' : 'statische QR-Code'
```
michael-markl, line 33: Wouldn't it be better to keep it in the only case that it is used? Maybe you need to introduce a variable scope as in
```ts
  case GraphQlExceptionCode.InvalidQrCodeSize: {
      const cardInfo = CardInfo.fromBinary(base64ToUint8Array(extensions.encodedCardInfoBase64!))
      const codeTypeText = extensions.codeType === CodeType.Dynamic ? 'Aktivierungscode' : 'statische QR-Code'
      ...
     }
```

administration/config/env.ts
```diff
@@ -92,10 +94,13 @@ function getClientEnvironment(publicUrl: string): {
...
-      env[key] = JSON.stringify(raw[key])
-      return env
-    }, {}),
+    'process.env': Object.keys(raw).reduce(
+      (env: Record<string, string>, key) => ({
+        ...env,
+        [key]: JSON.stringify(raw[key]),
+      }),
+      {}
+    ),
```
michael-markl, line 103: Please also revert 😅

administration/src/mui-modules/application/util/FormType.ts
michael-markl: Thanks for the explanation! (I guess one could argue whether `Options` has to be an actual object, but it's totally fine to restrict it to actual objects).

michael-markl (commented):
Somehow GitHub is extremely laggy for me on the diff-page of this PR 😄

administration/src/errors/DefaultErrorMap.tsx
steffenkleinle: TIL. And also that I did not read the docs of the rule properly. Thanks!

administration/config/webpack.config.ts
steffenkleinle: Yes, I fixed the fixed, is always guaranteed to be an object.

administration/.eslintrc.js
```diff
@@ -1,28 +1,27 @@
 module.exports = {
+  root: true,
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint'],
+  plugins: ['@typescript-eslint', 'jest', 'jsx-a11y', 'jsx-expressions', 'prefer-arrow', 'react', 'react-hooks'],
   extends: [
+    'airbnb',
+    'airbnb/hooks',
```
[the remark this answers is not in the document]
steffenkleinle: I didn't, but could be done in a follow-up cleanup issue.

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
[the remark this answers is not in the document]
michael-markl: Ah, true. I guess I was confused as some other programming languages expect an array of arrays as input to flatMap...

michael-markl (approved):
Did not test anything :)

steffenkleinle:
Any ideas why this is failing? I think I did not change anything in the CI/app-toolbelt...
```
/home/circleci/.npm-global/lib/node_modules/app-toolbelt/node_modules/ts-node/src/index.ts:859
    return new TSError(diagnosticText, diagnosticCodes, diagnostics);
           ^
TSError: ⨯ Unable to compile TypeScript:
error TS5109: Option 'moduleResolution' must be set to 'NodeNext' (or left unspecified) when option 'module' is set to 'NodeNext'.

    at createTSError (/home/circleci/.npm-global/lib/node_modules/app-toolbelt/node_modules/ts-node/src/index.ts:859:12)
    at reportTSError (/home/circleci/.npm-global/lib/node_modules/app-toolbelt/node_modules/ts-node/src/index.ts:863:19)
    at /home/circleci/.npm-global/lib/node_modules/app-toolbelt/node_modules/ts-node/src/index.ts:1379:34
    at Object.compile (/home/circleci/.npm-global/lib/node_modules/app-toolbelt/node_modules/ts-node/src/index.ts:1451:13)
    at Module.m._compile (/home/circleci/.npm-global/lib/node_modules/app-toolbelt/node_modules/ts-node/src/index.ts:1617:30)
    at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Object.require.extensions.<computed> [as .ts] (/home/circleci/.npm-global/lib/node_modules/app-toolbelt/node_modules/ts-node/src/index.ts:1621:12)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1024:12)
    at Module.require (node:internal/modules/cjs/loader:1233:19) {
  diagnosticCodes: [ 5109 ]
}

Exited with code exit status 1
```

administration/src/cards/extensions/KoblenzReferenceNumberExtension.tsx
```diff
@@ -14,6 +14,7 @@ const KoblenzReferenceNumberMaxLength = 15
 class KoblenzReferenceNumberExtension extends Extension<KoblenzReferenceNumberState, null> {
   public readonly name = KoblenzReferenceNumberExtension.name
 
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
```
michael-markl, line 17: Maybe this is not a great rule either :D But let's ignore it for now^^

steffenkleinle: Will be fixed in #1667.

michael-markl:
> Any ideas why this is failing? I think I did not change anything in the CI/app-toolbelt...

I feel like the way we deploy app-toolbelt is not great. I think that ts-node somehow grabs the wrong configuration file (ie. the one defined in the entitlementcard vs the one defined in app-toolbelt) and/or the wrong typescript version.

f1sh1918:
> > Any ideas why this is failing? I think I did not change anything in the CI/app-toolbelt...
> 
> I feel like the way we deploy app-toolbelt is not great. I think that ts-node somehow grabs the wrong configuration file (ie. the one defined in the entitlementcard vs the one defined in app-toolbelt) and/or the wrong typescript version.

Probably yes but the `moduleResolution` should be node in both and nothing was changed about this in that pr

steffenkleinle:
> > Any ideas why this is failing? I think I did not change anything in the CI/app-toolbelt...
> 
> I feel like the way we deploy app-toolbelt is not great. I think that ts-node somehow grabs the wrong configuration file (ie. the one defined in the entitlementcard vs the one defined in app-toolbelt) and/or the wrong typescript version.

@michael-markl thanks! Any ideas how to fix that? Otherwise I'll try to manually dissect what causes this new failure...

michael-markl:
I guess there are two problems:
a) app-toolbelt does not seem to properly support typescript v5.2  (as the error message you reference appears since this version) https://www.totaltypescript.com/concepts/option-module-must-be-set-to-nodenext-when-option-moduleresolution-is-set-to-nodenext
b) somehow typescript 5.5.4 is installed in the root directory of the entitlementcard as the npm workspaces setup seems somehow broken. (although the typescript version is restricted to v5.0.4 in both workspaces). I tried to fix the npm workspaces setup here (by deleting all lock files and node_modules directories), but now some tests are failing (potentially due to some unwanted package up or downgrade, but I haven't investigated that). But: The app-toolbelt error message has gone :D see this commit https://github.com/digitalfabrik/entitlementcard/commit/a81b59dcb11f4cb5473f6ff9f04eefdd82e11f76

steffenkleinle:
> I guess there are two problems: a) app-toolbelt does not seem to properly support typescript v5.2 (as the error message you reference appears since this version) https://www.totaltypescript.com/concepts/option-module-must-be-set-to-nodenext-when-option-moduleresolution-is-set-to-nodenext b) somehow typescript 5.5.4 is installed in the root directory of the entitlementcard as the npm workspaces setup seems somehow broken. (although the typescript version is restricted to v5.0.4 in both workspaces). I tried to fix the npm workspaces setup here (by deleting all lock files and node_modules directories), but now some tests are failing (potentially due to some unwanted package up or downgrade, but I haven't investigated that). But: The app-toolbelt error message has gone :D see this commit [a81b59d](https://github.com/digitalfabrik/entitlementcard/commit/a81b59dcb11f4cb5473f6ff9f04eefdd82e11f76) and [CircleCI](https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/6475/workflows/0c626808-eee7-4a45-adc0-039c7ba360f3)

Thank you! I'll have a look and try to get the tests working again.

michael-markl:
After some struggling, I get the feeling that npm workspaces is a bad mess... Maybe we should swap to pnpm (or simply remove the root npm package, as it's mostly unnecessary?)

steffenkleinle:
> After some struggling, I get the feeling that npm workspaces is a bad mess... Maybe we should swap to pnpm (or simply remove the root npm package, as it's mostly unnecessary?)

I have the same feeling. yarn also works flawlessly in Integreat. There is already an issue to clean up npm workspaces, perhaps you could add your thoughts there? #1659

steffenkleinle:
I think I got everything working again. Only had to change some types from ReactElement to ReactNode and reset some mocks in the tests.

f1sh1918:
> After some struggling, I get the feeling that npm workspaces is a bad mess... Maybe we should swap to pnpm (or simply remove the root npm package, as it's mostly unnecessary?)

I mean we added this to able to make imports from frontend to administration.
In general the plan was to have a general config package and import that to frontend, backend and administration. I also think yarn monorepo is the better solution. Even we have some properly configured repos with that solution

administration/package.json
```diff
@@ -101,7 +112,7 @@
     "ts:check": "tsc --build",
     "format": "NODE_ENV=test eslint . --fix && prettier . --write",
     "generate-graphql": "graphql-codegen --config graphql-codegen.yml",
-    "generate-protobuf": "protoc -I ../specs --es_out=src/generated --es_opt=target=ts --plugin=protoc-gen-es=./node_modules/.bin/protoc-gen-es ../specs/card.proto",
+    "generate-protobuf": "protoc -I ../specs --es_out=src/generated --es_opt=target=ts --plugin=protoc-gen-es=../node_modules/.bin/protoc-gen-es ../specs/card.proto",
```
seluianova, line 115: if it only for me the new command doesn't work?

steffenkleinle: Hmm. I don't remember changing this, maybe this was just a mistake or done automatically for some reason? Please feel free to revert this, sorry.

seluianova: well, it's failing in the ci otherwise

steffenkleinle: Did you try removing node_modules and reinstalling?

seluianova: no, I decided to complain first 😄
ok, I see now, the folder has been moved from administration to the project root, thanks

Issue #1415, opened 2024-04-29
Revoke existing cards via HTTP endpoint


ztefanie:
**Is your feature request related to a problem? Please describe.**
Adjust the existing HTTP PUT endpoint, so it will revoke existing cards, when the expiration date is set to a date in the past. 

**Describe the solution you'd like**
- Check if a entry in the csv has expiration date in the past
- find the card that is associated with this entry
- revoke the card

**Describe alternatives you've considered**

**Additional context**
For testing:
- Upload a csv with a valid user info
- activate the card in the app
- upload a csv with the same user info, but an expiration date in the past
- open the app: The card should no longer work

PR #1674, opened 2024-10-08
1415: Revoke existing cards via user import endpoint

seluianova:
### Short description
Revoke cards by entitlementId, when user entitlements are revoked or the expiration date is updated to a past date

### Proposed changes
- when user entitlements are updated, check whether they have been revoked or whether the endDate has been moved back in time and revoke existing cards
- add new tests

### Side effects / open questions
- do we want to update cards if the startDate (in the user entitlements) has been moved to the future?
- do we want to handle the case, when regionId (in the user entitlements) is updated?

### Testing
Full testing is currently blocked, until the app config is ready.
Currently the following can be checked:

1. Open administration --> login as a koblenz project admin --> go to settings --> create token
2. Try to import users using curl or Postman
`curl -H "Authorization: Bearer <my_token>" -F file=@import_testdata_local.csv http://0.0.0.0:8000/users/import`

**CSV content:**
```
```

3. Go to http://localhost:3000/ and select Koblenz
4. Go to http://localhost:3000/erstellen, create a card with a birthday 10.06.2003 and a reference number 123K
5. Update the CSV file to import users, change revoked to ‘true’ or endDate to past
6. Import users again --> check in the database that previously created cards (dynamic and static) have been revoked

### Resolved issues
Fixes: #1415


f1sh1918:
Just a general question. Does it make sense to also revoke cards where the DateEnd is in the past.
I think we we agreed on that koblenz revokes cards by setting `revoked: true` as this is a clean process.
To give another opportunity to revoke cards by changing the `EndDate` makes it just more complex i think.
I can't remember that there is a reason why we need that

ztefanie:
I updated base branch, so I can fully test it.

f1sh1918:
> I updated base branch, so I can fully test it.

I think the removal of the end date check is missing. So not sure if its ready for review

ztefanie (changes requested):
I tested this and it only partly works as expected:

Use cases:
1) If I import non revoked user data, then create a card and activate it in the app, then upload a new csv where the card is revoked, the card is revoked in the app and it is displayed properly to the user
2) If I import non revoked user data, then create a pdf, then upload a new csv where the card is revoked and then try to activate the card in the app, with the pdf previously created, i get the error message "Fehler beim lesen des codes", here should be a message: "Your card was revoked"
3) it i go to self service (/erstellen) and try to create a card for a revoked user entitlement i get the error "Sie sind scheinbar nicht berechtigt einen KoblenzPass zu erstellen. Bitte prüfen Sie Ihre Eingaben" i think this is missleading, i think the error message should be "You are no longer entitlet for a koblenzpass" or something like thi.

seluianova:
Updated:
* removed the condition for revoking cards when the user's entitlement endDate is in the past
* refactored importData() function a bit to reduce complexity, not sure we need even more splitting

f1sh1918:
Some thoughts to 3) @seluianova 
The "Sie sind scheinbar nicht berechtigt...." was not final and the message will be adjusted to
```js
Wir konnten Ihre Angaben nicht im System finden. Bitte überprüfen Sie Ihre Angaben und versuchen Sie es erneut.
```
[design](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=736-697&node-type=frame&t=JlupTMijnkeoB2Cs-0)

Due to this upcoming adjustments it will make sense to throw a different error and show a different message to the user as steffi mentioned
@hauf-toni 
So we have another state that should be handled and we need a proper error message for that.
If the user data can be found but the user is no longer entitled.
This should be reflected in the design

seluianova:
> The "Sie sind scheinbar nicht berechtigt...." was not final and the message will be adjusted to

Updated the message when the user entitlement is not found. Will add an additional message once provided.
Do we also want to separate errors when the user entitlement is revoked and when it has expired?

f1sh1918:
i think the expiration info will be fetched from another endpoint when the app was started and there should already be an error message.

seluianova:
@f1sh1918 you mean we should allow card creation when user entitlement is expired?

f1sh1918:
Ah okay i thought the issue is about what to show if an activated card is expired in the app.

I think we don't need separate errors but this depends on the error message. I think the user does not have to know if its revoked or the entitlement is expired the important thing is that the entitlement is not valid anymore

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -145,9 +145,9 @@ class CardMutationService {
         )
         val userHash = Argon2IdHasher.hashKoblenzUserData(user)
 
-        val userEntitlements = transaction { UserEntitlementsRepository.findUserEntitlements(userHash.toByteArray()) }
+        val userEntitlements = transaction { UserEntitlementsRepository.findByUserHash(userHash.toByteArray()) }
         if (userEntitlements == null || userEntitlements.revoked || userEntitlements.endDate.isBefore(LocalDate.now())) {
```
f1sh1918, line 149: I think we have to throw different exceptions here. At least we should create a TODO and ticket if it will not be done here.
We have to distinguish between the input data does not match the entitlement hash then we throw `UserEntitlementNotFoundException()`

If revoked or expired i would throw
`UserEntitlementExpiredException` or sth similar with a separate message

f1sh1918 (approved):
Works great. Tested on ios.
Please address the one issue i mentioned

ztefanie (approved):
not tested again

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
seluianova: I will need to associate a new exception with some error message.
Do you think `Sie sind nicht mehr berechtigt, einen KoblenzPass zu erstellen` is fine?

hauf-toni:
[Here](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=778-663&t=7BmRX1tjzp8daWp1-1) is the error display for the third variant that @ztefanie mentioned. 
Here is also the [updated version](https://www.figma.com/design/e4cHjTPVjrAVoYYLp1zfC5/KoblenzPass-ShareFile?node-id=85-704&t=elNwZlgmKpllQR4B-1) (only the text has been changed), which includes the changes from Koblenz.

PR #1675, opened 2024-10-09
Ignore intellij settings and use prettier code style

steffenkleinle:
### Short description

I frequently have changes to my intellij setting files which is rather annoying as I sometimes commit them accidentally.

### Proposed changes

- Use code style from prettierrc
- Add inspection profiles to gitignore

### Side effects

None.

### Testing

None.

### Resolved issues

Fixes: None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 node_modules
-.idea/inspectionProfiles
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+.idea/inspectionProfiles
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,5 +1,18 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <HTMLCodeStyleSettings>
+      <option name="HTML_SPACE_INSIDE_EMPTY_TAG" value="true" />
+    </HTMLCodeStyleSettings>
+    <JSCodeStyleSettings version="0">
+      <option name="USE_SEMICOLON_AFTER_STATEMENT" value="false" />
+      <option name="FORCE_SEMICOLON_STYLE" value="true" />
+      <option name="SPACE_BEFORE_FUNCTION_LEFT_PARENTH" value="false" />
+      <option name="USE_DOUBLE_QUOTES" value="false" />
+      <option name="FORCE_QUOTE_STYlE" value="true" />
+      <option name="ENFORCE_TRAILING_COMMA" value="Remove" />
+      <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
+      <option name="SPACES_WITHIN_IMPORTS" value="true" />
+    </JSCodeStyleSettings>
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value />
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1 +1,2 @@
-node_modules
\ No newline at end of file
+node_modules
+.idea/inspectionProfiles
```
ztefanie, line 2: Why do we not want to share settings of this? Can't we agree on some settings and keep sharing this file?

steffenkleinle: Hmmm, I am okay with adding this, but I think this just defines whether eslint is run by IntelliJ and errors are shown in the editor. I just figured that sometimes manual configuration is needed as automatic eslint config might not work for everybody.

.idea/dataSources.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="DataSourceManagerImpl" format="xml" multifile-model="true">
-    <data-source source="LOCAL" name="ehrenamtskarte@localhost" uuid="d83188d7-bad5-4b46-b4d7-13cabcdaf7c7">
+    <data-source source="LOCAL" name="ehrenamtskarte@db-postgis" uuid="d83188d7-bad5-4b46-b4d7-13cabcdaf7c7">
```
ztefanie, line 4: Please undo this.
If you want to avoid this, maybe try this:
1) open database view in intellij
2) rightclick on "ehrenamtskarte@db-postgis" and properties
3) adjust the name in the first line

ztefanie (commented):
I know you are waiting for reviews on this, but I am not really happy with the solution you provided and i do currently have no time / priority in having a deeper look into this.

The solution i would prefer is
- telling intellij to use code styled defined in es lint / prettier
- autoformat on save
- share all these settings for intellij with the whole team

.idea/dataSources.xml
steffenkleinle: Nice, thank you! That works :)

steffenkleinle:
> I know you are waiting for reviews on this, but I am not really happy with the solution you provided 

What exactly do you not like about this? This is just a small PR for me to finally get rid of all those uncommitted changes which are really annoying. I don't think this worsens behavior in comparison to the current state.

> The solution i would prefer is
> 
>     * telling intellij to use code styled defined in es lint / prettier
> 
>     * autoformat on save
> 
>     * share all these settings for intellij with the whole team

I definitely see the point and I think this might make sense, but I also see why someone would not want this. In any case, if you'd like additional things to be done, please create an issue for it. It does not really have anything to do with what I did here (except its also IntelliJ configuration).

f1sh1918 (approved):

ztefanie:
> > I know you are waiting for reviews on this, but I am not really happy with the solution you provided
> 
> What exactly do you not like about this? This is just a small PR for me to finally get rid of all those uncommitted changes which are really annoying. I don't think this worsens behavior in comparison to the current state. Adding `inspectionProfiles` to `.gitignore` does not change IntelliJ settings for anyone. The adjusted code style is just for generated code from IntelliJ, e.g. auto imports and so on and would be needed in any case.
> 
> > The solution i would prefer is
> > ```
> > * telling intellij to use code styled defined in es lint / prettier
> > 
> > * autoformat on save
> > 
> > * share all these settings for intellij with the whole team
> > ```
> 
> I definitely see the point and I think this might make sense, but I also see why someone would not want this. In any case, if you'd like additional things to be done, please create an issue for it. It does not really have anything to do with what I did here (except its also IntelliJ configuration).

I do not want any additional things to be done. But adding `inspectionProfiles` to .gitignore is a step in the wrong direction. I would rather enforce more shared settings, than less. This PR makes it "worse" in this point.

ztefanie:
Andy and I both do not have the problem with the inspection profiles being created, so this seems to be more a local problem of your setup that should be fixed locally.

steffenkleinle:
> > > I know you are waiting for reviews on this, but I am not really happy with the solution you provided
> > 
> > 
> > What exactly do you not like about this? This is just a small PR for me to finally get rid of all those uncommitted changes which are really annoying. I don't think this worsens behavior in comparison to the current state. Adding `inspectionProfiles` to `.gitignore` does not change IntelliJ settings for anyone. The adjusted code style is just for generated code from IntelliJ, e.g. auto imports and so on and would be needed in any case.
> > > The solution i would prefer is
> > > ```
> > > * telling intellij to use code styled defined in es lint / prettier
> > > 
> > > * autoformat on save
> > > 
> > > * share all these settings for intellij with the whole team
> > > ```
> > 
> > 
> > I definitely see the point and I think this might make sense, but I also see why someone would not want this. In any case, if you'd like additional things to be done, please create an issue for it. It does not really have anything to do with what I did here (except its also IntelliJ configuration).
> 
> I do not want any additional things to be done. But adding `inspectionProfiles` to .gitignore is a step in the wrong direction. I would rather enforce more shared settings, than less. This PR makes it "worse" in this point.

Alright, I'll just revert the change. Atm I don't have any inspection profiles, but will push them if they ever reappear.

ztefanie (approved):

Issue #1688, opened 2024-10-16
Custom scheme deeplinks possibly open wrong app


f1sh1918:
**Describe the bug**
When we have installed several apps that have the `berechtigungskarte` custom scheme like "Ehrenamtskarte" or a Sozialpass App it will open one of these apps and not a particular app with a custom deeplink like:
```js
berechtigungskarte://koblenz.sozialpass.app/activation/code#ClcKLQoNS2FybGEgS29ibGVuehD5nAEaGAoCCF8SBAjqvgEqBAiLmgEyBgoEMTIzSxIQLGlG5it%2FngAjzKPiUGOjnhoUXzANF5DJmEmvQUjD1CAdqf5vks0%3D/ --ios

```

**To Reproduce**
1. Install sozialpass nuernberg or ehrenamtskarte, then install koblenzPass
2. Create a card and open a custom scheme deeplink
3. Check that a wrong app like ehrenamt can be assoziated

**Expected behavior**
The correct app will be opened

**Additional context**
This may only happen for ios but we have to ensure that the domain is unique so we should adjust our custom deeplink f.e.
koblenzpass://koblenz.sozialpass.app/activation/code#ClcKLQoNS2FybGEgS29ibGVuehD5nAEaGAoCCF8SBAjqvgEqBAiLmgEyBgoEMTIzSxIQLGlG5it%2FngAjzKPiUGOjnhoUXzANF5DJmEmvQUjD1CAdqf5vks0%3D/ --ios
or
nuernbergpass://nuernberg.sozialpass.app/activation/code#ClcKLQoNS2FybGEgS29ibGVuehD5nAEaGAoCCF8SBAjqvgEqBAiLmgEyBgoEMTIzSxIQLGlG5it%2FngAjzKPiUGOjnhoUXzANF5DJmEmvQUjD1CAdqf5vks0%3D/ --ios

We should use sth really specific in the domain part

PR #1690, opened 2024-10-16
1688: Fix custom scheme deeplinks

f1sh1918:
### Short description

Currently we use the same custom scheme for the deeplinks of all our apps `berechtigungskarte://`
If a user has more than just one of these apps installed it is not clear which app should be opened by clicking a deeplink.
While in android there would be a app selection (which is also should be avoided), in ios there will be just one of the apps proposed to open the link. Therefore each app should have its own custom scheme to prevent that

### Proposed changes

- add a custom scheme for each project
- adjust the info.plist file for ios to use this value from the buildConfig instead of a hardcoded string
- adjust tests
- adjust custom deeplink creation to use the correct scheme

### Side effects

- none, since currently custom deeplinks only exist on the /activation route of the administration

### Testing
- Create a card. Copy the code from the console.
- Create a koblenz pass and check the activation link
- install koblenzpass and bavaria on your device. 
- open deeplink:
```js
npx uri-scheme open koblenzpass://koblenz.sozialpass.app/activation/code#<activationCode>/ --ios
npx uri-scheme open koblenzpass://koblenz.sozialpass.app/activation/code#<activationCode>/ --android
npx uri-scheme open ehrenamtbayern://bayern.ehrenamtskarte.app/activation/code#<activationCode>/ --ios
npx uri-scheme open ehrenamtbayern://bayern.ehrenamtskarte.app/activation/code#<activationCode>/ --android
```
- correct app should be opened

### Resolved issues

Fixes: #1688

### DIFF
--- a/administration/src/project-configs/showcase/config.ts
+++ b/administration/src/project-configs/showcase/config.ts
@@ -1,7 +1,8 @@
 import BavariaCardTypeExtension from '../../cards/extensions/BavariaCardTypeExtension'
 import RegionExtension from '../../cards/extensions/RegionExtension'
-import bayern, { applicationJsonToCardQuery, applicationJsonToPersonalData } from '../bayern/config'
+import { applicationJsonToCardQuery, applicationJsonToPersonalData } from '../bayern/config'
 import { DataPrivacyBaseText, dataPrivacyBaseHeadline } from '../bayern/dataPrivacyBase'
+import pdfConfiguration from '../bayern/pdf'
 import { ProjectConfig } from '../getProjectConfig'
 
 const config: ProjectConfig = {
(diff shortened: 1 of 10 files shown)


seluianova (approved):
Tested android only for eak, koblenz, nuernberg using the given instruction - works good!

steffenkleinle (approved):
Tested on android, works as expected :tada:

f1sh1918:
this pr is a fix for ios so it would be nice if someone can test it there @steffenkleinle or @ztefanie

ztefanie:
i tested it on bayern ios and the deeplink opens the card preview properly, but when i click "activate" i get the error "Fehler beim Lesen des Codes".

f1sh1918:
> i tested it on bayern ios and the deeplink opens the card preview properly, but when i click "activate" i get the error "Fehler beim Lesen des Codes".

is your card still active? i faced this issue once with revoked card. If not i will retest it next week. card data shouldn't have changed though

f1sh1918:
> i tested it on bayern ios and the deeplink opens the card preview properly, but when i click "activate" i get the error "Fehler beim Lesen des Codes".

Hm are you sure that the card is still active? I faced this issue for a revoked card which is okay.
I tested this with different cards

I updated the docs since they were deprecated

Issue #1618, opened 2024-09-04
Write Tests for api token feature


ztefanie:
Write Tests for the api token feature. 

Related: https://github.com/digitalfabrik/entitlementcard/pull/1615

PR #1694, opened 2024-10-18
1618: Add tests for api token feature

ztefanie:
### Short description
Added tests for the api token feature

### Resolved issues

Fixes: #1618


backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,165 @@
...
+
+            ApiTokenService().deleteApiToken(1, mockDfe)
+
+            val tokenNoLongerExists = ApiTokens.select { ApiTokens.id eq 1 }.singleOrNull() != null
+            assertTrue(tokenNoLongerExists)
+        }
+    }
+
+    @Test
+    fun deleteApiToken_deletesNoTokenFormOtherProjekt() {
```
f1sh1918, line 128:
```suggestion
    fun deleteApiToken_deletesNoTokenFromOtherProjekt() {
```

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        every { mockJwtPayload.adminId } returns TestAdministrators.KOBLENZ_PROJECT_ADMIN.id
+
+        transaction {
+            val numberOfTokensBefore = ApiTokens.selectAll().count()
+            ApiTokenService().deleteApiToken(3, mockDfe)
```
f1sh1918, line 133: shouldn't here also be an exception thrown if an admin from another project tries to delete a token.
is there even a token with the `id=3` ?

f1sh1918 (commented):
Only small findings. Rest looks good

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
ztefanie: You are right, there is no token with id 1, i adjusted it. With the current implementation there is no exception thrown for this case, it is just handled as "this token does not exist", as in the database request we filter for project matching. I fixed the test to test what it is expected to do.

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                it[tokenHash] = ByteArray(1)
+                it[expirationDate] = LocalDate.now()
+            }
+
+            val tokenExists = ApiTokens.select { ApiTokens.id eq 1 }.count() > 0
+            assertTrue(tokenExists)
+
+            ApiTokenService().deleteApiToken(1, mockDfe)
+
+            val tokenNoLongerExists = ApiTokens.select { ApiTokens.id eq 1 }.singleOrNull() != null
```
seluianova, line 99: `ApiTokens.select { ApiTokens.id eq 1 }.singleOrNull() != null` means, the token is not null, so, it exists.
the correct check would be == null, but then the test will fail.
I assume this is because the token is created with projectId = 1, this means that only admins from eak can delete this token.

seluianova, line 99: :wrench: Optional, but we can use the helper class TestData to create test objects.

```suggestion
        every { mockJwtPayload.adminId } returns TestAdministrators.KOBLENZ_PROJECT_ADMIN.id
        
        val tokenId = TestData.createApiToken(creatorId = TestAdministrators.KOBLENZ_PROJECT_ADMIN.id)

        transaction {
            val tokenExists = ApiTokens.select { ApiTokens.id eq tokenId }.count() > 0
            assertTrue(tokenExists)

            ApiTokenService().deleteApiToken(tokenId, mockDfe)

            val tokenNoLongerExists = ApiTokens.select { ApiTokens.id eq tokenId }.singleOrNull() != null
```

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,172 @@
...
+            assertTrue(tokenNoLongerExists)
+        }
+    }
+
+    @Test
+    fun deleteApiToken_deletesFromOtherAdminTokenSuccessfully() {
```
seluianova, line 105: it should be possible only inside one project. between Koblenz and Nuernberg is should fail

seluianova (commented):
Added a few suggestions.
Do you see a need to add tests for getApiTokenMetaData() as well?

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
ztefanie: Yes correct, that is checked by the next test: deleteApiToken_deletesNoTokenFormOtherProject

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
ztefanie: Fixed it.

ztefanie: good point, thanks. Added it

ztefanie:
> Added a few suggestions. Do you see a need to add tests for getApiTokenMetaData() as well?
 Will add.

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
seluianova: I mean, this test is named - “deletes token successfully”, but in this test the Koblenz admin tried to delete a token created by the Nürnberg admin, which should not be successful.
The test is currently green because of the same typo in the check condition as it was above: 
`val tokenNoLongerExists = ApiTokens.select { ApiTokens.id eq 1 }.singleOrNull() == null`
I think if we want to check that a token can be deleted by other admin, both admins (creating and deleting) should be from the same project.

ztefanie: Hi Tory, I understand your confusion here.

The expected behavior is, that:
1) An admin can delete all tokens of his project (even tokens of other admins)
2) An admin cannot delete tokens of other projects

This test deletes the token successfully as expected, as the token is created for the koblenz project (by the nuernberg admin). This would not be possible of course in the real software, but as we do not have another koblenz admin i just used the nuernberg admin id here. But to avoid confusion i will create a second koblenz project admin

seluianova: > the token is created for the koblenz project (by the nuernberg admin)

in the latest version of the test, the token was created for the nuernberg project by the nurnberg admin.
and it wasn't removed at the end.
because `ApiTokens.select { ApiTokens.id eq 1 }.singleOrNull() != null` means, the token is not null, so it's still there

ztefanie: ok thanks for explaining and you patience. I now fixed it.

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,211 @@
...
+
+        transaction {
+            TestData.createApiToken(creatorId = TestAdministrators.KOBLENZ_PROJECT_ADMIN.id)
+            val tokenExists = ApiTokens.select { ApiTokens.id eq 1 }.count() > 0
+            assertTrue(tokenExists)
+
+            ApiTokenService().deleteApiToken(1, mockDfe)
+
+            val tokenNoLongerExists = ApiTokens.select { ApiTokens.id eq 1 }.singleOrNull() == null
+            assertTrue(tokenNoLongerExists)
```
seluianova, line 97:
```suggestion
            val tokenId = TestData.createApiToken(creatorId = TestAdministrators.KOBLENZ_PROJECT_ADMIN.id)
            val tokenExists = ApiTokens.select { ApiTokens.id eq tokenId }.count() > 0
            assertTrue(tokenExists)

            ApiTokenService().deleteApiToken(tokenId, mockDfe)

            val tokenNoLongerExists = ApiTokens.select { ApiTokens.id eq tokenId }.singleOrNull() == null
            assertTrue(tokenNoLongerExists)
```
fix for the failing test

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,211 @@
...
+        every { mockJwtPayload.adminId } returns TestAdministrators.KOBLENZ_PROJECT_ADMIN.id
+
+        transaction {
+            TestData.createApiToken(creatorId = TestAdministrators.KOBLENZ_PROJECT_ADMIN.id)
+
+            val metaData = ApiTokenQueryService().getApiTokenMetaData(mockDfe)
+            assertEquals(1, metaData.size)
+            assertEquals(1, metaData[0].id)
+            assertEquals(TestAdministrators.KOBLENZ_PROJECT_ADMIN.email, metaData[0].creatorEmail)
+            assertEquals(LocalDate.now().plusYears(1).toString(), metaData[0].expirationDate)
```
seluianova, line 169:
```suggestion
            val tokenId = TestData.createApiToken(creatorId = TestAdministrators.KOBLENZ_PROJECT_ADMIN.id)

            val metaData = ApiTokenQueryService().getApiTokenMetaData(mockDfe)
            assertEquals(1, metaData.size)
            assertEquals(tokenId, metaData[0].id)
            assertEquals(TestAdministrators.KOBLENZ_PROJECT_ADMIN.email, metaData[0].creatorEmail)
            assertEquals(LocalDate.now().plusYears(1).toString(), metaData[0].expirationDate)
```
fix for the failing test

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,211 @@
...
+
+    @Test
+    fun deleteApiToken_deletesFromOtherAdminTokenSuccessfully() {
+        every { mockJwtPayload.adminId } returns TestAdministrators.KOBLENZ_PROJECT_ADMIN.id
+
+        transaction {
+            TestData.createApiToken(creatorId = TestAdministrators.KOBLENZ_PROJECT_ADMIN_2.id)
+
+            val tokenBefore = ApiTokens.select { ApiTokens.id eq 1 }.count() > 0
+            assertNotNull(tokenBefore)
```
seluianova, line 109:
```suggestion
            assertTrue(tokenBefore)
```
tokenBefore is boolean

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    @Test
+    fun deleteApiToken_deletesNoTokenFormOtherProject() {
+        every { mockJwtPayload.adminId } returns TestAdministrators.KOBLENZ_PROJECT_ADMIN.id
+        transaction {
+            TestData.createApiToken(creatorId = TestAdministrators.NUERNBERG_PROJECT_ADMIN.id)
+
+            val numberOfTokensBefore = ApiTokens.selectAll().count()
+            ApiTokenService().deleteApiToken(1, mockDfe)
```
seluianova, line 125:
```suggestion
            val tokenId = TestData.createApiToken(creatorId = TestAdministrators.NUERNBERG_PROJECT_ADMIN.id)

            val numberOfTokensBefore = ApiTokens.selectAll().count()
            ApiTokenService().deleteApiToken(tokenId, mockDfe)
```
every token created in this test set will get an incremented id.
It will be equal to '1' only in the first running test. So we need to make sure that the real id is passed to delete() function.

backend/src/test/kotlin/app/ehrenamtskarte/backend/auth/webservice/ApiTokenServiceTest.kt
```diff
@@ -0,0 +1,210 @@
...
+    @Test
+    fun deleteApiToken_deletesFromOtherAdminTokenSuccessfully() {
+        every { mockJwtPayload.adminId } returns TestAdministrators.KOBLENZ_PROJECT_ADMIN.id
+        TestData.createApiToken(creatorId = TestAdministrators.KOBLENZ_PROJECT_ADMIN_2.id)
+
+        transaction {
+            val tokenBefore = ApiTokens.select { ApiTokens.creatorId eq TestAdministrators.KOBLENZ_PROJECT_ADMIN_2.id }.singleOrNull()
+            assertNotNull(tokenBefore)
+
+            ApiTokenService().deleteApiToken(tokenBefore!!.get(ApiTokens.id).value, mockDfe)
```
f1sh1918, line 110:
```suggestion
            ApiTokenService().deleteApiToken(tokenBefore!![ApiTokens.id].value, mockDfe)
```
Got a warning here to use indexing

f1sh1918 (approved):
Looks good to me :)
Just a minor find

seluianova (approved):
Thanks, lgtm 👍🏻

PR #1695, opened 2024-10-18
Downgrade secure storage

f1sh1918:
### Short description

Due to this issue we should lock the version to 9.0.0
https://github.com/mogol/flutter_secure_storage/issues/762

### Proposed changes

- lock version 9.0.0

### Side effects

- hopefully none
-

### Testing
1. Activate card and test if the data will be kept in storage after app restart

I tested this with ios 17.5 und 18.0


ztefanie (approved):
Tested on mac, still builds the app.

steffenkleinle (approved):
LGTM, not tested

Issue #1696, opened 2024-10-21
Add custom font per project


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As koblenz requested to use their custom font in the native app, we should provide a local font and integrate it

**Describe the solution you'd like**
- provide local font for koblenz (TeX Gyre Heros (free licensed font) ) done in #1702 
- provide the font in the buildConfig config of koblenz done in #1702 
- add the standard font to the build dine in #1702 
- check all texts if theme is used

PR #1702, opened 2024-10-22
1696: Custom font

f1sh1918:
### Short description

Koblenz wants to have their custom font in the app.

### Proposed changes

- add font assets
- define `fontFamily` in buildConfig (set default font `Roboto` for other projects)
- apply font family from build config to typo of dark and light theme

### Side effects

- none

### Testing

You can just set in `themes.dart` a different font family but roboto and texgyreheroes do not differ that much, so hard to see. I checked several times that spacing differ a bit and so on

If you find a way in flutter inspector to show the exact font. Let me know
### Resolved issues

Fixes partially : #1696


michael-markl (commented):
I'm not sure if it's easily possible, but it would be great if we would only bundle the font for the koblnez pass. Have you tried something along the lines of this guide: https://docs.flutter.dev/deployment/flavors#conditionally-bundling-assets-based-on-flavor?

frontend/pubspec.yaml
```diff
@@ -121,17 +121,14 @@ flutter:
...
-  #
+  fonts:
+    - family: TexGyreHeroes
+      fonts:
+        - asset: assets/koblenz/fonts/texgyreheros-regular.otf
+        - asset: assets/koblenz/fonts/texgyreheros-italic.otf
+          style: italic
+        - asset: assets/koblenz/fonts/texgyreheros-bold.otf
+          weight: 700
+          #
```
steffenkleinle, line 132:
```suggestion

```

frontend/assets/koblenz/fonts/GUST e-foundry License.txt
```diff
@@ -0,0 +1,29 @@
+This is a preliminary version (2006-09-30), barring acceptance from
```
steffenkleinle, line 1: :upside_down_face: Does this file name has to be this way? I'd prefer it if there would be no whitespaces/special characters.

steffenkleinle (commented):
Tested on Android, I believe I see a small difference in the fonts, not that sure about it though. Everything works still though.

The text `Weitere Aktionen` is way thinner though, is that intended (2nd is koblenz)?
Screenshot_20241024-151946_Ehrenamt_1
Screenshot_20241024-152225_KoblenzPass_1

seluianova:
> If you find a way in flutter inspector to show the exact font. Let me know

have you tried this?
https://stackoverflow.com/questions/74262829/how-can-i-see-the-style-applied-to-a-widget
it looked convincing, I couldn't make it work though 😀 but maybe I have some local issue

seluianova (commented):
> roboto and texgyreheroes do not differ that much, so hard to see

I think I can see the difference between these fonts and it seems to me, that the new font is not actually applied in the Koblenz app.
For example, the letter G is pretty much different in TexGyreHeroes.
I enlarged the screenshot for the better visibility.
Above is the Koblenz app, below is what the font should look like.
image

seluianova (changes requested):
I guess I had to 'request changes' here

f1sh1918:
> I guess I had to 'request changes' here

Alright i will check that. No clue why it doesn't apply :/

frontend/assets/koblenz/fonts/GUST e-foundry License.txt
f1sh1918: i just copied everything from an downloaded archieve

seluianova (approved):
Allright, I can see changes now, thanks!

> But i will create a separate issue and base it on this one if you are fine since this pr fixed the general issue

I don't mind creating a new task, but just as an idea, maybe sometimes we can just keep the current task open when it's partially fixed. Up to you

steffenkleinle (approved):
Tested on Android

Issue #1685, opened 2024-10-15
Add the ability to delete an old card during deeplink activation


seluianova:
**Is your feature request related to a problem? Please describe.**
When a user already has a card in the app and tries to activate a new card via a deep link, he receives the message `Ihr Kartenlimit wurde erreicht. Bitte entfernen Sie vorher eine Karte, damit Sie eine neue aktivieren können.` and is forced to first delete the old card and then repeat the activation.

It would be nice to remove the old card during the deeplink activation, instead of forcing the user to do it manually.

**Describe the solution you'd like**
- display the message like 'You already have an activated card on this device. If you proceed with the actvation, the old card will be removed.'
- remove the old card, if the user wants to proceed, then add a new one.

**Describe alternatives you've considered**
Leave as is

**Additional context**

f1sh1918:
@seluianova
We may consider that this can also happen for activation via scanning activation code. So maybe this can be implemented for both cases or a seperate issue  should be created

seluianova:
> We may consider that this can also happen for activation via scanning activation code. So maybe this can be implemented for both cases or a seperate issue should be created

Scanning of the new code is currently not available until you remove the old card, and you don't have to switch between pdf/app/browser to do that, so not sure if that makes sense.

seluianova:
Cards amount will be increased instead https://github.com/digitalfabrik/entitlementcard/issues/1741

seluianova:
oops, thanks @steffenkleinle

PR #1709, opened 2024-10-23
1685: Add the ability to remove an old card during deep link activation

seluianova:
### Short description
Currently when a user already has a card in the app and tries to activate a new card via a deep link, he receives the message `Ihr Kartenlimit wurde erreicht. Bitte entfernen Sie vorher eine Karte, damit Sie eine neue aktivieren können.` and is forced to first delete the old card and then repeat the activation.

This PR adds the ability to delete the old card during activation.

### Proposed changes
- change the displayed message to `Sie haben bereits einen aktivierten Pass auf diesem Gerät. Wenn Sie fortfahren, wird der alte Pass gelöscht.`
- if the user proceeds, first delete the old card and then activate the new card
- this only applies to projects, where only 1 card is allowed (so, Nürnberg shouldn't be affected)
- small refactoring of RemoveCardConfirmationDialog to be able to reuse it

### Side effects
If the user proceeds and deletes the old card, but then some error occurs (`Fehler beim Lesen des Codes` for example), the user will have neither the old nor the new card in the app. Not sure if this is an issue though. If he keeps the old QR code, he should be able to activate the old card again.

### Testing
Instruction for Koblenz:

1. Open administration --> login as a koblenz project admin --> go to settings --> create token
2. Import users using curl or Postman
`curl -H "Authorization: Bearer <my_token>" -F file=@import_testdata_local.csv http://0.0.0.0:8000/users/import`

**CSV content:**
```
```

3. Go to http://localhost:3000/ and select Koblenz
4. Go to http://localhost:3000/erstellen, create a card with a birthday 10.06.2003 and a reference number 123K
5. Activate the card in the app
6. Go to http://localhost:3000/erstellen again, create another card with a birthday 10.06.2003 and a reference number 123K, _but enter some different name_ (required to produce a different cardInfo, otherwise you get the `Dieser Pass wurde bereits auf dem Gerät aktiviert.` message; alternatively some other data should differ)
7. Try to activate the new card via deep link:
`npx uri-scheme open berechtigungskarte://koblenz.sozialpass.app/activation/code#<activation-code> --android`
**It should be possible to remove the old card during the activation.**

### Resolved issues
Fixes: #1685

### DIFF
--- a/frontend/lib/identification/identification_page.dart
+++ b/frontend/lib/identification/identification_page.dart
@@ -49,7 +49,7 @@ class IdentificationPageState extends State<IdentificationPage> {
               startVerification: () => _showVerificationDialog(context, settings, userCodeModel),
               startActivation: () => _startActivation(context),
               startApplication: _startApplication,
-              openRemoveCardDialog: () => _openRemoveCardDialog(context),
+              openRemoveCardDialog: () => _openRemoveCardDialog(context, userCodeModel),
             ));
           }
 
@@ -105,10 +105,12 @@ class IdentificationPageState extends State<IdentificationPage> {
     );
   }
 
-  Future<void> _openRemoveCardDialog(BuildContext context) async {
-    final userCodeModel = Provider.of<UserCodeModel>(context, listen: false);
-    await RemoveCardConfirmationDialog.show(
-        context: context, userCode: userCodeModel.userCodes[cardIndex], carouselController: carouselController);
+  Future<void> _openRemoveCardDialog(BuildContext context, UserCodeModel userCodeModel) async {
+    final removed =
+        await RemoveCardConfirmationDialog.show(context: context, userCode: userCodeModel.userCodes[cardIndex]);
+    if (removed && userCodeModel.userCodes.isNotEmpty) {
+      carouselController.previousPage(duration: Duration(milliseconds: 500), curve: Curves.linear);
+    }
   }
 
   void _moveCarouselToLastPosition() {
--- a/frontend/lib/identification/verification_workflow/dialogs/remove_card_confirmation_dialog.dart
+++ b/frontend/lib/identification/verification_workflow/dialogs/remove_card_confirmation_dialog.dart
@@ -8,18 +8,15 @@ import 'package:provider/provider.dart';
 
 class RemoveCardConfirmationDialog extends StatefulWidget {
   final DynamicUserCode userCode;
-  final CarouselController carouselController;
 
-  const RemoveCardConfirmationDialog({super.key, required this.userCode, required this.carouselController});
+  const RemoveCardConfirmationDialog({super.key, required this.userCode});
 
-  static Future<void> show(
-          {required BuildContext context,
-          required DynamicUserCode userCode,
-          required CarouselController carouselController}) =>
-      showDialog(
+  static Future<bool> show(
+          {required BuildContext context, required DynamicUserCode userCode, CarouselController? carouselController}) =>
+      showDialog<bool>(
         context: context,
-        builder: (_) => RemoveCardConfirmationDialog(userCode: userCode, carouselController: carouselController),
-      );
+        builder: (_) => RemoveCardConfirmationDialog(userCode: userCode),
+      ).then((removed) => removed ?? false);
 
   @override
   RemoveCardConfirmationDialogState createState() => RemoveCardConfirmationDialogState();
@@ -33,7 +30,6 @@ class RemoveCardConfirmationDialogState extends State<RemoveCardConfirmationDial
       userCodeModel.removeCodes();
     } else {
       userCodeModel.removeCode(widget.userCode);
-      widget.carouselController.previousPage(duration: Duration(milliseconds: 500), curve: Curves.linear);
     }
     Navigator.of(context).pop(true);
   }
--- a/frontend/assets/koblenz/l10n/override_de.json
+++ b/frontend/assets/koblenz/l10n/override_de.json
@@ -38,6 +38,7 @@
     "headline": "Pass aktivieren",
     "description": "Wollen Sie diesen Pass aktivieren?",
     "limitReached": "Ihr Passlimit wurde erreicht. Bitte entfernen Sie vorher einen Pass, damit Sie einen neuen aktivieren können.",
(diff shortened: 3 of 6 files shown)


frontend/lib/identification/verification_workflow/dialogs/remove_card_confirmation_dialog.dart
```diff
@@ -8,18 +8,15 @@ import 'package:provider/provider.dart';
 
 class RemoveCardConfirmationDialog extends StatefulWidget {
```
michael-markl, line 9: I guess, this can be converted to a StatelessWidget

frontend/lib/identification/identification_page.dart
```diff
@@ -105,10 +105,12 @@ class IdentificationPageState extends State<IdentificationPage> {
...
-  Future<void> _openRemoveCardDialog(BuildContext context) async {
-    final userCodeModel = Provider.of<UserCodeModel>(context, listen: false);
-    await RemoveCardConfirmationDialog.show(
-        context: context, userCode: userCodeModel.userCodes[cardIndex], carouselController: carouselController);
+  Future<void> _openRemoveCardDialog(BuildContext context, UserCodeModel userCodeModel) async {
+    final removed =
+        await RemoveCardConfirmationDialog.show(context: context, userCode: userCodeModel.userCodes[cardIndex]);
+    if (removed && userCodeModel.userCodes.isNotEmpty) {
+      carouselController.previousPage(duration: Duration(milliseconds: 500), curve: Curves.linear);
+    }
```
michael-markl, line 113: 👍🏽

michael-markl (commented):
> If the user proceeds and deletes the old card, but then some error occurs (Fehler beim Lesen des Codes for example), the user will have neither the old nor the new card in the app. Not sure if this is an issue though. If he keeps the old QR code, he should be able to activate the old card again.

To avoid this scenario, couldn't we only remove the previous code if the activation was successfull?

michael-markl:
>  If we activate the new card first and it fails and the old card is not removed, it will also be a bit strange because the user has confirmed the removal.

I see ...

Another idea would be to adjust the phrasing to make it less confusing. Maybe if we phrase it like "If you continue, the previous card will be replaced." it would be less confusing? I am also fine with your implementation, though.

seluianova:
> Another idea would be to adjust the phrasing to make it less confusing. Maybe if we phrase it like "If you continue, the previous card will be replaced." it would be less confusing?

You mean, swap deletion and activation AND change the message text, right?

I'm not sure my point is plausible, but I feel like it would also be more error-prone:
What if an error occurred during deletion?
What if the app stopped (the device turned off or something) after the new card was activated but before the old card was deleted?

The current order is at least straightforward.

michael-markl:
> You mean, swap deletion and activation AND change the message text, 

Yes

> I'm not sure my point is plausible, but I feel like it would also be more error-prone:
What if an error occurred during deletion?
What if the app stopped (the device turned off or something) after the new card was activated but before the old card was deleted?

I guess the proper way would be to pass an optional `cardToReplace` to `activateCard` and (if `cardToReplace` is not null) instead of simply adding the newly activated card to the list of cards, replace `cardToReplace` with the new one and persist this array in one step. Then we would never violate the maxNumberOfCards  "constraint".

> The current order is at least straightforward.

It's easier to implement, but can be more frustrating for users in the error case (e.g. if the user wants to activate his new card, but his old card would still be valid for some period of time). As I said, I'm fine with both options, though :D

michael-markl:
Maybe, the question rather is why do we implement this, instead of just allowing to have multiple cards activated at the same time?

f1sh1918:
> Maybe, the question rather is why do we implement this, instead of just allowing to have multiple cards activated at the same time?

 i don't see any issues to increase the number of cards if its getting too complicated.
 I'm a big fan of easy solutions and i didn't expect this getting so tricky.
F.e. with a BahnCard you can also have multiple cards. And if someone has even exceeded the limit then he/she should delete one card.

michael-markl:
> i don't see any issues to increase the number of cards if its getting too complicated.

Yes, but I don't quite recall why we did set the card limit to 1 for Bayern. Maybe we wanted to sell this feature of having multiple Ehrenamtskarten activated (@dkehne) ?

michael-markl:
> but because Bavaria wanted to make sure that two people cannot use the card in parallel.

I think this is a different kind of restriction @dkehne . 

* The number of devices that can be activated by a single card at the same time (the one you mentioned  - this is restricted to 1 for all projects currently).

vs
 
* The number of activated cards on a single device (this is different between Bayern & Nürnberg currently; we restricted this for Bayern to 1)

Can we lift the latter restriction in Bayern to 5 (or whatever number we have for Nürnberg)?

dkehne:
Yes we can lift it to the limit we have in. Nuremberg. I will inform bavaria

seluianova:
So, I will just close this PR, right?

michael-markl:
I mean, maybe this should be decided in the entitlementcard dev group / weekly, I just wanted to share my thoughts 😬 
Didn't want to enforce any particular solution here :D

f1sh1918:
I mean we can ask for opinions on monday. I won't be here but i vote for extending the amount of cards

michael-markl:
> Just one concern from my side:
If we just allow multiple cards, and a Koblenz user issues a new card (e.g., via card extension), the old card will be automatically revoked but will stay in the app, however it can no longer be used, and there is basically no point in keeping it.

But we probably want to allow multiple cards on a single device for Koblenz for the same reasons we allow it for Nürnberg: E.g., parents might want to have the card for their child and for themselves on their phone

I am not sure how we design(ed) the card extension workflow. But maybe, if the process is started from within the app, it could be possible to add the information (for example through the deeplink) which card should be replaced when activating the new card.

ztefanie:
I think this could get quite complicated when someone has two cards (the own one and the one of her/his child) activated and want to add new one. Which of the two activated cards is then deleted?
Therefore I think I also prefer to increase the number of cards and the user needs to delete manually first. Not really a nice flow though.

f1sh1918:
i created an issue for increasing card amount here
https://github.com/digitalfabrik/entitlementcard/issues/1741
i think this pr can be closed afaics

steffenkleinle:
Can we delete this branch? :eyes:

seluianova:
maybe :)

Issue #1699, opened 2024-10-22
Adjust displayes card ui to koblenz design


ztefanie:
See discussion here: https://chat.tuerantuer.org/digitalfabrik/pl/epjkucbp6bbyxgzgsytc3wnhfo

PR #1711, opened 2024-10-23
1699: Card design koblenz

f1sh1918:
### Short description

Adjust the koblenz card design

### Proposed changes

- replace logos
- set theme color
- adjust card branding
- add extra upscale for header logo

Note: I adjusted the header right logo a bit because fine adjustments with the current card branding are difficult. At some point we gonna have to create separate card template widgets per project

### Side effects

- none

### Testing

1. Start koblenz local and activate card
2. Check the other apps if the card design doesn't break

### Resolved issues

Fixes: #1699 
Fixes: #1741


frontend/build-configs/koblenz/index.ts
```diff
@@ -18,8 +18,8 @@ export const koblenzCommon: CommonBuildConfigType = {
     },
     categories: [9, 10, 11, 12, 13, 14, 15, 16, 17, 18],
     theme: {
-        primaryLight: "#8377A9",
-        primaryDark: "#8377A9",
+        primaryLight: "#d10074",
```
steffenkleinle, line 21: :thought_balloon: I want a theme color :D Not related to your PR though

steffenkleinle (approved):
Tested on android, looks good

seluianova (commented):
Is a pink background intended here? (I know we get it from the theme, but anyway)
image

f1sh1918:
> Is a pink background intended here? image

The issue here is that material3 themes use surfaceTintColor. Usually a very light version of the primary color is used.
Since for koblenz the color is darker its still visible here.
The issue ist that if i override this color in general on lots of positions it will not be used like
image

So probably i have to override this color just for the dialogs

seluianova:
> I will talk to toni in the afternoon

ready for review or you still want to clarify/update anything?

seluianova:
just found another issue:
the card is not rendered correctly on the card verification.
image

```
======== Exception caught by rendering library =====================================================
The following assertion was thrown during layout:
A RenderFlex overflowed by 0.767 pixels on the bottom.

The relevant error-causing widget was: 
  Column Column:file:///home/tory/IdeaProjects/entitlementcard/frontend/lib/identification/id_card/card_content.dart:181:32
```

ztefanie (commented):
Did just a quick review and did not test as two people already did. lgtm. Would have liked to have a background (image) on the card, but just personal preference.

f1sh1918:
> Did just a quick review and did not test as two people already did. lgtm. Would have liked to have a background (image) on the card, but just personal preference.

Yes same for me but we don't have a proper design for that. This also leads to the issue that page background and card are both white and don't have a good contrast :(
Maybe we can improve that at a later point

f1sh1918:
@seluianova i added a tint color to override this slight pink color for dialogs.
Fixed also the pixel overflow.
Additional design optimization like colors and contrasts will be done here
https://github.com/digitalfabrik/entitlementcard/issues/1742

Issue #1713, opened 2024-10-24
Fix wrong validation of empty mail notification for csv import


f1sh1918:
**Describe the bug**
During the csv import of bavarian bulk data, the mail notification field will be validated with an error if no mail address for notification is provided. 

**To Reproduce**
1. Go to http://localhost:3000/cards/import
2. Import bulk data
3. See the mail notification field error validation while cards can be created

**Expected behavior**
Since mail notification field is optional and can be empty it should not be error validation

PR #1715, opened 2024-10-24
1713: Fix MailNotification validation in CSV import

steffenkleinle:
### Short description

Fix MailNotification validation in CSV import.

### Proposed changes

- Use default extension state if not provided
- Show `-` if string is empty

### Side effects

None.

### Testing

1. Go to http://localhost:3000/cards/import
2. Import bulk data
3. See the mail notification field error validation while cards can be created

### Resolved issues

Fixes: #1713


administration/src/cards/CardBlueprint.ts
```diff
@@ -125,8 +127,8 @@ export const isValueValid = (cardBlueprint: CardBlueprint, cardConfig: CardConfi
       return isExpirationDateValid(cardBlueprint) || hasInfiniteLifetime(cardBlueprint)
     default: {
       const extensionName = getExtensionNameByCSVHeader(cardConfig, columnHeader)
-      const extension = getExtensions(cardBlueprint).find(({ extension }) => extension.name === extensionName)
-      return extension?.extension.isValid(extension.state) ?? false
+      const extension = cardConfig.extensions.find(extension => extension.name === extensionName)
+      return extension?.isValid(cardBlueprint.extensions) ?? false
```
steffenkleinle, line 131: This is the relevant change, we need to check among all extensions configured in the card config, not just the ones actually set.

f1sh1918 (approved):
Looks good. tested on chrome

f1sh1918:
> Could you please provide a csv example to reproduce the issue? I tried this, MailNotification is empty, but I don't get a validation error on the main branch.
> 
> ```
> Name,Ablaufdatum,Kartentyp,MailNotification
> Qwerty,,Goldkarte,
> ```

you can use the one in the repo on the main branch (resources/bulk-bavaria) it should be marked red

steffenkleinle:
> > Could you please provide a csv example to reproduce the issue? I tried this, MailNotification is empty, but I don't get a validation error on the main branch.
> > ```
> > Name,Ablaufdatum,Kartentyp,MailNotification
> > Qwerty,,Goldkarte,
> > ```
> 
> you can use the one in the repo on the main branch (resources/bulk-bavaria) it should be marked red

Exactly. The error was shown if `mailNotifications` is not present at all in the csv.

seluianova (commented):
> The error was shown if MailNotification is not present at all in the csv.

:thought_balloon: okay, than it works as requested, but tbh I would expect 'missing column' error in such case, because this column is specified in csv format:
image
there is no clue, that this column can be ommited.
also I think it would be nice to mark mandatory columns with an * in the table view, when not all of them are mandatory.

seluianova (approved):
not sure if the thought above should block this PR, so adding my approval anyway

steffenkleinle:
> > The error was shown if MailNotification is not present at all in the csv.
> 
> 💭 okay, than it works as requested, but tbh I would expect 'missing column' error in such case, because this column is specified in csv format

Yes, it is specified since it can also be passed to allow for automatic email notifications.

> there is no clue, that this column can be ommited. also I think it would be nice to mark mandatory columns with an * in the table view, when not all of them are mandatory.

Sounds like a good addition. I'll create a new issue for this. We could also do something like `MailNotification (optional)`.

Issue #1573, opened 2024-08-14
Rate limiting for createCard mutation


seluianova:
**Is your feature request related to a problem? Please describe.**
Follow up to https://github.com/digitalfabrik/entitlementcard/issues/1421

_Since this endpoint is not protected, it should be ensured that IPs with too many requests should be blocked, especially because the argon2id calculation is expensive._

**Describe the solution you'd like**
- add a logger statement for failed cardCreation (hash does not match) 
- add the failregex pattern in salt config 
Example: https://git.tuerantuer.org/DF/salt/pulls/187/files#issuecomment-1538

**Describe alternatives you've considered**
Maybe there is a possibility to implement limiting at the infrastructure level?

**Additional context**

f1sh1918:
Probably we only want to ban failed requests
Currently we log these failed attempts and ban the ip.
Currently we have a Limit of 5 attempts.
Not sure if this is a proper solution here.
I think setting a general rate limit is difficult to find since there may be people creating 30 cards in an hours (internal card creation)

Any other suggestions here?

seluianova:
> Currently we log these failed attempts and ban the ip for endpoints that don't require login.

@f1sh1918 could you please tell me how to reproduce that exactly? If I just send requests that fail, I don't get a ban

f1sh1918:
> > Currently we log these failed attempts and ban the ip for endpoints that don't require login.
> 
> @f1sh1918 could you please tell me how to reproduce that exactly? If I just send requests that fail, I don't get a ban

we don't do this right now for this mutation.
You can check `SignInMutationService` for example. We log there failed logins. 
And with fail2ban we check the logs if the same ip addresses with a certain log pattern occur frequently and ban the ip address.
https://git.tuerantuer.org/DF/salt/src/branch/master/states/entitlementcard/files/fail2ban/filter.conf
But i dunno exactly how to configure this, f.e. set after how many requests someone will be banned. Probably we should use different values depending on the log pattern
Maybe @maxammann or @svenseeberg can help here.

maxammann:
> > > Currently we log these failed attempts and ban the ip for endpoints that don't require login.
> > 
> > @f1sh1918 could you please tell me how to reproduce that exactly? If I just send requests that fail, I don't get a ban
> 
> we don't do this right now for this mutation. But you can test with failed login attempts on production f.e.
> 
> You can check `SignInMutationService` for example. We log there failed logins. 
[...]

PR #1717, opened 2024-10-25
1573: Add logging info for rate limiting of card creation endpoint

steffenkleinle:
### Short description

Add logging info for rate limiting of card creation endpoint.

### Proposed changes

- Add logging info for rate limiting of card creation endpoint

### Side effects

None.

### Testing

?

### Resolved issues

Fixes: #1573.
Salt issue: https://git.tuerantuer.org/DF/salt/pulls/187


f1sh1918:
Local testing is not possible. I think the salt regex should be tested. Not sure how we can achieve this. 
For staging the threshold is 5 attempts within one or 2 minutes i think until your ip address will be for 10minutes. (have to check the exact values) We should discuss if that's okay

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -147,6 +147,7 @@ class CardMutationService {
 
         val userEntitlements = transaction { UserEntitlementsRepository.findUserEntitlements(userHash.toByteArray()) }
         if (userEntitlements == null || userEntitlements.revoked || userEntitlements.endDate.isBefore(LocalDate.now())) {
+            logger.info("${context.remoteIp} ${userEntitlements?.userHash} failed to create a new card")
```
f1sh1918, line 150: Hm as @seluianova described in the issue the rate limiting (and logging) should apply if no cardHash can be found
so
`if (userEntitlements == null)` and not if expired or revoked.
I don't think we should also log and possibly ban user that were revoked or expired.

When we optimize our user messages we will even differ between not found or expired and will throw different exceptions. At least i think we should do this.
So we will need a separate `== null` condition

f1sh1918 (commented):
I think we should only log and ban if no hash was found

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
```diff
@@ -147,6 +147,9 @@ class CardMutationService {
 
         val userEntitlements = transaction { UserEntitlementsRepository.findUserEntitlements(userHash.toByteArray()) }
         if (userEntitlements == null || userEntitlements.revoked || userEntitlements.endDate.isBefore(LocalDate.now())) {
+            if (userEntitlements == null) {
+                logger.info("${context.remoteIp} failed to create a new card")
```
seluianova, line 151: :thought_balloon: any chance that, without knowing the context, someone could remove or edit this?
Maybe a comment is needed explaining why this was added, since it's not clear from the code itself?

seluianova (approved):
as I understood, we can't test the ban itself until it's deployed in the env, right?
lgtm so far 👍

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
f1sh1918: yes your are right. its not a particular issue in this pr but we should in general add a comment to all ban related logs. But could be done in a different task but please create one (someone :))

f1sh1918 (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardMutationService.kt
steffenkleinle: Done for all found occurrences.

Issue #435, opened 2021-12-03
Enable use_build_context_synchronously linting rule


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
The rule [use_build_context_synchronously](https://dart-lang.github.io/linter/lints/use_build_context_synchronously.html) seems quite important and may prevent bugs.

**Describe the solution you'd like**
Enable the rule, set its severity to `error` and fix all occurrences.

maxammann:
Looks promising, right now this is experimental lint though. So I would wait.

f1sh1918:
still experimental

PR #1727, opened 2024-10-30
435: Use build context synchronously.

michael-markl:
### Short description

Enable the linting rule ["use_build_context_synchronously"](https://dart.dev/tools/linter-rules/use_build_context_synchronously) and fix errors in our code.

### Proposed changes

This PR sprinkles the codebase with some `mounted`  guards that check whether the widget is still mounted and the BuildContext can still be used, e.g. for displaying a dialog, etc.

### Notes

I noticed that the linting rule is not perfect, ie. it has both false positives and false negatives. However, as we have had some error reports due to incorrect usage of context, I think it's still a good idea to activate it.

### Testing

* Sorting by location in map & search
* Card Activation & verification
* Deeplink card activation

### Resolved issues

Fixes: #435, #1725, #1726, #1729, #1730, #1731, #1732, #1733, #1734, #1735

### DIFF
--- a/frontend/analysis_options.yaml
+++ b/frontend/analysis_options.yaml
@@ -142,6 +142,7 @@ analyzer:
     unnecessary_this: error
     unrelated_type_equality_checks: error
     unsafe_html: error
+    use_build_context_synchronously: error
     use_full_hex_values_for_flutter_colors: error
     use_function_type_syntax_for_parameters: error
     use_named_constants: error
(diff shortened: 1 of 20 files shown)


ztefanie (approved):
Good idea to add this rule. :+1: 
Code lgtm
tested on android emulator, everything seems to work fine

f1sh1918 (approved):
Tested on android and ios.
works as expected 👍

PR #1728, opened 2024-10-30
1696: Apply fonts

f1sh1918:
### Short description

Since we enabled a custom font for koblenz which is applies on the theme. we have to strictly style our texts and text buttons with our theme to apply the fonts everywhere.

### Proposed changes

- style text with textTheme where they don't inherit from standard widget
- add missing styles to flutter standard widgets in theme
- fix AppBar text color for ios, now clock and symbols are readable on every page (light theme)
- added style overrides to standard widgets styles where needed
- cleanup theme declarations in files

### Side effects

- some text sizes may differ slightly

### Testing
- Smoke test as much as possible

### Resolved issues

Fixes: #1696


michael-markl:
Have you considered adding a [DefaultTextStyle](https://api.flutter.dev/flutter/widgets/DefaultTextStyle-class.html) widget instead? This one would be added to the "root" widget and then all Text Widgets would use this text style as their default.

f1sh1918:
> DefaultTextStyle

Hm i could set the `fontFamily` and `fontSize` (sth similar as you style html body) with this widget but then i also would have to apply different `fontSizes` and `fontWeights` for the particular text.
So i just see one advantage if you forget to set the particular style for a text at least this defaultStyle including `fontFamily` will apply. 
I'm not sure about if you use f.e. ListTile that has title and subtile then different `fontSizes` and `fontWeights` apply.

I see your point and i may add this additionally but i also see the risk that styles will be forgotten. With this approach we have a standard way to style text elements

michael-markl:
> Hm i could set the fontFamily and fontSize (sth similar as you style html body) with this widget but then i also would have to apply different fontSizes and fontWeights for the particular text.
So i just see one advantage if you forget to set the particular style for a text at least this defaultStyle including fontFamily will apply.
> 
> I'm not sure about if you use f.e. flutter ListTile widget that has title and subtile then different fontSizes and fontWeights apply.

After diving a little bit into the theming / styling system built into Flutter, these were my findings:

* Most widgets (e.g. TextButton, ListView, AppBar, ...) apply default styles (by "rendering" a DefaultTextStyle widget) that is **always** based on the theme that we provide via `themes.dart`.
* If you only want to adjust a few styles for a Text, but keep the otherwise inherited defaults, you can just create a new TextStyle object. Properties that are omitted in the constructor will be inherited from the next default text style in the tree hierarchy above.
* Thus, when rendering any *standard* Material widget (such as buttons, ListTiles, AppBars, etc.) we do not need to adjust any styles (as the appropriate styles are already taken from the theme -- for the details read the docs / implementation of TextButton, ListTile, etc). If we want to adjust some text color for a specific widget, we can simply use `Text("my text", style: TextStyle(color: theme.colorScheme.whatever))` and do not need to check which text style is appropriate in the context as the other styles are inherited.

If we render more "custom" widgets  - such as the Tappable Cards in the Ausweisen Tab when no card is active - we still should supply theme with the appropriate text style. I believe, however, that even then a reasonable default will be set which is also taken from the theme.

> I see your point and i may add this additionally but i also see the risk that styles will be forgotten. With this approach we have a standard way to style text elements.

By now, I believe, we should not render this DefaultTextStyle widget at the root of our app. The flutter MaterialApp widget and friends will supply the correct defaults already. We only have to take care of the ThemeData object.

> In the end from a design system view you should have a typography that defines all text styles that should be used.

Yes, I agree. But this design system is already in place and specified by the Material design guidelines. The widgets use these design guidelines when defining their defaults.

f1sh1918:
Alright thx for this great explanation.
I think a big problem here is the flutter inspector and how to properly check if the theme/style has applied, since these inherited styles often will be shown as `null` at least I recognized this very often. So I can check which styles I can remove but then it's very difficult to check if f.e. fonts applied correctly if you don't have trained eyes though..

michael-markl:
>  I think a big problem here is the flutter inspector and how to properly check if the theme/style has applied, since these inherited styles often will be shown as null at least I recognized this very often. So I can check which styles I can remove but then it's very difficult to check if f.e. fonts applied correctly if you don't have trained eyes though..

I see... have you tried to use the tips discussed here? https://stackoverflow.com/a/74263526/3245533

f1sh1918:
> > I think a big problem here is the flutter inspector and how to properly check if the theme/style has applied, since these inherited styles often will be shown as null at least I recognized this very often. So I can check which styles I can remove but then it's very difficult to check if f.e. fonts applied correctly if you don't have trained eyes though..
> 
> I see... have you tried to use the tips discussed here? https://stackoverflow.com/a/74263526/3245533

Yes i always use this hovering mode. I really find this very difficult and some inherited styles are not should.
If i f.e. remove the text style from an `ListTile` in about it looks like the correct font family is applied.
If i check the inspector its null and even the dependencies don't give any info about what style is applied from the theme.

image

image

For tiles with proper styles applied everything looks fine
image

I see the point in removing unneeded styles but its super difficult to keep all the docs you mentioned in mind and to double check if everything applied correctly. So i'm not sure which way to go tbh.
Maybe this inspector will be improved in the future.
For now i think my solution is consistent and can be verified.

michael-markl:
> I see the point in removing unneeded styles but its super difficult to keep all the docs you mentioned in mind and to double check if everything applied correctly.

I think my argument would be the exact opposite: It's much harder to always remember the correct style (for e.g. ListTile title, ListTile subtitle, etc.). Therefore, it's much easier to be able to rely on the defaults that are set by ListTile. The fact that they are always based on the theme, makes it easy to know that it always uses the correct font family.

But I agree that the experience of the inspector is quite unpleasant when it comes to inspecting text styles. You have to dig through this dependencies set (expand every "bucket") and eventually you can find the DefaultTextStyle dependency which then tells you the text style it applies.

f1sh1918:
> > I see the point in removing unneeded styles but its super difficult to keep all the docs you mentioned in mind and to double check if everything applied correctly.
> 
> I think my argument would be the exact opposite: It's much harder to always remember the correct style (for e.g. ListTile title, ListTile subtitle, etc.). Therefore, it's much easier to be able to rely on the defaults that are set by ListTile. The fact that they are always based on the theme, makes it easy to know that it always uses the correct font family.
> 
> But I agree that the experience of the inspector is quite unpleasant when it comes to inspecting text styles. You have to dig through this dependencies set (expand every "bucket") and eventually you can find the DefaultTextStyle dependency which then tells you the text style it applies. It seems to be a known issue but it has been open for quite some time, unfortunately: [flutter/devtools#3084](https://github.com/flutter/devtools/issues/3084) [flutter/devtools#2272](https://github.com/flutter/devtools/issues/2272)

Okay i think then a good solution would be to define the typography classes like

```
headline 1
headline xx
title x
body x
```
 And apply them on textTheme and on f.e. subtitle and title of the ListTile f.e. and the other themes we adjust. Then it should automatically rely on that

f1sh1918:
@michael-markl i think i moved everything that should be defined in the theme to the `themes.dart` and removed unneeded styles from the other widgets

michael-markl:
One more remark: Everything inside a Material widget (such as everything inside a Scaffold or a Dialog) uses `bodyMedium` as DefaultTextStyle (unless it's overwritten by another widget further down the tree).
This means, that unless you're doing something very sketchy (which we don't), everything is themed "correctly" according to the theme. In particular, it's often fine to omit the text style even in "custom" widgets, if we intend to use this default.

ztefanie (approved):
deleted. sorry got lost in my browser tabs.

ztefanie (approved):
Code looks good. couldn't test as i need to fix protobuf before.

seluianova (commented):
Text of the info items is currently white
image

frontend/lib/widgets/app_bars.dart
```diff
@@ -101,10 +104,10 @@ class SearchSliverAppBarState extends State<SearchSliverAppBar> {
         focusNode: focusNode,
         decoration: InputDecoration.collapsed(
           hintText: t.search.searchHint,
-          hintStyle: TextStyle(color: foregroundColor?.withOpacity(0.8)),
+          hintStyle: theme.textTheme.bodyLarge,
```
seluianova, line 107: I think the opacity was appropriate here
image

frontend/lib/themes.dart
```diff
@@ -3,25 +3,47 @@ import 'package:ehrenamtskarte/util/color_utils.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
+/// Typography classes
```
seluianova, line 6: 👍

frontend/lib/store_widgets/detail/contact_info_row.dart
```diff
@@ -44,6 +45,7 @@ class ContactInfoRow extends StatelessWidget {
         Expanded(
           child: Text(
             _description,
+            style: theme.textTheme.bodyMedium,
```
seluianova, line 48: not sure if that's important, but it was grey before, while now it's black

frontend/lib/widgets/app_bars.dart
seluianova: yes, I agree 🙈
(I haven't checked all the views yet, maybe something else will arise)

frontend/lib/store_widgets/detail/contact_info_row.dart
f1sh1918: hm i think thats okay

seluianova: I personally think it looks more appealing when colors are different, but maybe it's too subjective

f1sh1918:
@seluianova thx for proper testing. I addressed the issues.
I smoke tested on ios light and dark theme.
Could you please check again on android :)

frontend/lib/widgets/app_bars.dart
seluianova: :upside_down_face: I don't want to annoy with nit-picky comments, so please leave it as is if you want to, but I can't resist to at least mentioning that 😄
but I think it was better with transparency. because currently it's like a label, and not like a background hint.
image

f1sh1918: I decided not use opacity due to bad readibility but i can adjust that

seluianova: up to you.
I think it's standard for such text to be grayed out.
image

frontend/lib/identification/verification_workflow/dialogs/negative_verification_result_dialog.dart
```diff
@@ -15,7 +15,7 @@ class NegativeVerificationResultDialog extends StatelessWidget {
       title: t.identification.notVerified,
       icon: Icons.error,
       iconColor: Colors.red,
-      child: Text(reason),
+      child: Text(reason, style: Theme.of(context).textTheme.bodyLarge),
```
seluianova, line 18: The content is bigger than the title. Can we apply some style to the title as well?
image

seluianova (approved):
So, I tried to test as much as possible and noted some minor findings. Maybe I missed something because it was a lot to test...
Anyway if something else comes up I believe it can be fixed later.
Thanks for tackling this huge task 👍

:thought_balloon: I think it would be nice to have a default text style (bodyMedium, for example), that doesn't need to be set explicitly to each text

UPD. ok, I see the discussion in the beginning about that

frontend/lib/search/filter_bar_button.dart
```diff
@@ -87,7 +87,7 @@ class _FilterBarButtonState extends State<FilterBarButton> with SingleTickerProv
       builder: (context, child) {
         final color = Color.lerp(theme.colorScheme.background, selectedColor, colorTween.value);
         return ConstrainedBox(
-          constraints: BoxConstraints.tightFor(width: width, height: 70),
+          constraints: BoxConstraints.tightFor(width: width, height: 74),
```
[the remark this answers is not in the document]
f1sh1918: Indeed. the font exceeds line height. Maybe we could also decrease lineHeight

Issue #1283, opened 2024-02-08
Show distance to user location for stores


f1sh1918:
As a user i would like to know the distance from my position to the acceptance store. This is an indicator how relevant a search result is for me

## Acceptance Criteria:
 - On the search page is a distance displayed for every list item if the location is enabled f.e. "1,0km entfernt"

## Technical Tasks:
 - add a distance label for every list item if user location is enabled

michael-markl:
This is already implemented, but I just noticed, that there is a bug in the backend right now that prohibits that the distance is shown...

PR #1736, opened 2024-11-02
1283: Fix physical store by store loader

michael-markl:
### Short description

The `searchAcceptingStoresInProject` query never returns physical stores.
Therefore, the functionality to show the distance (or location) of an accepting store, that is already implemented in Flutter in the search page, never worked (or rather stopped working at some point).

### Proposed changes

Fix the data loader responsible for querying the physical store given a store.

### Side effects

The functionality will kick in once the backend is deployed without needing to release the app again.

### Testing

Check out the search page.

### Resolved issues

Fixes: #1283

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/webservice/dataloader/PhysicalStoreByStoreIdLoader.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/webservice/dataloader/PhysicalStoreByStoreIdLoader.kt
@@ -11,7 +11,7 @@ import org.jetbrains.exposed.sql.transactions.transaction
 val physicalStoreByStoreIdLoader = newNamedDataLoader<Int, _>("PHYSICAL_STORE_BY_STORE_ID_LOADER") { ids ->
     transaction {
         PhysicalStoreEntity.find { PhysicalStores.storeId inList ids }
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/webservice/dataloader/PhysicalStoreByStoreIdLoader.kt
```diff
@@ -11,7 +11,7 @@ import org.jetbrains.exposed.sql.transactions.transaction
 val physicalStoreByStoreIdLoader = newNamedDataLoader<Int, _>("PHYSICAL_STORE_BY_STORE_ID_LOADER") { ids ->
     transaction {
         PhysicalStoreEntity.find { PhysicalStores.storeId inList ids }
-            .sortByKeys({ it.storeId }, ids)
+            .sortByKeys({ it.storeId.value }, ids)
```
michael-markl, line 14: A bit annoying that Kotlin did not catch this as a type error... Probably the generic was resolved to some common super type.

f1sh1918 (approved):
first time i see this great feature :D

steffenkleinle (approved):
Tested on android, works as expected. Nice!

Issue #1692, opened 2024-10-17
Add ‘This card has been revoked’ error message when activating a card in the app


seluianova:
**Is your feature request related to a problem? Please describe.**
Currently when a user tries to activate a card, that has been revoked, he gets the message `Der Inhalt des eingescannten Codes kann nicht verstanden werden. Vermutlich handelt es sich um einen QR-Code, der nicht für diese App generiert wurde.`
It might be nice to add a separate message for this case, like 'This card was revoked'

**Describe the solution you'd like**
Steps to reproduce:
1. Create a card as a region admin or via the self service portal (for the Koblenz project)
2. Revoke the created card ([revoking-card.md](https://github.com/digitalfabrik/entitlementcard/blob/main/docs/revoking-card.md))
3. Open the app and try to activate the revoked card

Expected error message should include the information that the card has been revoked.

**Describe alternatives you've considered**
Leave as is

**Additional context**
This is the current error message:
![image](https://github.com/user-attachments/assets/14695543-7c40-48fd-bbba-f23ea2abdd67)

PR #1737, opened 2024-11-04
1692: Add this card has been revoked error message

seluianova:
### Short description
Add a separate error message when the user tries to activate a revoked card in the app

### Proposed changes
- add additional activation state = revoked & corresponding error message
- I also renamed the QrParsingErrorDialog to ActivationErrorDialog and updated the title of the dialog to make it more generic, since it is not only used when scanning code, but also in the deep link activation

### Side effects
no

### Testing
1. create any card (for any project)
2. update revoked = true in the database
3. try to activate the revoked card via qr code scanning & via deep link activation
--> **"This card could not be activated as it has been revoked." error message should be displayed**

### Resolved issues
Fixes: #1692


f1sh1918 (approved):
Works as expected! Good job 👍

steffenkleinle (approved):

Issue #1653, opened 2024-09-30
Finalize koblenz app config


ztefanie:
- Adjust the koblenz map_zoom.png and the search_with_location.png when first stores are imported and the app shows data: use this for initial data: https://nextcloud.tuerantuer.org/index.php/apps/files/files/6518906?dir=/Projects/Digitale%20Berechtigungskarte/Stadt%20Koblenz/Akzeptanzpartner&openfile=true

- add disclaimer text: https://nextcloud.tuerantuer.org/index.php/apps/files/files/6520586?dir=/Projects/Digitale%20Berechtigungskarte/Stadt%20Koblenz/Releasedokumente/Herausgebertext&openfile=true
- ~~replace app logos with logos of correct size once we get svg from koblenz~~
- ~~radjust theme color for koblenz~~r
- ~~rreplace icon.png from assets with icon without white background~~r

PR #1740, opened 2024-11-05
1653: Add intro slides with stores and disclaimer

ztefanie:
### Short description
Added screenshots with initials stores to intro slides
Added disclaimer text

Fixes: #1653


f1sh1918:
@ztefanie
did you check if the correct font is used in the screenshots?
You can also create screenshots from my open pr
Even not sure if anyone will notice that

ztefanie:
> @ztefanie did you check if the correct font is used in the screenshots? You can also create screenshots from my open pr Even not sure if anyone will notice that

i think it is good enough for now, we can still update this, especially if more stores are published

f1sh1918 (approved):

f1sh1918:
@ztefanie 
I think in the finalization ticket is the `applicationUrl` missing
To avoid creating a separate task for adjusting a string could you please provide
` applicationUrl: "https://koblenz.sozialpass.app/erstellen"` in the buildConfig

steffenkleinle (approved):
LGTM

Issue #1623, opened 2024-09-09
Add i18n for applications in administration


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a dev i want to use a i18n library instead of hardcoding labels. This feature is also needed to resolve application data labels

**Describe the solution you'd like**
- implement i18n library
- create a translation file for keys and values for german
- replace hardcoded labels by i18n keys for applications

Note: For now we need the i18n for the verein360 interface, since they don't deliver all the application data labels. So we have to resolve them by the keys to be able to show incoming applications to the service users with the correct labels. In general we want to replace all labels by i18n keys but it doesn't have to be done in one ticket

PR #1744, opened 2024-11-05
1623: Add i18n for administration

ztefanie:
### Short description
Move labels from objects in the database to translation.json

### Proposed changes
- [x] Added i18n to administration
- [x] Use labels from translation file instead of database/object-json
- [x] remove adding labels before saving new application to the database

### Side effects
- May break everything especially things related to applications

### Testing
- Open application for `/beantragen` and check if all labels for all cases are displayed properly
- create applications for different type, e.g. blue entitlement card for work at organization, julica, ... and view if all labels are correctly displayed in the digitale druckerei "Eingehende Anträge" view.
- check if older applications created without these changes here are still displayed properly
- check if sent emails still look as expected

### Todo for follow up ticket
- replace all german strings with labels from translation file

### Resolved issues
Fixes: #1623


f1sh1918:
For the applications i would store the i18n labels in the json structure and resolve them in the frontend instead of saving the particular values (as it is now). i don't see the whole database redesign in this verein360 project to be honest even it will be a bit ugly to reconstruct the application data json after receiving the data from the verein360 endpoint. I think we then need proper tests for this reconstructing function.
I think we may extract the personal data (that will be more or less the same for all projects) at a later point and only store application depending data in the json format if we also want to maintain and store user data for long term.
Its even not sure if frankfurt will come and at the moment its not clear if we then need additional functionality which can not be provided with the current (json) solution.

The initial i18n implementation looks fine to me

steffenkleinle (commented):
Is there any reason you did not stick to the format/structure we use in integreat, i.e. using one file to store translations for all languages? That way, we could reuse existing tools and scripts and therefore our whole translation process.

ztefanie:
Thanks for your feedback.

@steffenkleinle 
> Is there any reason you did not stick to the format/structure we use in integreat, i.e. using one file to store translations for all languages? That way, we could reuse existing tools and scripts and therefore our whole translation process.
> 

Valid question. I thought about this, too and decided to stick to the solutions described in the documentation rather then to the integreat solution. The reasons for this is the integreat solution is a lot complexer, e.g. needing functions to merge the language documents into one translation file. This makes sense for integreat where we have this whole import / export translations logic, but not really for entitlement card furthermore the out of the box functionality of i18next is absolutely sufficient for the use cases of the entitlementcard and needs no enhancements like we use in integreat.

> One other question: Do you know why it is not necessary anymore to wrap the app in a `I18nProvider` or similar? Is that just an unnecessary thing we do in Integreat or it became obsolete with some update in the past?
> 

The I18nextProvider does take an i18next instance and passes that down using contexts. As we do not consume the translations for context, but by using the useTranslation() hook, it is not needed. I am not quite sure, if we still have occurences in Integreat where we use translations from context, if not, it should be save to delete this Provider.

administration/src/mui-modules/application/forms/WorkAtOrganizationForm.tsx
```diff
@@ -63,56 +64,59 @@ const WorkAtOrganizationForm: Form<State, ValidatedInput, AdditionalProps> = {
...
+        />
+        <SubForms.payment.Component
+          state={state.payment}
+          setState={useUpdateStateCallback(setState, 'payment')}
+          label={t('payment')}
+          options={paymentOptions}
+        />
+        <h4>Tätigkeitsnachweis</h4>
+        <p>
+          Falls vorhanden, hängen Sie hier bitte einen eingescannten oder abfotografierten Tätigkeitsnachweis an.{' '}
```
steffenkleinle, line 110: :wrench: Missed translation

administration/src/util/translations/de.json
```diff
@@ -0,0 +1,72 @@
...
+      "name": "Name der Organisation bzw. des Vereins",
+      "category": "Einsatzgebiet"
+    },
+    "organizationContact": {
+      "title": "Kontaktperson der Organisation",
+      "name": "Vor- und Nachname der Kontaktperson in der Organisation",
+      "telephone": "Telefon",
+      "email": "E-mail",
+      "hasGivenPermission": "Die Kontaktperson hat der Weitergabe ihrer Daten zum Zwecke der Antragsverarbeitung zugestimmt und darf zur Überprüfung kontaktiert werden"
+    },
```
steffenkleinle, line 39: :wrench: I really dislike the `getTranslationKey()` in `JsonFieldArray` as I think its quite hacky and error prone in the future. Getting rid of the namespace would make it redundant.
```suggestion
    "organization": "Angaben zur Organisation",
    "organizationName": "Name der Organisation bzw. des Vereins",
    "organizationCategory": "Einsatzgebiet"
    "organizationContact": "Kontaktperson der Organisation",
    "organizationContactName": "Vor- und Nachname der Kontaktperson in der Organisation",
    ...
    },
```

administration/src/util/translations/de.json
```diff
@@ -0,0 +1,72 @@
...
+    "certificate": "Tätigkeitsnachweis",
+    "hasAcceptedPrivacyPolicy": "Ich erkläre mich damit einverstanden, dass meine Daten zum Zwecke der Antragsverarbeitung gespeichert werden und akzeptiere die Datenschutzerklärung",
+    "hasAcceptedEmailUsage": "Ich stimme zu, dass ich von der lokalen Ehrenamtskoordination über Verlosungen und regionale Angebote informiert werden darf",
+    "givenInformationIsCorrectAndComplete": "Ich versichere, dass alle angegebenen Informationen korrekt und vollständig sind",
+    "title": "Bayerische Ehrenamtskarte beantragen",
+    "sendSuccessfully": "Erfolgreich gesendet",
+    "sendError": "Fehler beim Senden",
+    "blueCardEntitlementType": {
+      "WorkAtOrganizations": "Ich engagiere mich ehrenamtlich seit mindestens zwei Jahren freiwillig mindestens fünf Stunden pro Woche oder bei Projektarbeiten mindestens 250 Stunden jährlich.",
+      "Juleica": "Ich bin Inhaber:in einer JuLeiCa (Jugendleiter:in-Card).",
```
steffenkleinle, line 53: :upside_down_face: I don't really like these nested namespaces and think that at least `EntitlementType` (and just `Entitlement`
 below) would not be needed in the keys. I guess this is personal preference though so feel free to ignore this.
```suggestion
    "blueCardWorkAtOrganizations": "Ich engagiere mich ehrenamtlich seit mindestens zwei Jahren freiwillig mindestens fünf Stunden pro Woche oder bei Projektarbeiten mindestens 250 Stunden jährlich.",
    "blueCardJuleica": "Ich bin Inhaber:in einer JuLeiCa (Jugendleiter:in-Card).",
```

steffenkleinle: Another note: I think in the code we mostly use `standard` and `gold` as card types, perhaps we could follow this here as well?

steffenkleinle (commented):
Thanks a lot for the work on this tedious issue :pray: Looks pretty good already :)

administration/src/mui-modules/application/forms/WorkAtOrganizationForm.tsx
```diff
@@ -63,56 +64,59 @@ const WorkAtOrganizationForm: Form<State, ValidatedInput, AdditionalProps> = {
...
-            setState={useUpdateStateCallback(setState, 'amountOfWork')}
-            options={amountOfWorkOptions}
-            minWidth={250}
-          />
+  Component: ({ state, setState, onDelete, applicantName }: FormComponentProps<State, AdditionalProps>) => {
+    const { t } = useTranslation('application')
+    return (
+      <>
+        <ActivityDivider onDelete={onDelete} />
+        <h4>Angaben zu Ihrer ehrenamtlichen Tätigkeit</h4>
```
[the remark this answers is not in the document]
ztefanie: I planned to create a new PR for replacing all the labels to make this PR easier to review. Added this to the PR description.

administration/src/util/translations/de.json
ztefanie: See explanation above why this is not possible.

steffenkleinle (approved):
Smoke tested on firefox, looks good :)

administration/src/util/translations/de.json
ztefanie: if we decide we want to do this, yes. I will create one, then we can discuss this in the grooming.

steffenkleinle: Thanks!

ztefanie: https://github.com/digitalfabrik/entitlementcard/issues/1770

f1sh1918 (commented):
Looks good. One finding telephone label is missing in translations file
image

ztefanie:
> Looks good. One finding telephone label is missing in translations file 

Thanks for testing. Please approve if only small things needs fixing. Also maybe you do not want to publish your adress on github?

f1sh1918 (approved):

Issue #1430, opened 2024-04-29
Extend card


f1sh1918:
As a app user i want to be able to extend my card

## Acceptance Criteria:
 - If a card is extendable the user will get a notification
 - the card is extendable if its expiry date is earlier than the endDate of the associated user entitlement
 - on click "extend" the user gets forwarded to the self service portal and the user data form will be prefilled

## Technical Tasks:
- [x] add a notification area when a card is extendable https://github.com/digitalfabrik/entitlementcard/pull/1772
 - [x] provide a link with queryParams to the self service portal, fill the form with the passed data https://github.com/digitalfabrik/entitlementcard/pull/1749
 - [ ] add queryParams to applicationUrl in native app

seluianova:
Couple questions:

> if the card can be extended

how exactly do we define it? expirationDate >= today, or?

> If a card is extendable the user will get a notification

should it be a push-notification?
or should the user only see it when the app is opened?

should it be shown immediately when the application is opened - so, over the map, or on the Ausweisen tab?

> the user gets the new (updated) card

do we want to create a new card or update an existing one?
if we create a new card - what do we do with the static code?

f1sh1918:
i talked to daniel and card extension via self service portal via link with query params is okay. We can just implement this

seluianova:
@hauf-toni could you propose how best to display a notification in the app that a card extension is available?
it is better to display the notification immediately after the application is launched (so - over the map), or when the Ausweisen tab is opened? how should the notification look like?

hauf-toni:
You can find the design [📍 here](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=792-3416&t=kUofvHMqh07BbICq-1).

seluianova:
@hauf-toni thanks, I have a couple more questions:
what would you recommend for the case, when the card has already expired?
currently it looks like this:
![image](https://github.com/user-attachments/assets/b1b94c8d-d5ea-4c61-b393-ab46863dde87)
should we leave it as it is and just open the pre-filled form when the user clicks on "KoblenzPass beatragen oder verlängern"?
the only concern in this case I have:
the item says “beatragen ODER verlängern”, and if the user actually wants to request a pass with other data, he will still get a pre-filled form with the previous card's details.

ztefanie:
> 2. regarding the text "your card is about to expire".
[...]

PR #1749, opened 2024-11-06
1430: Add query params to card self service form

seluianova:
### Short description
When the user taps the “Extend Card” button in the app, the self-service portal should open with the pre-filled data.
In this PR, I added support for query parameters in the /erstellen route.

I feel a bit lost in React, so perhaps my implementation is not optimal, pls take a look.

### Proposed changes
- query params in the card self service form: name, geburtsdatum, ref

### Side effects
no?

### Testing
Open url with query params
~~`http://localhost:3000/erstellen?name=Karla%20Koblenz&ref=123K&geburtsdatum=2003-06-10`~~
`localhost:3000/erstellen?Name=Karla%20Koblenz&Referenznummer=123K&Geburtsdatum=10.06.2003`
The form should be pre-filled with the provided data

### Resolved issues
Part of: #1430


f1sh1918:
Some general ideas to your pr.
Instead of updating the card in a hook in the form component, i would suggest to add this functionality in the existing `useCardGeneratorSelfService` hook where the card will be initilized. That also avoids unexpected rerendering.

You can just add a fkt `handleQueryParams` and return the needed values there and then initlialize them here. Please also provide some Error handling if birthday can not be parsed and show a toast.

if you add Nullish Coalescingwhen getting the queryParam and return undefined you can even avoid null checking while initializing
```js
const fullName = cardQueryParams.get('name') ?? undefined
```

Example `useCardGeneratorSelfService.ts`
```js
  const [cardQueryParams] = useSearchParams()
  const {fullName, birthday, koblenzReferenceNumber} = handleQueryParams(cardQueryParams)
  const [selfServiceCard, setSelfServiceCard] = useState(
  initializeCard(projectConfig.card, undefined, { fullName, expirationDate: null, extensions: {birthday, koblenzReferenceNumber } })
  )
  )
```




After generating a card successfully the queryParams should be removed. You can get a set fkt from the `useSearchParams`hook and do sth like
```js
setSearchParams(undefined, { replace: true })
```

seluianova:
@f1sh1918 thanks for the feedback, I applied your suggestions, pls check if I understood you right.

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.ts
```diff
@@ -31,12 +33,46 @@ type UseCardGeneratorSelfServiceReturn = {
   downloadPdf: (code: CreateCardsResult, fileName: string) => Promise<void>
 }
 
+const parseDate = (dateString: string): PlainDate | undefined => {
+  try {
+    return PlainDate.from(dateString)
+  } catch {
+    return undefined
```
seluianova, line 40: haven't included this part yet 
> Please also provide some Error handling if birthday can not be parsed and show a toast to the user.
because if I add a toast here, it's displayed twice. have to think a while how to avoid that

seluianova: but actually I'm not sure if that's really needed. the users are not supposed to pass query params, and if it happens for some reason through the app - there will be a validation error for an empty field anyway

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -57,7 +57,7 @@ const BirthdayExtension: Extension<BirthdayExtensionState> = {
     },
   }),
   isValid: ({ birthday }: BirthdayExtensionState) => {
-    if (birthday === null) {
+    if (birthday == null) {
```
steffenkleinle, line 60: :wrench: Please revert this. I don't understand why the checks are still passing for this?

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.ts
```diff
@@ -31,12 +33,34 @@ type UseCardGeneratorSelfServiceReturn = {
...
-    initializeCard(projectConfig.card, undefined, { expirationDate: null })
+    initializeCard(projectConfig.card, undefined, {
+      fullName,
+      expirationDate: null,
+      extensions: { birthday, koblenzReferenceNumber },
+    })
   )
+  useEffect(() => {
+    setSearchParams(undefined, { replace: true })
+  }, [setSearchParams])
```
steffenkleinle, line 63: :upside_down_face: Could you please move this after the `useState` usages and add some newlines around it?

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.ts
```diff
@@ -31,12 +33,34 @@ type UseCardGeneratorSelfServiceReturn = {
...
+): {
+  fullName?: string
+  birthday?: PlainDate
+  koblenzReferenceNumber?: string
+} => {
+  const fullName = cardQueryParams.get('name') ?? undefined
+  const birthday = PlainDate.safeFrom(cardQueryParams.get('geburtsdatum')) ?? undefined
+  const koblenzReferenceNumber = cardQueryParams.get('ref') ?? undefined
+  return { fullName, birthday, koblenzReferenceNumber }
+}
```
steffenkleinle, line 47: :wrench: Please have a look at `AddCardsForm` and how card initialization is implemented there. I think the behavior there could be reused/generalized. I think it especially makes sense to reuse `initializeCardFromCSV` instead of implementing this and hardcoding the keys here again.

administration/src/cards/extensions/BirthdayExtension.tsx
seluianova: Why? I wanted to get false if birthday is null AND if birthday is undefined. How to do it better?

steffenkleinle: You can just do
``` ts
if (!birthday) {
...
}
```
Using only two equals `==` only performs loose type checking and therefore does implicit js type logic. Therefore, we only want strict equal `===`.

See e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness or https://stackoverflow.com/questions/16633947/in-javascript-is-it-better-to-use-two-equal-signs-or-three-when-comparing-varia

seluianova: Ok, thx 🙈

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.ts
f1sh1918: Yes you are probably right. I came to the same result during the birthday extension improvement task

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.ts
f1sh1918: yes this solution is even more flexible.
Nice find

seluianova: I assumed there is no difference, what to merge first.
I added to-do in the flutter PR.
But of course I will address this too

seluianova: Updated, but haven't tested card creation yet because of the issue https://github.com/digitalfabrik/entitlementcard/issues/1795

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
```diff
@@ -35,13 +37,21 @@ type UseCardGeneratorSelfServiceReturn = {
...
+    const values = headers.map(header => cardQueryParams.get(header))
+    return initializeCardFromCSV(projectConfig.card, values, headers, undefined, true)
+  })
   const [isLoading, setIsLoading] = useState<boolean>(false)
   const [selfServiceState, setSelfServiceState] = useState<CardSelfServiceStep>(CardSelfServiceStep.form)
   const [deepLink, setDeepLink] = useState<string>('')
   const [code, setCode] = useState<CreateCardsResult | null>(null)
+
+  useEffect(() => {
+    setSearchParams(undefined, { replace: true })
```
steffenkleinle, line 52: :wrench:
Could you perhaps move this inside the initializing method to avoid the extra use effect and possible side effects?

steffenkleinle (commented):
I think something is not quite right yet, when I open the URL as you suggested nothing happens except the query params being removed. None of the supplied properties is used to initialize the card.

seluianova:
> I think something is not quite right yet, when I open the URL as you suggested nothing happens except the query params being removed

I forgot to update the PR description. Param names should be the same as in card config

steffenkleinle (approved):
Tested on firefox, works with the correct link with correctly set query params. I updated the PR description.
Works as expected that way :)

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
```diff
@@ -35,13 +37,21 @@ type UseCardGeneratorSelfServiceReturn = {
...
   const appToaster = useAppToaster()
-  const [selfServiceCard, setSelfServiceCard] = useState(
-    initializeCard(projectConfig.card, undefined, { expirationDate: null })
-  )
+  const [cardQueryParams, setSearchParams] = useSearchParams()
+  const [selfServiceCard, setSelfServiceCard] = useState(() => {
+    const headers = getHeaders(projectConfig)
+    const values = headers.map(header => cardQueryParams.get(header))
+    return initializeCardFromCSV(projectConfig.card, values, headers, undefined, true)
+  })
```
steffenkleinle, line 45: :upside_down_face: I really like your implementation. If you feel motivated, feel free to add the same functionality to the `useCardGenerator` hook and remove the old implementation from `AddCardsFrom`

f1sh1918: i think we should merge this soon to create the next beta artifact and i would prefer to implement this in a separate task

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
f1sh1918: i think its better to move this to the `createKoblenzPass` task to only remove them if the card was created successfully, so the params will not be removed if there is an page refresh

f1sh1918 (approved):
Nicely done 👍 

One small comment should be resolved before merging

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
seluianova: @f1sh1918 pls check if that's what you mean

f1sh1918: i think you can just reuse the hook in `CardSelfServiceForm`
```js
const [_, setSearchParams] = useSearchParams()
```
and execute the function as you did. Then you don't have to pass down the function neither to `CardSelfServiceForm` nor adjust the `useCardSelfService` hook

seluianova: hopefully, it's good to go now, thx 🙈

Issue #1672, opened 2024-10-07
Improve the Geburtsdatum field in the card creation form


seluianova:
**Describe the bug**
As a koblenz pass user i want to type in my birthday as easy as possible on my mobile device

Acceptance Criteria:
- user can click in the date field and type in the date with the keyboard
- user can select the date from the datepicker if the icon was clicked
- user can clear the input via clear button or using backspace on desktop
- if the date is in the future the user gets "Error Message" below the field

Check here the MUI components (desktop version)
https://mui.com/x/react-date-pickers/date-picker/

f1sh1918:
@seluianova maybe we also think about a general improvement task for the birthday field.
Also placeholder is not used as it should because of styling issues
I think it would be better to also be able to type in the date, in mobile mode since picking the day, month and year might be annoying and some people reported not to know how to select the year. (Even this is just standard date picker html component)

seluianova:
> I think it would be better to also be able to type in the date

not sure if I understand this part, because I think I can currently type in the date 🤔

seluianova:
just an idea to consider:
we could also use 3 dropdowns to select birthday (day, month, year).
I would solve some of the issues I believe

seluianova:
just an idea to consider:
we could also use 3 dropdowns (day, month, year) to select birthday, with the ability to type in.
or just 3 text fields with the validation.
it would solve some of the issues I believe.
(_I just agree too that looking up the year on the calendar isn't as convenient. maybe we don't actually need a calendar there?_)

f1sh1918:
> It might make sense to use a date as default/placeholder that clearly indicates the date format, i.e. a day > 12, e.g. 31.01.1980 or something. Especially helpful if your device is using mm/dd/yyyy or dd/mm/yyyy.

this is now already applied `tt.mm.yyyy` for german localization

f1sh1918:
@hauf-toni we currently have the issue that the user on the mobile device needs a lot of effort to select the birthday due to scrolling the years, etc. Can you provide a proper proposal for that how to implement the "birthday" picker :)

hauf-toni:
@f1sh1918 I really like this one: https://mui.com/x/react-date-pickers/date-calendar/
What do you think? Would you like a design proposal as well?

f1sh1918:
@hauf-toni  this is just a calendar but we need also a form field.
This would be an option
https://mui.com/x/react-date-pickers/date-picker/

[...]

PR #1761, opened 2024-11-06
1670: Show form hints

f1sh1918:
### Short description

Show error messages/hints to the user for the self service form

### Proposed changes

- add error messages and validation for name input field (should this also apply to other projects? I think it makes sense)
- add validation to koblenz reference number
- adjust error messages
- add error message and untouched state to data privacy acceptance
- increase name length to 40

### not included
- error handling for Birthday extensions since not all requirements are clarified and the final implementation of the form field is not done yet (https://github.com/digitalfabrik/entitlementcard/issues/1672) 
- final text adjustments

### Side effects

- none

### Testing

1. Name field: Test special characters, name length and name includes 2 strings with at least 2chars. The particular error message should be displayed
2. Check that the reference number does not include special chars except of "." Check the min(4) and max length (15). The particular error message should be displayed
3. Check the data privacy accepting field: now it has a untouched state and if unchecked there should be an error message display
4. Check that the card creation button is not disabled but clicking on it with some invalid fields will lead to proper error message
5. Check the adjusted error message if you want to create a card for a revoked entitlement user

### Resolved issues

Fixes: #

### DIFF
--- a/administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
+++ b/administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
@@ -1,15 +1,18 @@
 import { Checkbox, FormGroup, InputGroup, Intent } from '@blueprintjs/core'
 import InfoOutlined from '@mui/icons-material/InfoOutlined'
-import { Alert, styled } from '@mui/material'
+import { styled } from '@mui/material'
 import React, { ReactElement, useContext, useState } from 'react'
 
-import { Card, isFullNameValid, isValid } from '../../cards/Card'
+import { Card, getFullNameValidationErrorMessage, isFullNameValid, isValid } from '../../cards/Card'
 import ClearInputButton from '../../cards/extensions/components/ClearInputButton'
 import useWindowDimensions from '../../hooks/useWindowDimensions'
 import BasicDialog from '../../mui-modules/application/BasicDialog'
 import { ProjectConfigContext } from '../../project-configs/ProjectConfigContext'
+import { useAppToaster } from '../AppToaster'
 import ExtensionForms from '../cards/ExtensionForms'
+import { DataPrivacyAcceptingStatus } from './CardSelfServiceView'
 import { ActionButton } from './components/ActionButton'
+import FormErrorMessage from './components/FormErrorMessage'
 import { IconTextButton } from './components/IconTextButton'
 import { UnderlineTextButton } from './components/UnderlineTextButton'
 
@@ -19,11 +22,6 @@ const StyledCheckbox = styled(Checkbox)`
   margin-left: 4px;
 `
 
-const StyledAlert = styled(Alert)`
-  margin-bottom: 24px;
-  white-space: pre-line;
-`
-
 const Container = styled('div')`
   margin-bottom: 24px;
 `
@@ -31,22 +29,11 @@ const Container = styled('div')`
 type CardSelfServiceFormProps = {
   card: Card
   updateCard: (card: Partial<Card>) => void
-  dataPrivacyAccepted: boolean
-  setDataPrivacyAccepted: (value: boolean) => void
+  dataPrivacyAccepted: DataPrivacyAcceptingStatus
+  setDataPrivacyAccepted: (status: DataPrivacyAcceptingStatus) => void
   generateCards: () => Promise<void>
 }
 
-const getTooltipMessage = (cardsValid: boolean, dataPrivacyAccepted: boolean): string => {
-  const tooltipMessages: string[] = []
-  if (!cardsValid) {
-    tooltipMessages.push('Mindestens eine Ihrer Angaben ist ungültig.')
-  }
-  if (!dataPrivacyAccepted) {
-    tooltipMessages.push('Bitte akzeptieren Sie die Datenschutzerklärung.')
-  }
-
-  return tooltipMessages.join('\n')
-}
 const CardSelfServiceForm = ({
   card,
   updateCard,
@@ -59,7 +46,24 @@ const CardSelfServiceForm = ({
   const [openDataPrivacy, setOpenDataPrivacy] = useState<boolean>(false)
   const [openReferenceInformation, setOpenReferenceInformation] = useState<boolean>(false)
   const cardValid = isValid(card, { expirationDateNullable: true })
-  const cardCreationDisabled = !cardValid || !dataPrivacyAccepted
+  const appToaster = useAppToaster()
+
+  const createKoblenzPass = async () => {
+    if (dataPrivacyAccepted === DataPrivacyAcceptingStatus.untouched) {
+      setDataPrivacyAccepted(DataPrivacyAcceptingStatus.denied)
+    }
+    if (!cardValid || dataPrivacyAccepted !== DataPrivacyAcceptingStatus.accepted) {
+      appToaster?.show({
+        message: (
+          <FormErrorMessage style={{ color: 'white' }} errorMessage='Mindestens eine Ihrer Angaben ist ungültig.' />
+        ),
+        timeout: 0,
+        intent: 'danger',
+      })
+      return
+    }
+    await generateCards()
+  }
 
   return (
     <>
@@ -80,23 +84,30 @@ const CardSelfServiceForm = ({
             value={card.fullName}
             onChange={event => updateCard({ fullName: event.target.value })}
           />
+          <FormErrorMessage errorMessage={getFullNameValidationErrorMessage(card.fullName)} />
         </FormGroup>
         <ExtensionForms card={card} updateCard={updateCard} />
         <IconTextButton onClick={() => setOpenReferenceInformation(true)}>
           <InfoOutlined />
           Informationen zur Referenznummer
         </IconTextButton>
-        <StyledCheckbox checked={dataPrivacyAccepted} onChange={() => setDataPrivacyAccepted(!dataPrivacyAccepted)}>
+        <StyledCheckbox
+          checked={dataPrivacyAccepted === DataPrivacyAcceptingStatus.accepted}
+          onChange={() =>
+            setDataPrivacyAccepted(
+              dataPrivacyAccepted === DataPrivacyAcceptingStatus.accepted
+                ? DataPrivacyAcceptingStatus.denied
+                : DataPrivacyAcceptingStatus.accepted
+            )
+          }>
           Ich akzeptiere die{' '}
           <UnderlineTextButton onClick={() => setOpenDataPrivacy(true)}>Datenschutzerklärung</UnderlineTextButton>.
         </StyledCheckbox>
+        {dataPrivacyAccepted === DataPrivacyAcceptingStatus.denied && (
+          <FormErrorMessage errorMessage='Bitte akzeptieren sie die Datenschutzerklärung' />
+        )}
       </Container>
-      {cardCreationDisabled && (
-        <StyledAlert variant='outlined' severity='warning'>
-          {getTooltipMessage(cardValid, dataPrivacyAccepted)}
-        </StyledAlert>
-      )}
-      <ActionButton onClick={generateCards} variant='contained' disabled={cardCreationDisabled} size='large'>
+      <ActionButton onClick={createKoblenzPass} variant='contained' size='large'>
         Pass erstellen
       </ActionButton>
       <BasicDialog
--- a/administration/src/bp-modules/self-service/components/FormErrorMessage.tsx
+++ b/administration/src/bp-modules/self-service/components/FormErrorMessage.tsx
@@ -1,0 +1,31 @@
+import InfoOutlined from '@mui/icons-material/InfoOutlined'
+import { styled } from '@mui/material'
+import React, { CSSProperties, ReactElement } from 'react'
+
+const Container = styled('div')`
+  margin: 6px 0;
+  color: #ba1a1a;
+  display: flex;
+  gap: 8px;
+  align-items: center;
+  font-size: 14px;
+`
+
+type FormErrorMessageProps = {
+  errorMessage: string | null
+  style?: CSSProperties
+}
+
+const FormErrorMessage = ({ errorMessage, style }: FormErrorMessageProps): ReactElement | null => {
+  if (!errorMessage) {
+    return null
+  }
+  return (
+    <Container style={style}>
+      <InfoOutlined />
+      {errorMessage}
+    </Container>
+  )
+}
+
+export default FormErrorMessage
--- a/administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
+++ b/administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
@@ -1,5 +1,5 @@
 import { ApolloError } from '@apollo/client'
-import { useCallback, useContext, useState } from 'react'
+import React, { useCallback, useContext, useState } from 'react'
 
 import { Card, generateCardInfo, initializeCard } from '../../../cards/Card'
 import { generatePdf } from '../../../cards/PdfFactory'
@@ -12,6 +12,7 @@ import { base64ToUint8Array, uint8ArrayToBase64 } from '../../../util/base64'
 import downloadDataUri from '../../../util/downloadDataUri'
 import getCustomDeepLinkFromQrCode from '../../../util/getCustomDeepLinkFromQrCode'
 import { useAppToaster } from '../../AppToaster'
+import FormErrorMessage from '../components/FormErrorMessage'
 
 export enum CardSelfServiceStep {
   form,
@@ -54,9 +55,9 @@ const useCardGeneratorSelfService = (): UseCardGeneratorSelfServiceReturn => {
       if (error instanceof ApolloError) {
         const { title } = getMessageFromApolloError(error)
         appToaster?.show({
-          message: title,
-          intent: 'danger',
+          message: <FormErrorMessage style={{ color: 'white' }} errorMessage={title} />,
           timeout: 0,
+          intent: 'danger',
         })
       } else {
         appToaster?.show({
--- a/administration/src/cards/Card.test.ts
+++ b/administration/src/cards/Card.test.ts
@@ -2,6 +2,7 @@ import { BavariaCardType } from '../generated/card_pb'
 import { Region } from '../generated/graphql'
 import PlainDate from '../util/PlainDate'
 import {
+  MAX_NAME_LENGTH,
   generateCardInfo,
   getValueByCSVHeader,
   initializeCard,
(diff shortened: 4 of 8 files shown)


administration/src/cards/extensions/KoblenzReferenceNumberExtension.tsx
```diff
@@ -11,6 +12,9 @@ type KoblenzReferenceNumberExtensionState = { [KOBLENZ_REFERENCE_NUMBER_EXTENSIO
 
 const KoblenzReferenceNumberMinLength = 4
 const KoblenzReferenceNumberMaxLength = 15
+const hasSpecialChars = (referenceNr: string): boolean => /[`!@#$%^&*()_+\-=\]{};':"\\|,<>?~]/.test(referenceNr)
```
steffenkleinle, line 15: :thinking: Is this a different regex compared to e.g. the name? If yes, what are the differences? Why?

administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
```diff
@@ -59,7 +46,24 @@ const CardSelfServiceForm = ({
...
+  const appToaster = useAppToaster()
+
+  const createKoblenzPass = async () => {
+    if (dataPrivacyAccepted === DataPrivacyAcceptingStatus.untouched) {
+      setDataPrivacyAccepted(DataPrivacyAcceptingStatus.denied)
+    }
+    if (!cardValid || dataPrivacyAccepted !== DataPrivacyAcceptingStatus.accepted) {
+      appToaster?.show({
+        message: (
+          <FormErrorMessage style={{ color: 'white' }} errorMessage='Mindestens eine Ihrer Angaben ist ungültig.' />
```
steffenkleinle, line 58: See above

administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
```diff
@@ -59,7 +46,24 @@ const CardSelfServiceForm = ({
...
+        ),
+        timeout: 0,
+        intent: 'danger',
+      })
+      return
+    }
+    await generateCards()
```
steffenkleinle, line 65: :upside_down_face: I find the following a lot more intuitive as no negated checks are needed. Personal preference though.
```suggestion
    if (cardValid && dataPrivacyAccepted === DataPrivacyAcceptingStatus.accepted) {
      await generateCards()
      return
    }
    if (dataPrivacyAccepted === DataPrivacyAcceptingStatus.untouched) {
      setDataPrivacyAccepted(DataPrivacyAcceptingStatus.denied)
    }
    appToaster?.show({
      message: (
        <FormErrorMessage style={{ color: 'white' }} errorMessage='Mindestens eine Ihrer Angaben ist ungültig.' />
      ),
      timeout: 0,
      intent: 'danger',
    })
```

administration/src/cards/Card.ts
```diff
@@ -65,11 +65,22 @@ export const getExtensions = ({ extensions }: Card): ExtensionWithState[] => {
...
+  /[`!@#$%^&*()_+\-=\]{};':"\\|,.<>?~0123456789]/.test(fullName)
+
+const hasValidNameLength = (fullName: string): boolean => {
   const encodedName = new TextEncoder().encode(fullName)
   return fullName.length > 0 && encodedName.length <= MAX_ENCODED_NAME_LENGTH && fullName.length <= MAX_NAME_LENGTH
 }
 
+const hasNameAndForename = (fullName: string): boolean => {
+  const names = fullName.split(' ')
+  return names.length > 1 && names.every(name => name.length > 1)
```
steffenkleinle, line 78: :wrench: This leads to failing checks if a user accidentally puts two spaces

steffenkleinle (commented):
Tested on firefox, nice work :)

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
```diff
@@ -54,9 +55,9 @@ const useCardGeneratorSelfService = (): UseCardGeneratorSelfServiceReturn => {
       if (error instanceof ApolloError) {
         const { title } = getMessageFromApolloError(error)
         appToaster?.show({
-          message: title,
-          intent: 'danger',
+          message: <FormErrorMessage style={{ color: 'white' }} errorMessage={title} />,
```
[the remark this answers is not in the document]
f1sh1918: yes because it doesn't work

administration/src/cards/extensions/KoblenzReferenceNumberExtension.tsx
f1sh1918: yes because "." is allowed here

administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
f1sh1918: does not work if you find a solution with styled components let me know

administration/src/cards/Card.ts
```diff
@@ -65,11 +65,22 @@ export const getExtensions = ({ extensions }: Card): ExtensionWithState[] => {
 export const hasInfiniteLifetime = (card: Card): boolean =>
   getExtensions(card).some(({ extension, state }) => extension.causesInfiniteLifetime(state))
 
-export const isFullNameValid = ({ fullName }: Card): boolean => {
+const containsNameSpecialCharacters = (fullName: string): boolean =>
+  /[`!@#$%^&*()_+\-=\]{};':"\\|,.<>?~0123456789]/.test(fullName)
```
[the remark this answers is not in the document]
f1sh1918: @hauf-toni do we really need this check in the frontend?

administration/src/cards/Card.ts
f1sh1918: i added a separate messages that multiple spaces between names are not allowed. i think just stripping them is difficult since the name on the card should be the same than entered in the form field

steffenkleinle: I see. How about solving this in a programmatic way when actually creating the card by just stripping multiple spaces then? I guess I'd prefer that over having a separate check with error messages here

steffenkleinle (approved):

administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
```diff
@@ -3,7 +3,13 @@ import InfoOutlined from '@mui/icons-material/InfoOutlined'
 import { styled } from '@mui/material'
 import React, { ReactElement, useContext, useState } from 'react'
 
-import { Card, getFullNameValidationErrorMessage, isFullNameValid, isValid } from '../../cards/Card'
+import {
+  Card,
+  getFullNameValidationErrorMessage,
+  isFullNameValid,
+  isValid,
+  removeMultipleSpaces,
```
steffenkleinle, line 11: :wrench: I think this should be moved to `util`/`helper`

administration/src/cards/Card.ts
```diff
@@ -76,18 +76,15 @@ const hasValidNameLength = (fullName: string): boolean => {
 }
 
 const multipleSpacePattern = /\s\s+/g
-const containsMultipleSpaces = (fullName: string): boolean => multipleSpacePattern.test(fullName)
+export const removeMultipleSpaces = (fullName: string): string => fullName.replace(multipleSpacePattern, ' ')
 
 const hasNameAndForename = (fullName: string): boolean => {
   const names = fullName.replace(multipleSpacePattern, ' ').split(' ')
```
steffenkleinle, line 82:
```suggestion
  const names = removeMulitpleSpaces(fullName).split(' ')
```

administration/src/cards/Card.test.ts
```diff
@@ -139,6 +140,28 @@ describe('Card', () => {
     })
   })
 
+  it.each(['$tefan Mayer', 'Karla K.', 'Karla 1234 '])(
```
ztefanie, line 143: When i enter a name with a space at the end i get an error message. I think this should just be valid but deleted internally

administration/src/cards/Card.ts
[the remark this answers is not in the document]
ztefanie: for example emojis will be allowed but break the app. I think it would make sense to restrict the name to unicode ranges that contain letters

administration/src/cards/extensions/KoblenzReferenceNumberExtension.tsx
ztefanie: Same as above, i would restrict this to latin-letters (or all letters) and numbers and dot unicode ranges

administration/src/util/helper.ts
```diff
@@ -10,3 +10,6 @@ export const updateArrayItem = <T>(array: T[], updatedItem: T, index: number): T
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) extends (x: infer I) => void ? I : never
+
+const multipleSpacePattern = /\s\s+/g
+export const removeMultipleSpaces = (value: string): string => value.replace(multipleSpacePattern, ' ')
```
ztefanie, line 15: Maybe this could be more general and also remove trailing or leading spaces

ztefanie (commented):
Looks already good. Some hints, especially the special char handling needs optimization, beside this it is fine.

Birthday field does not show a proper error message but only the box is outlined red, when e.g. entering 31.02.2003. 

I am sorry for trying to be the worst user, but entering emojis is allowed but will break the app:
image

administration/src/cards/Card.ts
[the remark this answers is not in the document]
f1sh1918: i just talked to tony and she said its good enough to restrict numbers ".?!,"

administration/src/util/helper.ts
f1sh1918: we would need a separate function then since if you trim leading spaces immediately after input no one can type in forename and name

administration/src/cards/Card.ts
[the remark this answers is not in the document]
ztefanie: Nothing that breaks the app can be allowed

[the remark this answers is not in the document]
f1sh1918: alright thx for the hint @ztefanie 
I added an additional check for disallowing emojis.
Maybe you can find a better regex for that.
I also thought about using https://xregexp.com/ to simply emoji restriction but not sure if we want to add an additional library.

Feel free to add a better proposal if you think its needed here

administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
```diff
@@ -59,7 +47,24 @@ const CardSelfServiceForm = ({
...
   const cardValid = isValid(card, { expirationDateNullable: true })
-  const cardCreationDisabled = !cardValid || !dataPrivacyAccepted
+  const appToaster = useAppToaster()
+
+  const createKoblenzPass = async () => {
+    if (cardValid && dataPrivacyAccepted === DataPrivacyAcceptingStatus.accepted) {
+      await generateCards()
+      return
+    }
+    if (dataPrivacyAccepted === DataPrivacyAcceptingStatus.untouched) {
```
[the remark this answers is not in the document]
f1sh1918: yes since its not implemented here

administration/src/cards/Card.test.ts
f1sh1918: since we can not trim it from the state i added this to the generateCardInfo and to the formField check.
Good find since mobile device auto completion would make this annoying

administration/src/util/helper.ts
f1sh1918: Since this function is used on state changes it shouldn't remove these single spaces since it would block the user input
The trimming for single spaces will be handled in the generateCardInfo to not impact on form typing

administration/src/bp-modules/self-service/components/FormErrorMessage.tsx
```diff
@@ -0,0 +1,31 @@
+import InfoOutlined from '@mui/icons-material/InfoOutlined'
+import { styled } from '@mui/material'
+import React, { CSSProperties, ReactElement } from 'react'
+
+const Container = styled('div')`
+  margin: 6px 0;
+  color: #ba1a1a;
```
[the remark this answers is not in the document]
f1sh1918: added a ticket to create a general colors file and define all colors there

administration/src/util/helper.ts
```diff
@@ -13,3 +13,10 @@ export type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) ex
...
...2\u2640]\uFE0F?))?|\uD83C(?:[\uDFFB-\uDFFF](?:\u200D(?:[\u2642\u2640]\uFE0F?))?)|\u200D(?:[\u2642\u2640]\uFE0F?)))?|[\u261D\u270C](?:(?:\uD83C[\uDFFB-\uDFFF]|\uFE0F))?|[\u270B\u270A](?:\uD83C[\uDFFB-\uDFFF])?|\u270D(?:(?:\uD83C[\uDFFB-\uDFFF]|\uFE0F))?|[\u2764\u2763\u26D1\u2618\u26F0\u26E9\u2668\u26F4\u2708\u23F1\u23F2\u2600\u2601\u26C8\u2602\u26F1\u2744\u2603\u2604\u26F8\u2660\u2665\u2666\u2663\u260E\u2328\u2709\u270F\u2712\u2702\u26CF\u2692\u2694\u2699\u2696\u26D3\u2697\u26B0\u26B1\u26A0\u2622\u2623\u2B06\u2197\u27A1\u2198\u2B07\u2199\u2B05\u2196\u2195\u2194\u21A9\u21AA\u2934\u2935\u269B\u267E\u2721\u2638\u262F\u271D\u2626\u262A\u262E\u25B6\u23ED\u23EF\u25C0\u23EE\u23F8-\u23FA\u23CF\u2640\u2642\u2695\u267B\u269C\u2611\u2714\u2716\u303D\u2733\u2734\u2747\u203C\u2049\u3030\u00A9\u00AE\u2122]\uFE0F?|[\u0023\u002A\u0030-\u0039](?:\uFE0F\u20E3|\u20E3)|[\u2139\u24C2\u3297\u3299\u25AA\u25AB\u25FB\u25FC]\uFE0F?|[\u2615\u26EA\u26F2\u26FA\u26FD\u2693\u26F5\u231B\u23F3\u231A\u23F0\u2B50\u26C5\u2614\u26A1\u26C4\u2728\u26BD\u26BE\u26F3\u267F\u26D4\u2648-\u2653\u26CE\u23E9-\u23EC\u2B55\u2705\u274C\u274E\u2795-\u2797\u27B0\u27BF\u2753-\u2755\u2757\u25FD\u25FE\u2B1B\u2B1C\u26AA\u26AB])).*/.test(
```
ztefanie, line 20: It is a lot shorted and easier to check if you use unicode ranges;
/(\u00a9|\u00ae|[\u2000-\u3300]|\ud83c[\ud000-\udfff]|\ud83d[\ud000-\udfff]|\ud83e[\ud000-\udfff])/g

Check e.g. this regex test cases someone build: https://regex101.com/r/0anB6Z/1

administration/src/util/helper.ts
```diff
@@ -13,3 +13,10 @@ export type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) ex
 
 const multipleSpacePattern = /\s\s+/g
 export const removeMultipleSpaces = (value: string): string => value.replace(multipleSpacePattern, ' ')
+export const containsSpecialCharacters = (value: string): boolean =>
+  /[`!@#$%^&*()_+\-=\]{};':"\\|,.<>?~0123456789]/.test(value)
+
+export const containsNoEmojis = (value: string): boolean =>
```
ztefanie, line 19: :wrench: I think it would be nice if the function is not negated. "containsEmojis" to avoid double negation when being used and being more consistent with "containsSpecialChars"

administration/src/util/helper.ts
ztefanie: But i think we should only allow supported chars and not try to catch all forbidden ones

ztefanie (commented):
Unfortunately this still does not catch unsupported chars, like arabic, cyrillic or greek letters. They will lead to unhandled errors

image


I would strongly recommend to only allow chars that can be handled by the app correctly, e.g. with this code:

```
  const nameValid  = "stefanie test"
  const nameInvalid = "فف فف"

  const supportedCodePoints = helveticaFont.getCharacterSet();
  const supportedSet = new Set(supportedCodePoints);

  const isStringSupportedByFont = (input: string): boolean => {
    for (const char of input) {
      const codePoint = char.codePointAt(0)!;
      if (!supportedSet.has(codePoint)) {
        return false;
      }
    }
    return true;
  }

  console.log(isStringSupportedByFont(nameValid))
  console.log(isStringSupportedByFont(nameInvalid))
```

f1sh1918:
> Unfortunately this still does not catch unsupported chars, like arabic, cyrillic or greek letters. They will lead to unhandled errors
> 
> image
> 
> I would strongly recommend to only allow chars that can be handled by the app correctly, e.g. with this code (needs to dynamically use the font that is active, in my example i just used helvetica to show what i mean)
> 
> ```
>   const nameValid  = "stefanie test"
>   const nameInvalid = "فف فف"
> 
>   const supportedCodePoints = helveticaFont.getCharacterSet();
>   const supportedSet = new Set(supportedCodePoints);
> 
>   const isStringSupportedByFont = (input: string): boolean => {
>     for (const char of input) {
>       const codePoint = char.codePointAt(0)!;
>       if (!supportedSet.has(codePoint)) {
>         return false;
>       }
>     }
>     return true;
>   }
> 
>   console.log(isStringSupportedByFont(nameValid))
>   console.log(isStringSupportedByFont(nameInvalid))
> ```

The big question is how i do this with font faces in a web app?
I mean i can get the font as a string from the input field using ref but that really doesn't help me to get a data type where i can get a character set
I absolutely see your issue but every person that ever created a card or used any other form for sth official in germany knows that latin letters have to be used. The person has official papers that has to be shown and it can not be expected that any acceptance store can read arabic letters. 
Even koblenz will not have any arabic letters in their database
So i really don't see that the effort is valid to implement this @ztefanie

f1sh1918:
I checked the implementation for that. To get the Charset we have to create a pdf doc (async) and then we would have pass that charset to all validation functions that means a lot of refactoring and this should be done in a separate issue that i created.
We also would have to adjust the csv import. We could provide the charset via hook. Lots of tests have to be adjusted
 @ztefanie

administration/src/cards/Card.ts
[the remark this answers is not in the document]
f1sh1918: ok i think xregexp is the lean way to go.

administration/src/cards/extensions/KoblenzReferenceNumberExtension.tsx
f1sh1918: i don't think we have to restrict that much here, because it has to fit to the hash and if it doesn't you can not create a pdf and it won't break

administration/src/util/helper.ts
f1sh1918: this is solved by latin charset restriction

administration/src/cards/extensions/KoblenzReferenceNumberExtension.tsx
f1sh1918: But you probably its general a good idea to add this check to all form fields. But then we also may to show a proper error message. I kept this issue open https://github.com/digitalfabrik/entitlementcard/issues/1776
There we should check all form fields and ensure that even card creation (nuernberg & bayern) can not break due to wrong charset

f1sh1918:
@ztefanie i added latin char restriction

ztefanie (approved):
looks good. tested an firefox

Issue #1422, opened 2024-04-29
Create PDF for Koblenz Pass


seluianova:
**Is your feature request related to a problem? Please describe.**
If a service user has to support the customer, it should be possible to download a pdf after card creation and provide it to the customer

**Describe the solution you'd like**
- fill pdf template for koblenz pass
- the pdf creation will be triggered here https://github.com/digitalfabrik/entitlementcard/issues/1419
- pdf template has to be provided by ui/ux
- use the koblenz custom font TeX Gyre Heros

**Additional context**
TBD

f1sh1918:
@hauf-toni 
you can base the pdf template on nuernberg
- replace Nürnberg with Koblenz
- download link will be "download.koblenz.sozialpass.app" instead
- get logos, card background and missing texts from daniel
- we don't need an address field and also not the D-Ticket area (talk to daniel about that)

dkehne:
@hauf-toni you can find the resources here: https://nextcloud.tuerantuer.org/index.php/f/6094223

hauf-toni:
[240811-Ehrenamtskarte-Koblenz.pdf](https://github.com/user-attachments/files/16578245/240811-Ehrenamtskarte-Koblenz.pdf)

@f1sh1918 please review the attached pdf

f1sh1918:
@hauf-toni there is the deeplink activation by clicking on the qr code missing. For Koblenz you can do both as we have in bavaria

hauf-toni:
@f1sh1918 yes, I'll be happy to add to it as soon as the content + design have been approved. do you have any feedback an that?

f1sh1918:
@hauf-toni for me the design looks good to me. The primary color for the project will be purple right?
I think @dkehne has to let the customer approve what will be printed on the pass f.e. "Amt für Existenzsicherung.." Probably this has to be adjusted

dkehne:
@hauf-toni just 2 things:
1. Can you remove the "Amt für Existenzsicherung und soziale Integration - Sozialamt" text
2. Is the gray background intentionally placed behind the logos?

hauf-toni:
@dkehne @f1sh1918 pdf now available [here](https://nextcloud.tuerantuer.org/index.php/apps/files/files?dir=/Projects/Digitale%20Berechtigungskarte/Stadt%20Koblenz/Releasedokumente), hyperlink and feedback are included

f1sh1918:
@hauf-toni 
[...]

PR #1764, opened 2024-11-08
1422: Create PDF for koblenz pass

ztefanie:
### Short description
Adjusted to koblenz PDF to the final design

### Side effects
None

### Testing
Create a card and check the layout of the activation PDF, check the deep-link area, check if long names are displayed properly

### Resolved issues
Fixes: #1422


f1sh1918:
I think Daniel mentioned that we should use the Koblenz font on the pdf. Will check the PR next week :)

ztefanie:
> I think Daniel mentioned that we should use the Koblenz font on the pdf. Will check the PR next week :)

I used the final pdf koblenz sent us today that we should use. If they would have wanted a different font on this pdf, they would have sent it to us.

administration/src/cards/PdfFactory.ts
```diff
@@ -29,18 +48,20 @@ const fillContentAreas = async (
...
-  const helveticaFont = await doc.embedFont(StandardFonts.Helvetica)
+  const font = await doc.embedFont(StandardFonts.Helvetica)
   pdfConfig.elements?.dynamicActivationQrCodes.forEach(configOptions =>
     pdfQrCodeElement(configOptions, { page: templatePage, qrCode: dynamicCode })
   )
 
-  const helveticaBoldFont = await doc.embedFont(StandardFonts.HelveticaBold)
+  const fontBold =
+    (pdfConfig.customBoldFont && (await loadCustomFont(pdfConfig.customBoldFont, doc))) ||
+    (await doc.embedFont(StandardFonts.HelveticaBold))
```
f1sh1918, line 58:
```suggestion
  const fontBold =
    pdfConfig.customBoldFont && (await loadCustomFont(pdfConfig.customBoldFont, doc)) ||
    await doc.embedFont(StandardFonts.HelveticaBold)
```
For me it also works with less parenthesis

f1sh1918 (commented):
Works as expected. Seems like there is leaner way to implement a custom font.
So good work 👍 

I'm not 100% sure of the error handling is good enough.
Its very unlikely that a local font can't be found but if i rename the font in the public folder, so that it can't be found i get an unhandled error

image

administration/src/cards/PdfFactory.ts
ztefanie: yes it will work with less parenthesis, because in js && is "stronger" than ||. But most humans do not know this on the first sight, so the parentheses increase readability as it is clear at the first look what belongs together

ztefanie:
> I'm not 100% sure of the error handling is good enough. Its very unlikely that a local font can't be found but if i rename the font in the public folder, so that it can't be found i get an unhandled error
>

improved the error handling

administration/src/cards/PdfFactory.ts
f1sh1918: alright i'm fine to keep it

f1sh1918:
> > I'm not 100% sure of the error handling is good enough. Its very unlikely that a local font can't be found but if i rename the font in the public folder, so that it can't be found i get an unhandled error
> 
> improved the error handling

I know get the using fallback message but still get uncaught runtime errors @ztefanie 
image

ztefanie:
> > > I'm not 100% sure of the error handling is good enough. Its very unlikely that a local font can't be found but if i rename the font in the public folder, so that it can't be found i get an unhandled error
> > 
> > 
> > improved the error handling


that is normal behaviour in dev environments. just close it, the pdf should be downloaded anyways

f1sh1918 (approved):
Tested on chrome

administration/src/cards/PdfFactory.ts
```diff
@@ -18,6 +19,22 @@ export class PdfError extends Error {
...
+      if (res.ok && res.headers.get('Content-Type')?.includes('font')) {
+        return res.arrayBuffer()
+      }
+      reportError(`Couldn't load custom font ${font}. Using fallback font.`)
+    })
+    .catch(error => {
+      reportError(error)
+    })
+  return fontBytes ? doc.embedFont(fontBytes) : doc.embedFont(fallbackFont)
+}
```
steffenkleinle, line 36: :wrench: Please don't mix async/await and then/catch syntax. I'd prefer if you'd just use async/await.

administration/src/cards/PdfFactory.ts
```diff
@@ -29,17 +46,20 @@ const fillContentAreas = async (
...
 ): Promise<void> => {
-  const helveticaFont = await doc.embedFont(StandardFonts.Helvetica)
+  const font = await doc.embedFont(StandardFonts.Helvetica)
   pdfConfig.elements?.dynamicActivationQrCodes.forEach(configOptions =>
     pdfQrCodeElement(configOptions, { page: templatePage, qrCode: dynamicCode })
   )
 
+  const fontBold = pdfConfig.customBoldFont
+    ? await loadCustomFontWithFallback(pdfConfig.customBoldFont, doc, StandardFonts.HelveticaBold)
+    : await doc.embedFont(StandardFonts.HelveticaBold)
```
steffenkleinle, line 56: :question: Do I understand this correctly that the normal font is the same for koblenz and only for the bold font they want some custom font?

steffenkleinle (approved):
Looks nice! Tested on firefox

administration/src/cards/PdfFactory.ts
ztefanie: Yes, we do only use bold fonts in koblenz pdf

administration/src/cards/PdfFactory.ts
ztefanie: good point. Adjusted it

Issue #1700, opened 2024-10-22
Refactor freinet CSV import


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
At the moment, the freinet CSV input is really hard to read and hacky, since the functions in it have a lot of side effects.
Arrays and objects are mutated and later it is just assumed that they are correctly modified before.

**Describe the solution you'd like**
Use functional programming and always explicitly return the updated value without implicitly assuming the necessary modifications happened.

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #1765, opened 2024-11-11
1700: Refactor and clean up csv import

steffenkleinle:
### Short description

Refactor and clean up csv import to avoid implicit side effects.

### Proposed changes

-
-

### Side effects

- None.

### Testing

Make sure that both normal and freinet CSV import still works.

### Resolved issues

Fixes: #1700


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
```
steffenkleinle, line 1: This appeared after setting the eslint configuration in IntelliJ to automatic, I guess we want this project wide then?

administration/src/project-configs/helper/index.ts
```diff
@@ -0,0 +1,7 @@
+import { ProjectConfig } from '../getProjectConfig'
+
+export const getCsvHeaders = (projectConfig: ProjectConfig): string[] => [
```
steffenkleinle, line 3: Do you think this is good naming? Or should we change this to `columnNames` or similar?

f1sh1918: hm for now i think its okay because it serves this particular value. But if we combine name, expiry date and extension into one array to make it more flexible and to enable better reordering of form fields we might not need to explicitly define here the columns names in the array but just loop through and then we could rename it to `getColumnNames`

f1sh1918 (approved):
Looks good!
Nice refactoring 👍 

Tested csv import for nuernberg and bavaria

ztefanie (commented):
Tested on Firefox:

Previously a data like this 10.01.25 was valid. Now this is transformed as "10.01.0025" and is invalid. This should be fixed to the previous behavior.

steffenkleinle:
> Previously a data like this 10.01.25 was valid. Now this is transformed as "10.01.0025" and is invalid. This should be fixed to the previous behavior.

I just tried this on main and the behavior seems to be the same for csv import, seems like this was broken before. I'll create an issue for this.

Issue #1766, opened 2024-11-11
Unify Datepicker components


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Currently we are using different datepicker components in the application process, birthday extensions and other date extensions.

**Describe the solution you'd like**
- We should try to unify that and use one component for that.
- Maybe we have to use different stylings

andrew8er:
Just for clarification: 
- I found two date related components used in the frontend, `<CustomDatePicker/>` and `<TextField/>` in `DateForm.tsx`.
- `<CustomDatePicker/>` should be used throughout the code base.

Is this correct?

f1sh1918:
> Just for clarification: 
> - I found two date related components used in the frontend, `<CustomDatePicker/>` and `<TextField/>` in `DateForm.tsx`.
> - `<CustomDatePicker/>` should be used throughout the code base.
> 
> Is this correct?

Exactly. I guess we need to support some styling differences for the application form of bavaria

PR #1767, opened 2024-11-11
1672: Improve birthday field

f1sh1918:
### Short description

As a koblenz pass applicant i want to provide my birthday easily on a mobile device

### Proposed changes

- added a custom mui datepicker in desktop variant that enables date input with (virtual) keyboard
- improved datepicker usage by choosing date -> month -> day on calendar icon click and dropdown
- add error hint
- adjusted the clearButton slightly to fit better to material icons

### Note
- it makes sense to reuse this component for other date inputs to unify behavior and especially for the application process for bavaria we should reuse this approach even the styling is a bit different #1766
- the style for the calendar differs to the design, since its a lot of effort to customize that and i think not worth the effort
- the under 16 years old check is not really clear. I don't think this makes sense because no one can create a card for their children if we disallow that.

### Side effects

- none

### Testing

1. Please start the application on a real mobile device to check if everything works fine
2. Create a card and activate it and check if birthday value is correct

### Resolved issues

Fixes: #1672


administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onClear,
+  isValid,
+  minDate,
+  maxDate,
+  disableFuture,
+}: CustomDatePickerProps): ReactElement => {
+  const { viewportSmall } = useWindowDimensions()
+  const formStyle = viewportSmall ? { fontSize: 16, padding: '9px 10px' } : { fontSize: 14, padding: '6px 10px' }
+  const textFieldBoxShadow =
+    '0 0 0 0 rgba(205, 66, 70, 0), 0 0 0 0 rgba(205, 66, 70, 0), inset 0 0 0 1px #cd4246, inset 0 0 0 1px rgba(17, 20, 24, 0.2), inset 0 1px 1px rgba(17, 20, 24, 0.3)'
```
f1sh1918, line 29: just copy the box shadow from the standard TextField components of MUI.
If someone know how to improve that, just let me know

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -13,35 +13,38 @@ const minBirthday = new PlainDate(1900, 1, 1)
...
-        }}
-        onChange={event => {
-          const date = PlainDate.safeFrom(event.target.value)
-          if (date !== null) {
-            setValue({ birthday: date })
-          }
-        }}
+      <CustomDatePicker
+        date={birthday ? birthday.toLocalDate() : null}
+        onChange={date => changeBirthday(date)}
```
steffenkleinle, line 41: :upside_down_face: 
```suggestion
        date={birthday?.toLocalDate() ?? null}
        onChange={changeBirthday}
```

administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <DesktopDatePicker
+      views={['year', 'month', 'day']}
+      value={date}
+      format='dd.MM.yyyy'
```
steffenkleinle, line 34: :question: How does this correspond with the `LocalizationProvider` you wrapped the app in?

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -13,35 +13,38 @@ const minBirthday = new PlainDate(1900, 1, 1)
...
-  const birthdayDate = value.birthday?.toString() ?? ''
+  const changeBirthday = (date: Date | null) => {
+    if (date) {
+      try {
+        setValue({ birthday: PlainDate.fromLocalDate(date) })
+      } catch {
+        // Couldn't parse date, no need to catch
+      }
+    }
+  }
```
steffenkleinle, line 35: :wrench: Please use `PlainDate.safeFromCustomFormat(date)` to avoid the null check/try-catch block

steffenkleinle (approved):
Tested on firefox, works great :tada:

administration/src/bp-modules/components/CustomDatePicker.tsx
f1sh1918: This library needs the Localization provider you can check the documentation

ztefanie (commented):
(minor)
Currently the styling of the DateField and the other Text fields in same, when not selected, but different when selected:

Unbenannt

administration/src/bp-modules/components/CustomDatePicker.tsx
f1sh1918: thx for the hint. i even don't get this particular styling.
I think we should unify the form fields to mui to have the same styling. currently the text input is blueprint and the date fields are mui. I will create a ticket for that

f1sh1918:
> (minor) Currently the styling of the DateField and the other Text fields in same, when not selected, but different when selected:
> 
> Unbenannt

I think this issue will be solved with https://github.com/digitalfabrik/entitlementcard/issues/1779 and is as you said really minor.
To manually override the style now is really ugly and difficult to find. You can give it a try, i already spend couple of hours on that

administration/package.json
```diff
@@ -27,8 +28,8 @@
     "date-fns": "^2.30.0",
     "graphql": "^16.8.1",
     "localforage": "^1.10.0",
-    "normalize.css": "^8.0.1",
     "normalize-strings": "^1.1.1",
+    "normalize.css": "^8.0.1",
```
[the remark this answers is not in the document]
michael-markl: Easiest solution would be to remove the `normalize-strings` dependency altogether :D I think, `str.normalize('NFKD').replace(/[^\x00-\x7F]/g, "")` would be a good enough replacement for it (this decomposes umlauts etc. to unicode "base" character + modifier symbol and removes non ascii characters -- such as the modifier symbols).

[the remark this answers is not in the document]
steffenkleinle: Interesting approach to get rid of this problem haha, I would have spent hours at trying to set the sorting order someway. I think this is a good idea (also since the library did not receive an update in 4 years). I will add an issue to do the same for integreat.

PR #1769, opened 2024-11-13
Adjust name length to 40 in druckerei

ztefanie:
### Short description
Nürnberg wanted to increase the possible chars in the name field

### Proposed changes
- [x] Adjust PDF
- [ ] Adjustments in Backend or qr code generation needed?

### Side effects
- Maybe the qr codes won't work as expected

### Testing
Create Nürnberg Pass and check pdf and qrcode activation

This PR contains the changes we applied to the pdf and send Nürnberg for approval. Once they approve this can be continued / merged.


steffenkleinle:
Won't do at the moment.

PR #1772, opened 2024-11-13
1430: Extend card

seluianova:
### Short description
Add the ability to extend the Koblenz Pass in the app.
- The card is extendable if its expiry date is earlier than the endDate of the associated user entitlement
- The user can see a notification 3 months (90 days) before the current card expires
- When the user clicks the “Extend Card” button, he is redirected to the self-service portal

Design: https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=792-3416&t=kUofvHMqh07BbICq-1

### Proposed changes
- Adjust backend: add the 'extendable' field to the card verification result
- Add ExtendCardNotification widget in the app

**Not included:**
- In this PR an empty form to create card is opened (query params are not passed).
Query params will be added in (or after) https://github.com/digitalfabrik/entitlementcard/pull/1749
- Tests (separate tasks will be created)
- I have not implemented a separate view for the case when the card has already expired, as suggested in the design, because I think our current view is fine, but feel free to object.

### Side effects
No?

### Testing
TBD

### Resolved issues
Part of: #1430


frontend/assets/koblenz/l10n/override_de.json
```diff
@@ -10,6 +10,9 @@
     "cardNotYetValid": "Der Gültigkeitszeitraum Ihres Passes hat noch nicht begonnen.",
     "checkFailed": "Ihr Pass konnte nicht auf seine Gültigkeit geprüft werden. Bitte stellen Sie sicher, dass eine Internetverbindung besteht und prüfen Sie erneut.",
     "codeRevoked": "Dieser Pass konnte nicht aktiviert werden, da er widerrufen wurde.",
+    "extendCard": "Pass verlängern",
+    "extendCardNotificationTitle": "Hinweis zum Ablauf des Passes",
+    "extendCardNotificationDescription": "Ihr Pass läuft demnächst ab.\nVerlängern Sie sie jetzt, um Ihre Dienste weiterhin nutzen zu können.",
```
seluianova, line 15:
```suggestion
    "extendCardNotificationDescription": "Ihr Pass läuft demnächst ab.\nVerlängern Sie ihn jetzt, um Ihre Dienste weiterhin nutzen zu können.",
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardQueryService.kt
```diff
@@ -34,13 +34,14 @@ class CardQueryService {
         val context = dfe.getContext<GraphQLContext>()
```
seluianova, line 34: Do we need to keep the deprecated function above?

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardQueryService.kt
seluianova, line 1: Do we need to keep the deprecated function above?

frontend/assets/l10n/app_de.json
```diff
@@ -109,6 +109,9 @@
     "codeVerificationFailedConnection": "Der eingescannte Code konnte nicht verifiziert werden. Bitte stellen Sie sicher, dass eine Internetverbindung besteht und prüfen Sie erneut.",
     "compareWithID": "Gleichen Sie die angezeigten Daten mit einem amtlichen Lichtbildausweis ab.",
     "comparedWithID": "Ich habe die Daten mit einem amtlichen Lichtbildausweis abgeglichen.",
+    "extendCard": "Karte verlängern",
+    "extendCardNotificationTitle": "Hinweis zum Ablauf der Karte",
+    "extendCardNotificationDescription": "Ihre Karte läuft demnächst ab.\nVerlängern Sie sie jetzt, um Ihre Dienste weiterhin nutzen zu können.",
```
michael-markl, line 114: "Dienste" sounds a bit odd to me...

Maybe just "..., um sie weiterhin nutzen zu können." ?

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardQueryService.kt
michael-markl: I guess, unless it's too much overhead to maintain, we should keep it. Afaics, we can just make it a wrapper of the un-deprecated function which should minimize maintenance overhead.
If / When we have Matomo tracking, we can check if there are still devices with old app versions using it... If there are (almost) none left, we can remove it.

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/types/CardVerificationResultModel.kt
```diff
@@ -4,5 +4,6 @@ import java.time.Instant
 
 data class CardVerificationResultModel(
     val valid: Boolean,
+    val extendable: Boolean = false,
```
michael-markl, line 7: Hmm... I think that verifiers should not be able to retrieve this information.
Possibly, we should create a new endpoint that uses the totp secret as authentication.

f1sh1918:
@hauf-toni 
We should also provide a slightly adjusted message if the card was already expired, since the user may not open the app for 90 days

seluianova:
> We should also provide a slightly adjusted message if the card was already expired
@f1sh1918 currently it looks like this, do you think it requires adjustment?
image

f1sh1918:
> > We should also provide a slightly adjusted message if the card was already expired
> 
> @f1sh1918 currently it looks like this, do you think it requires adjustment? (there was a small discussion in the task regarding that) image

Hm so if you click on "Weitere Aktionen" and click on first Item the link will differ depending on the card can be extended or not?

seluianova:
> Hm so if you click on "Weitere Aktionen" and click on first Item the link will differ depending on the card can be extended or not?

That was the idea, yes. Keep the existing menu item (since we have it already, and it says ‘beatragen ODER verlängern’), just add query params if the card is extendable.

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/types/CardVerificationResultModel.kt
seluianova: I assumed it's not a sensitive value but more technical, and it's not handled in the app when requested from the verifier

seluianova:
When the pass expired, I adjusted the message to `Ihr Pass ist abgelaufen. Unter "Weitere Aktionen" können Sie Ihren Pass verlängern oder einen neuen Pass beantragen.` to make it suitable for when the card is expired AND extendable, and for when it is expired but NOT extendable.
I'm not sure if we need different messages (and views) for these two cases. But feel free to object.
image

frontend/assets/l10n/app_de.json
seluianova: what about 'um Ihre Services weiterhin nutzen zu können.' ? 
just thinking if the original meaning could be preserved

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/types/CardVerificationResultModel.kt
michael-markl: I mean, I agree that for Koblenz, it does not matter because (afaik) the self-service portal would also reveal this information... Feel free to resolve it.

frontend/assets/l10n/app_de.json
michael-markl: Maybe, "um Ihre Vorteile weiterhin erhalten/nutzen zu können."? Maybe in English "... to continue receiving your benefits."?

michael-markl: But maybe it only sounds odd to me and "Dienste" is also fine.

seluianova: I like benefits/Vorteile

f1sh1918:
> When the pass expired, I adjusted the message to `Ihr Pass ist abgelaufen. Unter "Weitere Aktionen" können Sie Ihren Pass verlängern oder einen neuen Pass beantragen.` to make it suitable for when the card is expired AND extendable, and for when it is expired but NOT extendable. I'm not sure if we need different messages (and views) for these two cases. But feel free to object. image

Alright thx. Gonna check your pr.
Sorry for not reading the whole pr description.
I always complain about that if other reviewers don't do that :/

frontend/lib/identification/card_detail_view/extend_card_notification.dart
```diff
@@ -0,0 +1,79 @@
...
+                    child: Column(
+                      crossAxisAlignment: CrossAxisAlignment.start,
+                      children: [
+                        Text(t.identification.extendCardNotificationTitle),
+                        SizedBox(height: 8),
+                        Text(
+                          t.identification.extendCardNotificationDescription,
+                          textScaler: TextScaler.linear(0.9),
+                        ),
+                        SizedBox(height: 8),
```
f1sh1918, line 44: Is there a reason why you don't use the `Padding` widgets for set padding on texts and buttons?

frontend/lib/identification/card_detail_view/extend_card_notification.dart
```diff
@@ -0,0 +1,79 @@
...
+                              borderRadius: BorderRadius.circular(4),
+                            ),
+                          ),
+                          onPressed: () => _openApplication(),
+                          child: Text(t.identification.extendCard.toUpperCase()),
+                        ),
+                      ],
+                    ),
+                  ),
+                  SizedBox(width: 16),
```
f1sh1918, line 59: please use padding widget if possible

frontend/lib/identification/card_detail_view/extend_card_notification.dart
```diff
@@ -0,0 +1,79 @@
...
+                        _isVisible = false;
+                      });
+                    },
+                    child: Icon(Icons.close, size: 16),
+                  ),
+                ],
+              ),
+            ),
+          )
+        : SizedBox.shrink();
```
f1sh1918, line 72:
```suggestion
        : Container();
```
I think this is the standard way to render an empty `Container()`

frontend/lib/identification/card_detail_view/extend_card_notification.dart
```diff
@@ -0,0 +1,79 @@
...
+                            ),
+                          ),
+                          onPressed: () => _openApplication(),
+                          child: Text(t.identification.extendCard.toUpperCase()),
+                        ),
+                      ],
+                    ),
+                  ),
+                  SizedBox(width: 16),
+                  GestureDetector(
```
f1sh1918, line 60: There is also a standard widget from flutter that you may use (optional)
```js
                  IconButton(
                    padding: EdgeInsets.all(0),
                    alignment: Alignment.topRight,
                    onPressed: () {
                      setState(() {
                        _isVisible = false;
                      });
                    },
                    icon: const Icon(Icons.close),
                  ),
```

f1sh1918 (commented):
Looks really good actually.

I couldn't finish the testing, will continue tomorrow. Just left you some comments

frontend/lib/identification/card_detail_view/extend_card_notification.dart
seluianova: I wondered about that, but couldn't find any reason why Padding should be prefered other than in case of personal preference.
https://stackoverflow.com/questions/50722011/sized-box-vs-padding-for-distances-in-columns-and-rows
I personally think, the SizedBox looks neater, because the Padding implies a nested structure.
It's also used in other widgets already.

seluianova: I wondered about that, but couldn't find any reason why Padding should be prefered other than in case of personal preference. 
If you mean using Padding as a wrapper - then I think SizedBox would be neater because it doesn't imply a nested structure.
https://stackoverflow.com/questions/50722011/sized-box-vs-padding-for-distances-in-columns-and-rows
We also use the SizedBox in other widgets.

f1sh1918: Ok I'm fine with that even I think it's common use to add a padding around texts and is for me easier to understand the boilerplate Syntax since it's also common Syntax for every frontend dev.
Even you could also create a div Container as a spacer between components

But yes it may be personal preference

frontend/lib/identification/card_detail_view/extend_card_notification.dart
seluianova: this is supposed to have an invisible performance benefit, but if it's confusing, I will change that

seluianova:
> i think the extension container should have a box shadow too

@f1sh1918 if we add a shadow, we'll have to do something with the background color. because right now I'm using primary color + opacity.
it doesn't look nice with the shadow.

frontend/lib/identification/card_detail_view/extend_card_notification.dart
seluianova: To make the IconButton look the same as it does now, it needs to be customized as follows:
```
                  IconButton(
                    onPressed: () {
                      setState(() {
                        _isVisible = false;
                      });
                    },
                    icon: const Icon(Icons.close),
                    iconSize: 16,
                    padding: EdgeInsets.zero,
                    constraints: const BoxConstraints(),
                    style: const ButtonStyle(
                      tapTargetSize: MaterialTapTargetSize.shrinkWrap,
                    ),
                  ),
```
While for the GestureDetector I only need to set 2 attributes.

seluianova:
I think I have addressed all the comments, @f1sh1918 pls take a look

frontend/lib/identification/card_detail_view/extend_card_notification.dart
f1sh1918: then keep it as it is

frontend/lib/identification/card_detail_view/extend_card_notification.dart
```diff
@@ -0,0 +1,79 @@
...
+              child: Row(
+                crossAxisAlignment: CrossAxisAlignment.start,
+                children: [
+                  Icon(Icons.info, color: theme.primary),
+                  SizedBox(width: 16),
+                  Expanded(
+                    child: Column(
+                      crossAxisAlignment: CrossAxisAlignment.start,
+                      children: [
+                        Text(t.identification.extendCardNotificationTitle),
```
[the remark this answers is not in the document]
seluianova: applied, but I assume it depends on your PR https://github.com/digitalfabrik/entitlementcard/pull/1728
because currently the fonts don't match the design

seluianova:
I have removed my previous comment and added a minimal elevation to the notification to add some volume (I feel like the shadow shouldn't be the same as for the card...)
image
I also accidentally messed up the commits a bit and did rebase, sorry if you have to reset the branch locally because of that 🙄

seluianova:
I removed my previous comment and added a minimal elevation to the notification to add some volume (I feel like the shadow shouldn't be the same as for the card...)
image
I also accidentally messed up the commits a bit and did rebase, sorry if you have to reset the branch locally because of that 🙄

f1sh1918 (approved):
Works fine on ios 👍. 
Just one small thing commented

If you don't want to wait until the url is final (your other pr) then please create another ticket and add a ticket nr to the todo

seluianova:
> If you don't want to wait until the url is final (your other pr) then please create another ticket

I was not going to close the current task, untill everything is done. I find it convenient, but maybe we should discuss this topic in the team, if task splitting is more preferable 🙈

f1sh1918:
> > If you don't want to wait until the url is final (your other pr) then please create another ticket
> 
> @f1sh1918 I was not going to close the current task, untill everything is done. I find it convenient, but maybe we should discuss this topic in the team, if task splitting is more preferable in such cases 🙈

Alright i think the best way would be to define the queryParam keys in the `index.ts(frontend)` and import it in the `config.ts(administration)`

ztefanie (commented):
I tried to test this, but I get this error:

```
lib/identification/card_detail_view/extend_card_notification.dart:92:33: Error: The argument type 'ApplicationUrl' can't be assigned to the parameter type 'String'.
 - 'ApplicationUrl' is from 'package:ehrenamtskarte/build_config/build_config.dart' ('lib/build_config/build_config.dart').
    launchUrlString(buildConfig.applicationUrl, mode: LaunchMode.externalApplication);
```

Happens both for bayern and koblenz flavor. I deleted build_config.dart file

seluianova:
> I tried to test this, but I get this error

@ztefanie broken after merged changed, should be fixed now

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardQueryService.kt
```diff
@@ -15,32 +15,22 @@ class CardQueryService {
...
-        var verificationResult = false
-
-        if (card.codeType == CodeType.STATIC) {
-            verificationResult = card.totp == null && CardVerifier.verifyStaticCard(project, cardHash, projectConfig.timezone)
-        } else if (card.codeType == CodeType.DYNAMIC) {
-            verificationResult = card.totp != null && CardVerifier.verifyDynamicCard(project, cardHash, card.totp, projectConfig.timezone)
-        }
-        Matomo.trackVerification(context.backendConfiguration, projectConfig, context.request, dfe.field.name, cardHash, card.codeType, verificationResult)
-        return false
+        return verifyCardInProjectV2(project, card, dfe).valid
```
ztefanie, line 18: Do we really need to update this or can we delete this function as it is deprecated for 1.5 years and we do not use it?

frontend/assets/koblenz/l10n/override_de.json
```diff
@@ -5,11 +5,14 @@
     "activateDescription": "Sie haben den KoblenzPass bereits beantragt und einen Aktivierungscode erhalten? Scannen Sie den Code hier ein.",
     "activateTitle": "Pass aktivieren",
     "applyDescription": "Sie haben noch keinen KoblenzPass? Hier können Sie Ihren KoblenzPass beantragen.",
-    "cardExpired": "Ihr Pass ist abgelaufen. Unter \"Weitere Aktionen\" können Sie einen Antrag auf Verlängerung stellen.",
+    "cardExpired": "Ihr Pass ist abgelaufen. Unter \"Weitere Aktionen\" können Sie Ihren Pass verlängern oder einen neuen Pass beantragen.",
     "cardInvalid": "Ihr Pass ist ungültig. Er wurde entweder widerrufen oder auf einem anderen Gerät aktiviert.",
     "cardNotYetValid": "Der Gültigkeitszeitraum Ihres Passes hat noch nicht begonnen.",
     "checkFailed": "Ihr Pass konnte nicht auf seine Gültigkeit geprüft werden. Bitte stellen Sie sicher, dass eine Internetverbindung besteht und prüfen Sie erneut.",
     "codeRevoked": "Dieser Pass konnte nicht aktiviert werden, da er widerrufen wurde.",
+    "extendCard": "Pass verlängern",
```
ztefanie, line 13: It should not be needed to add the translations both to the overrides and the normal translations. As this is quite duplicate, i suggest to remove it.

seluianova: They are different. Normal translation is ”Karte verlängern”, and here is “Pass verlängern”

backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/webservice/schema/CardQueryService.kt
seluianova: There was a discussion about that above. I decided to refactor and keep it. Alternatively we should check if we don’t have users with the old app version.

frontend/assets/koblenz/l10n/override_de.json
ztefanie: Ah i see, missed that

Issue #1670, opened 2024-10-07
Show form field hints for card extensions


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a user i want to get a clear impression why my field input is invalid, so i want to get a clear hint with the information

**Describe the solution you'd like**
- add a method that shows a hint if the input field is invalid
- add a container next to the field that shows that hint
- these hints should be added to all card creatiion form fields and extensions

**Describe alternatives you've considered**
Keep it as it is

We have to clarify with ui/ux where and how we show that hint

PR #1777, opened 2024-11-14
WIP:1776: restrict latin charset

f1sh1918:
WIP!!!

### Proposed changes

-
-

### Side effects

-
-

### Resolved issues

Fixes: #


ztefanie:
:thought_balloon: Just one thought: Maybe we can just call the supportedCharset function once "manually" and just save the output to a variable (hardcode the charset). But you solution seems way more mature. Just wanted to suggest this, if you do not want spent much time on it

f1sh1918:
Yes i also thought about that. Just have to check how so save this properly into a variable. I think that would be good enough

f1sh1918:
i will close this pr @ztefanie 
I found a leaner solution with `XRegExp`
An easy to use online liner and will add this to #1670

Issue #1678, opened 2024-10-14
Change filename of the activation-pdf


dkehne:
**Is your feature request related to a problem? Please describe.**
The pdf files are named generic "berechtigungskarte-13.pdf" and this leads to opening/renaming-effort during the activation-/card-creation process.

**Describe the solution you'd like**
Include the name in the filename, e.g. "maxmustermann.pdf" or "berechtigungskarte-maxmustermann.pdf"

**Describe alternatives you've considered**
none.

**Additional context**
this should be only valid for single card creations. mass creation cannot include the name.

Issue #1756, opened 2024-11-06
File name of the activation PDF should contain the name of the authorized person.


dkehne:
**Is your feature request related to a problem? Please describe.**
If many maps are created at once, this leads to confusion after the download. The reason: All files have the same name or are only numbered up by -1 -2 -3 -4.
They are then renamed manually by the clerks.

**Describe the solution you'd like**
The file name of the PDF file should contain the name of the authorized person when downloading. The name of the PDF should contain the current year and the name. Example:
2024-Max Mustermann.pdf

If several maps are downloaded (mass import), everything remains as usual.

ztefanie:
duplicate of #1678

ztefanie:
close as completed in this PR: https://github.com/digitalfabrik/entitlementcard/pull/1778

PR #1778, opened 2024-11-15
1678: Change filename of single card pdf

ztefanie:
### Short description
Adjusted pdf filenames for single cards to owner name for nuernberg and bayern

### Side effects
none

### Testing
Create a single card in nuernberg or bayern -> filename should be firstname lastname.pdf
create multiple cards in nuernberg or bayern -> filename should stay as it is berechtigungskarten.pdf
koblenz card creation should stay as it is

### Resolved issues

Fixes: #1678 https://github.com/digitalfabrik/entitlementcard/issues/1756


administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -174,11 +174,11 @@ const useCardGenerator = (region: Region): UseCardGeneratorReturn => {
     async (applicationIdToMarkAsProcessed?: number) => {
       await generateCards(
         (codes: CreateCardsResult[], cards: Card[]) => generatePdf(codes, cards, projectConfig.pdf, region),
-        'berechtigungskarten.pdf',
+        cards.length === 1 ? `${cards[0].fullName}.pdf` : 'berechtigungskarten.pdf',
```
steffenkleinle, line 177: :question: Should we replace spaces with underscores or hyphens?

steffenkleinle (approved):

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
michael-markl: I think spaces are not critical (from a security perspective -- maybe from a convenience perspective). But maybe (from a security perspective) we should remove non ASCII characters (and potentially convert umlauts etc. to their "base" character) for the file name.
Something like 
```js
  fullName.replace(/ /g, "_").normalize('NFKD').replace(/[^a-z,A-Z,_]/g, '')
```

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -170,15 +170,18 @@ const useCardGenerator = (region: Region): UseCardGeneratorReturn => {
     [cards, client, projectConfig, handleError, sendCardConfirmationMails, region.activatedForCardConfirmationMail]
   )
 
+  const filenameSaveString = (input: string) =>
+    input.replace(/ /g, "_").normalize('NFKD').replace(/[^a-z,A-Z,_]/g, '')
```
steffenkleinle, line 174: :wrench: Depending on which PR is merged first, it would be cool if we could reuse the functionality from #1782 instead of re-implementing it here.

ztefanie: good point, as the other PR is already merged, i will update mine

f1sh1918 (approved):
LGMT 👍

PR #1781, opened 2024-11-17
Fix number of licenses

michael-markl:
### Short description

The interpolation of the number of licenses is currently broken.

### Proposed changes

Fix it by using the correct interpolation syntax.

### Side effects

None

### Testing

Open the licenses page


steffenkleinle (approved):
Tested on android

f1sh1918 (approved):
not tested

PR #1782, opened 2024-11-17
Remove normalize-strings dependency

michael-markl:
### Short description

Remove normalize-strings dependency.

Very similar functionality can be achieved by using `String.normalize` ([caniuse.com](https://caniuse.com/?search=String.normalize)) without this dependency.

### Side effects

Possibly slightly different outcomes of `normalizeName` and `normalizeString`. However, it's only used for comparing two names, so it should not be critical.

### Testing

For enthuastic testers: test the organization form in the application form.

### DIFF
--- a/administration/src/util/normalizeString.ts
+++ b/administration/src/util/normalizeString.ts
@@ -1,6 +1,11 @@
-import normalizeStrings from 'normalize-strings'
+// eslint-disable-next-line no-control-regex
+const nonAsciiRegex = /[^\x00-\x7F]/g
 
-const normalizeString = (str: string): string => normalizeStrings(str).toLowerCase().trim()
-export const normalizeName = (name: string): string => normalizeString(name).replaceAll(/[^a-zA-Z0-9]/g, '')
+const normalizeToAscii = (str: string): string => str.normalize('NFKD').replace(nonAsciiRegex, '')
+
+const normalizeString = (str: string): string => normalizeToAscii(str).toLowerCase().trim()
+
+const disallowedCharsInNameRegex = /[^a-zA-Z0-9]/g
+export const normalizeName = (name: string): string => normalizeString(name).replace(disallowedCharsInNameRegex, '')
 
 export default normalizeString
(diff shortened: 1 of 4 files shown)


administration/src/util/normalizeString.ts
```diff
@@ -1,6 +1,11 @@
-import normalizeStrings from 'normalize-strings'
+// eslint-disable-next-line no-control-regex
```
steffenkleinle, line 1: :question: Could you explain why this needs to be disabled/why you are checking for this?

administration/src/util/normalizeString.ts
```diff
@@ -1,6 +1,11 @@
...
+const nonAsciiRegex = /[^\x00-\x7F]/g
 
-const normalizeString = (str: string): string => normalizeStrings(str).toLowerCase().trim()
-export const normalizeName = (name: string): string => normalizeString(name).replaceAll(/[^a-zA-Z0-9]/g, '')
+const normalizeToAscii = (str: string): string => str.normalize('NFKD').replace(nonAsciiRegex, '')
```
steffenkleinle, line 4: :question: Why do we want the decomposed form here?

steffenkleinle (approved):
LGTM

administration/src/util/normalizeString.ts
michael-markl: To convert, for example, "ä" to "a+modifier". (this just resembles the previous implementation of `normalize-strings`.

administration/src/util/normalizeString.ts
michael-markl: Because the rule wants to forbid mentioning control characters in JS regex'es.
However, the ASCII symbols have control characters :D
Seems like a weird rule to me, but yeah :D

administration/src/util/normalizeString.ts
steffenkleinle: Ahhh makes sense, I did not get this from the docs. Thanks for explaining.

f1sh1918 (approved):

PR #1783, opened 2024-11-17
Use dart run instead of flutter pub run

michael-markl:
### Short description

`flutter pub run` is deprecated in favor of `dart run`

### DIFF
--- a/.idea/runConfigurations/Set_build_config_bayern.xml
+++ b/.idea/runConfigurations/Set_build_config_bayern.xml
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Set build config bayern" type="ShConfigurationType" folderName="Frontend">
-    <option name="SCRIPT_TEXT" value="fvm flutter pub run build_runner build --define &quot;df_build_config=name=bayern&quot;" />
+    <option name="SCRIPT_TEXT" value="fvm dart run build_runner build --define &quot;df_build_config=name=bayern&quot;" />
     <option name="INDEPENDENT_SCRIPT_PATH" value="true" />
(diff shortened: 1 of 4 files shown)


ztefanie (commented):
Thanks for this PR. Do you know why this was changed and how we can use a defined dart version?

I have the problem that on my laptop dart is always updated to another version on restart and i need to downgrade it in order to get entitlementcard running. I even have a script for this :D

```
sudo apt remove dart -y && sudo rm -rf /usr/lib/dart && sudo rm -rf /usr/local/share/dart
sudo apt install dart=3.3.1-1 -y
dart pub global deactivate fvm && dart pub global deactivate protoc_plugin
dart pub global activate fvm && dart pub global activate protoc_plugin
dart --version
```

How can we avoid that the wrong dart version is used. Isn't there something like `venv` or `nvm`?

michael-markl:
I think your problem is not related to this PR. This PR just gets rid of a warning when running the frontend run configs in intellij.

However, I think I had the same problem as you: there was some problem with protobuf when running the build runner. The problem was/is that the protoc_plugin is compiled for the globally installed dart version, but it seems to necessary to use the dart version from fvm.
I made some changes locally to use protobuf_builder instead of our own dart pub for generating protobuf. I can open a PR later today.

ztefanie (approved):

michael-markl:
fvm does not (by default) set the global version of flutter (which is good imo).

All commands should be preceded by `fvm`, instead.

The problem was that the protoc_plugin was installed using the global dart version (instead of the one we use in the entitlementcard project via fvm) which frequently led/leads to conflicts...

steffenkleinle (approved):

Issue #1581, opened 2024-08-21
Use fvm dart version to run builds


ztefanie:
Currently the build runner and build tasks in the Intellij configs (and even when running the commands in the command line) are using the globally installed dart version. This is not very robust and leads to errors.

We should change the project run config stuff to use the fvm dart version not the globally installed one.

Maxs comment on this: "I think there was a reason to do it the other way around, but maybe only for CI. Not so sure sadly but worth trying! It makes way more sense!"

ztefanie:
linux workaround script to downgrade dart:`
```
sudo apt remove dart -y && sudo rm -rf /usr/lib/dart && sudo rm -rf /usr/local/share/dart
sudo apt install dart=3.3.1-1 -y
dart pub global deactivate fvm && dart pub global deactivate protoc_plugin
dart pub global activate fvm && dart pub global activate protoc_plugin
dart --version
```

PR #1784, opened 2024-11-18
Use protoc_builder instead of own implementation

michael-markl:
### Short description

Replace df_protobuf with protoc_builder to avoid compatability problems with global Dart version.

### Proposed changes

protoc_builder downloads the specified protoc and protoc_plugin, compiles protoc_plugin for the Dart version specified/used by fvm.

### Side effects

The frontend project no longer uses the globally installed protoc version, but instead downloads its own protoc version during the build_runner process.

One more dependency :/

### Testing

Check if the build runner (& running the app) still works on your machine.
You can deactivate the globally installed protoc_plugin:
```bash
dart pub global deactivate protoc_plugin
```

### Resolved issues

Fixes: #1581


frontend/build.yaml
```diff
@@ -15,11 +15,31 @@ targets:
...
-      df_protobuf:
-        generate_for:
-          include:
-            - "card.proto"
-        enabled: True
+      protoc_builder:
+        options:
+          # The version of the Protobuf compiler to use.
+          protobuf_version: "27.5"
+          # The version of the Dart protoc_plugin package to use.
```
ztefanie, line 22: Should we really keep all these comments?

.circleci/src/jobs/check_frontend.yml
```diff
@@ -29,13 +27,11 @@ steps:
       command: |
         fvm flutter analyze --fatal-infos --fatal-warnings
         fvm flutter analyze pubs/df_build_config --fatal-infos --fatal-warnings
-        fvm flutter analyze pubs/df_protobuf --fatal-infos --fatal-warnings
```
steffenkleinle, line 32: :question: This check is now also done by the build runner?

ztefanie (approved):
Nice :rocket: 
Thank you for this PR. 

Tested locally, after cleaning .pub-cache it worked as expected :+1:

frontend/pubspec.yaml
```diff
@@ -61,9 +61,12 @@ dependencies:
 
 dev_dependencies:
   build_runner: ^2.3.3
+  protoc_builder:
+    git:
+      url: https://github.com/fischerscode/dart-protoc-builder.git
```
steffenkleinle, line 66: :question: This repo has 0 stars with the last activity 9 months ago, but I guess you picked this on purpose? Could you quickly explain why you chose this over the original repo (which seems to be exactly the same)?

frontend/build.yaml
```diff
@@ -15,11 +15,31 @@ targets:
...
             - "lib/build_config/build_config.yaml"
         enabled: True
-      df_protobuf:
-        generate_for:
-          include:
-            - "card.proto"
-        enabled: True
+      protoc_builder:
+        options:
+          # The version of the Protobuf compiler to use.
```
steffenkleinle, line 20: :thinking: do we need all those comments? Arent they usually self explanatory/could be easily looked up in the documentation?

frontend/build.yaml
michael-markl: I prefer to keep them, but removed them, since steffen also suggested removing them :D

frontend/pubspec.yaml
michael-markl: Yeah, the `protoc_builder` is in general not a too active repo. I decided to take the state of the PR https://github.com/pikaju/dart-protoc-builder/pull/21 as it seems much more reasonable to specify the protoc_plugin in the pubspec.yaml instead of making protoc_builder download it.

frontend/build.yaml
michael-markl: I removed them

.circleci/src/jobs/check_frontend.yml
michael-markl: I completely removed `pubs/df_protobuf`

steffenkleinle (approved):
LGTM

PR #1787, opened 2024-11-19
Upgrade some pubs

michael-markl:
### Short description

Upgrades pubs that are easy to upgrade.

The replacement of artemis (PR incoming) is a bit more involved (but still not outrageous in size). Therefore, I made it in a separate PR.

### Side effects

Hopefully none.

### Testing

Some smoke testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -609,13 +609,12 @@ jobs:
         working_directory: ~/project/backend
     check_frontend:
         docker:
-            - image: cimg/node:20.18.0-browsers
+            - image: cimg/node:20.18.0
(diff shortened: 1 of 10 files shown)


f1sh1918 (approved):
Smoke tested the app on ios. Works fine 👍 
Geolocation, deeplink activation, search, i18n

I updated the pods @michael-markl to get it work on ios

f1sh1918:
@michael-markl 
some translation files were created by slang for me which were not commited? What about them?

michael-markl:
I believe, these should be local files that are no longer necessary; you can probably just delete them.

Issue #1352, opened 2024-03-18
Replace artemis


f1sh1918:
**Is your feature request related to a problem? Please describe.**
We build dart types from graphql schemas and queries using `artemis`. Since may 23 `artemis` is discontinued.
As an alternative `ferry` is proposed. Currently artemis is only supported until `dart 3.1.0` but it still works with `3.3.0`
**Describe the solution you'd like**
- Replace artemis by another graphql code generation library
Options are : 
- https://pub.dev/packages/graphql_codegen
- https://github.com/gql-dart/ferry

**Describe an alternative**
Find another library

**Additional Information**
https://github.com/comigor/artemis

PR #1788, opened 2024-11-19
Replace artemis with graphql_codegen

michael-markl:
### Short description

Replace the discontinued artemis package with graphql_codegen.

### Proposed changes

I went with graphql_codegen, as I believe using `ferry` would require much more changes, and has an (imo) worse API.

### Side effects

None

### Testing

Smoke test interactions with the server.

### Resolved issues

Fixes: #1352


f1sh1918 (approved):
Tested on ios. seems to work fine 👍

Issue #1792, opened 2024-11-20
[DEV] Application url depends on environment


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a tester/dev i want to open the application process in the app correctly depending on the environment (production/staging/local)

**Describe the solution you'd like**
- provide different `applicationUrl`
- open the correct url depending the given environment (production/staging/local)

PR #1793, opened 2024-11-20
1792: Application url depends environment

f1sh1918:
### Short description

Currently we use one particular applicationUrl (Pass beantragen) for the app. For better testing on staging and local we should provide this depending on environment.

### Proposed changes

- add different applicationUrls
- open application process depending environment

### Testing

1. Open "Beantragen"  on local, staging and production environment

### Resolved issues

Fixes: #1792

### DIFF
--- a/frontend/build-configs/nuernberg/index.ts
+++ b/frontend/build-configs/nuernberg/index.ts
@@ -1,4 +1,8 @@
-import { ACTIVATION_PATH, NUERNBERG_PRODUCTION_ID, NUERNBERG_STAGING_ID } from "../constants"
+import {
+    ACTIVATION_PATH,
+    NUERNBERG_PRODUCTION_ID,
+    NUERNBERG_STAGING_ID
+} from '../constants'
 import BuildConfigType, { CommonBuildConfigType } from "../types"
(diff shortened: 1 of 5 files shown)


frontend/build-configs/nuernberg/index.ts
```diff
@@ -72,7 +76,11 @@ export const nuernbergCommon: CommonBuildConfigType = {
         verification: true,
         favorites: false,
     },
-    applicationUrl: "https://beantragen.nuernberg.sozialpass.app",
+    applicationUrl: {
+        production: 'https://beantragen.nuernberg.sozialpass.app',
+        staging: 'https://beantragen.nuernberg.sozialpass.app',
+        local : 'https://beantragen.nuernberg.sozialpass.app'
```
steffenkleinle, line 82: :wrench: I think this still has to be adjusted

steffenkleinle (approved):

frontend/build-configs/nuernberg/index.ts
f1sh1918: no there is just one url from nuernberg

michael-markl (approved):

Issue #1624, opened 2024-09-09
Add role and option to create api tokens for verein 360


f1sh1918:
**Is your feature request related to a problem? Please describe.**
If the mutation to add applications is called with the "isAlreadyVerified" flag it needs to be protected with the api token. These api tokens should be created by a user with a specific role.

**Describe the solution you'd like**
- create a separate role "Verein 360" that can create api tokens
- check if the api token has the correct access rights (is role Verein360 and project eak)

f1sh1918:
@ztefanie not sure if we also have to add the possibility to create a user with this role via administration gui or it is sufficient to create it via runConfig and backend command, since we probably will only need one. Please check if there is anything to add in this ticket :)

seluianova:
@f1sh1918 @ztefanie just for my better understanding, could you please clarify why we decided to add a new role for that?
why not allow tokens to be created by project admin, for example?

seluianova:
@ztefanie right, makes sense, thanks for the reply!

PR #1794, opened 2024-11-20
1624: Add role for verein360 to create api tokens

ztefanie:
### Short description
Added a new role for verein360 admins

### Proposed changes
- added a new role to create api tokens for verein360
- added new column "type" for api tokens
- adjusted project settings
- new external-user role can only create tokens of type "verified_application"
- project admins of koblenz create tokens of type "user_import"
- only api tokens of type user-import are allowed to use koblenz csv endpoint

### Side effects
- may break anything releated to api token generation of koblenz project admins or csv upload endpoint

### Testing
- create and delete tokens for different roles, check if type is set correctly
- check if endpoint for koblenz csv shows forbidden exception when used with api token of type external-verified
- add and remove user with external-api role

### Resolved issues

Fixes: #1624


administration/src/bp-modules/NavigationBar.tsx
```diff
@@ -69,7 +69,7 @@ const Navigation = ({ onSignOut }: Props): ReactElement => {
             <Button minimal icon='path-search' text='Region verwalten' />
           </NavLink>
         ) : null}
-        {role === Role.ProjectAdmin && config.userImportApiEnabled ? (
+        {(role === Role.ProjectAdmin && config.userImportApiEnabled) || role === Role.ExternalVerifiedApiUser ? (
```
f1sh1918, line 72: :wrench:
I think it would improve the readability if we introduce a variable like
`canSeeProjectSettings` or sth similar

f1sh1918 (commented):
The concept looks really good to me 👍 
Just found some issues

image

1. In the dashboard the item for "Projekt verwalten" is missing when i login with external api user

2. When i click on "Projekt verwalten" i get a confusing headline
image

I think the headline should be somehow connected with the user role to indicate the purpose of the token dynamically

3. Is it intended that Koblenz can also create an external api user?
I mean it could make sense to separate the user import right from project admin to a specific role (external api users) but is in our scenario for koblenz not really needed
image

I think it could make sense to even a check in the backend that only certain ROLES and projects can `mayCreateExternalAPIUser` same for edit

ztefanie:
> 1. In the dashboard the item for "Projekt verwalten" is missing when i login with external api user

good find, fixed it.

> 2. When i click on "Projekt verwalten" i get a confusing headline
> 
fixed this by chaing the headling to "Api token"

> 
> 3. Is it intended that Koblenz can also create an external api user?
>    I mean it could make sense to separate the user import right from project admin to a specific role (external api users) but is in our scenario for koblenz not really needed
> 
> I think it would make sense to even add a check in the backend that only certain ROLES and projects can create external api user f.e. `mayCreateExternalAPIUser` fkt same for edit

Good point. I added this check to the existing "mayCreateUser" function.

administration/src/bp-modules/users/RoleSelector.tsx
```diff
@@ -12,14 +13,21 @@ const RoleSelector = ({
...
+  const config = useContext(ProjectConfigContext)
+
+  return (
+    <HTMLSelect fill onChange={e => onChange((e.target.value as Role | null) ?? null)} value={role ?? ''} required>
+      <option value='' disabled>
+        Auswählen...
+      </option>
+      {hideProjectAdmin ? null : <option value={Role.ProjectAdmin}>{roleToText(Role.ProjectAdmin)}</option>}
+      {hideProjectAdmin || !config.applicationFeature?.applicationUsableWithApiToken ? null : (
+        <option value={Role.ExternalVerifiedApiUser}>{roleToText(Role.ExternalVerifiedApiUser)}</option>
```
f1sh1918, line 26:
```suggestion
     {config.applicationFeature?.applicationUsableWithApiToken && currentUserRole === Role.ProjectAdmin ? (
        <option value={Role.ExternalVerifiedApiUser}>{roleToText(Role.ExternalVerifiedApiUser)}</option>
      ) : null}
```
I think your check is really hard to read and even we need a line code more i think we should add the context for the currentUserRole and check that
```js
const { role: currentUserRole } = useContext(WhoAmIContext).me!
```

f1sh1918 (approved):
One small change i would really recommend because i found it difficult to understand the code.
Works fine. Nicely done 👍

administration/src/bp-modules/project-settings/UserEndpointSettings.tsx
```diff
@@ -49,7 +49,10 @@ const DeleteIcon = styled(Delete)`
   cursor: pointer;
 `
 
-const UserEndpointSettings = (): ReactElement => {
+type UserEndpointSettingsProps = {
```
seluianova, line 52: :upside_down_face: I think it's not "user endoint settings" anymore
ProjectSettings / ApiTokenSettings ?

seluianova:
:upside_down_face: description of the new role is missing here
image
:thought_balloon: also, maybe we want to disable this role for projects where it's not needed? (not necessarily as part of this PR but in general)

seluianova (approved):
lgtm, just shared a couple of minor thoughts

Issue #1710, opened 2024-10-23
Handle empty language overrides


f1sh1918:
**Is your feature request related to a problem? Please describe.**
While initialization locale overrides will be applied for the defined languages
Currently for koblenz we only have one local `de` and `en` can't be found and an unhandled error is thrown

**Describe the solution you'd like**
- avoid setting overrides with language files that are not available

lear and concise description of any alternative solutions or features you've considered.

**Additional context**
```
flutter: Environment: local
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Unable to load asset: "assets/koblenz/l10n/override_en.json".
The asset does not exist or has empty data.
```

PR #1798, opened 2024-11-22
1710: Handle empty language overrides

seluianova:
### Short description
Get rid of the exception when starting the Koblenz app

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Unable to load asset: "assets/koblenz/l10n/override_en.json".
The asset does not exist or has empty data.
```

### Proposed changes
- override translations only if the asset exists

### Side effects
- no?

### Testing
Start the Koblenz app, check there is no exception in log

### Resolved issues
Fixes: #1710

### DIFF
--- a/frontend/lib/main.dart
+++ b/frontend/lib/main.dart
@@ -51,8 +51,12 @@ Future<String?> loadLocaleOverride(AppLocale locale) async {
   final path = '${buildConfig.localeOverridePath}/override_${locale.languageCode}.json';
   try {
     return await rootBundle.loadString(path);
-  } catch (e) {
-    return null;
+  } on FlutterError catch (e) {
+    if (e.message.contains('Unable to load asset')) {
+      debugPrint('Locale override not found at path: $path. The default translation will be used');
+      return null;
+    }
+    rethrow;
   }
 }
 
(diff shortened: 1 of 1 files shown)


frontend/lib/main.dart
```diff
@@ -52,3 +46,19 @@ Future<void> main() async {
     run();
   }
 }
+
+Future<String?> loadLocaleOverride(AppLocale locale) async {
+  final path = '${buildConfig.localeOverridePath}/override_${locale.languageCode}.json';
+  try {
+    return await rootBundle.loadString(path);
+  } catch (e) {
+    return null;
```
ztefanie, line 55: :question: Can this happen in production? Is it fine to just ignore this exception?

ztefanie (approved):
Looks good. Not tested

frontend/lib/main.dart
seluianova: Well, we could actually check the exception message, and only ignore if it's "asset doesn't exist". Maybe it would make sense

steffenkleinle: :question: I think we should definitely handle this, thanks. However, shouldn't we display at least a warning message to the console in case this is not missing on purpose?

steffenkleinle (approved):
Tested on android

frontend/lib/main.dart
seluianova: Added message check and log message, thanks for the ideas :)

PR #1801, opened 2024-11-26
1430: Add tests for the verifyCardInProjectV2 query

seluianova:
### Short description
Backend tests for verifyCardInProjectV2 query as part of https://github.com/digitalfabrik/entitlementcard/issues/1430

### Proposed changes
Added VerifyCardTest class to check valid & extendable fields in the verification result.
**Only static** cards are covered here, because for dynamic cards we need to pass the totp token as a param and and I haven't decided yet whether to generate it in the tests or mock.

### Side effects
No, just tests


backend/src/test/kotlin/app/ehrenamtskarte/backend/cards/VerifyCardTest.kt
```diff
@@ -0,0 +1,173 @@
...
+            val userId = TestData.createUserEntitlement(regionId = 95, endDate = LocalDate.now().plusMonths(2L))
+            return TestData.createDynamicCard(
+                entitlementId = userId,
+                expirationDay = LocalDate.now().plusMonths(1L).toEpochDay(),
+                totpSecret = null
+            )
+        }
+    }
+
+    @AfterEach
```
ztefanie, line 103: :question: Wouldn't it make more sense to do this in `@BeforeEach` to ensure also the first test starts with a clean setup?

backend/src/test/kotlin/app/ehrenamtskarte/backend/cards/VerifyCardTest.kt
```diff
@@ -0,0 +1,173 @@
...
+    fun cleanUp() {
+        transaction {
+            Cards.deleteAll()
+            UserEntitlements.deleteAll()
+        }
+    }
+
+    @ParameterizedTest
```
ztefanie, line 111: :+1: Nice i really like the parameterized tests :)

ztefanie (approved):
Nice, looks good. Thanks for adding these tests :rocket:

backend/src/test/kotlin/app/ehrenamtskarte/backend/cards/VerifyCardTest.kt
seluianova: actually yes, it would be more robust. will update that

Issue #1802, opened 2024-11-27
Fix more actions button not properly displayed


f1sh1918:
**Describe the bug**
If the card extension notification is displayed and multiple cards are activated on the device so the carousel indicator is shown, the "Weitere Aktionen" Button is not properly displayed and can't be used.
Currently a workaround is applied to add an additional sized box.

**To Reproduce**
1. Remove SizedBox() in `card_detail_view`
2. Activate multiple cards for koblenz and make one card extendable for koblenz
3. Scroll down to "More actions"
4. Button is not displayed and can't be used
5. May only happen for ios

**Expected behavior**
The button will be displayed properly

**Screenshots**
![image](https://github.com/user-attachments/assets/c09c4a65-52e1-436d-9904-deb2ff045240)

seluianova:
Doesn't seem to be reproducible on Android
![image](https://github.com/user-attachments/assets/0ced4327-5313-4f3c-8498-237007c95352)

bahaaTuffaha:
I can't reproduce it on android...
where did this pop up come from? it didn't show up..

![Image](https://github.com/user-attachments/assets/ff53e0c4-6721-4f8f-932e-f5e23c7d3e6a)


at my end it look like this (pixel 7 pro - 6.7'):

![Image](https://github.com/user-attachments/assets/b86e9a07-6e3b-4165-807f-46265189f657)

Tested also on (pixel 2 - 5.0') I can scroll to the button without issues:

![Image](https://github.com/user-attachments/assets/d2f6b399-6795-45f6-90b9-68438c275e2b)

f1sh1918:
@bahaaTuffaha this popup comes from the card extension and as tory mentioned it looks like its an ios only issue

PR #1803, opened 2024-11-27
1430: Extend card add params

f1sh1918:
### Short description

As a user i want the forms in the self service portal to be prefilled with my personal information if i click "Karte verlängern" with an existing card

#### Note
I'm not super happy with this solution, but since we can not generate proper functions from the `frontend/build-configs/index` to the `buildConfig.dart` to provide a function for each project to generate the applicationUrl with (or without) queryParams i didn't find a better solution. 

I mean we could maybe adjust the build config generation to be able to generate functions with parameters but this means lots of effort.

At least this solution avoids multiple linting ignore that would be needed if we provide a query param key object instead of single string values that can be null. Feel free to suggest a better solution but beware that you should run the buildConfig for all the projects to ensure that nothing breaks.

### Proposed changes

- pass query params keys to the app, ensure that they are the same as in administration
- provide a function that return the applicationUrl with queryParams
- open the self service portal with queryParams if a card exists
- add tests
- workaround for #1802 

### Side effects

- may create a wrong applicationUrl for other projects. so open "Beantragen" for other projects

### Testing
- Create a card for koblenz that is less than 90 days valid from now
- Activate this card
- Change the end date in "UserEntitlements" to a later date
- Click on the "Karte verlängern" Button or go to more actions and extend the card there
- All form fields should be prefilled with valid information
- Note: I also tested "Umlaute and dots" and everything worked fine. Please also test that properly

### Resolved issues

partially fixes: #1430


frontend/build-configs/types.ts
```diff
@@ -99,6 +99,9 @@ export type CommonBuildConfigType = {
         production: string
         local: string
     }
+    applicationUrlQueryKeyName: string | null,
```
ztefanie, line 102: I think it would be better to move this to separate object
```
applicationUrlQuery  {
name: string
birthday: string
referenceNumber: string} | null
```

would also make the checks easiert and we only need to set one attribute to null for the other projects

frontend/build-configs/constants/index.ts
```diff
@@ -7,4 +7,8 @@ export const BAYERN_STAGING_ID = 'staging.bayern.ehrenamtskarte.app'
 export const NUERNBERG_PRODUCTION_ID = 'nuernberg.sozialpass.app'
 export const NUERNBERG_STAGING_ID = 'staging.nuernberg.sozialpass.app'
 export const KOBLENZ_PRODUCTION_ID = 'koblenz.sozialpass.app'
-export const KOBLENZ_STAGING_ID = 'staging.koblenz.sozialpass.app'
\ No newline at end of file
+export const KOBLENZ_STAGING_ID = 'staging.koblenz.sozialpass.app'
+
+export const QUERY_PARAM_NAME = 'Name'
```
ztefanie, line 12: i would prefer english names for this, as i think this is quite standard to do so :thinking:

ztefanie (commented):
Tried to test this, the query params are set correctly, but the fields are not prefilled for me.

image

the url is: localhost:3000/erstellen?Name=Stefanie+Test&Geburtsdatum=10.06.2003&Referenznummer=123K

frontend/build-configs/types.ts
f1sh1918: Yes i tried this at the first place but i didn't work because this will result to an empty class for other projects in the build config dart file

frontend/build-configs/constants/index.ts
f1sh1918: We always used german names for columns in the project i think so i kept this

f1sh1918:
> Tried to test this, the query params are set correctly, but the fields are not prefilled for me.
> 
> image
> 
> the url is: localhost:3000/erstellen?Name=Stefanie+Test&Geburtsdatum=10.06.2003&Referenznummer=123K
> 

Interesting I tested this on ios and worked fine even this shouldn't be related to this pr

f1sh1918:
@ztefanie 
i can not reproduce it on android. even dunno why inapp browser is opening maybe this is the issue

Bildschirmfoto 2024-12-02 um 11 53 21

frontend/lib/util/get_application_url.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:ehrenamtskarte/build_config/build_config.dart';
+import 'package:ehrenamtskarte/configuration/definitions.dart';
+import 'package:ehrenamtskarte/constants/project_ids.dart';
+import 'package:ehrenamtskarte/identification/util/card_info_utils.dart';
+import 'package:ehrenamtskarte/proto/card.pb.dart';
+
+String getApplicationUrl(ApplicationUrl applicationUrl, bool isStagingEnabled) {
```
steffenkleinle, line 7: :wrench: Why not just pass the build context and handle everything in these two methods? Would make clearer code  and separate responsibilities better imo.
```suggestion
String getApplicationUrl(BuildContext context) {
  final isStagingEnabled = Provider.of<SettingsModel>(context, listen: false).enableStaging;
  final applicationUrl = buildConfig.applicationUrl;
```

frontend/lib/util/get_application_url.dart
```diff
@@ -0,0 +1,44 @@
...
+          : applicationUrl.local;
+}
+
+String getApplicationUrlWithParameters(
+    String applicationUrl,
+    CardInfo cardInfo,
+    String projectId,
+    String? applicationUrlQueryKeyName,
+    String? applicationUrlQueryKeyBirthday,
+    String? applicationUrlQueryKeyReferenceNumber) {
```
steffenkleinle, line 21:
```suggestion
String getApplicationUrlWithParameters(BuildContext context, CardInfo cardInfo) {
  final applicationUrl = getApplicationUrl(context);
  final projectId = buildConfig.projectId.production;
  final applicationUrlQueryKeyName = buildConfig.applicationUrlQueryKeyName;
  final applicationUrlQueryKeyBirthday = buildConfig.applicationUrlQueryKeyBirthday;
  final applicationUrlQueryKeyReferenceNumber = buildConfig.applicationUrlQueryKeyReferenceNumber;
```

frontend/build-configs/types.ts
steffenkleinle: ~~How about using an array, e.g.:~~
``` ts
// koblenz
applicationQueryParams: [QUERY_PARAM_NAME, QUERY_PARAM_BIRTHDAY, QUERY_PARAM_KOBLENZ_REFERENCE_NUMBER],
// bayern
applicationQueryParams: [],
```
Edit: I think not that easy to implement because then you need to check in dart code what each param key means/which field it refers to.

seluianova:
> Tried to test this, the query params are set correctly, but the fields are not prefilled for me.

it seems I have the same issue on the current branch.
but on main it works if I apply this: https://github.com/digitalfabrik/entitlementcard/pull/1806
I think we need to merge it and try to update the current branch..

frontend/lib/util/get_application_url.dart
```diff
@@ -0,0 +1,44 @@
...
+  if (projectId != koblenzProjectId) {
+    return applicationUrl;
+  }
```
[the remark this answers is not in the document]
f1sh1918: yes i thought about that. it was just meant to clarify that its particular for koblenz project but can remove it

frontend/lib/util/get_application_url.dart
f1sh1918: because it makes testing way more difficult to mock the whole  build context.

frontend/lib/util/get_application_url.dart
f1sh1918: if i change that, we need some linting ignores for null checks, since these values don't exist for all projects

ztefanie (approved):
Not tested again.

frontend/lib/identification/identification_page.dart
```diff
@@ -45,14 +45,25 @@ class IdentificationPageState extends State<IdentificationPage> {
           return Container();
         }
 
+        final isStagingEnabled = Provider.of<SettingsModel>(context, listen: false).enableStaging;
+        final applicationUrl = getApplicationUrl(buildConfig.applicationUrl, isStagingEnabled);
         if (userCodeModel.userCodes.isNotEmpty) {
           final List<Widget> carouselCards = [];
           for (var code in userCodeModel.userCodes) {
+            final applicationUrlWithParameters = getApplicationUrlWithParameters(
```
seluianova, line 53: :thought_balloon: I think we only wanted to set parameters if the card is extendable

steffenkleinle (approved):
I had to run `yarn tsc --build` once to get administration to pick up the changes in frontend/build-configs. Tested on firefox, works as expected! Thanks for putting time into this, seems like quite a nasty thing...

frontend/lib/util/get_application_url.dart
f1sh1918: added this in last commit

frontend/lib/identification/identification_page.dart
f1sh1918: i added this in the last commit and also adjusted the tests! thx for the hint
@seluianova 
Feel free to test

frontend/lib/util/get_application_url.dart
```diff
@@ -0,0 +1,38 @@
...
+        host: parsedApplicationUrl.host,
+        port: parsedApplicationUrl.port,
+        path: parsedApplicationUrl.path,
+        queryParameters: {
+          applicationQueryKeyName: cardInfo.fullName,
+          applicationQueryKeyBirthday: getFormattedBirthday(cardInfo),
+          applicationQueryKeyReferenceNumber: cardInfo.extensions.hasExtensionKoblenzReferenceNumber()
+              ? cardInfo.extensions.extensionKoblenzReferenceNumber.referenceNumber
+              : null
+        }).toString();
```
seluianova, line 35:
```suggestion
  final parsedUrl = Uri.parse(applicationUrl);
  final queryParams = {
    applicationQueryKeyName: cardInfo.fullName,
    applicationQueryKeyBirthday: getFormattedBirthday(cardInfo),
    applicationQueryKeyReferenceNumber: cardInfo.extensions.hasExtensionKoblenzReferenceNumber()
        ? cardInfo.extensions.extensionKoblenzReferenceNumber.referenceNumber
        : null,
  };
  return parsedUrl.replace(queryParameters: queryParams).toString();
```
:upside_down_face: small simplification

seluianova:
> ok i will merge recent main when you pr was merged

now I don't understand why it doesn't work 🤔
even if I open the link with params in the desktop browser, the form is not filled.
it works on main, but doesn't work on the current branch...

frontend/lib/identification/identification_page.dart
```diff
@@ -45,14 +46,21 @@ class IdentificationPageState extends State<IdentificationPage> {
           return Container();
         }
 
+        final applicationUrl = getApplicationUrl(context);
         if (userCodeModel.userCodes.isNotEmpty) {
           final List<Widget> carouselCards = [];
           for (var code in userCodeModel.userCodes) {
+            final applicationUrlForCardExtension = isCardExtendable(code.info, code.cardVerification)
```
seluianova, line 53:
```suggestion
            final applicationUrl = isCardExtendable(code.info, code.cardVerification)
```
:upside_down_face: I think it would make more sense. because it will be 'for extension' only is the card is extendable. but the same variable is also used when the card is not extendable.

seluianova:
Just added a couple of optional nit-picks, because overall it looks good 👍
But I'm concerned about the following:

> ok i will merge recent main when you pr was merged

The form is still not being filled out for me 🤔
Even if I open the link with params in the desktop browser, the form is not filled out, so it doesn't seem to be related to the app or emulator.
It works on main for me, but doesn't work on the current branch...
Does it work for you? Is it only reproduced for me and Steffi?

f1sh1918:
> Just added a couple of optional nit-picks, because overall it looks good 👍 But I'm concerned about the following:
> 
> > ok i will merge recent main when you pr was merged
> 
> The form is still not being filled out for me 🤔 Even if I open the link with params in the desktop browser, the form is not filled out, so it doesn't seem to be related to the app or emulator. It works on main for me, but doesn't work on the current branch... Especially strange if it's only reproduced for me and Steffi...

Did you run `npm i` on root folder? Because i set the queryParam keys in `frontend` buildConfig and only if you run node modules installation these new keys can be found in `administration` module.
Maybe thats the issue. I tested the queryParam init on all browser (chrome, firefox, safari)

image

seluianova:
> Did you run npm i on root folder? 

Ah, no I haven't. I will be available tomorrow only from 2 pm, so feel free to merge if you want to deploy it earlier, I basically tested everything except that. I believe it should be good to go.

PR #1807, opened 2024-12-02
Improve development setup docs

steffenkleinle:
### Short description

I cleaned up the development setup with new best practices.

### Proposed changes

- Remove duplicated fronted setup documentation
- Clean up development setup documentation

### Side effects

- None.

### Testing

None.

### Resolved issues

Fixes: None.


docs/development-setup.md
```diff
@@ -1,69 +1,147 @@
...
+
+### Run Frontend
+
+1. Forward ports. As there are several services running and interacting some ports need to be forwarded.
 This includes the backend port 8000 and the map tiles port at 5002.
-For this run:
+The command has to be run every time a device is connected.
+
+```shell
+adb reverse tcp:8081 tcp:8081 && adb reverse tcp:8000 tcp:8000 && adb reverse tcp:5002 tcp:5002
```
steffenkleinle, line 57: This includes port forwarding for integreat as I figured having one command in the history for this makes sense. However, I am also open to keep it entitlementcard only.

docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: "3.5"
```
steffenkleinle, line 1: This is obsolete and led to a warning on startup.
See https://docs.docker.com/reference/compose-file/version-and-name/.

docs/development-setup.md
f1sh1918: hm we have the `expose:ports` that can be triggered via `npm run expose:ports` i think you should even update this (it also checks for attached devices) and add the command here

f1sh1918 (approved):
Looks good to me

seluianova (approved):
looks nice 👍

PR #1809, opened 2024-12-02
1430: Show extend card notification only if the card is valid

seluianova:
### Short description
Just noticed a small issue:
After a user extends the card and activates the new one, the extend card notification is still displayed for the old (revoked) card, and this can be confusing.

### Proposed changes
Change the condition to display the notification: if the card is valid (instead of not expired) and extendable

### Side effects
Couldn’t think of any

### Testing
If needed, please refer to the detailed testing instructions from here:
https://github.com/digitalfabrik/entitlementcard/pull/1772
Once the extendable card has been created:
- click ‘Pass verlängern’ in the app notification
- submit the form
- activate the new card
- check that the extend card notification is not displayed anymore

### Resolved issues
Follow up for #1430


michael-markl:
Should the user be able to extend an expired card?

seluianova:
> Should the user be able to extend an expired card?

Yes, under "Weitere Aktionen" --> KonlenzPass beantragen oder verlängern

michael-markl:
I see. It's maybe a bit weird that the "notification" for extension is no longer shown once the card is expired (I think this is the app's behavior with and without your PR).
Feel free to merge, though.

seluianova:
@michael-markl when the card has expired, the message about the card extension is displayed below the card
IMG_0971

Issue #1791, opened 2024-11-20
create application verification automatically for verein360 requests


ztefanie:
- Create application verification automatically on applications that are sent with the isAlreadyVerified flag.
- migrate the database so the applicationverification has a "source" field, where we enter "Verein360"

PR #1810, opened 2024-12-03
1791: Create application verifications automatically for verein360 applications

ztefanie:
### Short description
Application verifications get source set for verein360 applications

### Proposed changes
- adjusted database to have column "automaticSource"
- refined check for incoming applications

### Side effects
None, maybe affecting normal application creation process

### Testing
create new application with one or more organizations. Check for correct error messages if isAlreadyVerified differs. Check if one or all are true, the applicationVerifications table column automaticSource is set properly to VEREIN360

Feel free to test the complete feature in this PR: https://github.com/digitalfabrik/entitlementcard/pull/1811

### Resolved issues
Fixes: #1791

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
@@ -2,6 +2,7 @@ package app.ehrenamtskarte.backend.application.database.repos
 
 import app.ehrenamtskarte.backend.application.database.ApplicationEntity
 import app.ehrenamtskarte.backend.application.database.ApplicationVerificationEntity
+import app.ehrenamtskarte.backend.application.database.ApplicationVerificationExternalSource
 import app.ehrenamtskarte.backend.application.database.ApplicationVerifications
 import app.ehrenamtskarte.backend.application.database.Applications
 import app.ehrenamtskarte.backend.application.webservice.schema.view.ApplicationView
@@ -60,6 +61,7 @@ object ApplicationRepository {
                     this.contactName = it.contactName
                     this.organizationName = it.organizationName
                     this.contactEmailAddress = it.contactEmailAddress
+                    this.automaticSource = ApplicationVerificationExternalSource.NONE
                 }
             }
 
@@ -151,6 +153,7 @@ object ApplicationRepository {
                 false
             } else {
                 applicationVerification.verifiedDate = Instant.now()
+                applicationVerification.automaticSource = ApplicationVerificationExternalSource.VEREIN360
                 true
             }
         }
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
@@ -116,7 +116,10 @@ object ApplicationRepository {
         }
     }
 
-    fun getApplicationByApplicationVerificationAccessKey(applicationVerificationAccessKey: String, dfe: DataFetchingEnvironment): ApplicationView {
+    fun getApplicationByApplicationVerificationAccessKey(
+        applicationVerificationAccessKey: String,
+        dfe: DataFetchingEnvironment
+    ): ApplicationView {
         val logger = LoggerFactory.getLogger(ApplicationRepository::class.java)
         val context = dfe.getContext<GraphQLContext>()
         return transaction {
@@ -146,14 +149,17 @@ object ApplicationRepository {
         return applicationVerification.verifiedDate != null || applicationVerification.rejectedDate != null
     }
 
-    fun verifyApplicationVerification(accessKey: String): Boolean {
+    fun verifyApplicationVerification(
+        accessKey: String,
+        automaticSource: ApplicationVerificationExternalSource = ApplicationVerificationExternalSource.NONE
+    ): Boolean {
         return transaction {
             val applicationVerification = getApplicationVerification(accessKey)
             if (isAlreadyVerified(applicationVerification)) {
                 false
             } else {
                 applicationVerification.verifiedDate = Instant.now()
-                applicationVerification.automaticSource = ApplicationVerificationExternalSource.VEREIN360
+                applicationVerification.automaticSource = automaticSource
                 true
             }
         }
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
@@ -14,6 +14,7 @@ import app.ehrenamtskarte.backend.exception.webservice.exceptions.RegionNotFound
 import app.ehrenamtskarte.backend.mail.Mailer
 import app.ehrenamtskarte.backend.regions.database.repos.RegionsRepository
 import graphql.execution.DataFetcherResult
+import io.javalin.http.BadRequestResponse
 import org.jetbrains.exposed.sql.transactions.transaction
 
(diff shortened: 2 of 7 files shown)


f1sh1918 (commented):
First of all a some general thoughts

I'm just thinking of how we "mark" this application in the overview view table and i'm not sure if we really have to define this "source" attribute on a verification but on an application level once. That would simplify that we don't need to loop through all organizations (Even the verification source values will be equal for a application).
Or do you think we will have the case where we have for one application different verification sources?
I would prefer to just add a "Source" column to applications to keep it simple because one application comes typically from one source.
We could then set "Internal" as default source or sth. similar

Maybe i've overseen sth that makes your implementation needed.

ztefanie:
> First of all some general thoughts
> 
> I'm just thinking of how we "mark" this application in the overview view table and i'm not sure if we really have to define this "source" attribute on a verification but on an application level once. That would simplify that we don't need to loop through all verifications (Even the verification source values will be equal for a application). Or do you think we will have the case where we have for one application different verification sources? I would prefer to just add a "Source" column to applications to keep it simple because one application comes typically from one source. We could then set "Internal" as default source or sth. similar
> 
> Maybe i've overseen sth that makes your implementation needed.

Yes of course I thought about this, too. I decided to do it this way for the following reasons:
1) The most important reasone for me to do so is single responsibility: It is where it logically belongs to, as the applicaiton-verifications are espeically for this use case. Adding the automaticVerified field to application-verifications ensures that each table and entity has a single, clear purpose. The applications table is responsible for storing general information about an application, while application-verifications focuses on tracking and managing the verification details. By adding automaticVerified to application-verifications, we avoid overloading the applications table with information that doesn't directly pertain to its core responsibility.

2) It is also on this level in the api-endpoint / the graphQL scheme. We cannot change this anymore, as this is already sent to verein360. I think it would be confusing if we handle this incosistently, and putting it on organization/application-verification level in api, but on application level in the datastructure.

3) If automaticVerified were placed in the applications table, it would create potential redundancy since the same attribute might need to be derived from or overridden by the values in application-verifications. Keeping the field in application-verifications ensures the data is stored in its relevant context, avoiding the need to synchronize changes between multiple levels of the system. Otherwise we would need to check, if application-verifications are consistent with the "already verified" stamp. So for example what would happen, if the application "aleadyVerified" field is set to true, but the applications-verification mail was sent and clicked "rejected". Sure this should never happen, but we somehow would need to catch this case in the code and I think this is more complex, than checking for all application-verifications of one application.

3) If we plan in the future to extend this feature, it will be easier if the verification step is linked to the organization, than the whole application. E.g. for the feature: Someone works 5 hours at a pre-verified Sportclub, gets an E-Mail: "please add your second volutarty work". Or if we want to add checks, for which organizations verein360 is allowed to set the "automaticVerified" flag, it would be easier to check and handle, e.g. if they set two organziations, one they are allowed to set the automaticVerified flag and for one they are not, we could easily accept the first organization and send and please-verifiy-this-email for the second organization and handle this gracefully.

f1sh1918:
Alright thx for the detailed explanation. :)
I will start to review

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
```diff
@@ -88,16 +89,16 @@ class ApplicationHandler(
...
-            // check if api token is set and valid, if not throw unauthorized exception
-            // Will be done in #1790
-            throw UnauthorizedException()
+        val isAlreadyVerifiedList =
+            application.applicationDetails.blueCardEntitlement?.workAtOrganizationsEntitlement?.list?.map { it.isAlreadyVerified }
+                ?: emptyList()
+        return when {
+            isAlreadyVerifiedList.all { it == false || it == null } -> false
+            isAlreadyVerifiedList.all { it == true } -> {
+                throw UnauthorizedException()
```
f1sh1918, line 98: Since you wrote in the description to test with `alreadyVerified= true` to test if `VEREIN360` is set.
If i try this i end up here, since its currently "blocked"

f1sh1918 (commented):
The code looks fine to me but i couldn't test if the external source will be set to "Verein360". I added a comment

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
```diff
@@ -88,16 +89,16 @@ class ApplicationHandler(
     }
 
     fun isValidPreVerifiedApplication(): Boolean {
-        val isAlreadyVerifiedSet =
-            application.applicationDetails.blueCardEntitlement?.workAtOrganizationsEntitlement?.list?.any {
-                it.isAlreadyVerified == true
-            } ?: false
-        if (isAlreadyVerifiedSet) {
-            // check if api token is set and valid, if not throw unauthorized exception
```
f1sh1918, line 96: this comment shouldn't be remobed since its not implemented here

ztefanie: You are right, but as the other PR is already open and based on this PR, i think it is not worth the effort to now readd this comment, rebase the other PR and then remove it there.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
ztefanie: You are right, but as I wanted to avoid that at any point we have this enpoint unprotected on main, i added this exeption. You can either remove the exception by commenting it out and test or you just do not test it here, but do all the testing here: https://github.com/digitalfabrik/entitlementcard/pull/1811/

f1sh1918 (approved):
Works fine. Nice 👍

seluianova:
Seems to be an issue:

1. create a normal application that requires confirmation from the organization
2. confirm it using the link from the email
--> check db: applicationverifcation.automaticSource is VEREIN360

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -151,6 +153,7 @@ object ApplicationRepository {
                 false
             } else {
                 applicationVerification.verifiedDate = Instant.now()
+                applicationVerification.automaticSource = ApplicationVerificationExternalSource.VEREIN360
```
seluianova, line 156: I think the source should be passed as param here, because this function is used for all verifications, not only for verein360

ztefanie: good catch, thanks. Fixed it.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
```diff
@@ -88,22 +90,25 @@ class ApplicationHandler(
...
+            isAlreadyVerifiedList.all { it == true } -> {
+                throw UnauthorizedException()
+            }
+            else -> throw BadRequestResponse("isAlreadyVerified must be the same for all entries")
         }
-        return false
     }
 
-    fun setApplicationVerificationToVerifiedNow(verificationEntities: List<ApplicationVerificationEntity>) {
+    fun setApplicationVerificationToPreVerifiedNow(verificationEntities: List<ApplicationVerificationEntity>) {
```
seluianova, line 105:
```suggestion
    fun setApplicationVerificationToPreVerified(verificationEntities: List<ApplicationVerificationEntity>) {
```
:upside_down_face: to make the name a bit shorter, not sure if 'now' is an important detail here

seluianova (approved):
thanks, looks good to me now!

Issue #1790, opened 2024-11-20
Verein360 application data endpoint protection


ztefanie:
- protect the verein360 endpoint with api token of this user
- send insufficient permissions error and log the failed attempts

blocked by: https://github.com/digitalfabrik/entitlementcard/issues/1624

Issue #1625, opened 2024-09-09
Show Verein360 icon in application overview


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a service user i want to get a quick overview if the application is from verein360

**Describe the solution you'd like**
- show an verein360 icon next to the application that have the `applicationType` value "Verein360"

f1sh1918:
One question from my side. Do we need a separate field in the application table like "Source" or is it sufficient to use the "applicationType"
I mean for the display of the icon in the frontend it should be sufficient but if we may want to use it for statistics or sth similar it might be easier to also save information in a "source" field or sth

f1sh1918:
@dkehne
We need a particular icon that we should display. It would be good to have a similar format
<img width="119" alt="image" src="https://github.com/user-attachments/assets/5b0dcf50-d850-42ce-9fb4-30116f81cfc0">

ztefanie:
I think this should definitely be a new separate "source" field. As if we reuse the "ApplicationType" field we cannot  save if an application is an first or renewal application. And it is just something completely different.

PR #1811, opened 2024-12-03
1790: Verein360 endpoint protection

ztefanie:
### Short description
Finalize the verein360 endpoint

### Proposed changes
- protect the applications endpoint correctly
- show verein360 in digitale druckerei

### Side effects
- May break the normal application flow

### Testing
Create an api token
create applications with and without the api token
create applications with and without the pre-verified flag set
create applications with one or multiple organizations

### Resolved issues
Fixes: #1790 #1625


backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/Utils.kt
```diff
@@ -0,0 +1,35 @@
+package app.ehrenamtskarte.backend.common.webservice
```
seluianova, line 1: :upside_down_face: I would suggest moving this to the auth package

seluianova:
Not directly related to the current PR but I have to ask:
The [endpoint documentation](https://github.com/digitalfabrik/entitlementcard/blob/main/docs/verein360-enpoint.md)  contains some requirements to the acceptable values, but some of them we don't check.
For example:
`"applicationType": "FIRST_APPLICATION",         // Must be FIRST_APPLICATION`
But I can submit RENEWAL_APPLICATION, and it will be accepted.
And for some values we have checks, but we return a 500 error with no meaningful description.
So I wonder if we actually need some error handling here?

seluianova:
:question: and one more general question:
when verein360 application is submitted, the emails to the organisation ("Bestätigung notwendig") and to the user ("Antrag erfolgreich eingereicht") are sent, but I feel like it's not actually needed, isn't it?

f1sh1918:
> ❓ and one more general question: when verein360 application is submitted, the emails to the organisation ("Bestätigung notwendig") and to the user ("Antrag erfolgreich eingereicht") are sent, but I feel like it's not actually needed, isn't it? At least when isAlreadyVerified: true

Yes i think we decided on not sending mails in this case. For organizations it makes no sense and for the users we would have to adjust the texts because they not really submitted the application

ztefanie:
> ❓ and one more general question: when verein360 application is submitted, the emails to the organisation ("Bestätigung notwendig") and to the user ("Antrag erfolgreich eingereicht") are sent, but I feel like it's not actually needed, isn't it? At least when isAlreadyVerified: true

You are right, this was a bug. I fixed it.

ztefanie:
> ❓ Not directly related to the current PR but I have to ask somewhere: The [endpoint documentation](https://github.com/digitalfabrik/entitlementcard/blob/main/docs/verein360-enpoint.md) contains some requirements to the acceptable values, but some of them we don't check. For example: `"applicationType": "FIRST_APPLICATION", // Must be FIRST_APPLICATION` But I can submit RENEWAL_APPLICATION, and it will be accepted. And for some values we have checks, but we return a 500 error with no meaningful description. For example, if I send `"givenInformationIsCorrectAndComplete": false, // Must be true` So I wonder if we actually need some user-friendly error handling here? (as a separate task maybe)

Ok good point. I would split this:
1) I created a new issue for general validation for this endpoint: https://github.com/digitalfabrik/entitlementcard/issues/1816
2) Validation that is specific for verein360 pre-verified requests: Will add this to this PR.

seluianova:
pls update the branch, I will re-test all together :)

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -113,14 +114,35 @@ const RightElement = ({ jsonField, application }: RightElementProps): ReactEleme
...
+        Array.isArray(entitlement.value) &&
+        entitlement.value.some(
+          (organizationField: GeneralJsonField) =>
+            organizationField.name === 'isAlreadyVerified' && organizationField.value === true
+        )
+    )
+    return isAlreadyVerified
+  }
+
+  let quickIndicator
```
f1sh1918, line 133: :wrench:
I think redeclaring variable with let should be avoid as far as possible and i think the redundant declaration of the `<PreVerifiedQuickIndicator/>`  react component could be avoided by first checking if `isPreVerified()` even only a different type will be passed to the same component

sth. like this look cleaner to me
```js
const isPreVerified = isJuleicaEntitlementType() || isPreVerifiedByOrganization()

  return (
    <RightElementContainer>
      {!!application.note && application.note.trim() && <Icon icon='annotation' intent='none' />}
      {isPreVerified ? (
        <PreVerifiedQuickIndicator
          type={
            isJuleicaEntitlementType() ? PreVerifiedQuickIndicatorType.Juleica : PreVerifiedQuickIndicatorType.Verein360
          }
        />
      ) : (
        <VerificationsQuickIndicator verifications={application.verifications} />
      )}
    </RightElementContainer>
  )
```

the function above could be renamed to `isVerifiedByOrganization` to make it clearer. Since juleica is also preVerified but not by organization

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -113,14 +114,35 @@ const RightElement = ({ jsonField, application }: RightElementProps): ReactEleme
     return !!blueCardJuleicaEntitlement
   }
 
+  const isPreVerified = (): boolean => {
```
f1sh1918, line 117:
```suggestion
  const isPreVerifiedByOrganization() = (): boolean => {
```

administration/src/util/translations/de.json
```diff
@@ -41,6 +41,7 @@
     "responsibility": "Tätigkeit",
     "amountOfWork": "Arbeitsstunden pro Woche (Durchschnitt)",
     "workSinceDate": "Tätig seit",
+    "isAlreadyVerified": "Die Angaben wurden bereits durch den zuständigen Verein bei Beantragung verifiziert",
```
f1sh1918, line 44: ❓ i couldn't find the place where this translation is used
Maybe in the `VerificationsView` below the headline? Not sure if it fits in the tooltip in the `PreverifiedQuickIndicator`

f1sh1918 (commented):
Works fine! Nicely done!

One thing i think we have to handle somehow
If i send an empty string for any value like `forename` i get a server response of

```js
{
  "title": "Server Error",
  "status": 500,
  "type": "https://javalin.io/documentation#internalservererrorresponse",
  "details": {}
}
```

And the api throws internally
```
Caused by: app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidJsonException: Exception for GraphQL error INVALID_JSON was thrown.
```

Just added some minor comments about code style in react

ztefanie:
> Works fine! Nicely done!
> 
> One thing i think we have to handle somehow If i send an empty string for any value like `forename` i get a server response of
> 
> ```js
> {
>   "title": "Server Error",
>   "status": 500,
>   "type": "https://javalin.io/documentation#internalservererrorresponse",
>   "details": {}
> }
> ```
> 
> And the api throws internally
> 
> ```
> Caused by: app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidJsonException: Exception for GraphQL error INVALID_JSON was thrown.
> ```
> 
> Just added some minor comments about code style in react

ztefanie:
> One thing i think we have to handle somehow
> If i send an empty string for any value like `forename` i get a server response of

This will be handled here, right? https://github.com/digitalfabrik/entitlementcard/issues/1816
or is there anything additional you noticed that needs to be fixed? I added your example to the issue description.

administration/src/util/translations/de.json
ztefanie: it is in the application detail view in digitale druckerei -> eingegangende Anträge -> expand details of the application

f1sh1918:
> > One thing i think we have to handle somehow
> > If i send an empty string for any value like `forename` i get a server response of
> 
> This will be handled here, right? #1816 or is there anything additional you noticed that needs to be fixed? I added your example to the issue description.

Sorry i just realized that there is already an issue! I think thats fine

administration/src/util/translations/de.json
f1sh1918: okay sorry not sure why i didn't find this via search. thx for the hint

f1sh1918 (commented):
Tested on chrome. Works fine. 👍 

I'm okay with merging it even i think the small proposals for the quick indicator would improve the code

f1sh1918 (approved):

ztefanie:
> 
> I'm okay with merging it even i think the small proposals for the quick indicator would improve the code

Added your suggestion

seluianova (approved):
lgtm! 👍

seluianova:
:thought_balloon: maybe would be nice to have more or less similar style for the icons. I assume we will have more of them in the future
image

ztefanie:
> 💭 maybe would be nice to have more or less similar style for the icons. I assume we will have more of them in the future. shall we create a ui/ux ticket for that? 

Good point. Yes may be an option. I will create ticket.

Issue #1817, opened 2024-12-04
Allow certain special chars in names


Docmale:
**Is your feature request related to a problem? Please describe.**
Special Character are not allwed in names, hyphens ("-") and ("'") should be allowed since double names or british names are common.

PR #1819, opened 2024-12-04
1817: Allow certain special characters

f1sh1918:
### Short description
Since some names may contain "-", "." or "'" we have to allow some certain special chars.
If you find anything thats missing here please let me know

### Proposed changes

- remove some special chars from the not allowed regex pattern
- add some missing special characters
- added some additional tests

### Side effects

- none

### Testing

- Create a koblenz card and test some names that include allowed special chars
- Test some names like "Dr. Vorname Nachname" , "Mac O'Connor", "Hans-Wilhelm Müller-Wohlfahrt"

### Resolved issues

Fixes: #1817

### DIFF
--- a/administration/src/cards/Card.test.ts
+++ b/administration/src/cards/Card.test.ts
@@ -140,13 +140,23 @@ describe('Card', () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


ztefanie (approved):

Issue #1821, opened 2024-12-04
Fix wrong map attribution


Docmale:
**Describe the bug**
In Map Data on map (c) LBE Baern shows up in Nürnberg and Koblenz App

**Expected behavior**
This should be removed or replaced

f1sh1918:
I think this information should be defined in the `buildConfig` of each project

PR #1823, opened 2024-12-04
1821: Fix wrong map attribution

f1sh1918:
### Short description

Currently the map attribution was hardcoded (with LBE Bayern) for all whitelabels but the map attribution depends on the particular project.

### Proposed changes

- show different map attribution per project

### Side effects

- none

### Testing

- Open the map attribution for each project (left corner of map)
- Check the text and if the link is working

### Resolved issues

Fixes: #1821

### DIFF
--- a/frontend/build-configs/koblenz/index.ts
+++ b/frontend/build-configs/koblenz/index.ts
@@ -34,6 +34,10 @@ export const koblenzCommon: CommonBuildConfigType = {
         showcase: "https://api.entitlementcard.app/project/showcase.entitlementcard.app/map",
         local: "http://localhost:8000/project/koblenz.sozialpass.app/map",
     },
+    mapAttribution: {
+        text: 'Stadt Koblenz',
+        url: 'https://koblenz.de/',
+    },
     mapInitialCoordinatesLat: 50.3575886,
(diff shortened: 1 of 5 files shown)


ztefanie (approved):
Looks good, not tested

frontend/build-configs/koblenz/index.ts
```diff
@@ -34,6 +34,10 @@ export const koblenzCommon: CommonBuildConfigType = {
         showcase: "https://api.entitlementcard.app/project/showcase.entitlementcard.app/map",
         local: "http://localhost:8000/project/koblenz.sozialpass.app/map",
     },
+    mapAttribution: {
+        text: 'Stadt Koblenz',
+        url: 'https://koblenz.de/',
+    },
```
steffenkleinle, line 40: :question: Why do we even show that as map attribution? Its not like they created the map or anything, right?

steffenkleinle (approved):

frontend/build-configs/koblenz/index.ts
f1sh1918: Lbe even not but they are responsable for the content

Issue #1824, opened 2024-12-04
Card extension jumps to old card after activation


f1sh1918:
**Describe the bug**
If the user activated a card and then extend the card and activate the new one after shortly showing the new card the old "invalid" card will be shown. In general we should add some kind of hint that this card can be deleted and the user should be somehow informed that it can be swiped between cards (maybe separated issue)

**To Reproduce**
1. Create and activate a card
2. Send the enddate after the current endDate in the database
3. Extend the card by creating a new one and activating it
4. Card will be activated correctly but then jumps to the old card

**Expected behavior**
The view doesn't jump to the old card

PR #1828, opened 2024-12-05
1824: Card extension jumps to old card

f1sh1918:
### Short description

When a user activates a second card via deeplink, the app directly moves to the first card after activation. This may confuse the user

### Proposed changes

- introduce a `cardIndex` for Home route be able to initialize the identification page  (child of home route) with the particular card index
- add the card index to the route push when activating card via deeplink

### Note
Since we have to wait until the identification page widget is initialized its not possible to just execute a `moveTo` function after activating the deeplink, so i think thats the best way to do.

### Side effects

- don't think so, worst thing could happen to initialize at cardIndex 0

### Testing

1. Activate a card via deeplink
2. Activate a second card via deeplink
3. Check the after activation the latest card will be displayed

### Resolved issues

Fixes: #1824

### DIFF
--- a/frontend/lib/activation/deeplink_activation.dart
+++ b/frontend/lib/activation/deeplink_activation.dart
@@ -103,39 +103,41 @@ class _DeepLinkActivationState extends State<DeepLinkActivation> {
                 child: Column(mainAxisSize: MainAxisSize.min, children: [
                   if (_state == _State.waiting) _WarningText(status, userCodeModel),
                   ElevatedButton.icon(
-                    onPressed:
-                        activationCode != null && _state == _State.waiting && status == DeepLinkActivationStatus.valid
-                            ? () async {
+                    onPressed: activationCode != null &&
+                            _state == _State.waiting &&
+                            status == DeepLinkActivationStatus.valid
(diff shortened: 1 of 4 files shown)


frontend/lib/activation/deeplink_activation.dart
```diff
@@ -103,39 +103,41 @@ class _DeepLinkActivationState extends State<DeepLinkActivation> {
...
+                            status == DeepLinkActivationStatus.valid
+                        ? () async {
+                            setState(() {
+                              _state = _State.loading;
+                            });
+                            try {
+                              final activated = await activateCard(context, activationCode);
+                              if (!context.mounted) return;
+                              if (activated) {
+                                final cardAmount = Provider.of<UserCodeModel>(context, listen: false).userCodes.length;
```
f1sh1918, line 117: Just added/adjusted these two lines (adding the card amount as cardIndex)

ztefanie (approved):
Tested works as expected. Thanks for fixing

steffenkleinle (approved):
Well done, tested on android, works as expected.

PR #1829, opened 2024-12-05
1827: Surnames single digit

f1sh1918:
### Short description

Since in some countries exists name with only one digit, we should allow them

### Proposed changes

- reduce minimum name character count to `>0`

### Side effects

- none

### Testing

1. Go to /erstellen
2. Type in name with just one single character like "A F" or "Karla K"
3. Name shouldn't be marked as invalid anymore

### Resolved issues

Fixes: #1827

### DIFF
--- a/administration/src/cards/Card.test.ts
+++ b/administration/src/cards/Card.test.ts
@@ -160,7 +160,7 @@ describe('Card', () => {
     }
   )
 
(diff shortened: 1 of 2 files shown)


steffenkleinle (approved):

ztefanie (approved):
Not tested as I am quite sure this works :D

PR #1831, opened 2024-12-09
Adjust verein 360 docs

ztefanie:
### Short description

Just added the endpoint urls


f1sh1918 (approved):
Not sure if you also want to add the graphiql link so they can finde the documentation

https://api.staging.entitlementcard.app/graphiql/ for testing

But for me its also fine without

steffenkleinle (approved):

Issue #1833, opened 2024-12-09
Change filename of PDF output files (add Berechtigungskarte_)


dkehne:
**Is your feature request related to a problem? Please describe.**
We have changed the filname output (activation pdf) with #1778. This causes problems in Nuremberg.

**Describe the solution you'd like**
Related to #1778 we need to add the word "Berechtigungskarte_" in front of the current file name content to not break automation processes in Nuremberg.

PR #1834, opened 2024-12-09
1833: Fix filename for nuernberg

ztefanie:
### Short description

Add prefix to filename as this breaks automation on nuernberg side

Fixes: #1833


administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -173,10 +173,11 @@ const useCardGenerator = (region: Region): UseCardGeneratorReturn => {
 
   const generateCardsPdf = useCallback(
     async (applicationIdToMarkAsProcessed?: number) => {
+      // "Berechtigungskarte_" prefix needs to always in the filename to ensure Nuernberg automation will not break
```
f1sh1918:
```suggestion
      // "Berechtigungskarte_" prefix needs to always be in the filename to ensure Nuernberg automation will not break
```

f1sh1918 (approved):
small issue in the commenting, rest looks fine.

Issue #1835, opened 2024-12-10
Accepting store list item not readable


f1sh1918:
**Describe the bug**
If you want to read the short description of the last store entry of the list it is hidden behind the sort FAB. 

**To Reproduce**
1. Open the app and go to search
2. Scroll down the list to the last item
3. Check that the button hides the short description

**Expected behavior**
There should be some space after the last item to be able to read the short description

**Screenshots**
<img src="https://github.com/user-attachments/assets/b196dcae-9db8-45ca-920e-ffea4da80b4d" width="480">

PR #1838, opened 2024-12-10
1835: Fix last store item hidden

f1sh1918:
### Short description
Last store item is hidden in list because its behind the sorting FAB button

### Proposed changes

- add a sized box as spacer

### Side effects

- none

### Testing

Open search page and scroll down to last item. Check if its readable

### Resolved issues

Fixes: #1835


ztefanie (approved):
Tested on Android, works as expected :rocket:

steffenkleinle (approved):

Issue #1840, opened 2024-12-17
fix broken activity log


ztefanie:
**Describe the bug**
When going to `https://staging.nuernberg.sozialpass.app/user-settings` the view is broken.

**To Reproduce**
1. Go to staging
2. Create a card
3. go to user settings
4. See error:
![image](https://github.com/user-attachments/assets/6e8fcf5d-a647-4e90-8b25-8e9e98260511)

**Expected behavior**
Activity log is displayed correctly

PR #1844, opened 2024-12-18
1840: Properly serialize activity log and fix it

steffenkleinle:
### Short description

The activity log was broken for a while due to serializing not happening properly. This introduces proper typing and serializing.

### Proposed changes

- Add `fromSerialized` and `serialize` methods to each extension and for cards
- Properly serialize activity log and use iso8601

### Side effects

- Old activity logs break the serializing (but this should not happen as they are only session based)

### Testing

See issue.

### Resolved issues

Fixes: #1840

### DIFF
--- a/administration/src/cards/extensions/AddressFieldExtensions.ts
+++ b/administration/src/cards/extensions/AddressFieldExtensions.ts
@@ -24,6 +24,8 @@ const getAddressFieldExtension = <T extends AddressFieldExtension>(
   isValid: () => true,
   fromString: (value: string) => ({ [name]: value } as AddressFieldExtensionState<T>),
   toString: (state): string => state[name],
+  fromSerialized: (value: string) => ({ [name]: value } as AddressFieldExtensionState<T>),
+  serialize: (state): string => state[name],
 })
 
 export const AddressLine1Extension = getAddressFieldExtension(ADDRESS_LINE_1_EXTENSION)
--- a/administration/src/bp-modules/cards/hooks/useCardGenerator.ts
+++ b/administration/src/bp-modules/cards/hooks/useCardGenerator.ts
@@ -15,7 +15,7 @@ import getDeepLinkFromQrCode from '../../../util/getDeepLinkFromQrCode'
 import { updateArrayItem } from '../../../util/helper'
 import normalizeString from '../../../util/normalizeString'
 import { useAppToaster } from '../../AppToaster'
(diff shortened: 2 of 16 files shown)


steffenkleinle:
The additional `fromSerialized` and `serialize` methods for extensions are only necessary if user display strings are different form serialized values. Currently, this is only the case for dates (dd.MM.yyyy vs yyyy-MM-dd (ISO8601)). Obviously, we could also serialize dates as dd.MM.yyyy but I think in the long term this is better and provides more flexibility. Let me know if you disagree.

administration/src/cards/extensions/AddressFieldExtensions.ts
```diff
@@ -24,6 +24,8 @@ const getAddressFieldExtension = <T extends AddressFieldExtension>(
   isValid: () => true,
   fromString: (value: string) => ({ [name]: value } as AddressFieldExtensionState<T>),
   toString: (state): string => state[name],
+  fromSerialized: (value: string) => ({ [name]: value } as AddressFieldExtensionState<T>),
```
f1sh1918, line 27: i think you could also predefine `toString()` and `fromString()` and reuse it like in the other extensions

f1sh1918 (approved):
General note: This should be deployed as a hotfix. so please do not merge anything from main

One small improvement. Rest looks good to me! 👍 
Thx for the quick fix

Tested on chrome.
Also tested card creation for koblenz, nuernberg and bayern (also bulks)

administration/src/cards/extensions/AddressFieldExtensions.ts
f1sh1918: Tried to adjust that but couldn't get it work easily. so nevermind.
For me its good to go

bahaaTuffaha (approved):
Tested on Brave and Firefox. At nuernberg
It works great 👌
Interesting use of `reduce` at  serializeCard/deserializeCard .
- Unrelated point: BavariaCardTypeExtension is weird TBH are we inserting a component within an object?

steffenkleinle:
>     * Unrelated point: BavariaCardTypeExtension is weird TBH are we inserting a component within an object?

Yes, kind of. We set a component as object property. In the end, a component is just a render function that returns a react element in any case. That way, we can have generic components for each extension (without needing to know anything about the actual implementation/look).

f1sh1918:
@steffenkleinle since we do a new beta this week and no one recognized the issue yet, i would just merge it and include it in the next regular version

PR #1845, opened 2024-12-20
Disable backup of app data and delete secure storage if a wrong secret is used on Android

michael-markl:
### Proposed changes

-
-

### Side effects

-
-

### Resolved issues

Maybe fixes: #1265


frontend/lib/identification/user_code_store.dart
```diff
@@ -54,11 +41,31 @@ class UserCodeStore {
...
 }
+
+/// This function deletes the secure storage in the situation that the app data was backed up (e.g. on a new device),
+/// but the secret is no longer available.
+/// Upstream issue: https://github.com/juliansteenbakker/flutter_secure_storage/issues/354
+Future<String?> _safeRead(FlutterSecureStorage storage, String key) async {
+  try {
+    return await storage.read(key: key);
+  } on PlatformException catch (e, stackTrace) {
+    if (Platform.isAndroid && e.toString().contains('javax.crypto.BadPaddingException')) {
```
maxammann, line 60: The contains check might be a bit too unreliable. It is a bit weird that it throws a padding exception because the passing should remain valid actually.

Is there an issue with just deletingAll on any error?

frontend/android/app/src/main/res/xml/data_extraction_rules.xml
maxammann, line 1: What is the purpose of these files?

michael-markl: It declares which files may be backed up for Android >= 12. The other xml file is for Android <= 11. And the flag in AndroidManifest is just for cloud backups (on many devices) :D very messy stuff...

frontend/lib/identification/user_code_store.dart
michael-markl: > It is a bit weird that it throws a padding exception because the passing should remain valid actually.

Yeah it's a weird error message; but it also says BAD_DECRYPT at some point. Maybe the padding somehow depends on the secret?

> Is there an issue with just deletingAll on any error?

I did not want to do that, as there might potentially be other errors and I am fearing data loss. If there are other errors, we at least get reported via Sentry and we can then include them here.

michael-markl: I'll leave it like this for now, as I succesfully reproduced it, and it worked with this filter :)

michael-markl:
@f1sh1918 Should we make this a hotfix?

f1sh1918:
> @f1sh1918 Should we make this a hotfix?

In my opinion we can make a normal release next week including the two or 3 other bugfixes, f.e. card deletion is also not working properly directly after deep link activation which should also be fixed very soon.
And there is currently a workaround by deleting app data.

f1sh1918:
@michael-markl i tested scenario a) and it works fine
I'm not really sure what the difference between bullet point 1 of scenario a) and b) is. If i install from current main i can't install different version
Probably you meant in scenario b) to directly install your fixed version from this branch here? @michael-markl ?

f1sh1918 (approved):
Thx for this great fix 👍 
I think i tested all scenarios and it seem to work fine.

What i realized is that it doesn't work with the staging switch for me so i used production but it should even be possible with local build, because on staging the data is always cleared if you update or reinstall but thats okay

michael-markl:
>  Probably you meant in scenario b) to directly install your fixed version from this branch here?

Yes, that's what I meant!

> What i realized is that it doesn't work with the staging switch for me so i used production but it should even be possible with local build, because on staging the data is always cleared if you update or reinstall but thats okay

For me, if I install the apk from main, switch to staging, and then install the apk from this branch, I am still on staging. I never tried the local backend server.

Also: It would be great, if someone could test the fix on iPhone, just to make sure that I didn't break anything there.
Basically: Install the old app, activate a card, install the app from this branch, make sure that the card is still there.

seluianova (approved):
Tested 3 scenarios & device reboot, haven't found any issues :+1:

f1sh1918:
i will test on ios @michael-markl so please do not merge it yet

f1sh1918:
it works fine on ios @michael-markl  you can merge

Issue #1843, opened 2024-12-17
Label/Translation missing in Bavarian Volunteer Card


dkehne:
**Describe the bug**
One Label is missing in the applicaton details. goldenCardWorkAtOrganizationsEntitlement
See Screenshot: 

![Screenshot 2024-12-17 174203](https://github.com/user-attachments/assets/fc182c71-345b-45f2-8f02-f56cd1357640)

PR #1855, opened 2025-01-08
1843: Missing application labels

f1sh1918:
### Short description

Some application labels were not resolved properly

### Proposed changes

- rename i18n keys according to how they will stored in the database to be able to resolve them properly,
- adjust these keys in the form states where they will be set
- fix key of `goldenCardHonoredByMinisterPresidentEntitlement`, 
- consider parents for resolving keys for attachments, 
- fix wrong label (#1852)

### Side effects

- Maybe some attachment labels may be affected so please check all labels of an application card properly

### Testing

1. Create applications with the different requirements, all possible for the blue card and all possible for the golden card
2. Open the application list and check if all application labels were resolved correctly, due to the `de.json`, there were some keys like `certificate` resolved wrong because the parents wasn't used

### Resolved issues

Fixes: #1843 
Fixes: #1852

### DIFF
--- a/administration/src/bp-modules/applications/JsonFieldArray.tsx
+++ b/administration/src/bp-modules/applications/JsonFieldArray.tsx
@@ -60,7 +60,7 @@ const JsonFieldArray = ({
       'organizationContact',
       'organization',
       'volunteerServiceEntitlement',
-      'honoredByMinisterPresidentEntitlement',
+      'goldenCardHonoredByMinisterPresidentEntitlement',
     ].includes(jsonField.name)
       ? `${jsonField.name}.title`
       : jsonField.name
@@ -68,7 +68,16 @@ const JsonFieldArray = ({
   const children = jsonField.value.map((jsonFieldIt, index: number) => (
     <JsonFieldView
       jsonField={jsonFieldIt}
-      parentName={['organizationContact', 'organization'].includes(jsonField.name) ? jsonField.name : undefined}
+      parentName={
+        [
+          'organizationContact',
+          'organization',
+          'volunteerServiceEntitlement',
+          'goldenCardHonoredByMinisterPresidentEntitlement',
+        ].includes(jsonField.name)
+          ? jsonField.name
+          : undefined
+      }
       baseUrl={baseUrl}
       // This is the best key we have as jsonField.name is not unique
       // eslint-disable-next-line react/no-array-index-key
--- a/administration/src/bp-modules/applications/JsonFieldElemental.tsx
+++ b/administration/src/bp-modules/applications/JsonFieldElemental.tsx
@@ -26,7 +26,7 @@ const PrintOnlySpan = styled.span`
   }
 `
 
-// Some fields name are equals therefore the parents is needed for resolving the label
+// Some field names are equal therefore the parents are needed for resolving the label
 const getTranslationKey = (fieldName: string, parentName?: string) =>
   parentName ? `${parentName}.${fieldName}` : fieldName
 
--- a/administration/src/bp-modules/applications/JsonFieldElemental.tsx
+++ b/administration/src/bp-modules/applications/JsonFieldElemental.tsx
@@ -26,8 +26,12 @@ const PrintOnlySpan = styled.span`
   }
 `
 
+// Some fields name are equals therefore the parents is needed for resolving the label
+const getTranslationKey = (fieldName: string, parentName?: string) =>
+  parentName ? `${parentName}.${fieldName}` : fieldName
+
 const JsonFieldAttachment = memo(
-  ({ jsonField, baseUrl, attachmentAccessible }: JsonFieldViewProps<JsonField<'Attachment'>>) => {
+  ({ jsonField, baseUrl, attachmentAccessible, parentName }: JsonFieldViewProps<JsonField<'Attachment'>>) => {
     const appToaster = useAppToaster()
     const token = useContext(AuthContext).data?.token
     const { t } = useTranslation('application')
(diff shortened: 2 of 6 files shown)


administration/src/bp-modules/applications/JsonFieldView.tsx
```diff
@@ -16,7 +16,7 @@ export type JsonFieldValueByType = {
   String: string
   Number: number
   Boolean: boolean
-  Attachment: { fileIndex: number }
+  Attachment: { fileIndex: number; parentName?: string }
```
michael-markl, line 19: I don't see why `parentName`  should be added here...

michael-markl (approved):
The architecture of these translations keys seems quite fragile to me...

I haven't checked if you caught every translation key. I also did not test anything.

administration/src/util/translations/de.json
```diff
@@ -20,11 +20,16 @@
...
+    "blueCardMilitaryReserveEntitlement": "Ich habe in den vergangenen zwei Kalenderjahren als Reservist regelmäßig aktiven Wehrdienst in der Bundeswehr geleistet, indem ich insgesamt mindestens 40 Tage Reservisten-Dienstleistung erbracht habe oder ständige:r Angehörige:r eines Bezirks- oder Kreisverbindungskommandos war",
+    "blueCardWorkAtOrganizationsEntitlement": "Ich engagiere mich ehrenamtlich seit mindestens zwei Jahren freiwillig mindestens fünf Stunden pro Woche oder bei Projektarbeiten mindestens 250 Stunden jährlich",
     "juleicaNumber": "Kartennummer",
     "juleicaExpiration": "Karte gültig bis",
     "copyOfJuleicaFront": "Kopie der Karte (1)",
     "copyOfJuleicaBack": "Kopie der Karte (2)",
-    "blueCardWorkAtOrganizationsEntitlement": "Ich engagiere mich ehrenamtlich seit mindestens zwei Jahren freiwillig mindestens fünf Stunden pro Woche oder bei Projektarbeiten mindestens 250 Stunden jährlich",
+    "goldenCardWorkAtOrganizationsEntitlement": "Ich bin seit mindestens 25 Jahren mindestens 5 Stunden pro Woche oder 250 Stunden pro Jahr bei einem Verein oder einer Organisation ehrenamtlich tätig",
```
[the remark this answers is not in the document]
f1sh1918: The problem is that the keys were stored like that in the database (example)
I believe this happens in the `GoldenCardEntitlement.kt` 
If i change this structure i would also have to replace the keys in jsonField in the production database via migration or sth :/
```
{
          "name": "goldenCardHonoredByMinisterPresidentEntitlement",
          "type": "Array",
          "value": [
            {
              "name": "certificate",
              "type": "Attachment",
              "value": {
                "fileIndex": 0
              }
            }
          ]
        },
```
i don't see any other way now to resolve the other structure.
If you have any idea let me known @michael-markl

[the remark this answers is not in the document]
f1sh1918: Update: Okay i adjusted the keys that were used for the state (these keys will also be stored like that in the database, so we can properly resolve it again)
Removed the duplicates

f1sh1918:
@seluianova and @bahaaTuffaha this pr needs really proper testing, since we will have bad issues with wrong i18n keys in the database, if there is a mess. 
So i suggest to create all types of applications /golden and blue with all possible requirements options before checking out this pr.
Then checkout this branch and create all types of applications again to check if old applications work and new applications (resolve all i18n labels)

administration/src/bp-modules/applications/JsonFieldArray.tsx
```diff
@@ -60,15 +60,24 @@ const JsonFieldArray = ({
...
+      'goldenCardHonoredByMinisterPresidentEntitlement',
     ].includes(jsonField.name)
       ? `${jsonField.name}.title`
       : jsonField.name
 
   const children = jsonField.value.map((jsonFieldIt, index: number) => (
     <JsonFieldView
       jsonField={jsonFieldIt}
-      parentName={['organizationContact', 'organization'].includes(jsonField.name) ? jsonField.name : undefined}
+      parentName={
```
f1sh1918, line 71: since these keys are objects we need also the parent to resolve them properly

seluianova:
Duplicated ":"
image

administration/src/bp-modules/applications/JsonFieldElemental.tsx
```diff
@@ -26,8 +26,12 @@ const PrintOnlySpan = styled.span`
   }
 `
 
+// Some fields name are equals therefore the parents is needed for resolving the label
```
seluianova, line 29:
```suggestion
// Some field names are equal therefore the parents are needed for resolving the label
```

seluianova (approved):
I checked every application type (or I hope so)
Just a couple nitpicks from my side :+1:

administration/src/util/translations/de.json
```diff
@@ -20,11 +20,25 @@
     "wantsDigitalCard": "Ich beantrage eine digitale Ehrenamtskarte",
     "wantsPhysicalCard": "Ich beantrage eine physische Ehrenamtskarte",
     "blueCardJuleicaEntitlement": "Ich bin Inhaber:in einer JuLeiCa (Jugendleiter:in-Card)",
+    "blueCardWorkAtDepartmentEntitlement": "Ich bin aktiv in der Freiwilligen Feuerwehr mit abgeschlossener Truppmannausbildung bzw. abgeschlossenem Basis-Modul der Modularen Truppausbildung (MTA), oder im Katastrophenschutz oder im Rettungsdienst mit abgeschlossener Grundausbildung.",
```
seluianova, line 23: :upside_down_face: maybe we don't need the "." at the end of the entitlement descriptions.
or at least it should be the same for all items.
currently there is no "." in blueCardJuleicaEntitlement and goldenCardHonoredByMinisterPresidentEntitlement, but we have the "." for other items.

f1sh1918: well i think the requirements are entire sentences so they should have a dot when the user selects a requirement.
I think its not really needed to strip that. But you are right it should be equal for all requirements

Issue #1755, opened 2024-11-06
Make email addresses in the application object clickable


dkehne:
**Is your feature request related to a problem? Please describe.**
To speed up the process of making contact, the e-mail addresses should be clickable so that they do not have to be copied and pasted.

**Describe the solution you'd like**
The e-mail addresses in the application details should be clickable so that you can quickly send an e-mail to the applicant or organization.

<img width="791" alt="Screenshot 2024-11-06 182113" src="https://github.com/user-attachments/assets/723a2a7f-5b7f-4391-a1fd-3a797ecfe0fe">

f1sh1918:
unfortunately it looks like lots of labels are missing especially for golden card requirements

PR #1856, opened 2025-01-08
1755: clickable email addresses

bahaaTuffaha:
### Short description

The e-mail addresses should be clickable so that they do not have to be copied and pasted.

### Proposed changes

- Modifying `JsonFieldElemental` to check string if its a valid email address to show is as a `mail:to` link.
- `verificationHelper.tsx` having the regex for email verification function at `bp-modules/applications/utils`

### Side effects

- `JsonFieldElemental.tsx` modified.

### Testing
- Create an application for a specific region
- Login with regionAdmin or regionManager of that specific region
- Check application if mail addresses are clickable

Fixes: #1755

### DIFF
--- a/administration/src/bp-modules/EmailLink.tsx
+++ b/administration/src/bp-modules/EmailLink.tsx
@@ -1,6 +1,5 @@
 import React, { ReactElement } from 'react'
 
-import { isEmailValid } from './applications/utils/verificationHelper'
+const EmailLink = ({ email }: { email: string }): ReactElement => <a href={`mailto:${email}`}>{email}</a>
 
-export const EmailLink = (email: string): ReactElement =>
-  isEmailValid(email) ? <a href={`mailto:${email}`}> {email}</a> : <span>{email}</span>
+export default EmailLink
--- a/administration/src/bp-modules/EmailLink.tsx
+++ b/administration/src/bp-modules/EmailLink.tsx
@@ -1,0 +1,6 @@
+import React, { ReactElement } from 'react'
+
+import { isEmailValid } from './applications/utils/verificationHelper'
+
+export const EmailLink = (email: string): ReactElement =>
+  isEmailValid(email) ? <a href={`mailto:${email}`}> {email}</a> : <span>{email}</span>
--- a/administration/src/bp-modules/applications/JsonFieldElemental.tsx
+++ b/administration/src/bp-modules/applications/JsonFieldElemental.tsx
@@ -8,6 +8,7 @@ import downloadDataUri from '../../util/downloadDataUri'
 import { useAppToaster } from '../AppToaster'
 import { printAwareCss } from './ApplicationCard'
 import { GeneralJsonField, JsonField, JsonFieldViewProps } from './JsonFieldView'
+import { isEmailValid } from './utils/verificationHelper'
 
 const extensionByContentType = new Map([
   ['application/pdf', 'pdf'],
@@ -96,7 +97,12 @@ const JsonFieldElemental = ({
     case 'String':
       return (
         <p>
-          {t(getTranslationKey())}: {jsonField.value}
+          {t(getTranslationKey())}:
+          {isEmailValid(jsonField.value) ? (
+            <a href={`mailto:${jsonField.value}`}>{jsonField.value}</a>
+          ) : (
+            jsonField.value
+          )}
         </p>
       )
     case 'Date':
--- a/administration/src/bp-modules/applications/JsonFieldElemental.tsx
+++ b/administration/src/bp-modules/applications/JsonFieldElemental.tsx
@@ -99,7 +99,7 @@ const JsonFieldElemental = ({
         <p>
           {t(getTranslationKey())}:
           {isEmailValid(jsonField.value) ? (
-            <a href={`mailto:${jsonField.value}`}>{jsonField.value}</a>
+            <a href={`mailto:${jsonField.value}`}> {jsonField.value}</a>
           ) : (
             jsonField.value
           )}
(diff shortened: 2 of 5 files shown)


administration/src/bp-modules/applications/utils/verificationHelper.tsx
```diff
@@ -0,0 +1,4 @@
+export const isEmailValid = (email: string): boolean => {
```
f1sh1918, line 1:
```suggestion
export const isEmailValid = (email: string): boolean => /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(email)
```
🔧 Since you are not reusing the pattern, this one liner should be enough

administration/src/bp-modules/applications/utils/verificationHelper.tsx
```diff
@@ -0,0 +1,4 @@
...
+  const pattern = /^[^\s@]+@[^\s@]+\.[^\s@]+$/
+  return pattern.test(email)
+}
```
f1sh1918, line 4: ❌ Please rename the file to `.ts` since you are not using any tsx syntax like templating

f1sh1918 (commented):
Congratulation to your first PR! Very nice work! 👍 
Works like a charm. Tested on chrome

Just found some small improvements, i commented

One thing that was probably not mentioned in the issue was that in the organization section (screenshot) the mail address should also be clickable since the service workers may contact the organization for further questions. You can either add it in this pr or create a new issue for it.
image

administration/src/bp-modules/applications/JsonFieldElemental.tsx
```diff
@@ -96,7 +97,12 @@ const JsonFieldElemental = ({
     case 'String':
       return (
         <p>
-          {t(getTranslationKey())}: {jsonField.value}
+          {t(getTranslationKey())}:
+          {isEmailValid(jsonField.value) ? (
+            <a href={`mailto:${jsonField.value}`}>{jsonField.value}</a>
```
f1sh1918, line 102:
```suggestion
            <a href={`mailto:${jsonField.value}`}> {jsonField.value}</a>
```
i think there is an empty space missing after the label and before the mail address

bahaaTuffaha: I noticed it but I kept the same format as possible...

administration/src/bp-modules/applications/JsonFieldElemental.tsx
```diff
@@ -96,7 +97,7 @@ const JsonFieldElemental = ({
     case 'String':
       return (
         <p>
-          {t(getTranslationKey())}: {jsonField.value}
+          {t(getTranslationKey())}:{EmailLink(jsonField.value)}
```
seluianova, line 100: :question: it seems a bit confusing to me that we return an “EmailLink” component for each field with type String.
But I don't have a better suggestion though. Maybe someone else could offer some advice here.

administration/src/bp-modules/applications/JsonFieldElemental.tsx
```diff
@@ -96,7 +97,7 @@ const JsonFieldElemental = ({
     case 'String':
       return (
         <p>
-          {t(getTranslationKey())}: {jsonField.value}
```
seluianova, line 99: :upside_down_face: it may have been better to add a space here rather than inside EmailLink. Because it looks like a part of the link currently
image

seluianova (approved):
Congrats with your first PR! 👍🏻
As far as I tested looks good, just left some optional comments

administration/src/bp-modules/applications/JsonFieldElemental.tsx
seluianova: Maybe introduce new type of the field - email

f1sh1918: i think a new field type is not that easy to implement since this type is stored in the database and has to backward compatible.

According the naming i think i agree with tory.
I think its cleaner to create a "MailLink" component that is a clickable anchor and don't put that `isEMailValid` logic inside of the component but keep it dump.
So this component in any case delivers what it says
You can keep that logic in the `JsonFieldElemental` component

administration/src/bp-modules/EmailLink.tsx
```diff
@@ -0,0 +1,6 @@
+import React, { ReactElement } from 'react'
+
+import { isEmailValid } from './applications/utils/verificationHelper'
+
+export const EmailLink = (email: string): ReactElement =>
```
f1sh1918, line 5: please convert this in a proper react functional component with a default export as we do for other functional components

administration/src/bp-modules/applications/VerificationsView.tsx
```diff
@@ -99,7 +100,7 @@ const VerificationsView = ({ verifications }: { verifications: Application['veri
                 </tr>
                 <tr>
                   <td>Email:</td>
-                  <td>{verification.contactEmailAddress}</td>
+                  <td>{EmailLink(verification.contactEmailAddress)}</td>
```
f1sh1918, line 103: Please us here proper react tsx syntax 
<EmailLink email={...} />

f1sh1918 (changes requested):
Your last commit contains changes that violates the react syntax and our coding conventions :)
Please check my comments

administration/src/bp-modules/EmailLink.tsx
bahaaTuffaha: Nice catch ...

f1sh1918 (approved):
Nicely done

Issue #1804, opened 2024-11-27
Add "Hint"-button to self-service-portal birth date input field


dkehne:
**Is your feature request related to a problem? Please describe.**
When entering the birth date in our self service portal there should be a hint: "Bei Minderjährigen unter 16 Jahren darf der KoblenzPass nur mit Einverständnis der Erziehungsberechtigten abgerufen werden."

**Describe the solution you'd like**
Maybe a hint-button or "question mark" symbol next to the input field?

**Describe alternatives you've considered**

**Additional context**
is not time critical - can be implemented in first part of 2025

PR #1858, opened 2025-01-14
1804: Adding hint for birth date input for koblenz

bahaaTuffaha:
### Short description

When entering the birth date in our self service portal there should be a hint: "Bei Minderjährigen unter 16 Jahren darf der KoblenzPass nur mit Einverständnis der Erziehungsberechtigten abgerufen werden."

### Proposed changes

- Added an optional `sideComponent` prop for customDatePicker.
- I used HelpOutline icon wrapped with Tool Tip from bluePrint.

### Side effects

- CustomDatePicker.tsx, BirthdayExtension.tsx

### Testing

- I just test it from `http://localhost:3000/erstellen`

Fixes: #1804


administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -45,6 +61,17 @@ const BirthdayForm = ({
         isValid={isValid || !showErrorMessage}
         maxDate={new Date()}
         disableFuture
+        sideComponent={
+          <>
+            {projectConfig.projectId.includes('koblenz') && (
```
f1sh1918, line 66: this is not a very secure check, i would prefer to add a flag in `projectConfig` like
`showBirthdayMinorHint` or sth similar and set it true or false

f1sh1918 (commented):
The tooltip is slightly misaligned (sorry for the nit-pick)
image

this could be fixed by setting the width and the margin on the tooltip component and not on the icon
image

administration/src/cards/extensions/extensions.tsx
```diff
@@ -53,7 +53,6 @@ const Extensions = [
   AddressPlzExtension,
   AddressLocationExtension,
   NuernbergPassIdExtension,
-  NuernbergPassIdExtension,
```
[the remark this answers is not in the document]
steffenkleinle: Thank you for this, it was definitely not on purpose!

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -45,6 +60,17 @@ const BirthdayForm = ({
...
+          <>
+            {projectConfig.showBirthdayMinorHint && (
+              <StyledToolTip
+                className={Classes.TOOLTIP_INDICATOR}
+                content='Bei Minderjährigen unter 16 Jahren darf der KoblenzPass nur mit Einverständnis der Erziehungsberechtigten abgerufen werden.'>
+                <StyledHelpOutlineIcon fontSize='small' />
+              </StyledToolTip>
+            )}
+          </>
+        }
```
steffenkleinle, line 73: :thinking: I think we should not pass this to the DatePicker component, this is not a thing it should be concerned with. Instead, we should make sure the styling is correct here and render the hint directly here.
I see three options:
1. Render it next to the label of the FormGroup (label accepts also React nodes)
2. Use FormGroups `helperText` prop and just display a text, possibly only if the birthday is actually underage
3. Render it here next to the date input and fix the styling here. I don't really like it though as it makes the inputs have different lengths

@f1sh1918 what do you think?

administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -24,38 +30,42 @@ const CustomDatePicker = ({
...
-      onChange={onChange}
-    />
+        }}
+        localeText={deDE.components.MuiLocalizationProvider.defaultProps.localeText}
+        disableFuture={disableFuture}
+        minDate={minDate}
+        maxDate={maxDate}
+        onChange={onChange}
+      />
+      {sideComponent}
```
steffenkleinle, line 67: The icon is sadly not correctly vertically centered/aligned:
image

administration/src/project-configs/koblenz/config.ts
```diff
@@ -36,6 +36,7 @@ const config: ProjectConfig = {
   selfServiceEnabled: true,
   storesManagement: storesManagementConfig,
   userImportApiEnabled: true,
+  showBirthdayMinorHint: true,
```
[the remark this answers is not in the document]
seluianova: showBirthdayInfoHint, if the text is customizable per project?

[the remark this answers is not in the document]
f1sh1918: hm since we have no real configuration structure for our extension i think we may include the extension name in the flag. Maybe we want to refactor the extensions structure in the future and this flag can move then to the extension itself

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -45,6 +60,17 @@ const BirthdayForm = ({
         isValid={isValid || !showErrorMessage}
         maxDate={new Date()}
         disableFuture
+        sideComponent={
+          <>
+            {projectConfig.showBirthdayMinorHint && (
+              <StyledToolTip
+                className={Classes.TOOLTIP_INDICATOR}
+                content='Bei Minderjährigen unter 16 Jahren darf der KoblenzPass nur mit Einverständnis der Erziehungsberechtigten abgerufen werden.'>
```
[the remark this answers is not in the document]
f1sh1918: yes it should be in the translations file. we may create a namespaces "extensions" for this.
Currently i think we have no need to create project specific translations in administration but it could be necessary for the future. I wouldn't put effort in this currently but maybe create overrides in the future

administration/src/bp-modules/components/CustomDatePicker.tsx
f1sh1918: looks like the parent is not flex

administration/src/cards/extensions/BirthdayExtension.tsx
f1sh1918: Hm i think you should ask @hauf-toni 
We have already an explanation to the "Aktenzeichen" form field which i opens a modal with explanations
i'm not pretty sure if we want to implement the same behavior but it makes it easier for the user to have sth. consistent

image

administration/src/bp-modules/components/CustomDatePicker.tsx
f1sh1918: but first check where we want to put it in the end (see other comment)

administration/src/cards/extensions/BirthdayExtension.tsx
f1sh1918: i would even ask toni

hauf-toni: Would it be possible to include real-time feedback? This would allow us to inform users immediately if they enter a date indicating an age below 16, improving usability. If real-time validation isn’t feasible, I also like @seluianova 's suggestion of adding an extra confirmation after form submission.

f1sh1918: not sure if this screenshot is the current state. But this red color and the highlighting of the textfield indicates its an error. But it shouldn't be an error but a hint. So it should be just black and not red indicated both. @bahaaTuffaha

bahaaTuffaha: Sorry I pushed before reading Tori's comment...

f1sh1918:
@bahaaTuffaha there are still tests failing.
I will start reviewing and testing after this is fixed :)

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -23,13 +24,27 @@ const BirthdayForm = ({
...
+    const today = PlainDate.fromLocalDate(new Date())
+
     if (!birthday) {
       return t('birthdayMissingError')
     }
-    if (birthday.isAfter(PlainDate.fromLocalDate(new Date()))) {
+    if (birthday.isAfter(today)) {
       return t('birthdayFutureError')
     }
+    if (isBirthdayHintEnabled()) {
```
f1sh1918, line 44: :wrench:
i think this hint should be handled separately, since its not an error

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -45,10 +60,16 @@ const BirthdayForm = ({
...
+        isHint={isBirthdayHintEnabled()}
         maxDate={new Date()}
         disableFuture
       />
-      {showErrorMessage && <FormErrorMessage errorMessage={getErrorMessage()} />}
+      {showErrorMessage && (
+        <FormErrorMessage
+          style={{ color: isBirthdayHintEnabled() ? 'black' : undefined }}
+          errorMessage={getErrorMessage()}
+        />
```
f1sh1918, line 71: :wrench:
i would add here a `showBirthdayHint` and separate hint and error handling

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -45,10 +60,16 @@ const BirthdayForm = ({
         onChange={changeBirthday}
         onClear={() => setValue({ birthday: null })}
         isValid={isValid || !showErrorMessage}
+        isHint={isBirthdayHintEnabled()}
         maxDate={new Date()}
         disableFuture
       />
-      {showErrorMessage && <FormErrorMessage errorMessage={getErrorMessage()} />}
+      {showErrorMessage && (
+        <FormErrorMessage
```
f1sh1918, line 68: If we want to use this for a hint(info) and error we could also use a more neutral name like `Alert` which is also using mui for naming and the severity can be different. The severity the can define the message color

f1sh1918 (commented):
Looks quite good already. Just added some small suggestions

administration/src/cards/extensions/BirthdayExtension.tsx
bahaaTuffaha: I will add a prop called `isHint` becouse `FormErrorMessage` is used in multiple places

administration/src/bp-modules/self-service/CardSelfServiceForm.tsx
```diff
@@ -63,7 +63,7 @@ const CardSelfServiceForm = ({
     }
     if (!cardValid || dataPrivacyAccepted !== DataPrivacyAcceptingStatus.accepted) {
       appToaster?.show({
-        message: <FormErrorMessage style={{ color: 'white' }} errorMessage={t('atLeastOneInputIsInvalid')} />,
+        message: <FormAlert style={{ color: 'white' }} errorMessage={t('atLeastOneInputIsInvalid')} />,
```
f1sh1918, line 66: :upside_down_face:  i think we may create a separate ticket to create a `ToastAlertMessage` component, since it will never be null and the text color will be white and its not a "form alert"

Alternatively we could just set sth is like `isToast = true` and set a defined color to avoid inline styling

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -23,13 +24,31 @@ const BirthdayForm = ({
...
     }
-    if (birthday.isAfter(PlainDate.fromLocalDate(new Date()))) {
+    if (birthday.isAfter(today)) {
       return t('birthdayFutureError')
     }
+
+    return null
+  }
+
+  const getHintMessage = (): string | null => {
```
f1sh1918, line 48: ❓ why do you need this additional function?
I posted a suggestion how to simplify that

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -48,7 +67,8 @@ const BirthdayForm = ({
         maxDate={new Date()}
         disableFuture
       />
-      {showErrorMessage && <FormErrorMessage errorMessage={getErrorMessage()} />}
+      {showErrorMessage && <FormAlert severity='error' errorMessage={getErrorMessage()} />}
+      {!!getHintMessage() && <FormAlert severity='info' errorMessage={getHintMessage()} />}
```
f1sh1918, line 71:
```suggestion
       {showBirthdayHint() && <FormAlert severity='info' errorMessage={t('birthdayHint')} />}
```

administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -23,13 +24,31 @@ const BirthdayForm = ({
   const [touched, setTouched] = useState(false)
   const { birthday } = value
   const showErrorMessage = touched || showRequired
+  const projectConfig = useContext(ProjectConfigContext)
+
+  const isBirthdayHintEnabled = (): boolean => {
```
f1sh1918, line 29: 🔧 
```suggestion
  const showBirthdayHint = (): boolean => {
```
I think "enabled" is not a good naming, since it may be enabled by the projectConfig but not shown because the age conditions are false

f1sh1918 (commented):
Looks good so far. Just added some small suggestions to improve the code

administration/src/cards/extensions/BirthdayExtension.tsx
bahaaTuffaha: You suggested to split it up from errors?

f1sh1918 (approved):
Good work 👍 

I would suggest to squash merge your pr since we had couple of changes and a lot of commit though

Issue #1816, opened 2024-12-04
Input validation for application endpoint


ztefanie:
**Is your feature request related to a problem? Please describe.**
Currently sending bad requests to the application endpoint and results in `500 server errors`. This should not be the case and proper error handling should be added and proper error messages should be sent.

**Describe the solution you'd like**
Check all values of the application endpoint. What happens if they are set to "wrong" values. E.g. if setting `"givenInformationIsCorrectAndComplete" to false a 500 error is thrown or if an empty string is sent for any value (like forename) the server sends a 500 error.
Also if sending multiple query / muntations at once leads to an 500 error

**Additional context**
As this endpoint is open it should handle all kinds of requests that it gets properly.

ztefanie:
I already moved this to next up, as I think when verein360 starts to use this endpoint, it would make sense we soonish validate input properly

PR #1863, opened 2025-01-15
1816: Input validation for application endpoint

seluianova:
### Short description
Currently, sending invalid requests to an application endpoint results in 500 server errors, and no error description provided to the user.
This happens because we do validation in init blocks of the data classes, and the exceptions that occur get wrapped up in InvocationTargetException, which are not handled.

### Proposed changes
- refactor CustomDataFetcherExceptionHandler to get rid of the deprecated function
- unwrap InvocationTargetException to return GraphQLBaseException to the user

### Side effects
No?

### Testing
1. Login to administration as Verein360 admin, create API token
2. Send addEakApplication request using the created token
3. Set some invalid data in the request, for example, givenInformationIsCorrectAndComplete = false
4. Check error response

Valid request example:
```
mutation AddEakApplication {
    addEakApplication(
        regionId: 1,
application: {
  applicationDetails: {
    applicationType: FIRST_APPLICATION,
    cardType: BLUE,
    givenInformationIsCorrectAndComplete: true,
    hasAcceptedEmailUsage: true,
    hasAcceptedPrivacyPolicy: true,
    wantsDigitalCard: true,
    wantsPhysicalCard: false,
    blueCardEntitlement: {
      entitlementType: WORK_AT_ORGANIZATIONS,
      juleicaEntitlement: null,
      workAtOrganizationsEntitlement: {
        list: {
          amountOfWork: 7.5,
          organization: {
            address: {
              country: {
                shortText: "Deutschland"
              },
              houseNumber: {
                shortText: "123"
              },
              location: {
                shortText: "München"
              },
              postalCode: {
                shortText: "80331"
              },
              street: {
                shortText: "Example Street"
              }
            },
            category: { shortText: "Sport" },
            contact: {
              email: { email: "jane.doe@sportverein.de" },
              hasGivenPermission: true,
              name: { shortText: "Jane Doe" },
              telephone: { shortText: "0150123456789" }
            },
            name: {
              shortText: "Sportverein Augsburg-Nord"
            }
          },
          payment: false,
          responsibility: { shortText: "Trainer" },
          workSinceDate: { date: "2020-10-06" },
          isAlreadyVerified: true
        }
      }
    }
  },
  personalData: {
    address: {
      country: {
        shortText: "Deutschland"
      },
      houseNumber: {
        shortText: "123"
      },
      location: {
        shortText: "München"
      },
      postalCode: {
        shortText: "80331"
      },
      street: {
        shortText: "Example Street"
      }
    },
    dateOfBirth: {
      date: "2000-01-01"
    },
    emailAddress: {
      email: "johndoe@example.com"
    },
    forenames: {
      shortText: "John"
    },
    surname: {
      shortText: "Doe"
    },
    telephone: {
      shortText: "123456789"
    }
  }
}
        project: "bayern.ehrenamtskarte.app"
    )
}
```

### Resolved issues
Fixes: #1816

### DIFF
--- a/administration/src/util/translations/de.json
+++ b/administration/src/util/translations/de.json
@@ -56,7 +56,7 @@
     "amountOfWork": "Arbeitsstunden pro Woche (Durchschnitt)",
     "workSinceDate": "Tätig seit",
     "isAlreadyVerified": "Die Angaben wurden bereits durch den zuständigen Verein bei Beantragung verifiziert",
-    "payment": "Für diese ehrenamtliche Tätigkeit wurde eine Aufwandsentschädigung gewährt, die über den jährlichen Freibetrag hinaus geht (840 Euro Ehrenamtspauschale bzw. 3000 Euro Übungsleiterpauschale):",
+    "payment": "Für diese ehrenamtliche Tätigkeit wurde eine Aufwandsentschädigung gewährt, die über den jährlichen Freibetrag hinaus geht (840 Euro Ehrenamtspauschale bzw. 3000 Euro Übungsleiterpauschale)",
     "certificate": "Tätigkeitsnachweis",
     "hasAcceptedPrivacyPolicy": "Ich erkläre mich damit einverstanden, dass meine Daten zum Zwecke der Antragsverarbeitung gespeichert werden und akzeptiere die Datenschutzerklärung",
     "hasAcceptedEmailUsage": "Ich stimme zu, dass ich von der lokalen Ehrenamtskoordination über Verlosungen und regionale Angebote informiert werden darf",
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
@@ -13,6 +13,7 @@ import app.ehrenamtskarte.backend.auth.webservice.TokenAuthenticator
 import app.ehrenamtskarte.backend.common.webservice.GraphQLContext
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidFileSizeException
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidFileTypeException
+import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidJsonException
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.MailNotSentException
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.RegionNotActivatedForApplicationException
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.RegionNotFoundException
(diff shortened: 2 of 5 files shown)


administration/src/util/translations/de.json
```diff
@@ -56,7 +56,7 @@
     "amountOfWork": "Arbeitsstunden pro Woche (Durchschnitt)",
     "workSinceDate": "Tätig seit",
     "isAlreadyVerified": "Die Angaben wurden bereits durch den zuständigen Verein bei Beantragung verifiziert",
-    "payment": "Für diese ehrenamtliche Tätigkeit wurde eine Aufwandsentschädigung gewährt, die über den jährlichen Freibetrag hinaus geht (840 Euro Ehrenamtspauschale bzw. 3000 Euro Übungsleiterpauschale):",
+    "payment": "Für diese ehrenamtliche Tätigkeit wurde eine Aufwandsentschädigung gewährt, die über den jährlichen Freibetrag hinaus geht (840 Euro Ehrenamtspauschale bzw. 3000 Euro Übungsleiterpauschale)",
```
seluianova, line 59: Otherwise the ":" symbol is duplicated
image

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
```diff
@@ -128,24 +129,24 @@ class ApplicationHandler(
...
-            require(!organizations.isNullOrEmpty()) {
+            require(organizations.isNotEmpty()) {
                 "Work at organizations list cannot be empty if application is already verified"
             }
             require(organizations.all { it.organization.category.shortText == "Sport" }) {
                 "All organizations must be of category Sport if application is already verified"
             }
         } catch (e: IllegalArgumentException) {
-            throw BadRequestResponse(e.message!!)
+            throw InvalidJsonException(e.message!!)
```
seluianova, line 149: Replaced with InvalidJsonException to stick to the common error format.
(we have a custom handler for GraphQLBaseExceptions to convert them to a specific response format)

f1sh1918 (approved):
Works great 👍 
Thx for the refactoring.

i'm not that super deep into error handling of request so maybe someone with better knowledge should review that, too
For me it looks fine and i also tested some other errors and they still work as expected

administration/src/util/translations/de.json
steffenkleinle: I think it might be good to do this the other way round though, keep the symbol here and remove it otherwise. The reason for this is that some languages handle punctuation marks differently. For example, in French there is a space between the sentence and the colon, e.g. `C'est vrai : j'ai ...`.
Even though we don't have any languages atm that do this, I think it would not hurt.

steffenkleinle (approved):
Tested in firefox as mentioned, works as expected.

administration/src/util/translations/de.json
seluianova: Would make sense, but this translation is also used on the application form, where ":" is not needed at all:
image

steffenkleinle: okay, then lets just keep it like this :)

Issue #1774, opened 2024-11-14
Add all german texts to translation file


ztefanie:
**Is your feature request related to a problem? Please describe.**
Add all german text to translation file in digitale druckerei

**Testing**
Test all features of digitale druckerei and check all labels / texts are displayed correctly

Follow up ticket for #1623

ztefanie:
All translations of `administration/bp-modules` are already moved i this pr #1841

Issue #1869, opened 2025-01-16
Finalize translations for administration


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Implement the missing translation for administration, especially some validation errors need some adjustments to be able to use i18n there.

**Describe the solution you'd like**
Replace the missing hardcoded labels with translation keys and move values to `de.json`:
- validation error messages that are outside of components (application form) -> `i18next.t` can be used
- cleanup: check where passing of `TFunction` can be avoided and use `i18next.t` instead if possible ( f.e.`getMessageFromApolloError` and `graphqlErrorMap`). 
- fix test with `TFunction` (deleteCards) if still needed after cleanup

PR #1870, opened 2025-01-16
1774: Move german translation mui components

f1sh1918:
### Short description

Move translation of mui-components and related components to translation file.
I decided to split this issue, since its really huge (see todo) and adding i18n in some form validation function may not be that trivial and needs some refactoring

### Proposed changes

- add i18n for graphql error map
- add translations for most of the mui-components (some validation errors are missing)
- added readme to point out that the keys in "application" namespaces should not be edited, thats why i created a new namespace "applicationForms" for labels that could be changed

### TODO
- finalize i18n for administration (#1869)

### Side effects

- may break labels for application form, activation, application applicant view, application verification view, card creation, self service portal (koblenz)

### Testing

1. Create an application and create a card from the application for bavarian entitlementcard
2. Click on the mail links for application creation, application verification and card creation (card was created from application)
3. Create a koblenz pass within the self service portal
4. Check all the parts that were mentioned in side effects

### Resolved issues

Fixes: #1774


administration/src/util/translations/Readme.md
```diff
@@ -0,0 +1,6 @@
+# Important Notice
+
+- Translation keys inside the `application` namespace should not be edited!!!
+- These keys are stored in the database and have to be resolvable.
+- If you change these keys, labels of old applications can not be resolved anymore.
+- Labels that can be changed because they are not stored in the database were put in other namespaces f.e. "applicationForms".
```
seluianova, line 6: :thought_balloon: just as an idea, because I'm pretty sure that if someone wants to change the lable, they won't look at the readme before doing so.
what if we add an json entry to the "application" like:
"_comment" : "Important! These labels can't be changed because ...."
it will be ignored by the app but serve as a visible comment

f1sh1918: thats a good idea, i will do that

administration/src/bp-modules/stores/StoresImportController.tsx
```diff
@@ -54,7 +55,7 @@ const StoresImport = ({ fields }: StoreImportProps): ReactElement => {
       setAcceptingStores([])
     },
     onError: error => {
-      const { title } = getMessageFromApolloError(error)
+      const { title } = getMessageFromApolloError(error, t)
```
seluianova, line 58: :question: might be a dumb question, but I see several places where we create const { t } = useTranslation('errors') and only use it to pass as a param into getMessageFromApolloError() function.
so I wonder why const { t } is not created inside getMessageFromApolloError() ?

same with getQueryResult(), we pass a parameter there, but it is not used and only passed to getMessageFromApolloError().

f1sh1918: thats because `useTranslation` is a hook and you can't use them in regular functions but in react functional components, so we pass them from functional components to the function
https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-from-react-functions

seluianova: Ah, thanks for the clarification

administration/src/util/translations/de.json
```diff
@@ -65,9 +75,62 @@
...
+    "honoredByMinisterCertificate": "Urkunde",
+    "honoredByMinisterCertificateDescription": "Hängen Sie hier bitte Ihre eingescannte oder abfotografierte Urkunde an.",
+    "goldenCardRequirementsTitle": "Ich erfülle folgende Voraussetzung für die Beantragung einer goldenen Ehrenamtskarte:",
+    "blueCardRequirementsTitle": "Ich erfülle folgende Voraussetzung für die Beantragung einer blauen Ehrenamtskarte:",
+    "applicationStepPersonalData": "Persönliche Angaben",
+    "applicationStepRequirements": "Voraussetzungen",
+    "applicationStepSubmit": "Antrag Senden"
   },
   "applications": {
+    "_comment": "!Important! These keys should not be edited, since they are stored in the database and have to remain resolvable. If you change these keys, labels of old applications can not be resolved anymore.",
```
seluianova, line 133: it should be under "application" I guess, not "applications"

f1sh1918: Ich could be applicationOverview maybe

f1sh1918: hm i checked the content of the namespaces and overview is not fitting. seems like its kind of mixture unfortunately `applicationCard` could be good

f1sh1918: https://github.com/digitalfabrik/entitlementcard/issues/1887

steffenkleinle (approved):
Thanks for this PR, must have been a lot of work. :pray:  Looks good, only smoke tested though as I think this PR is way to big and therefore very hard to review. For the next PR it would be better to further split things like this into smaller tasks/PR.

Issue #1683, opened 2024-10-15
An application is still created when no email could be sent to contact person of organization


juliankehne:
**Describe the bug**
Entering an invalid email, e.g. "a@de", which is allowed to be set in the input text field, but cannot be sent when sending the application will give an error message that no email could be sent to the address. The application however will still be created.
The user might or might not fix the email and send the application again, which will result in duplicate applications.

**To Reproduce**
1. Go to https://staging.bayern.ehrenamtskarte.app/beantragen
2. Enter all mandatory information
3. For the email of the contact person of an organization enter an invalid email, e.g. "a@de"
4. See that input text field will allow it
5. Click "Send Application"
6. See error message and that there is no confirmation that an application was created
7. Go to: https://staging.bayern.ehrenamtskarte.app/applications and see that a new application was created

**Expected behavior**
No application should be created when the email wasn't able to be sent to the contact person.

**Screenshots**

![2024-10-15 10_09_34-Window](https://github.com/user-attachments/assets/bc69698b-7249-4fd8-a089-bf3e30d82f0b)
![2024-10-15 10_10_08-Window](https://github.com/user-attachments/assets/968a0c82-a260-4da3-adc0-7cf42777ddc4)
![2024-10-15 10_23_03-Window](https://github.com/user-attachments/assets/676998f7-fdb6-4eae-9c69-3752e2c831a1)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Firefox
 - Version: 131.0.3

**Additional context**
- Version 24.9.0

PR #1874, opened 2025-01-20
1683: Adding email validation for email forms

bahaaTuffaha:
### Short description

Entering an invalid email, e.g. "a@de", which is allowed to be set in the input text field, but cannot be sent when sending the application will give an error message that no email could be sent to the address. The application however will still be created.

### Proposed changes

- Used `isEmailValid` function from utils `bp-modules/applications/utils`

### Side effects

- `primitive-inputs/EmailForm.tsx`

### Suggestion:
- We could use https://www.usercheck.com api to prevent temporary emails also ... it has 1000 check per month free then when its done we can fallback to normal behavior.

### Testing

1. Go to https://staging.bayern.ehrenamtskarte.app/beantragen
2. Enter all mandatory information
3. Under WEITERE ANGABEN for email address enter an invalid email, e.g. "a@de".
4. For the email of the contact person of an organization enter an invalid email, e.g. "a@de".

### Resolved issues

Fixes: #1683

### DIFF
--- a/administration/src/mui-modules/application/primitive-inputs/EmailForm.tsx
+++ b/administration/src/mui-modules/application/primitive-inputs/EmailForm.tsx
@@ -1,6 +1,7 @@
 import { TextField } from '@mui/material'
 import React, { useContext, useState } from 'react'
 
+import { isEmailValid } from '../../../bp-modules/applications/utils/verificationHelper'
 import { EmailInput } from '../../../generated/graphql'
 import { FormContext } from '../SteppedSubForms'
 import { Form, FormComponentProps } from '../util/FormType'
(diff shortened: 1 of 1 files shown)


administration/src/mui-modules/application/primitive-inputs/EmailForm.tsx
```diff
@@ -22,6 +23,12 @@ const EmailForm: Form<State, ValidatedInput, AdditionalProps> = {
         message: `Text überschreitet die maximal erlaubten ${MAX_SHORT_TEXT_LENGTH} Zeichen.`,
       }
     }
+    if (!isEmailValid(email)) {
+      return {
+        type: 'error',
+        message: `Ungültige E-Mail.`,
```
f1sh1918, line 29:
```suggestion
        message: `E-Mailadresse ist ungültig.`,
```
Since we use proper sentences in the messages before we should keep this

connium: The Duden dictionary recommends writing the term as "E-Mail-Adresse" (see https://www.duden.de/rechtschreibung/E_Mail_Adresse). The rule for this is that foreign words and German words are joined with a hyphen.

seluianova (approved):
Apart from the suggestion already mentioned, looks good to me!

Issue #1850, opened 2025-01-07
Card can be added multiple times on the same device


seluianova:
**Describe the bug**
If the user scans the QR code of a card that has already been added to the device, the user will receive the message "Your card has already been activated on another device. If you continue, your card will be automatically deactivated on the other device if it is still active there."
If the user proceeds, he gets the same card 2 times on the same device (first one is revoked, second one is active).

**To Reproduce**
The issue can be reproduced for projects that allow more than 1 card (Nürnberg, Koblenz)
1. Create PDF for the card activation
2. Open the app --> Ausweisen --> Pass aktivieren
3. Scan the QR code
4. When the card is added, scan the same code again.

**Expected behavior**
I think there should be the message "This card has already been activated on this device"

f1sh1918:
good find. i think currently we only check this for deeplink activation

seluianova:
@f1sh1918 
> i think currently we only check this for deeplink activation

actually it seems to be broken by this change https://github.com/digitalfabrik/entitlementcard/pull/1721

PR #1876, opened 2025-01-21
1850: Fix multi card activation

f1sh1918:
### Short description

This pr fixes the issue that the same card could be activated twice on the same device, even the first card is changing to an invalid state. Basically if the same card is already activated on this device, the user shouldn't be able to activate it again.

### Proposed changes

- add a deep equality check for the peppers lists ton ensure same cards are correctly identified

### Side effects

- none

### Testing

1. Create a card for nuernberg or koblenz (where multiple cards can be activated)
2. Activate the card once. Then use the same qr code and try to activate the same card
3. There should be a message, that the card is already activated on your device

### Resolved issues

Fixes: #1850


seluianova (approved):
Tested deeplink and normal activation on android emulator, looks good!
Thanks for the fix 👍

Issue #1859, opened 2025-01-14
Misleading error message when required fields are missing


connium:
**Describe the bug**
When applying for an Ehrenamtskarte and I click on the button "Antrag abschicken" without having filled all required inputs on page 3, the message "Ungültige bzw. fehlende Eingaben entdeckt. Bitte prüfen Sie die rot markierten Felder." is shown, but there aren't any inputs marked red.

**To Reproduce**
1. Apply for an Ehrenamtskarte. Enter valid data for page 1 and 2.
2. On page 3 enter only some data and leave at least one required input field empty.
3. Scroll down and click on step 4.
4. Tick the necessary checkboxes.
5. Click on "Antrag abschicken".
6. See the error message. The indicator for step 3 shows a checkmark, which indicates, that everything is correct. Even if I reopen the page for step 3, nothing is marked red.

**Expected behavior**
If the message tells me there are red marked fields, there should be some.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser Firefox 134

**Smartphone (please complete the following information):**
 - Device: LG Q7
 - OS: Android 9
 - Browser stock browser (Chrome)

**Additional context**
IMO there are multiple solutions we could consider:

1. Mark all fields red, when clicking the "Nächster Schritt" button.
2. Think of a different error indicator at least for Android, because the tooltip vanishes immediately when the corresponding field receives the focus.
3. Think of a proper error indicator for the drop down "Einsatzgebiet" because on Android it was not obvious to me, that this field was missing.
4. Do not mark step 3 with a check mark, if there are still errors in the form.
5. Rephrase the error message to refer to "required inputs" instead of "red marked inputs".
6. ...

michael-markl:
Navigating to step n should only be possible if steps 1,..., n-1 are valid. This logic used to work, but it seems to have been broken somehow...

michael-markl:
Found the culprit: 63167b11c2412952f22013df8ee00ee2c2f30a02

Seems to an artifact of a refactor.

PR #1877, opened 2025-01-21
Fix Application Form: Do not allow jumping to next step if a previous…

michael-markl:
… step contains invalid data

### Short description

Fix Application Form: Do not allow jumping to next step if a previous…

### Proposed changes

Fix the bug that was introduced by a huge refactor :D

### Side effects

None

### Testing

Navigating to a step in the application form should no longer be possible if there is invalid data.

### Resolved issues

Fixes: #1859


seluianova (approved):
Works like a charm!

f1sh1918 (approved):
thx for the fix 👍 
works as expected

Issue #1660, opened 2024-10-02
Handle apollo errors correctly 


ztefanie:
**Is your feature request related to a problem? Please describe.**
the function `getMessageFromApolloError` returns wrong error messages to the user. As a user I want to see correct error messages.

**Describe the solution you'd like**
The getMessageFromApolloError return the correct error message

**Additional context**

- Go to ProjectSettingsController.tsx and removee the if case that is return an "missing rights" section.
- Now login as bayern project admin
- open http://localhost:3000/project
- create a token
- see the error message "Server nicht erreichbar."
the backend return the correct error code "403 Forbidden". The user should get an appropriate information about the error that occured. This should not only happen for this example but everywhere.

PR #1878, opened 2025-01-21
1660: Handle apollo errors

f1sh1918:
### Short description

We have the issue that all non `GraphlBaseExceptions` will result in the same message to the user and 403 exceptions will be handled as a network error and just result in `Server nicht erreichbar`
### Proposed changes

- remove the check for `networkError` but handle all errors which are not graphql errors in a `defaultErrorMap` and check for the particular error code and send the matching error message
- if non of these error codes were found just then return 'Server nicht erreichbar', because we didn't get any response from the backend (the most probably reason i guess)

### Side effects

- some error messages may change or not be displayed

### Testing

I'm not pretty sure how to test this on staging because, most of the errors were handled before by not showing the function like `PepperSettings` or `ApiTokenSettings``
As a dev you can follow the instructions in the related issue to test the error messages

### Resolved issues

Fixes: #1660

### DIFF
--- a/administration/src/errors/getMessageFromApolloError.tsx
+++ b/administration/src/errors/getMessageFromApolloError.tsx
@@ -2,27 +2,22 @@ import { ApolloError } from '@apollo/client'
 import { ReactElement } from 'react'
 
 import defaultErrorMap from './DefaultErrorMap'
+import graphQlErrorMap from './GraphQlErrorMap'
 
-type GraphQLErrorMessage = {
+export type GraphQLErrorMessage = {
   title: string
   description?: string | ReactElement
 }
 
 const getMessageFromApolloError = (error: ApolloError): GraphQLErrorMessage => {
-  const defaultMessage = 'Etwas ist schief gelaufen.'
-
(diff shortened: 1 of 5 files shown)


administration/src/errors/getMessageFromApolloError.tsx
```diff
@@ -2,27 +2,22 @@ import { ApolloError } from '@apollo/client'
...
-type GraphQLErrorMessage = {
+export type GraphQLErrorMessage = {
   title: string
   description?: string | ReactElement
 }
 
 const getMessageFromApolloError = (error: ApolloError): GraphQLErrorMessage => {
-  const defaultMessage = 'Etwas ist schief gelaufen.'
-
-  if (error.networkError) {
```
f1sh1918, line 14: not sure if everything is still catched. All i tested was catched by `DefaultErrorMap`
Maybe there is also sth misconfigured so that 403 were classified as `NetworkErrors` accidently?

steffenkleinle (approved):

Issue #1864, opened 2025-01-15
Integration tests for addEakApplication mutation (Verein360)


seluianova:
**Is your feature request related to a problem? Please describe.**
Follow up for https://github.com/digitalfabrik/entitlementcard/pull/1863

To be done:
- [x] positive scenarios - valid requests are sent and processed
- [x] authorization testing - api token missing / invalid
- [x] request validation specific for Verein360 - validateAllAttributesForPreVerifiedApplication() function
- [ ] ~~general request validation applicable to all application requests~~ --> moved to #1932

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1879, opened 2025-01-21
1864: Integration tests for Verein360 applications

seluianova:
### Short description
Integration tests for Verein360 applications

### Proposed changes
Some tests

Not included:
- general testing of application creation (not related to Verein360)
- some branches in validateAllAttributesForPreVerifiedApplication() because them seem to be unreachable

### Side effects
no, only tests

### Testing
n/a

### Resolved issues
Fixes: #1864


backend/src/test/kotlin/app/ehrenamtskarte/backend/application/webservice/Verein360ApplicationTest.kt
```diff
@@ -0,0 +1,208 @@
...
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+
+internal class Verein360ApplicationTest : GraphqlApiTest() {
+
+    data class ValidationErrorTestCase(val application: Application, val error: String)
+
+    companion object {
+        @JvmStatic
+        fun validationErrorTestCases(): List<ValidationErrorTestCase> {
```
f1sh1918, line 35: i would add here the policy opt in
`givenInformationIsCorrectAndComplete` to check that its set

backend/src/test/kotlin/app/ehrenamtskarte/backend/util/GraphQLRequestSerializer.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.full.declaredMembers
+
+object GraphQLRequestSerializer {
+
+    /**
+     * Serialize object into graphql request format
+     * e.g. { stringProperty: "value", intProperty: 123, objectProperty: { booleanProperty: true } }
+     */
+    fun serializeObject(obj: Any): String {
```
f1sh1918, line 14: 👍 Super nice

f1sh1918 (approved):
Very good work! 👍 
Lots of important reusable parts for future tests.

I couldn't find any issue but could image to add one or two small test cases

steffenkleinle (approved):
LGTM

backend/src/test/kotlin/app/ehrenamtskarte/backend/application/webservice/Verein360ApplicationTest.kt
seluianova: I will open a separate PR for that, it requires some additional work

seluianova: The problem with these fields is that they are validated in init blocks, which means, it's not possible to create an application object with invalid values, such as givenInformationIsCorrectAndComplete = false.
I think I would have to re-work the current approach and build an application as a string rather than as an object.
Or create a default object, serialize it to a string, and then replace the values in the string with invalid values.
Not sure yet what is less ugly.
I will create another task for that.

PR #1883, opened 2025-01-27
1882: Fix translations override

seluianova:
### Short description
There is an error in the log when launching the Nürnberg app
```
TypeError: Null check operator used on a null value
  File "singleton.dart", line 504, in LocaleSettingsExt.overrideTranslations
  File "translations.g.dart", line 174, in LocaleSettings.overrideTranslations
  File "main.dart", line 66, in overrideLocale
```
This also means that one of the languages was not loaded correctly and the default translation was used.

### Proposed changes
I guess it’s related to asynchronous processing.
I have added ‘await’ to the overrideLocale() call and it seems to help.

### Side effects
Hopefully none

### Testing
- Check that the translations for each application are loaded properly (overrides are applied in the app)
- Check the language change for Nürnberg (About -> Settings -> Change language)

### Resolved issues
Fixes: #1882

### DIFF
--- a/frontend/lib/main.dart
+++ b/frontend/lib/main.dart
@@ -31,7 +31,9 @@ Future<void> main() async {
   // Use override locales for whitelabels (e.g. nuernberg)
   // ignore: unnecessary_null_comparison
   if (buildConfig.localeOverridePath != null) {
-    AppLocale.values.forEach(overrideLocale);
+    await Future.forEach(AppLocale.values, (locale) async {
+      await overrideLocale(locale);
+    });
   }
 
   debugPrint('Environment: $appEnvironment');


steffenkleinle (approved):
Tested on android, thanks for the fix! Works as expected :)

f1sh1918 (commented):
I'm still getting this error
I think the issue here is that the main widget is not initialized correctly, so we probably have to use some lifecycle functions and execute the locale init after this. I thought that `ensureInitialized` will do that but it seem not to work

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Null check operator used on a null value
#0      LocaleSettingsExt.overrideTranslations (package:slang/src/api/singleton.dart:504:51)
#1      LocaleSettings.overrideTranslations (package:ehrenamtskarte/l10n/translations.g.dart:174:137)
#2      overrideLocale (package:ehrenamtskarte/main.dart:68:20)
<asynchronous suspension>
#3      main.<anonymous closure> (package:ehrenamtskarte/main.dart:35:7)
<asynchronous suspension>
#4      Future._kTrue (dart:async/future.dart:660:3)
<asynchronous suspension>

```

seluianova:
@f1sh1918 I've merged you proposal, please review again

f1sh1918 (approved):
Tested on ios looks good 👍

Issue #1407, opened 2024-04-22
Add test for juleica verification / application card component


seluianova:
**Is your feature request related to a problem? Please describe.**
Add tests to cover https://github.com/digitalfabrik/entitlementcard/pull/1394

**Additional context**
TBD

PR #1886, opened 2025-01-29
1407: Application verification tests

f1sh1918:
### Short description

I added here some view and component tests for the application verification. I excluded application (issue description) card, since its getting too big.
This can be done in a separate task, since lots of unit tests are missing in general.

### Proposed changes

- Created unit tests for application verification related components
- created a separate component for `VerificationListItem` to improve testability
- set utc timezone globally for jest

### Side effects

- Even i just moved code to a separate component it should be checked if everything is fine

### Testing

- npm run test
- check the verifications in the application card, if they still look ok ("Bestätigungen der Organisation...")

### Resolved issues

Fixes: #1407


administration/src/bp-modules/applications/VerificationsView.tsx
```diff
@@ -84,47 +77,11 @@ const VerificationsView = ({ verifications }: { verifications: Application['veri
       <H5>{t('confirmationsByOrganizations')}</H5>
       <VerificationContainer>
         {verifications.map(verification => {
-          const status = getStatus(verification)
```
f1sh1918, line 87: moved to separate component

administration/src/bp-modules/applications/__tests__/ApplicationStatusBar.test.tsx
```diff
@@ -118,11 +119,11 @@ describe('ApplicationStatusBar', () => {
         activeBarItem={barItems[0]}
       />
     )
-
+    await act(async () => null) // Blueprint5.Icon update
```
f1sh1918, line 122: there is a general issue with the Blueprint icons that lead to a console.error
Not sure if we can find a general solution for that instead of adding this in each test

administration/src/bp-modules/applications/__tests__/VerificationsQuickIndicator.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+describe('VerificiationQuickIndicator', () => {
+  it('should show only awaiting verifications', async () => {
+    const { getByTestId } = renderWithTranslation(<VerificationQuickIndicator verifications={verificationsAwaiting} />)
+    await act(async () => null) // Blueprint5.Icon update
+    expect(getByTestId(`indicator-${VerificationStatus.Verified}`).textContent).toBe(': 0')
+    expect(getByTestId(`indicator-${VerificationStatus.Awaiting}`).textContent).toBe(': 2')
+    expect(getByTestId(`indicator-${VerificationStatus.Rejected}`).textContent).toBe(': 0')
+  })
+
+  it('should show only mixed verifications', async () => {
```
seluianova, line 22:
```suggestion
  it('should show mixed verifications', async () => {
```
:upside_down_face: or: should show verifications in all statuses

seluianova (approved):
Thanks for taking on this old task that I created and abandoned 🙈
Looks good to me.
Verifications in the applications overview are also fine.

administration/src/bp-modules/applications/PreVerifiedQuickIndicator.tsx
```diff
@@ -26,7 +26,7 @@ const PreVerifiedQuickIndicator = memo(({ type }: { type: PreVerifiedQuickIndica
       }>
       <UnFocusedDiv>
         <Indicator status={VerificationStatus.Verified} />
-        <img src={logo} alt={type.toString()} height='100%' />
+        <img data-testid='quick-indicator-logo' src={logo} alt={type.toString()} height='100%' />
```
steffenkleinle, line 29: :wrench: Please use `getByAltText` instead of images if possible (see https://testing-library.com/docs/queries/about/#priority). Same for other occurrences.

administration/src/bp-modules/applications/__tests__/ApplicationStatusBar.test.tsx
steffenkleinle: Mocking Icon/Button worked for me to avoid this error:
``` ts
jest.mock('@blueprintjs/core', () => ({
  ...jest.requireActual('@blueprintjs/core'),
  Icon: () => 'icon',
  Button: () => 'button',
}))
```
You could replace these strings with actual mocks to still be able to assert/test on specific button behavior. You could then even put this to the `jest.setup.ts` file to avoid having to mock it separately in every file.

One note: For me actually only the first test in this file logged an error, everything else in this file had a clean output.

steffenkleinle (commented):
There are lots of  tests asserting on test ids, this is not recommend (if there are other possibilities). Please avoid that and try to find other ways for asserting.

administration/src/bp-modules/applications/__tests__/VerificationsView.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+    expect(getByTestId('no-verifications-hint').textContent).toBe('(keine)')
+  })
+
+  it('should render a list of verification items', async () => {
+    const { getAllByTestId, getByText } = renderWithTranslation(
+      <VerificationsView verifications={verificationsMixed} />
+    )
+    await act(async () => null) // Blueprint5.Icon update
+    expect(getByText('Bestätigung(en) durch Organisationen:')).toBeTruthy()
+    expect(getAllByTestId('verification-list-item')).toHaveLength(3)
```
f1sh1918, line 21: @steffenkleinle i know that it also should be possible to use here 
```
 expect(queryAllByRole('li')).toHaveLength(3)
 ```
 But for me often it does not work, even the items are in the dom

File: administration/src/bp-modules/applications/__tests__/ApplicationStatusBar.test.tsx (hunk 2 above)
Comment (f1sh1918) [reply administration/src/bp-modules/applications/__tests__/ApplicationStatusBar.test.tsx]: i had to mock this in the particular test where its needed. 
If you put it in the `jest.setup` won't work since all buttons (where you have tests for) will only be a 'button' string. So no texts or anything can be checked

Comment (steffenkleinle) [reply administration/src/bp-modules/applications/__tests__/ApplicationStatusBar.test.tsx]: > You could replace these strings with actual mocks

See my comment above. The code above was just an example how to handle this in general. To be able to normally assert you have to mock the button with some react component that actually calls the on click method.

File: administration/src/bp-modules/applications/components/VerificationListItem.tsx [hunk 6]
```diff
@@ -0,0 +1,65 @@
...
+    ? `${t('rejectedOn')} ${new Date(verification.rejectedDate).toLocaleString('de')}`
+    : t('pending')
+  const text = verification.verifiedDate
+    ? `${t('verifiedOn')} ${new Date(verification.verifiedDate).toLocaleString('de')}`
+    : unverifiedText
+  const unverifiedColor = verification.rejectedDate ? Colors.RED2 : Colors.ORANGE2
+  const color = verification.verifiedDate ? Colors.GREEN2 : unverifiedColor
+
+  return (
+    <ListItem $color={color} data-testid='verification-list-item'>
```
steffenkleinle, line 36: I think this is not used anymore
```suggestion
    <ListItem $color={color}>
```

administration/src/bp-modules/applications/VerificationsQuickIndicator.tsx
```diff
@@ -28,7 +28,7 @@ const VerificationQuickIndicator = memo(({ verifications }: { verifications: App
           {t('verified/pending/rejected')}
         </div>
       }>
-      <UnFocusedDiv>
+      <UnFocusedDiv data-testid='indicator-container'>
```
steffenkleinle, line 31: I think this is not used anymore 
```suggestion
      <UnFocusedDiv>
```

Issue #1871, opened 2025-01-20
Remove logging for applications


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Related to this issue we added some logging
https://github.com/digitalfabrik/entitlementcard/issues/1223
Now the error disappeared due to improved mail template (links can be better resolved)
We don't need this logging anymore and it just pollutes our backend log

**Describe the solution you'd like**
- remove the logging we added here https://github.com/digitalfabrik/entitlementcard/pull/1553/files

PR #1890, opened 2025-02-03
1871: Remove logging applications

f1sh1918:
### Short description

We added some logging for investigating some error messages. It appeared that there were some broken links f.e. a particular mail link couldn't be resolved. We added proper html mail templates which seem to have fixed that. We didn't receive any error messages in the backend so far.

### Proposed changes

- remove logging

### Side effects

- none

### Testing

1. Test if application creation still works and the particular mails for application and verification are still be send

### Resolved issues

Fixes: #1871


seluianova (approved):

steffenkleinle (approved):
LGTM

Issue #1868, opened 2025-01-16
CSV bulk card creation works without mandatory extensions


f1sh1918:
**Describe the bug**
If we import csv card data for bavaria and not all mandatory card extensions are set f.e. card type, the csv import still works and just indicating an error

**To Reproduce**
1. Login to bavaria card administration
2. use the sozialpass nuernberg bulk resource (administration/resources) or delete the column "Kartentyp"
4. See in table but card be created

**Expected behavior**
As the import file doesn't include mandatory extensions like card type, there should be an error message, that the user should check for the correct column format

**Screenshots**
<img width="801" alt="image" src="https://github.com/user-attachments/assets/115bd325-9b06-4deb-8ee8-02f79c08d130" />

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] macos
 - Browser [e.g. chrome, safari] chrome

PR #1891, opened 2025-02-04
1868: Check mandatory extensions

f1sh1918:
### Short description

Currently we accept creating cards from csv even if some mandatory extensions are not included in the csv (like startDay or cardType)

### Proposed changes

- add a mandatory flag to each extension
- add a check in `isValid()` for cards that checks if cards has all mandatory extensions of the project
- add units tests for extensions in card.ts
- add a unit test for the csv freinet import
- allow null values in `startDay` since this could happen due to csv import of empty values and some validation checks would break

**Note:** In the future it may make sense to set the mandatory flags not on the extension directly but in the cardConfig for each project. So far i don't see the use case. Just to keep that in mind that we may refactor our extension config in the project config if its getting more complex

### Side effects

- csv import may break for the particular project
- card creation in general could break since another check for `isValid()` was added

### Testing

1. Test the csv import for nuernberg and for bayern (normal & freinet)
2. Check that cards with missing extensions can't be created. Check that the values that can't be resolved will be marked red in the table
Bayern: `Kartentyp`
NBG: `Startdatum`, `Geburtsdatum`, `Pass-ID`
3. Test if single card creation works for all projects
### Resolved issues

Fixes: #1868
Fixes: #1893


f1sh1918:
I've realized that the `initializeCardFromCsv` only initializes fields and extensions that are valid.
Thats why i added a check if all required extensions are there and then the error disappears.

The problem is that the issue in the csv will not be shown to the user and tell what to fix.

administration/src/bp-modules/cards/AddCardForm.tsx
```diff
@@ -28,6 +28,8 @@ const AddCardForm = ({ card, onRemove, updateCard }: CreateCardsFormProps): Reac
   const today = PlainDate.fromLocalDate(new Date())
   const { t } = useTranslation('cards')
 
+  console.log(card)
```
steffenkleinle, line 31:
```suggestion
```

administration/src/cards/Card.ts
```diff
@@ -131,10 +131,28 @@ export const isExpirationDateValid = (card: Card, { nullable } = { nullable: fal
...
+export const isValid = (
+  card: Card,
+  cardConfig: CardConfig,
+  { expirationDateNullable } = { expirationDateNullable: false }
+): boolean => {
+  console.log('fullname:', isFullNameValid(card))
+  console.log(
+    'expirationDate:',
+    isExpirationDateValid(card, { nullable: expirationDateNullable }) || hasInfiniteLifetime(card)
+  )
```
steffenkleinle, line 148:
```suggestion
```

administration/src/cards/Card.ts
```diff
@@ -131,10 +131,28 @@ export const isExpirationDateValid = (card: Card, { nullable } = { nullable: fal
   )
 }
 
-export const isValid = (card: Card, { expirationDateNullable } = { expirationDateNullable: false }): boolean =>
-  isFullNameValid(card) &&
-  getExtensions(card).every(({ extension, state }) => extension.isValid(state)) &&
-  (isExpirationDateValid(card, { nullable: expirationDateNullable }) || hasInfiniteLifetime(card))
+export const cardHasAllMandatoryExtensions = (card: Card, cardConfig: CardConfig): boolean => {
+  const mandatoryExtensions = cardConfig.extensions.filter(extension => extension.isMandatory)
+  return mandatoryExtensions.every(extension => Object.keys(card.extensions).includes(extension.name))
```
steffenkleinle, line 136: :thinking: I guess our current architecture prevents it, but in theory it would be possible that the value is still null or undefined as you are only checking the key, which would not really count as `has all mandatory extensions` imo. Works fine atm though.

steffenkleinle (approved):
Tested on firefox, works as expected. Nice work :)

seluianova (approved):
lgtm

Issue #1739, opened 2024-11-04
Add error message when user tries to activate an expired card via deep link


seluianova:
**Is your feature request related to a problem? Please describe.**
When a user tries to activate an expired card via deep link, the following error message is displayed:
`Die Aktivierung befindet sich in einem ungültigen Zustand`
I believe would be nice to display a message like `This card has been expired`
Currently we have a separate message when a user activates an expired card via QR code scanning, but it’s not applied to deep link activation.

**Describe the solution you'd like**
- add a separate error message (some refactoring might be required)

**Describe alternatives you've considered**
Leave as is

**Additional context**

PR #1895, opened 2025-02-06
1739: Show an error message for expired card via deep link

bahaaTuffaha:
### Short description

When a user tries to activate an expired card via deep link, the following error message is displayed:
Die Aktivierung befindet sich in einem ungültigen Zustand
I believe would be nice to display a message like This card has been expired.

### Proposed changes

- I added a new `DeepLinkActivationStatus` called `cardExpired`
- I investigated the `getActivationCode` method because it keeps catching on `QrCodeParseException` so I commented the code responsible for checking the expiration date at `verification_qr_code_processor.dart`.

### Side effects

- `verification_qr_code_processor.dart` assertConsistentCardInfo method.

### Testing

- Create a card for bayern with expired date (you can return true at `isExpirationDateValid` - `card.ts` just to pass)
- Copy the the deeplink from the pdf to the mobile browser.
- Click on Karte aktivieren.. should show the card with gray button and an error message saying:
`Ihre Karte ist abgelaufen. Unter "Weitere Aktionen" können Sie einen Antrag auf Verlängerung stellen.`

### Resolved issues

Fixes: #1739


f1sh1918:
We can not just remove the consistency of the cardInfo check, because its also used for the verification process
I think have to refactor some more stuff.

Currently you could just show an error dialog if an error occurs here:
```
on QrCodeParseException catch (e, _)
```
in the `deeplink_activation`.

Since this is not very consistent, we could think about setting an error message as a state while catching.
Instead of showing this "random" status:
`DeepLinkActivationStatus.invalidLink => t.deeplinkActivation.activationInvalid`
Which should be called "invalidCardInfo" not `invalidLink` we could show what is in the state and just if the state is null, we can fallback on this random `activationInvalid` message

frontend/lib/activation/deeplink_activation.dart
```diff
@@ -193,6 +202,11 @@ DynamicActivationCode? getActivationCode(BuildContext context, String base64qrco
   try {
     final activationCode = const ActivationCodeParser().parseQrCodeContent(const Base64Decoder().convert(base64qrcode));
     return activationCode;
+  } on CardExpiredException catch (e,_){
+    debugPrint('Die Karte ist abgelaufen: ${e.expiry}');
+    final expirationDate = DateTime.utc(e.expiry.year, e.expiry.month, e.expiry.day).millisecondsSinceEpoch ~/1000;
+    final activationCode = DynamicActivationCode(info: CardInfo(fullName: e.fullName,expirationDay:expirationDate));
+    return activationCode;
```
f1sh1918, line 209: i wouldn't return the activationCode on exception but null.
You can pass an `updateErrorMessage` function that updates on a particular exception the error message

frontend/lib/identification/verification_workflow/verification_qr_code_processor.dart
```diff
@@ -97,6 +97,7 @@ extension DateComparison on DateTime {
 
 class CardExpiredException extends QrCodeParseException {
   final DateTime expiry;
+  final String fullName;
```
f1sh1918, line 100: :question: i think fullname is not needed here

f1sh1918 (commented):
👍 Good work. Works great on ios.
Just some minor things

frontend/lib/identification/verification_workflow/verification_qr_code_processor.dart
bahaaTuffaha: I can see why eslint is needed now 😅

frontend/assets/l10n/app_de.json
```diff
@@ -77,7 +77,10 @@
     "invalidCode": "Ihr Aktivierungscode ist ungültig.",
     "alreadyExists": "Diese Karte wurde bereits auf diesem Gerät aktiviert.",
     "activationSuccessful": "Die Aktivierung war erfolgreich.",
-    "activationInvalid": "Die Aktivierung befindet sich in einem ungültigen Zustand."
+    "activationInvalid": "Die Aktivierung befindet sich in einem ungültigen Zustand.",
+    "cardExpiredExceptionMessage": "Ihre Karte kann nicht aktiviert werden, weil sie bereits abgelaufen ist",
```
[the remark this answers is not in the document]
bahaaTuffaha: Aha.. I got overwhelmed at the beginning from the generated translation files... nice translation package.🙂

f1sh1918 (approved):
Looks great. Tested on ios

frontend/lib/activation/deeplink_activation.dart
```diff
@@ -8,15 +8,17 @@ import 'package:ehrenamtskarte/identification/id_card/id_card_with_region_query.
...
 import 'package:ehrenamtskarte/proto/card.pb.dart';
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
+import 'package:intl/intl.dart';
 import 'package:provider/provider.dart';
 
 enum DeepLinkActivationStatus {
   /// Link is invalid
-  invalidLink,
+  invalidCardInfo,
```
steffenkleinle, line 21: :upside_down_face:  I think there are more reasons why the activation could be invalid, not just due to invalid card infos
``` suggestion
// Activation is invalid
invalid,
```

frontend/lib/activation/deeplink_activation.dart
```diff
@@ -189,15 +199,20 @@ class _WarningText extends StatelessWidget {
...
 
-DynamicActivationCode? getActivationCode(BuildContext context, String base64qrcode) {
+DynamicActivationCode? getActivationCode(
+    BuildContext context, String base64qrcode, Function(String message) updateErrorMessage) {
   try {
     final activationCode = const ActivationCodeParser().parseQrCodeContent(const Base64Decoder().convert(base64qrcode));
     return activationCode;
+  } on CardExpiredException catch (e, _) {
+    updateErrorMessage(
+        t.deeplinkActivation.cardExpiredExceptionMessage(expiryDate: DateFormat('dd.MM.yyyy').format(e.expiry)));
```
steffenkleinle, line 209: :thought_balloon: We should avoid to use hardcoded date formats like this. However, since we do this in a lot of places already, I created a separate [issue](https://github.com/digitalfabrik/entitlementcard/issues/1942) so no need to change anything here.

steffenkleinle (approved):
Nice work, tested on android!

frontend/lib/activation/deeplink_activation.dart
bahaaTuffaha: @f1sh1918 what do you think?

f1sh1918: i think its a valid suggestion, since we also have a fallback message if an unknown error happens

Issue #1762, opened 2024-11-07
Add freinet agency information


ztefanie:
For this feature #1751
To be able to push data to freinet (application/card data) we need an `agencyId` and `accessKey`
We have an api from freinet where we can fetch this information and map it via the `regionIdentifiers` (ars) to our regions

- Fetch the freinet `agencyId` and `accessKey` in regular cycles and the maybe (or not) save them to our system. Evaluate if caching is enough or not. Updates in the freinet system are possible and need to lead to updates in our system, e.g. if a region is added to freinet.
- show the `agencyId` and the `accessKey` in the region settings if they are available

f1sh1918:
Maybe I overseen sth but why do we have to create an input field for the Client id? We fetch the regions including a region identifier (not available yet), access Key, mandantenId(freinet) with that we can map the Region to our RegionId and have All required information to push the card and user data. 
I thought we keep this information in memory.

f1sh1918:
Since we need the `agencyId` for mapping accepting stores and for pushing data application data to freinet i think it makes sense to call the freinet api on backend start and store the `agencyId` in the regions table
This id will not change frequently, so i think on release to refetch it, it should be enough

f1sh1918:
@ztefanie can you explain what bullet point 2 and 3 mean?
I think we just fetch this ids and i see no point in adding them manually?
Or do you mean if we cannot find any match?

ztefanie:
yes, i think the idea was to map the freinet region ids with the eak region id's manually. But sure, if this can be done automatically, it is even better.

f1sh1918:
Currently there are lots of wrong `ars` numbers so we can only map 42 regions, so far. I contacted freinet to check that

Issue #1899, opened 2025-02-12
Show freinet agency information in region settings


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a region admin i want to be able to see my freinet agency information in regions settings and be able to activate the data transfer

**Describe the solution you'd like**
- show a section in the regions settings that show the freinet agency information if they are available
- show a checkbox to activate the data transfer (will be used for sending card a application data in the future)
- create a mutation that updates the activated value and stores it in the database table

~~blocked by #1762~~

PR #1902, opened 2025-02-13
1762: Add agency information to region

f1sh1918:
### Short description

Here we add required agency information from freinet to our regions. We need the `agencyId` and `accessKey` to push applicant and card data to freinet in a separate issue

### Proposed changes

- add new database table with freinet agency information
- add FreinetAgencyLoader that fetches agency information on backend startup
- store the agency information in the database and map it to a region
- add a query to get agency information by regionId
- add api tests

### Note
Here is a separate pr to provide backend config and the freinet api access key on the server
https://git.tuerantuer.org/DF/salt/pulls/309/files
Frontend implementation will be done in #1899 

### Side effects

- none

### Testing

1. First you have to add the freinet information & api access key to your `config.local.yml`
You can find the accessKey here: https://passbolt.tuerantuer.org/app/passwords/view/0b693665-7176-4524-a6dd-206fad7c4d8d
2. Run migration
3. Run backend start
4. Agency data should be stored in the database

### Resolved issues

Fixes: #1762


backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetAgenciesLoader.kt
```diff
@@ -0,0 +1,85 @@
...
+
+            return transformAndFilterAgencyData(xmlMapper.readValue(response, XMLAgencies::class.java))
+        } catch (e: Exception) {
+            logger.error("Couldn't fetch agency information: ", e)
+            return emptyList()
+        }
+    }
+
+    private fun transformAndFilterAgencyData(agencies: XMLAgencies): List<FreinetApiAgency> {
+        return agencies.agencies.filter { it.agencyId != null && it.ars != null && it.accessKey != null && it.agencyName !== null }.map {
```
steffenkleinle, line 76: :upside_down_face: Could shorten this with [mapNotNull](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/map-not-null.html)

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetAgencyQueryService.kt
```diff
@@ -0,0 +1,40 @@
...
+                ?: throw UnauthorizedException()
+        val projectEntity = ProjectEntity.find { Projects.project eq project }.firstOrNull()
+            ?: throw ProjectNotFoundException(project)
+
+        if (projectEntity.project != EAK_BAYERN_PROJECT) {
+            throw NotEakProjectException()
+        }
+        if (!Authorizer.mayViewFreinetAgencyInformationInRegion(admin, regionId)) {
+            throw ForbiddenException()
+        }
```
steffenkleinle, line 36: :thinking: Shouldn't we move this to the start of the function to avoid doing anything the user is not allowed to do?

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/schema/types/FreinetApiAgency.kt
```diff
@@ -0,0 +1,8 @@
+package app.ehrenamtskarte.backend.freinet.webservice.schema.types
+
+data class FreinetApiAgency(
+    val arsList: List<String>,
```
steffenkleinle, line 4: :wrench: Can we rename this to something meaningful? What does `ars` stand for?

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/database/Schema.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.jetbrains.exposed.dao.IntEntity
+import org.jetbrains.exposed.dao.IntEntityClass
+import org.jetbrains.exposed.dao.id.EntityID
+import org.jetbrains.exposed.dao.id.IntIdTable
+
+object FreinetAgencies : IntIdTable() {
+    val regionId = reference("regionId", Regions)
+    val agencyId = integer("agencyId").uniqueIndex()
+    val agencyName = varchar("agencyName", 255)
+    val apiAccessKey = varchar("apiAccessKey", 10)
```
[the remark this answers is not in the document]
f1sh1918: good point. i will increase it. what is a good length?

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/database/Setup.kt
```diff
@@ -0,0 +1,25 @@
...
+        FreinetAgenciesEntity.new {
+            regionId = regionEntity.id
+            agencyId = agency.agencyId
+            agencyName = agency.agencyName
+            apiAccessKey = agency.apiAccessKey
+        }
+    } else {
+        dbAgency.agencyId = agency.agencyId
+        dbAgency.agencyName = agency.agencyName
+        dbAgency.apiAccessKey = agency.apiAccessKey
```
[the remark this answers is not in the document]
f1sh1918: Not this is a flag that will be manually set by the user and is `false` per default ` #1899

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/schema/types/FreinetApiAgency.kt
f1sh1918: AmtlicherRegionalSchlüssel

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -47,19 +51,25 @@ fun insertOrUpdateRegions() {
...
                     name = eakRegion[1]
                     prefix = eakRegion[0]
                     regionIdentifier = eakRegion[2]
                     website = eakRegion[3]
                 }
             } else {
+                regionEntity = dbRegion
                 dbRegion.name = eakRegion[1]
                 dbRegion.prefix = eakRegion[0]
                 dbRegion.website = eakRegion[3]
```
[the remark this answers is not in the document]
f1sh1918: hm i do not get why the order matters here and the `dbRegion` has to be moved to the end. Can you explain that to me @steffenkleinle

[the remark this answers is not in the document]
steffenkleinle: See https://kotlinlang.org/docs/control-flow.html#if-expression (under `Branches of an if expression can be blocks. In this case, the last expression is the value of a block`). Basically, with assignments with an if, the last value in the block is assigned to the variable.

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/database/Setup.kt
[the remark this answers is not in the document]
steffenkleinle: Okay, thanks for explaining.

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/schema/types/XMLAgencies.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement
+
+@JacksonXmlRootElement(localName = "agencies")
+data class XMLAgencies(
+    @JsonProperty("mandant")
+    @JacksonXmlElementWrapper(useWrapping = false)
+    var agencies: ArrayList<XMLAgency> = ArrayList()
```
steffenkleinle, line 11: :question: Is there any reason this needs to be `var` instead of `val`?
```suggestion
    var agencies: ArrayList<XMLAgency> = ArrayList()
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/schema/types/XMLAgency.kt
```diff
@@ -0,0 +1,19 @@
+package app.ehrenamtskarte.backend.freinet.webservice.schema.types
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlCData
+
+data class XMLAgency(
+    @JsonProperty("agenturid")
+    @JacksonXmlCData
+    var agencyId: String?,
```
steffenkleinle, line 9: Same as above and below.

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetAgenciesLoader.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private fun transformAndFilterAgencyData(agencies: XMLAgencies): List<FreinetApiAgency> {
+        return agencies.agencies.mapNotNull {
+            FreinetApiAgency(
+                agencyId = it.agencyId!!.toInt(),
+                agencyName = it.agencyName!!,
+                apiAccessKey = it.accessKey!!,
+                officialRegionalKeys = it.officialRegionalKeys?.split(",") ?: emptyList()
+            )
```
steffenkleinle, line 82: :x: Sorry, maybe my previous comment was misunderstandable. This does not really work/will crash if a property is null since you never check whether the properties are null and just use a non-null assertion (`!!`) (even though it seems like nothing is actually null).
`mapNotNull` only filters out `null` after mapping, therefore it does act exactly like `map` in this case. You could do the following (if changing `var` to `val` as mentioned above):
```suggestion
        return agencies.agencies.mapNotNull {
            return@mapNotNull if (it.agencyId != null && it.agencyName != null && it.accessKey != null) {
                FreinetApiAgency(
                    agencyId = it.agencyId.toInt(),
                    agencyName = it.agencyName,
                    apiAccessKey = it.accessKey,
                    officialRegionalKeys = it.officialRegionalKeys?.split(",") ?: emptyList()
                )
            } else null
```

steffenkleinle (changes requested):
Tested, works as expected. Nice work!

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/schema/types/XMLAgencies.kt
f1sh1918: no i think not but your suggested changes is wrong :)

f1sh1918:
> Tested, works as expected. Nice work! The api key is quite short.... 🤔

I know. but thats what we got from freinet. You can tell them your concern about security issues

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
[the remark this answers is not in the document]
f1sh1918: thx for the reference!

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetAgenciesLoader.kt
steffenkleinle: Just takes some time to get used to kotlin magic I guess :D

steffenkleinle (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetAgencyQueryService.kt
```diff
@@ -0,0 +1,40 @@
...
+        val projectEntity = ProjectEntity.find { Projects.project eq project }.firstOrNull()
+            ?: throw ProjectNotFoundException(project)
+
+        if (projectEntity.project != EAK_BAYERN_PROJECT) {
+            throw NotEakProjectException()
+        }
+        val agency = FreinetAgencyRepository.getFreinetAgencyByRegionId(regionId)
+        agency?.let { FreinetAgency(agencyId = it.agencyId, apiAccessKey = agency.apiAccessKey, agencyName = agency.agencyName, dataTransferActivated = agency.dataTransferActivated) }
```
andrew8er, line 38: 🔧 These two lines can be pulled together using a safe navigation operator, declaring a function level variable is not necessary. I would however, place all arguments of the `FreinetAgency()` ctor on separate lines.

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetAgencyQueryService.kt
```diff
@@ -0,0 +1,40 @@
...
+                ?: throw UnauthorizedException()
+        if (!Authorizer.mayViewFreinetAgencyInformationInRegion(admin, regionId)) {
+            throw ForbiddenException()
+        }
+        val projectEntity = ProjectEntity.find { Projects.project eq project }.firstOrNull()
+            ?: throw ProjectNotFoundException(project)
+
+        if (projectEntity.project != EAK_BAYERN_PROJECT) {
+            throw NotEakProjectException()
+        }
```
andrew8er, line 36: 🔧 The `if` block below can be pulled up into a `?.let {}` block before the Elvis operator, thus no variable needs to be declared.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0026_AddFreinetAgenciesTable.kt
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Adds freinetagencies table
+ */
+@Suppress("ClassName")
+internal class V0026_AddFreinetAgenciesTable : Migration() {
+    override val migrate: Statement = {
+        exec(
+            """
+                CREATE TABLE freinetagencies (
+                    id SERIAL PRIMARY KEY ,
```
andrew8er, line 15: 💭 The recommended way to declare sequence ID columns for modern PostgreSQL is using
[identity columns](https://www.postgresql.org/docs/17/ddl-identity-columns.html).

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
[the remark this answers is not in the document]
andrew8er: 🔧 Since this is an expression, what about:

```kt
val regionEntity: RegionEntity = dbRegion
    ?.apply {
        name = eakRegion[1]
        prefix = eakRegion[0]
        website = eakRegion[3]
    }
    ?: RegionEntity.new {
        projectId = eakProject.id
        name = eakRegion[1]
        prefix = eakRegion[0]
        regionIdentifier = eakRegion[2]
        website = eakRegion[3]
    }
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/Setup.kt
```diff
@@ -59,6 +63,11 @@ fun insertOrUpdateRegions() {
...
+                dbRegion
+            }
+            val agency = agencies.find { agency -> agency.officialRegionalKeys.any { it.startsWith(eakRegion[2]) } }
+            if (agency != null) {
+                insertOrUpdateFreinetRegionInformation(agency, dbFreinetRegionInformation, regionEntity)
```
andrew8er, line 70: 🔧 The code in the `if` block can be conditionally applied with `?.let {}`. No function level variable necessary here.

andrew8er (approved):
Some stylistic suggestions.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -152,6 +152,11 @@ object Authorizer {
         return user.projectId.value == projectId && user.role == Role.PROJECT_STORE_MANAGER.db_value
     }
 
+    fun mayViewFreinetAgencyInformationInRegion(user: AdministratorEntity, regionId: Int): Boolean {
+        return user.role == Role.REGION_ADMIN.db_value && ProjectEntity.find { Projects.project eq EAK_BAYERN_PROJECT }
+            .single().id.value == user.projectId.value && user.regionId?.value == regionId
+    }
+
```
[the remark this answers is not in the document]
steffenkleinle: Good suggestions, I already opened a PR to improve this file though: #1944
I think it would therefore be nice to do further changes there to avoid merge conflicts :)

backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
```diff
@@ -35,7 +36,8 @@ data class ProjectConfig(
     val timezone: ZoneId,
     val selfServiceEnabled: Boolean,
     val smtp: SmtpConfig,
-    val matomo: MatomoConfig?
+    val matomo: MatomoConfig?,
+    val freinet: FreinetConfig?
```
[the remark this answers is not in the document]
steffenkleinle: Nice, is there a linting/formatter to automate/force these changes?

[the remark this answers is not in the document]
andrew8er: `ktlint` can do this, as well as the IntelliJ IDEA formatter, but this will generate a lot of changes. Ideally, this would be done, when there are no (or at least very few) PRs contributin to `backend`.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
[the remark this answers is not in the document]
f1sh1918: i will keep the current state in my pr

backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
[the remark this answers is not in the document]
f1sh1918: I'm not sure if the gradle ktlint plugin can do that but maybe i didn't find the correct package
https://github.com/ankidroid/Anki-Android/pull/14997#issuecomment-1946306075
But maybe we can just provide some additional rules in the `.editorconfig`
Feel free to open a ticket for that.

Currently adding trailing commas will not help because ktlint formating will remove it. So i will keep it as it is

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/database/repos/FreinetAgencyRepository.kt
```diff
@@ -0,0 +1,10 @@
+package app.ehrenamtskarte.backend.freinet.database.repos
+
+import app.ehrenamtskarte.backend.freinet.database.FreinetAgencies
+import app.ehrenamtskarte.backend.freinet.database.FreinetAgenciesEntity
+
+object FreinetAgencyRepository {
+    fun getFreinetAgencyByRegionId(regionId: Int): FreinetAgenciesEntity? {
+        return FreinetAgenciesEntity.find { FreinetAgencies.regionId eq regionId }.singleOrNull()
```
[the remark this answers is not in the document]
f1sh1918: Alright i guess this is what you mean?
```js
fun getFreinetAgencyByRegionId(regionId: Int): FreinetAgenciesEntity? =
        FreinetAgenciesEntity.find { FreinetAgencies.regionId eq regionId }.singleOrNull()
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetAgencyQueryService.kt
f1sh1918: hm can you give me an example how to do that @andrew8er 
When i put the let after `firstOrNull()` i dunno how i can include both exceptions in the expression

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetAgenciesLoader.kt
```diff
@@ -0,0 +1,85 @@
...
+                if (response.status != HttpStatusCode.OK) {
+                    throw Exception(response.status.toString())
+                }
+            }
+        }
+    }
+
+    fun loadAgenciesFromXml(projectConfigs: List<ProjectConfig>): List<FreinetApiAgency> {
+        val bayernConfig =
+            projectConfigs.find { it.id == EAK_BAYERN_PROJECT } ?: throw NotFoundException("Project config not found")
```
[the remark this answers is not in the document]
f1sh1918: hm i think the line break is not really needed here

f1sh1918:
> Some stylistic suggestions.

Thanks for the great review. I think i addressed all comments.
On one or two i was a bit unsure i addressed it correct.
So it would be good to check my last commit

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetAgencyQueryService.kt
andrew8er: This should be equivalent, if I'm not mistaken:
```kt
        ProjectEntity.find { Projects.project eq project }.firstOrNull()
            ?.let {
                if (it.project != EAK_BAYERN_PROJECT) {
                    throw NotEakProjectException()
                }        
            }
            ?: throw ProjectNotFoundException(project)
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0026_AddFreinetAgenciesTable.kt
andrew8er: Yes, an identity column can be (but is not necessarily) a primary key. I don't know if it is a good idea to change the way primary keys are defined mid-project, although I do not see any larger problems with it either. I was just trying to give a hint here.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0026_AddFreinetAgenciesTable.kt
```diff
@@ -12,7 +12,7 @@ internal class V0026_AddFreinetAgenciesTable : Migration() {
         exec(
             """
                 CREATE TABLE freinetagencies (
-                    id SERIAL PRIMARY KEY ,
+                     id bigint GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
```
andrew8er, line 15: 🙃 There's an extra space here.

andrew8er (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetAgencyQueryService.kt
f1sh1918: ah alright i thought in can be solved without the using the `if`condition. thx

Issue #1610, opened 2024-09-02
Possible resource limitation for very large import files


Docmale:
**Describe the bug**
Error: 
![image (2)](https://github.com/user-attachments/assets/d27b89d6-9eb1-4e87-8f91-c9aaeacdf74d)
I got the error with two files with 299 and 300 entries. When I tried again with 150 entries and later with 300 entries, I could no longer reproduce the behaviour. The error only occurs again when I start another generation in parallel. So it has something to do with resources?

Docmale:
The effect occurs with Freinet and standard CSV import

ztefanie:
Does not occur every time, just sometimes.

f1sh1918:
created cards from csv (300 entries) in two browser tabs in parallel. Couldn't reproduce the error with chrome

seluianova:
Couldn't reproduce locally, but the issue is reproduced on staging.
The error occurs because of the timeout (it seems to be 1 minute).
`POST https://api.staging.entitlementcard.app/ 504 (Gateway Timeout)`
I guess it's limited on the infrastructure level (?)
@f1sh1918 do you know if we can (and want to) adjust it?

f1sh1918:
> Couldn't reproduce locally, but the issue is reproduced on staging.
> The error occurs because of the timeout (it seems to be 1 minute).
> `POST https://api.staging.entitlementcard.app/ 504 (Gateway Timeout)`
> I guess it's limited on the infrastructure level (?)
> @f1sh1918 do you know if we can (and want to) adjust it?

Can you please ask that in our tech channel. I do not know that.

seluianova:
> This can also be an issue for larger csv files maybe? Not sure

Indirectly yes. The error occurs when it takes more than 1 minute to process a request. Which is most likely to happen when the file is large.

seluianova:
> This can also be an issue for larger csv files maybe? Not sure

@f1sh1918 it could be, but we don't allow to upload more than 300 entries in one file

seluianova:
Another analysis round:
There seems to be a significant performance degradation on the staging env.
Processing of a single request with 300 entries takes ~40 seconds locally:
![Image](https://github.com/user-attachments/assets/a25b71b2-d0a4-45d6-93fa-78f55e553579)
While the exact same request on staging takes ~60 seconds:
![Image](https://github.com/user-attachments/assets/6bbcf938-5d86-4554-b731-fc5c24bc75e9)

[...]

PR #1904, opened 2025-02-16
1610: Optimize cards creation performance

seluianova:
### Short Description
We have a performance issue in the staging environment.
When creating cards in bulk (Karten erstellen --> Mehrere Karten importieren), if a csv file contains 300 entries, the request processing time takes almost a minute.
If a user imports multiple large files in parallel, processing may take more than a minute, resulting in a timeout error.

### Proposed Changes
97% of the processing time during the card creation is spent hashing the activation code.
I suggest reducing the hash cost to 10 (this is the default value), this will cut the query processing time in half.

Pls check out the discussion in the issue comments for more details.
The discussion there is still open, so I'm not sure if everyone agrees, but I'm opening this PR as a suggestion to move the task forward anyway.

### Side Effects
The hashing complexity is reduced, but since we're hashing a randomly generated array of bytes, I tend to think it's not a problem.

### Testing
The initial issue from the ticket description can only be reproduced in the staging env.
Locally we can test the following:

1. Login as region admin to bayern 
2. Go to Karten erstellen --> Mehrere Karten importieren
3. Try uploading a large csv file (300 records), see if the processing time is reduced compared to the main branch.

Also can be checked:
- Old cards (created on the previous release version or main branch) can be activated and verified in the app.
- New cards can be created, activated and verified in the app.

### Resolved Issues
Fixes: #1610

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/service/CardActivator.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/service/CardActivator.kt
@@ -6,7 +6,7 @@ import com.eatthepath.otp.TimeBasedOneTimePasswordGenerator
 import javax.crypto.KeyGenerator
 
 object CardActivator {
-    private const val cost = 11
+    private const val cost = 10
 
     @Synchronized
     fun generateTotpSecret(): ByteArray {


f1sh1918 (approved):
I tested the card creation and activation of old and new cards and it still works.

I didn't stop the time if bulk creation now goes faster :)
Thx for the investigation 👍

steffenkleinle:
Perhaps @maxammann or @michael-markl could have a look at this?

maxammann:
imo the fix is to add background processing. Generally the project will need something like that in the future either way.

The PR is a good workaround thought.

With that fix the processing of cards is likely twice as fast as before.

Issue #1909, opened 2025-02-17
Using project config in extension breaks the app


steffenkleinle:
### Describe the Bug

Using project config in extension breaks the app due to a circular dependency issue:
extension -> ProjectConfigProvider -> getProjectConfig -> project config -> extension
This prevents using the project config in an extension (e.g. in #1858).

### Steps to Reproduce

1. Add `  const projectConfig = useContext(ProjectConfigContext)` to an extension
2. Run the tests
3. See them fail 

### Expected Behavior

Everything works.

### Related Issues

Occurred in #1858.

### Additional Information

N/A

PR #1910, opened 2025-02-17
1909: Fix circular dependency to use of project config in extensions

steffenkleinle:
### Short Description

Fix circular dependency to allow for use of project config in extensions (including for testing).

### Proposed Changes

- Pass a project config to ProjectContextProvider instead of using getProjectConfig directly
- Adjust the tests to easily set a project config

### Side Effects

The `setOverrideProject` has been removed as it is unsafe and should be avoided (e.g. for tests).

### Testing

Everything should still work for the different projects.

### Resolved Issues

Fixes: #1909


f1sh1918 (approved):
Thx for the quick fix
👍  works nice 
Tested card creation for all projects

bahaaTuffaha (approved):
I tested it with the context included.. it works without issues
Thanks for the fix 🙂

PR #1916, opened 2025-02-18
Fix self service form test

f1sh1918:
### Short Description

fixes the unit test by providing correct project config


seluianova (approved):

steffenkleinle (approved):
You could just use the new functionality to set the project config instead of manually wrapping in the ProjectConfigProvider:
``` ts
    const { getByLabelText, getByPlaceholderText, getByTestId, getByText } = renderWithTranslation(
      Component,
      { wrapper, projectConfig: koblenzConfig }
    )
```

Issue #1799, opened 2024-11-25
Adjust useCardGenerator hook


seluianova:
**Is your feature request related to a problem? Please describe.**
Follow up to https://github.com/digitalfabrik/entitlementcard/pull/1749

**Describe the solution you'd like**
Add the same functionality for the card initialization as in useCardGeneratorSelfService to the useCardGenerator hook and remove the old implementation from AddCardsFrom

**Describe alternatives you've considered**
Leave as is

**Additional context**
n/a

PR #1917, opened 2025-02-18
1799: Refactor useCardGenerator and useCardGeneratorSelfService

steffenkleinle:
### Short Description

Refactor useCardGenerator and useCardGeneratorSelfService hooks and other utilities.

### Proposed Changes

- Move card initialization from query params to useCardGenerator
- Extract code from hooks to new files useSendConfirmationMail and cardGenerationError and some old files
- Reuse code between useCardGenerator and useCardGeneratorSelfService
- Refactor createCards, CsvFactory, deleteCards, PdfFactory

### Side Effects

- Pdf download name for self service is now `KoblenzPass.pdf` instead of `KoblenzPass`
- Opening a csv in card import, navigating back (accepting the popup) leads to an error (however, this is the same on main and I was not able to reproduce this on staging).

### Testing

Test everything related to card generation (normal + self service) and csv and pdf generation.

### Resolved Issues

Fixes: #1799

### DIFF
--- a/administration/src/bp-modules/cards/hooks/useCardGenerator.ts
+++ b/administration/src/bp-modules/cards/hooks/useCardGenerator.ts
@@ -46,10 +46,10 @@ const useCardGenerator = (region: Region): UseCardGeneratorReturn => {
   const projectConfig = useContext(ProjectConfigContext)
   const [searchParams] = useSearchParams()
   const [state, setState] = useState(CardActivationState.input)
-  const [createCardsService] = useCreateCardsMutation()
-  const [deleteCardsService] = useDeleteCardsMutation()
+  const [createCardsMutation] = useCreateCardsMutation()
+  const [deleteCardsMutation] = useDeleteCardsMutation()
   const appToaster = useAppToaster()
-  const sendCardConfirmationMails = useSendCardConfirmationMails()
+  const sendConfirmationMails = useSendCardConfirmationMails()
   const { t } = useTranslation('errors')
   const [cards, setCards] = useState<Card[]>(() => {
     const headers = getCsvHeaders(projectConfig)
@@ -71,20 +71,20 @@ const useCardGenerator = (region: Region): UseCardGeneratorReturn => {
       setState(CardActivationState.loading)
 
       try {
-        codes = await createCards(createCardsService, projectConfig, cards, t, applicationId)
+        codes = await createCards(createCardsMutation, projectConfig, cards, t, applicationId)
         const dataUri = await generateFunction(codes, cards, projectConfig, region)
         downloadDataUri(dataUri, filename)
         cards.forEach(saveActivityLog)
 
         if (region.activatedForCardConfirmationMail) {
-          await sendCardConfirmationMails(codes, cards)
+          await sendConfirmationMails(codes, cards)
         }
 
         setState(CardActivationState.finished)
       } catch (error) {
         if (codes) {
           // Rollback
-          await deleteCards(deleteCardsService, region.id, codes, t).catch(reportErrorToSentry)
+          await deleteCards(deleteCardsMutation, region.id, codes, t).catch(reportErrorToSentry)
         }
         if (appToaster) {
           showCardGenerationError(appToaster, error, t)
@@ -96,11 +96,11 @@ const useCardGenerator = (region: Region): UseCardGeneratorReturn => {
     },
     [
       cards,
-      createCardsService,
-      deleteCardsService,
+      createCardsMutation,
+      deleteCardsMutation,
       projectConfig,
       appToaster,
-      sendCardConfirmationMails,
+      sendConfirmationMails,
       region,
       applicationId,
       t,
--- a/administration/src/bp-modules/cards/hooks/useCardGenerator.ts
+++ b/administration/src/bp-modules/cards/hooks/useCardGenerator.ts
@@ -1,7 +1,8 @@
 import { useCallback, useContext, useState } from 'react'
 import { useTranslation } from 'react-i18next'
+import { useSearchParams } from 'react-router-dom'
 
-import { Card, updateCard as updateCardObject } from '../../../cards/Card'
+import { Card, initializeCardFromCSV, updateCard as updateCardObject } from '../../../cards/Card'
 import { generateCsv, getCSVFilename } from '../../../cards/CsvFactory'
 import { generatePdf, getPdfFilename } from '../../../cards/PdfFactory'
 import createCards, { CreateCardsResult } from '../../../cards/createCards'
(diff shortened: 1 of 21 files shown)


administration/src/bp-modules/self-service/CardSelfServiceActivation.tsx
```diff
@@ -17,15 +20,21 @@ const StyledIconTextButton = styled(IconTextButton)`
...
   const { t } = useTranslation('selfService')
+  const deepLink = getCustomDeepLinkFromQrCode(projectConfig, {
+    case: 'dynamicActivationCode',
+    value: code.dynamicActivationCode,
+  })
+
   return (
     <Container>
-      <StyledIconTextButton onClick={downloadPdf}>
+      <StyledIconTextButton onClick={() => downloadPdf(code, `${projectConfig.name}.pdf`)}>
```
steffenkleinle, line 37: :question: So far the pdf name was only `projectConfig.name`, was that intended?

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -1,202 +1,119 @@
...
-    },
+  const sendCardConfirmationMails = useSendCardConfirmationMails()
+  const { t } = useTranslation('errors')
+  const [cards, setCards] = useState<Card[]>(() => {
+    const headers = getCsvHeaders(projectConfig)
+    const values = headers.map(header => searchParams.get(header))
+    return [initializeCardFromCSV(projectConfig.card, values, headers, region, true)]
   })
+  const rawApplicationId = searchParams.get('applicationIdToMarkAsProcessed')
+  const applicationId = rawApplicationId ? parseInt(rawApplicationId, 10) : null
```
steffenkleinle, line 59: These are the actual changes related to the issue :D

administration/src/bp-modules/self-service/hooks/useCardGeneratorSelfService.tsx
```diff
@@ -1,34 +1,22 @@
...
 import { useAppToaster } from '../../AppToaster'
-import FormErrorMessage from '../components/FormErrorMessage'
+import { showCardGenerationError } from '../../util/cardGenerationError'
 
-export enum CardSelfServiceStep {
-  form,
-  information,
-  activation,
-}
+export type SelfServiceCardGenerationStep = 'input' | 'loading' | 'information' | 'activation'
```
steffenkleinle, line 15: I faced the following issue with the enum and could not figure out how to solve this except changing it to a string type: https://github.com/webpack/webpack/issues/16146

administration/src/bp-modules/self-service/CardSelfServiceActivation.tsx
f1sh1918: hm if you test on recent staging for koblenz, you download a file that is a working pdf with the correct type extension

steffenkleinle: Hmm, maybe the browser automatically adds the extension based on the file type? Anyway, I think it does not hurt to make it explicit right?

The previous code:
``` ts
await downloadPdf(code, projectConfig.name)
```
with
```
  const downloadPdf = async (code: CreateCardsResult, fileName: string): Promise<void> => {
    const blob = await generatePdf([code], [selfServiceCard], projectConfig.pdf)
    downloadDataUri(blob, fileName)
  }
```

f1sh1918: looks good to me

administration/src/bp-modules/self-service/hooks/__tests__/useCardGeneratorSelfService.test.tsx
```diff
@@ -50,10 +50,7 @@ describe('useCardGeneratorSelfService', () => {
     expect(result.current.selfServiceCard).toEqual(exampleCard)
     await act(async () => result.current.generateCards())
     expect(toasterSpy).not.toHaveBeenCalled()
-    expect(result.current.selfServiceState).toBe(CardSelfServiceStep.information)
-    expect(result.current.deepLink).toBe(
-      'koblenzpass://koblenz.sozialpass.app/activation/code#ClcKLQoNS2FybGEgS29ibGVuehDmnwEaGAoCCF8SBAjqvgEqBAiLmgEyBgoEMTIzSxIQL%2Fle3xYGNIKS50god4ITmxoUOUYGq%2FjAE99bK4edMPo2I3ojr78%3D/'
-    )
+    expect(result.current.cardGenerationStep).toBe('information')
```
f1sh1918, line 53: hm i think if the check for the correct deeplink is removed, there should be a separate test for that case

administration/src/bp-modules/util/cardGenerationError.tsx
```diff
@@ -0,0 +1,43 @@
...
+  } else if (error instanceof ApolloError) {
+    const { title } = getMessageFromApolloError(error, t)
+    appToaster.show({
+      message: <FormErrorMessage style={{ color: 'white' }} errorMessage={title} />,
+      timeout: 0,
+      intent: 'danger',
+    })
+  } else if (error instanceof PdfError) {
+    appToaster.show({
+      message: t('pdfCreationError'),
```
f1sh1918, line 28: 🔧  just a general suggestion.
I think we could avoid passing all the `TFunction` by using `i18next.t('errors:csvCreationError')`
I also used to pass the `t` down to everywhere and the amount of function parameters just increase.

f1sh1918 (changes requested):
I just started reviewing but then faced an issue i wanted to report before finishing because it may require some changes

Bulk creation of pdf does not work work anymore for nuernberg
image

administration/src/bp-modules/util/cardGenerationError.tsx
steffenkleinle: I think this was enough refactoring in this PR and not done by me, therefore I would really not do this here anymore :thinking: However, I planned to do a follow up PR anyway to address this issue. I would do it in a different way though by only returning the key without translating it any only translating it where it is actually shown to the user.

administration/src/bp-modules/self-service/hooks/__tests__/useCardGeneratorSelfService.test.tsx
f1sh1918: did you add sth? i couldn't see an update

f1sh1918 (approved):
Looks good to me. 👍 
Tested card bulk creation for bayern and nuernberg
Tested single card creation for all projects
Tested queryParams resolving

We definitely need some tests for the seachParams resolvement, since it was broken once for nuernberg and i feel quite unsafe with that.
I will create an issue for that

administration/src/bp-modules/self-service/hooks/__tests__/useCardGeneratorSelfService.test.tsx
steffenkleinle: See `administration/src/util/__tests__/getCustomDeepLinkFromQrCode.test.ts`

bahaaTuffaha:
At Nuernberg I created two card and exported them as CSV and when I imported the file this is the preview:
image

heres the file:
[SozialpassMassExport.csv](https://github.com/user-attachments/files/19088423/SozialpassMassExport.csv)

f1sh1918:
> At Nuernberg I created two cards and exported them as CSV and when I imported the file this is the preview: image
> 
> Here's the file: [SozialpassMassExport.csv](https://github.com/user-attachments/files/19088423/SozialpassMassExport.csv)

@bahaaTuffaha the exported csv is for an external print service provider and not for reimporting it here

bahaaTuffaha (approved):
Tested Card generation/multi-card also PDF and CSV exporting is working as expected.

Issue #1644, opened 2024-09-24
Add validation to importAcceptingStores mutation in the backend


seluianova:
**Is your feature request related to a problem? Please describe.**
Follow up for #1571

Currently we have csv validation on the administration side, but no validation on the backend.
This means that if the request is sent directly to the backend, our validation rules can be violated.

For example:

In the administration we check that the mandatory fields are not null (name, street, houseNumber, postalCode, location).
But with a direct request to the backend I can create a store with a null-name.

Or:
We validate in the administration, that the categoryId refers to the current project.
But in the backend I can save a store with any existing category id.

Or:
In the administration we check duplicates by name & address.
But the backend allows to create a store with the same name & address, but with a difference in some other fields (like a different description, for example), because in the backend we only consider 2 records as duplicates if they are completely the same.

**Describe the solution you'd like**
Add csv validation to the importAcceptingStores mutation

**TODO**:

- [x] check that mandatory fields are not blank: name, street, houseNumber, postalCode, location, coordinates
- [ ] check that the categoryId refers to the current project
- [x] unify working with duplicates
- [x] clean data (remove trailing whitespaces)

**Describe alternatives you've considered**
Alternatively we could move the validation entirely to the backend (in order not to implement validation 2 times).
If validation errors occur, they can be returned from the backend in the response and displayed in the administration.

seluianova:
Also we don't "clean" data in the backend (remove trailing spaces)
Probably this should be moved from administration component to the backend.

seluianova:
Why "street" in the database is nullable, but in the administration this field is mandatory for Nürnberg?

Do we want to create separate validation rules for each project in the backend like in the admistration? 
Maybe the rules like mandatory and optional fields should be the same for all projects.

connium:
For CSV the fields are mandatory. For the XML import we are not sure whether these fields are filled.

seluianova:
> check that the categoryId refers to the current project

We don't have this check in the current import process, because we don't store the mapping between projects and categories in the backend.
But we configure that separately in the administration and in the frontend module.
We probably need to store this data in one place, for example, in the backend, so that other components could query that from the backend.

Issue #1915, opened 2025-02-18
Move the project category configuration to the backend


seluianova:
### Describe the Problem
Currently we have a category configuration per project in the frontend module --> it is used to display category filters in the app.
~~We also have a category configuration per project in the administration~~ (_upd_. actually we don't have a separate config there, we read the one from the frontend) --> it is used to validate the csv file during import of accepting points.
We don't have a category configuration in the backend, but I think we need it there too to add category validation when importing stores.

### Describe Your Preferred Solution
- Instead of having a category configuration in each module, I think it should be moved to the backend.
For example, we can store the mapping in a database and provide a graphql query to retrieve that data.
- provide which categories are available per project in backend project config.
- extend category table by icon (material icon name)
- use category shortName as name and not long name
- create category descriptions table with relation to categories
- create query to fetch category information (will be extended by further projectConfig information later -> should be called getProjectInformation or sth. similar) 

### Related Issues
[#1644](https://github.com/digitalfabrik/entitlementcard/pull/1928)
#2882

Issue #1591, opened 2024-08-26
Remove import pipeline nuernberg


f1sh1918:
**Is your feature request related to a problem? Please describe.**
When we finished the management of the accepting stores in the administration module, the import pipeline can be deleted in the backend module

**Describe the solution you'd like**
- remove the nuernberg import pipeline and related code that is not needed anymore 
- ~~adjust the data class `CSVAcceptingStore` to match with the fieldConfiguration in the frontend. Mandatory and optional fields.~~ will be done in #1644 

~~blocked by #1571 , blocked by #1570~~

PR #1928, opened 2025-02-19
1644: Add validation to importAcceptingStores mutation in the backend

seluianova:
### Short Description
Currently we have csv validation on the administration side, but no validation on the backend.
This means that if the request is sent directly to the backend, our validation rules can be violated.

### Proposed Changes
- add validation that the mandatory fields are not empty (name, street, houseNumber, postalCode, location)
- return an error, if request contains duplicates (consider records as duplicate if the name, street, house number, zip code, and location are the same)
- add data cleaning (remove trailing whitespaces)
- add new tests for new validation checks

**Not included**:
Check that the categoryId refers to the current project.
I think it would be better to do it after #1915 ❓

### Side Effects
no?

### Testing
This can be checked by direct requests to the API or by disabling/commenting the corresponding checks in the code in the administration.

### Resolved Issues
Fixes: #1644


f1sh1918:
@seluianova please also add #1591 as fixed

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/database/repos/AcceptingStoresRepository.kt
```diff
@@ -118,35 +111,12 @@ object AcceptingStoresRepository {
             }.firstOrNull()?.let { it[AcceptingStores.id].value }
     }
 
-    fun importAcceptingStores(stores: List<CSVAcceptingStore>, project: ProjectEntity, dryRun: Boolean): StoreImportReturnResultModel {
```
seluianova, line 121: I refactored and moved this function to the service layer because I think conceptually the repository should only handle database operations (CRUD).
Validation and building a response are more like business logic.
But feel free to object.

f1sh1918:
@seluianova i'm a little bit confused about the branch name of this pr. was this intended?

f1sh1918:
> > i'm a little bit confused about the branch name of this pr. was this intended?
> 
> @f1sh1918 jein, I messed them up a bit because this PR originally included (and started with) the pipeline deletion. after I moved the pipeline deletion to a separate PR, I haven't recreated the current branch.
> 
> I can actually recreate the PR with a new branch name, but since the commits are correct and include the correct task numbers, I was not sure if that makes sence. does it?

hm but you can not use the filter by branch function properly.
If you don't mind it would be great to adjust the branch name

Issue #1875, opened 2025-01-21
Replace pdf-lib


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Currently we use a very unsupported pdf-lib for our pdf generation (last release 4years+), so we should switch to better maintained fork.

**Describe the solution you'd like**
- replace current library by sth well supported f.e. this fork. ttps://github.com/cantoo-scribe/pdf-lib
- test if all current pdf generations still work

PR #1931, opened 2025-02-19
1875: Replace pdf-lib

bahaaTuffaha:
### Short Description

Currently we use a very unsupported pdf-lib for our pdf generation (last release 4years+), so we should switch to better maintained fork.

### Proposed Changes

- Replaced the old one with its fork [@cantoo/pdf-lib](https://www.npmjs.com/package/@cantoo/pdf-lib)
- Its also compatible with `@pdf-lib/fontkit`
Note: I tested the this package on pdf creation at bayern and koblenz.

### Side Effects

- None.

### Testing

- Card creation for Koblenz, Bayern and Nuremberg and generate pdfs.
### Resolved Issues

Fixes: #1875


f1sh1918 (approved):
Tested single and bulk pdf creation for all projects
Works great 👍

Issue #1935, opened 2025-02-20
Sentry sends error reports for staging environment


f1sh1918:
### Describe the Problem

Currently for sentry error reporting only the `REACT_APP_IS_PRODUCTION` of the environment file is checked. Unfortunately the flag is true for staging and production but not for local development which leads to reports also for staging environment.
This happens for web only.

### Describe Your Preferred Solution

- add an additional check of the domain if its staging or production

PR #1936, opened 2025-02-20
1935: Sentry sends reports for staging

f1sh1918:
### Short Description

Since we build the same bundles for staging and production on web, we have to distinguish between staging and production on the basis of the domain. Thats why we currently send reports also for staging.

### Proposed Changes

- add a missing hostname check
- unify function names

### Side Effects

- none

### Testing

N/A

### Resolved Issues

Fixes: #1935

### DIFF
--- a/administration/src/util/helper.ts
+++ b/administration/src/util/helper.ts
@@ -3,7 +3,6 @@ import XRegExp from 'xregexp'
 export const isStagingEnvironment = (): boolean => !!window.location.hostname.match(/staging./)
 export const isProductionEnvironment = (): boolean =>
   process.env.REACT_APP_IS_PRODUCTION === 'true' && !isStagingEnvironment()
-export const isDevEnvironment = (): boolean => window.location.hostname === 'localhost'
 
 export const updateArrayItem = <T>(array: T[], updatedItem: T, index: number): T[] => {
   if (index >= array.length || index < 0) {
--- a/administration/src/util/helper.ts
+++ b/administration/src/util/helper.ts
@@ -1,8 +1,9 @@
 import XRegExp from 'xregexp'
 
-export const isProductionEnvironment = (): boolean => process.env.REACT_APP_IS_PRODUCTION === 'true'
-export const isDevMode = (): boolean => window.location.hostname === 'localhost'
-export const isStagingMode = (): boolean => !!window.location.hostname.match(/staging./)
+export const isStagingEnvironment = (): boolean => !!window.location.hostname.match(/staging./)
+export const isProductionEnvironment = (): boolean =>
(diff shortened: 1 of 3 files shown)


administration/src/util/helper.ts
```diff
@@ -1,8 +1,9 @@
 import XRegExp from 'xregexp'
 
-export const isProductionEnvironment = (): boolean => process.env.REACT_APP_IS_PRODUCTION === 'true'
-export const isDevMode = (): boolean => window.location.hostname === 'localhost'
-export const isStagingMode = (): boolean => !!window.location.hostname.match(/staging./)
+export const isStagingEnvironment = (): boolean => !!window.location.hostname.match(/staging./)
+export const isProductionEnvironment = (): boolean =>
+  process.env.REACT_APP_IS_PRODUCTION === 'true' && !isStagingEnvironment()
+export const isDevEnvironment = (): boolean => window.location.hostname === 'localhost'
```
steffenkleinle, line 6: :thinking: I think a better solution would be to replace the boolean `REACT_APP_IS_PRODUCTION` with a string `REACT_APP_ENVIRONMENT` which can be set to `production`, `staging` or `development`. Otherwise using the ip of localhost might not work for example.

:upside_down_face: I think the naming of the environment variables is not really great as the `REACT_APP` is very redundant (and does not even follow our other naming).

f1sh1918: I'm not sure if environment variables without `REACT_APP_` will be recognized without additional implementation. At least for `create-react-app` projects it was the case. 

As i mentioned in the description. Setting a variable to `staging` won't help us, since we build the packages for staging and just promote them. Thats why we can only distinguish in case of build variables between development or not. If we want to change that, we would have to build different packages (staging/production) with i would like to avoid.

steffenkleinle, line 6:
```suggestion
export const isDevEnvironment = (): boolean => process.env.REACT_APP_IS_PRODUCTION !== 'true'
```

f1sh1918: well the question is then how we distinguish between staging and production also via the hostname.
But we could turn the variable to `_IS_DEVELOPMENT` and then check for false if the hostname contains staging. Then also for other ip addresses than localhost development will be used

steffenkleinle: I don't get that. We can keep the current checks for staging and production and just alter the check for development?

f1sh1918: in the end it turned out that we don't need this helper function. i just adjusted some checks and fixed the tests

steffenkleinle (approved):

bahaaTuffaha (approved):

Issue #1943, opened 2025-02-24
Make user checks reusable


steffenkleinle:
### Describe the Problem

At the moment, code to check a users region, project and role are duplicated over lots of places when they could be easily reused. This worsens code maintainability as well as readability.

### Describe Your Preferred Solution

Extract functionality to check a users
- region
- project
- role(s)

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #1944, opened 2025-02-24
1943: Fix permission checks and reuse user checks

steffenkleinle:
### Short Description

Create reusable functions for user checks for project, region and role.

### Proposed Changes

- Add user functions to check for project, region and role
- Refactor Authorizer to use reusable functions
- Add lots of tests
- Fix a few permission checks, see side effects

### Side Effects

- Added check to ensure that EXTERNAL_VERIFIED_API_USER is only available for bayern (align with create)
- Delete user even if role is `NO_RIGHTS`

Hopefully no other changes. I ran the tests both on the code before and after and all tests succeed in both versions (exception: changing a role to EXTERNAL_VERIFIED_API_USER in koblenz.

### Testing

Test that the user permissions still work.

### Resolved Issues

Fixes: #1943


backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/database/Schema.kt
```diff
@@ -54,6 +55,16 @@ class AdministratorEntity(id: EntityID<Int>) : IntEntity(id) {
...
     var notificationOnVerification by Administrators.notificationOnVerification
     var deleted by Administrators.deleted
+
+    fun isProject(project: String): Boolean =
+        this.projectId == ProjectEntity.find { Projects.project eq project }.single().id
+
+    fun isProject(projectId: Int): Boolean = this.projectId.value == projectId
+
+    fun isRole(role: Role, vararg other: Role): Boolean =
+        this.role == role.db_value || other.any { this.role == it.db_value }
```
michael-markl, line 65: @wouldn't `hasRole` make more sense?
Similarly, I'd prefer `isInProject` and `isInRegion`?

Or maybe `roleIs`, `projectIs` and `regionIs`?

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -80,30 +60,19 @@ object Authorizer {
...
-        if (actingAdmin.role == Role.PROJECT_ADMIN.db_value && newAdminRole == Role.EXTERNAL_VERIFIED_API_USER) {
-            return transaction {
-                ProjectEntity.find { Projects.project eq EAK_BAYERN_PROJECT }
-                    .single().id.value == actingAdmin.projectId.value
+        if (actingAdmin.isRole(Role.PROJECT_ADMIN)) {
+            return newAdminRole != Role.EXTERNAL_VERIFIED_API_USER || transaction {
+                actingAdmin.isProject(
+                    EAK_BAYERN_PROJECT
+                )
             }
```
michael-markl, line 71: I feel like negating statements, using disjunctions, and packing everything into a return statement makes it harder to read (at least for me).

```suggestion
            if (newAdminRole == Role.EXTERNAL_VERIFIED_API_USER && transaction { actingAdmin.isProject(EAK_BAYERN_PROJECT) }) {
              // Currently, we only allow external API users for the EAK Bayern project.
              return false
            }
            return true
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -113,74 +82,54 @@ object Authorizer {
...
-            return true
-        }
-        return false
     }
 
-    fun mayDeleteUser(
-        actingAdmin: AdministratorEntity,
-        existingAdmin: AdministratorEntity
-    ): Boolean {
-        if (actingAdmin.projectId.value != existingAdmin.projectId.value && existingAdmin.role != Role.NO_RIGHTS.db_value) {
```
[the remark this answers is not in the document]
michael-markl: As far as I know, a user has the NO_RIGHTS if and only if the user is deleted. (This might have changed though).
I think, it does not matter too much, whether we allow a deleted user to be deleted again.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -113,74 +82,54 @@ object Authorizer {
...
-                    .single().id.value == user.projectId.value
-                ) ||
-                (
-                    user.role == Role.EXTERNAL_VERIFIED_API_USER.db_value && ProjectEntity.find { Projects.project eq EAK_BAYERN_PROJECT }
-                        .single().id.value == user.projectId.value
-                    )
+            (user.isRole(Role.PROJECT_ADMIN) && user.isProject(KOBLENZ_PASS_PROJECT)) ||
+                (user.isRole(Role.EXTERNAL_VERIFIED_API_USER) && user.isProject(EAK_BAYERN_PROJECT))
         }
     }
```
michael-markl, line 124: This is wild :D (not your changes, but the fact that we actually make such distinctions between projects for the roles. I feel like this is going to get confusing very quickly, when we have more than 3 projects)

michael-markl (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
steffenkleinle: Agreed, negating should be avoided as possible. Will adjust accordingly.

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
steffenkleinle: Agreed. @f1sh1918 do you have any more insights whether we actually have to distinguish between projects here?

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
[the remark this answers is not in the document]
steffenkleinle: Thanks for clarifying :+1:

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
f1sh1918: Yes i'm also not super happy with this implementation. I think the main aspect was not to create a separate role for that because it would be more effort. Which we in the end did for verein360 since we had to ensure that they don't get project admin rights as a third party provider, but that wasn't planned so far when we implemented that for koblenz..

I think it would be easier and cleaner to also use this separate role for koblenz (like for eak)
The problem is that we soon go live with koblenz and we should not break the current implementation.

My suggestion would be to add the ability for koblenz to create `EXTERNAL_VERIFIED_API_USER` and allow them to use the particular endpoint (entitleduserdata). Then give them the possibility to create a token and adjust their import so everything still work.
If that was done i would remove the token creation from the koblenz project admin role.

As a third step it would probably make sense to add RBAC and also be able to have different roles as a user so someone could be `Project_Admin` and `External_API_User`

steffenkleinle: Hmm, I don't know enough about the project. Could you perhaps create an issue for this if you think it should be done? Thanks :)

f1sh1918: i created #1965 and #1966

steffenkleinle: Thank you!

f1sh1918 (changes requested):
Great refactoring.

I found one issue when i want to delete a user
Login as `regionAdmin` and try to delete a regionManager
I get an internal error and the following backend log
```js
Caused by: org.postgresql.util.PSQLException: ERROR: new row for relation "administrators" violates check constraint "deletedifandonlyifnorights"
```
Looks like deleting users is broken in general

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/service/Authorizer.kt
```diff
@@ -80,30 +60,18 @@ object Authorizer {
...
             newAdminRegion != null && actingAdmin.regionId == newAdminRegion.id &&
             newAdminRole in setOf(Role.REGION_ADMIN, Role.REGION_MANAGER)
-        ) {
-            return true
-        }
-
-        return false
     }
 
     fun mayEditUser(
```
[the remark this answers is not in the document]
steffenkleinle: This could be refactored in this instance and in a few other places, we have lots of duplicated checks. I chose not to as the changes here are already enough and this would have made it even harder to review. I am also not sure what best practices are in this case. In any case, I would not do this in this PR/issue.

steffenkleinle:
> I found one issue when i want to delete a user Login as `regionAdmin` and try to delete a regionManager I get an internal error and the following backend log
> Looks like deleting users is broken in general

@f1sh1918 I think I found the issue, I found and replace one usage I should not have. Please test again :)

f1sh1918 (approved):
Works great 👍

PR #1947, opened 2025-02-26
1945: Unit test card search params

f1sh1918:
### Proposed Changes

-

### Side Effects

-

### Resolved Issues

Fixes: #


administration/src/bp-modules/cards/hooks/useCardGenerator.test.tsx
```diff
@@ -11,19 +11,44 @@ import createCards, { CreateCardsError, CreateCardsResult } from '../../../cards
...
+      <AppToasterProvider>
+        <ApolloProvider>{children}</ApolloProvider>
+      </AppToasterProvider>
+    </ProjectConfigProvider>
   </MemoryRouter>
 )
 
+const withCustomWrapper =
+  (initialRoutes: string[], projectConfig: ProjectConfig) =>
+  // @ts-expect-error any is okay here
```
f1sh1918, line 44: didn't find a way to set a proper type here

administration/src/bp-modules/cards/hooks/useCardGenerator.test.tsx
```diff
@@ -11,19 +11,44 @@ import createCards, { CreateCardsError, CreateCardsResult } from '../../../cards
...
+  ({ children }) =>
```
steffenkleinle, line 45: How about
``` suggestion
  ({ children }: { children: ReactNode}) =>
```

administration/src/bp-modules/cards/hooks/useCardGenerator.test.tsx
```diff
@@ -11,19 +11,44 @@ import createCards, { CreateCardsError, CreateCardsResult } from '../../../cards
...
 import downloadDataUri from '../../../util/downloadDataUri'
 import { AppToasterProvider } from '../../AppToaster'
 import useCardGenerator from './useCardGenerator'
 
-const wrapper = ({ children }: { children: ReactNode }) => (
-  <MemoryRouter>
-    <AppToasterProvider>
-      <ApolloProvider>{children}</ApolloProvider>
-    </AppToasterProvider>
+const wrapper = ({
```
steffenkleinle, line 24: :upside_down_face: Not changed in this PR and not important at all but usually mocks are directly below imports in our projects. I like that for consistency.

administration/src/bp-modules/cards/hooks/useCardGenerator.test.tsx
```diff
@@ -11,19 +11,44 @@ import createCards, { CreateCardsError, CreateCardsResult } from '../../../cards
...
+    <ProjectConfigProvider projectConfig={projectConfig ?? showcaseConfig}>
+      <AppToasterProvider>
+        <ApolloProvider>{children}</ApolloProvider>
+      </AppToasterProvider>
+    </ProjectConfigProvider>
   </MemoryRouter>
 )
 
+const withCustomWrapper =
+  (initialRoutes: string[], projectConfig: ProjectConfig) =>
```
steffenkleinle, line 43: :upside_down_face: Since we never set multiple initial routes in this file, we could simplify this to
```suggestion
  (initialRoues: string, projectConfig: ProjectConfig) =>
```

administration/src/bp-modules/cards/hooks/useCardGenerator.test.tsx
```diff
@@ -11,19 +11,44 @@ import createCards, { CreateCardsError, CreateCardsResult } from '../../../cards
...
+  // @ts-expect-error any is okay here
+  ({ children }) =>
+    wrapper({
+      children,
+      initialRoutes,
```
steffenkleinle, line 48:
```suggestion
      initialRoutes: [initialRoute],
```

steffenkleinle (approved):
Nice, I like it :+1: You want to add the same tests for `useSelfServiceCardGenerator`? Also, there is no test yet for initializing `applicationIdToMarkAsProcessed` if you feel motivated.

administration/src/bp-modules/cards/hooks/useCardGenerator.test.tsx
f1sh1918: oh dear yes. i added the typing at the wrong position by mistake ^^

administration/src/bp-modules/cards/hooks/useCardGenerator.test.tsx
f1sh1918: hm the reason why i used an array was that it can be empty without an additional check.
Since passing this prop to the wrapper is optional, i then would have to check if the value is undefined before passing it to the `MemoryRouter`
You can test your suggested changes in the code

f1sh1918:
> applicationIdToMarkAsProcessed

I moved the test for koblenz to the selfService hook. Not sure if we need both tests, since normal card creation is disabled for koblenz.

I dunno how i should test this applicationId since it will not be set in the cards object.

administration/src/bp-modules/self-service/hooks/__tests__/useCardGeneratorSelfService.test.tsx
```diff
@@ -40,6 +30,24 @@ jest.mock('../../../../cards/PdfFactory', () => ({
...
+  </MemoryRouter>
+)
+
+const withCustomWrapper =
+  (initialRoute: string) =>
+  ({ children }: { children: ReactNode }) =>
+    wrapper({
+      children,
+      initialRoutes: [initialRoute],
+    })
```
steffenkleinle, line 49: :upside_down_face: I guess we should make this reusable at some point to a shared testing file.

steffenkleinle:
> I dunno how i should test this applicationId since it will not be set in the cards object.

You could mock the mutation and ensure that it was called with the application id from the query params.

administration/src/bp-modules/self-service/hooks/__tests__/useCardGeneratorSelfService.test.tsx
steffenkleinle: You could also just straight move it to a helper? :D

bahaaTuffaha (approved):
Ran tests .. Looks good 👍️

PR #1950, opened 2025-02-28
Allow parallel installation of development and production builds

steffenkleinle:
### Short Description

Allow parallel installation of development and production builds on android.

### Proposed Changes

- Add `.development` to application id for development builds
- Add ` DEV` suffix to app name for development builds

### Side Effects

- Previous installations will not be used anymore, so e.g. cards have to be added again

### Testing

Install both a production build from the store and a development build in parallel.

### Resolved Issues

Fixes: N/A

### DIFF
--- a/frontend/android/app/build.gradle
+++ b/frontend/android/app/build.gradle
@@ -38,7 +38,7 @@ idea {
 ext.setupApplication = { ApplicationProductFlavor flavor, buildConfigName ->
     var buildConfig = createBuildConfig(buildConfigName)
     flavor.applicationId buildConfig.applicationId
-    flavor.manifestPlaceholders = [*:determineIcon(buildConfig)]
+    flavor.manifestPlaceholders = [*:determineAppName(buildConfig), *:determineIcon(buildConfig)]
 
     setupResourceValues(buildConfigName, {
         String type, String name, String value -> flavor.resValue(type, name, value)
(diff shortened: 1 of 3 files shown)


frontend/android/app/build.gradle
```diff
@@ -38,7 +38,7 @@ idea {
 ext.setupApplication = { ApplicationProductFlavor flavor, buildConfigName ->
     var buildConfig = createBuildConfig(buildConfigName)
     flavor.applicationId buildConfig.applicationId
-    flavor.manifestPlaceholders = [*:determineIcon(buildConfig)]
+    flavor.manifestPlaceholders = [*:determineAppName(buildConfig), *:determineIcon(buildConfig)]
```
andrew8er, line 41: Could we just declare the manifest placeholders here explicitly, instead of calling several functions? We do not use those anywhere else.

andrew8er (approved):
This PR will conflict with #1956. I suggest merging 1956 first and then this one, since it will be easier than the other way around.

michael-markl:
> This PR will conflict with https://github.com/digitalfabrik/entitlementcard/pull/1956. I suggest merging 1956 first and then this one, since it will be easier than the other way around.

I based my PR on this branch, so we can merge this PR first. Maybe your comment is already resolved in #1956 .

andrew8er:
Ah I just noticed that you are doing PRs into non-`main` branches also. I have a look at #1956 then 👍

f1sh1918 (approved):
Great improvement 👍 
Especially if you have an entitlementcard on your real device

I was just thinking if it would be an improvement to use a prefix since some android launchers just show the first x chars in the app list and for me the "dev" is not displayed but with prefix the order changes. So i think its okay like that

steffenkleinle:
> I was just thinking if it would be an improvement to use a prefix since some android launchers just show the first x chars in the app list and for me the "dev" is not displayed but with prefix the order changes. So i think its okay like that

Yeah, I was thinking the same. However, I decided against it since then the apps are in very different places if sorted alphabetically. We could think of truncating the app name after some chars but I think for now this is good enough. Feel free to create a new PR to improve this.

Issue #1953, opened 2025-02-28
Avoid non-null assertions for whoami context


steffenkleinle:
### Describe the Problem

At the moment we do `useContext(WhoAmIContext).me!` in quite a lot of places. This is not ideal and makes me feel bad while developing wondering whether this is actually okay.

### Describe Your Preferred Solution

Create a new hook exposing the same properties as the context with `me` guaranteed not to be null.

### Describe Possible Alternatives

N/A

### Related Issues

N/A

### Additional Information

N/A

PR #1954, opened 2025-02-28
1953: Avoid non-null assertions for WhoAmI

steffenkleinle:
### Short Description

Avoid non-null assertions for who am i by creating a hook.

### Proposed Changes

- Create a hook to expose the context non-nullable

### Side Effects

- Using the hook is only possible if `me` is actually not null, so when a user is logged in

### Testing

Ensure that everything still works.

### Resolved Issues

Fixes: #1953

### DIFF
--- a/administration/src/WhoAmIProvider.tsx
+++ b/administration/src/WhoAmIProvider.tsx
@@ -6,6 +6,7 @@ import { AuthContext } from './AuthProvider'
 import StandaloneCenter from './bp-modules/StandaloneCenter'
 import { WhoAmIQuery, useWhoAmIQuery } from './generated/graphql'
 import { ProjectConfigContext } from './project-configs/ProjectConfigContext'
+import { hasProp } from './util/helper'
 
 type WhoAmIContextType = {
   me: WhoAmIQuery['me'] | null
@@ -20,11 +21,11 @@ export const WhoAmIContext = createContext<WhoAmIContextType>({
 type UseWhoAmIReturn = WhoAmIContextType & { me: WhoAmIQuery['me'] }
 
 export const useWhoAmI = (): UseWhoAmIReturn => {
-  const { me, ...context } = useContext(WhoAmIContext)
-  if (!me) {
+  const context = useContext(WhoAmIContext)
+  if (!hasProp(context, 'me')) {
     throw new Error('WhoAmI context is not available')
   }
-  return { me, ...context }
+  return context
 }
 
 const WhoAmIProvider = ({ children }: { children: ReactNode }): ReactElement => {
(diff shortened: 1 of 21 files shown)


administration/src/WhoAmIProvider.tsx
```diff
@@ -7,14 +7,23 @@ import StandaloneCenter from './bp-modules/StandaloneCenter'
...
+
+export const WhoAmIContext = createContext<WhoAmIContextType>({
   me: null,
   refetch: () => undefined,
 })
 
+type UseWhoAmIReturn = WhoAmIContextType & { me: WhoAmIQuery['me'] }
+
+export const useWhoAmI = (): UseWhoAmIReturn => {
+  const { me, ...context } = useContext(WhoAmIContext)
```
michael-markl, line 23: I think we should throw an error, if `me` is null.
Also, I think it's not necessary to create a new object, we can just return whatever `useContext(WhoAmIContext)` returns.

I am also a bit confused about the `@ts-expect-errror` on line 58. 🤔

steffenkleinle: > I think we should throw an error, if `me` is null.

:+1: 

> Also, I think it's not necessary to create a new object, we can just return whatever `useContext(WhoAmIContext)` returns.
> 
> I am also a bit confused about the `@ts-expect-errror` on line 58. 🤔

Typescript seems to have problems when checking object properties for nullish values and raises an error in that case. Same for directly returning the context without destructuring/creating a new object like this:

``` ts
  const context = useContext(WhoAmIContext)
  if (context.me === null) {
    throw new Error('')
  }
  // TS 2322: Type null is not assignable to type ...
  return context
```

michael-markl: Yeah, this seems to be a known limitation of typescript 🤔 . Probably, it's better to do 
```ts
  const context = useContext(WhoAmIContext)
  if (context.me === null) {
    throw new Error('WhoAmI context is not available')
  }
  return context as UseWhoAmIReturn
```
(and below we can probably do `value as WhoAmIContextType` instead of ignoring the error)

steffenkleinle: Hmm, I really don't like typecasts if we can avoid them. I agree for below, doesn't really matter if we have a ` @ts-expect-error` or a typecast, but looks cleaner.

michael-markl: You could also abstract away this typecast in a type guard like 
```ts
const hasProp= <
  P extends string,
  // We check whether the value of property P is the non-nullish type V.
  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
  V extends {},
  O extends { [p in P]: V | null }
>(obj: O, p: P): obj is (O & { [p in P]: V }) => !!obj[p]

export const useWhoAmI = (): UseWhoAmIReturn => {
  const context = useContext(WhoAmIContext)
  if (!hasProp(context, 'me')) {
    throw Error("WhoAmI context is not available")
  }
  return context
}
```
Maybe, that's the cleaniest, but slightly overwhelming solution...

bahaaTuffaha (approved):
Tested Administration LGTM 👍️

administration/src/WhoAmIProvider.tsx
steffenkleinle: I really like the solution :tada: I'll use it for integreat as well :)

f1sh1918 (approved):
Nice refactoring 👍 
Works fine

PR #1955, opened 2025-02-28
1919: Update typescript and eslint

bahaaTuffaha:
### Short Description

We are still using typescript 5.0x due to lacking eslint support.

### Proposed Changes

- Updated to latest typescript and its plugins then I migrated to eslint v9 by using:
``` npx  @eslint/migrate-config .eslintrc.*``` it will create a new config file called : `eslint.config.mjs`
- After running that command it asked me to install the following packages to devDependences: `@eslint/compat`, `@eslint/js`, `@eslint/eslintrc`
- I solved some problems with the following plugins so I wrapped them with `fixupPluginRules()`:
```
1- ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "jsx-a11y".
2- Error: Error while loading rule 'jsx-expressions/strict-logical-expressions'
```
- I used for jest version `version: require('jest/package.json').version,` as it was before and added these two lines to enable require:
```
const require = createRequire(import.meta.url)
import { createRequire } from 'module'
```
- Changed the old config file's name to `.eslintrc.js.old`
- Eslint showed some errors so I had to fix them here's the following errors I fixed:
```
/home/coldtea/Repos/entitlementcard/administration/config/env.ts
  59:13  error  Unexpected value in conditional. A boolean expression is required  @typescript-eslint/strict-boolean-expressions

/home/coldtea/Repos/entitlementcard/administration/src/bp-modules/applications/ApplicationsOverview.tsx
  69:33  error  Unexpected value in conditional. A boolean expression is required  @typescript-eslint/strict-boolean-expressions

/home/coldtea/Repos/entitlementcard/administration/src/mui-modules/application/forms/StepRequirementsForm.tsx
  25:13  error  Switch is not exhaustive. Cases not matched: null  @typescript-eslint/switch-exhaustiveness-check

/home/coldtea/Repos/entitlementcard/administration/src/util/PlainDate.ts
  30:5  error  Nested block is redundant  no-lone-blocks

✖ 4 problems (4 errors, 0 warnings)
```

### Side Effects

- Any Eslint/typescript related should be in this list.

### Testing

- Test: Formating , linting, jest test, and give the administration a run.
- Try to trigger an eslint issue like `nest ternary expressions` at any component.
- Check if what I have done to fixing these Eslint problems are correct or need any adjustments.
- Give `eslint.config.mjs` a look and see if there is something the automatic tool (@eslint/migrate) has gotten wrong.

### Resolved Issues

Fixes: #1919

### CODE REVIEW

File: administration/config/env.ts [hunk 1]
```diff
@@ -56,7 +56,7 @@ const getClientEnvironment = (
   const appDirectory = fs.realpathSync(process.cwd())
   process.env.NODE_PATH = (process.env.NODE_PATH || '')
     .split(path.delimiter)
-    .filter(folder => folder && !path.isAbsolute(folder))
+    .filter(folder => !!folder && !path.isAbsolute(folder))
```
andrew8er, line 59: I think you are testing for zero length strings here? If that is the case, I would use `folder.length > 0`. It is slightly longer but way more explicit.

administration/.eslintrc.js.old
andrew8er, line 1: Do we still need this file?

frontend/build-configs/package.json
andrew8er, line 1: Indentation is off by two spaces.

andrew8er (approved):
LGTM, with a few small suggestions.

administration/src/mui-modules/application/forms/StepRequirementsForm.tsx
```diff
@@ -37,6 +37,8 @@ const StepRequirementsForm: Form<StepRequirementsFormState, ValidatedInput, Addi
         }
         return { type: 'valid', value: { type: BavariaCardType.Golden, value: goldenCardEntitlement.value } }
       }
+      case null:
+        return { type: 'error' }
```
steffenkleinle, line 41: :upside_down_face: 
```suggestion
      case null:
```

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -68,7 +68,7 @@ const ApplicationsOverview = ({ applications }: { applications: Application[] })
     () =>
       sortedApplications.filter(application => {
         if (activeBarItem.status === undefined) {
-          return application
+          return Boolean(application)
```
steffenkleinle, line 71: :wrench: Please avoid typecasts like this. It is really implicit in its behavior and usually there is a reason why we get a ts/linting error.
```suggestion
          return true
```

administration/package.json
```diff
@@ -64,8 +67,8 @@
     "@types/react-router-dom": "^5.3.3",
     "@types/resolve": "^1.20.2",
     "@types/styled-components": "^5.1.26",
-    "@typescript-eslint/eslint-plugin": "^8.8.0",
-    "@typescript-eslint/parser": "^8.8.0",
+    "@typescript-eslint/eslint-plugin": "^8.25.0",
+    "@typescript-eslint/parser": "^8.25.0",
```
steffenkleinle, line 71: :wrench: These dependencies should not be necessary anymore as typescript-eslint exports these:
https://typescript-eslint.io/packages/typescript-eslint / https://typescript-eslint.io/packages/typescript-eslint/#migrating-from-legacy-eslintrc-configs
```suggestion
```

administration/package.json
```diff
@@ -77,18 +80,19 @@
...
     "eslint-plugin-jsx-expressions": "^1.3.2",
     "eslint-plugin-prefer-arrow": "^1.2.3",
-    "eslint-plugin-react": "^7.34.1",
-    "eslint-plugin-react-hooks": "^4.6.0",
-    "eslint-plugin-unicorn": "^52.0.0",
+    "eslint-plugin-react": "^7.37.4",
+    "eslint-plugin-react-hooks": "^5.1.0",
+    "eslint-plugin-unicorn": "^57.0.0",
     "file-loader": "^6.2.0",
+    "globals": "^16.0.0",
```
steffenkleinle, line 95: :question: Do we actually need this package?

administration/eslint.config.mjs
```diff
@@ -0,0 +1,217 @@
...
+import { fileURLToPath } from 'node:url'
+
+const require = createRequire(import.meta.url)
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+const compat = new FlatCompat({
+  baseDirectory: __dirname,
+  recommendedConfig: js.configs.recommended,
+  allConfig: js.configs.all,
```
steffenkleinle, line 23: :wrench: We were not using these configs before, was there any reason you added these? I think using `typescript-eslint` recommended config provides better defaults (including useful rules for typescript) already, so these should not be necessary (https://eslint.org/docs/latest/use/migrate-to-9.0.0#-eslintrecommended-and-eslintall-no-longer-accepted-in-flat-config)

administration/package.json
```diff
@@ -46,6 +46,9 @@
     "@babel/core": "^7.22.5",
     "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
     "@bufbuild/protoc-gen-es": "^1.2.1",
+    "@eslint/compat": "^1.2.7",
+    "@eslint/eslintrc": "^3.3.0",
+    "@eslint/js": "^9.21.0",
```
steffenkleinle, line 51: :x: Please remove, see comment above for `@eslint/js`. For `@eslint/eslintrc` from the [docs](https://www.npmjs.com/package/@eslint/eslintrc):
> This repository contains the legacy ESLintRC configuration file format for ESLint. This package is not intended for use outside of the ESLint ecosystem. It is ESLint-specific and not intended for use in other programs.
Note: This package is frozen except for critical bug fixes as ESLint moves to a new config system.

administration/eslint.config.mjs
```diff
@@ -0,0 +1,217 @@
...
+})
+
+export default [
```
steffenkleinle, line 26: :wrench: Please readd the comments from the old config here :)

steffenkleinle (changes requested):
Thanks for having a go at this, great work already. I think we should get rid of unnecessary libraries though.

administration/package.json
bahaaTuffaha: I checked again ... I don't think it's needed

administration/package.json
bahaaTuffaha: The migrator asked me to install them after the conversion... I will remove them and see if that works.

administration/.eslintrc.js.old
bahaaTuffaha: I will delete it.

frontend/build-configs/package.json
bahaaTuffaha: I think from Prettier... does it effect how it works?

administration/eslint.config.mjs
bahaaTuffaha: Do you mean adding the old comments here?

administration/package.json
bahaaTuffaha: Not all configs are supported yet by Eslint v9 look at : https://github.com/eslint/eslint/issues/18391
so I think we need a compatibility tool ..
https://eslint.org/docs/latest/use/configure/migration-guide#using-eslintrc-configs-in-flat-config

frontend/build-configs/package.json
andrew8er: No, just code style issue…

administration/eslint.config.mjs
```diff
@@ -0,0 +1,219 @@
+import { fixupConfigRules, fixupPluginRules } from '@eslint/compat'
+import { FlatCompat } from '@eslint/eslintrc'
+import typescriptEslint from '@typescript-eslint/eslint-plugin'
+import tsParser from '@typescript-eslint/parser'
```
steffenkleinle, line 4: :wrench: These should be replaced with the corresponding imports from 'typescript-eslint'. Same for all other imports/usages of '@typescript-eslint`.

administration/eslint.config.mjs
```diff
@@ -0,0 +1,219 @@
...
+      'airbnb',
+      'airbnb/hooks',
+      'plugin:@typescript-eslint/recommended',
+      'plugin:react/recommended',
+      'plugin:react-hooks/recommended',
+      'prettier',
+      'plugin:jest/recommended',
+      'plugin:jest/style'
+    )
+  ),
```
steffenkleinle, line 38: :wrench: I think it would be nice to only fixup plugins/configs where it is actually required to keep track what is still missing and what already supports the new config.

administration/package.json
steffenkleinle: Do we actually need both fixupConfigRules as well as FlatCompat to support legacy configs? Sounds to me like this is a duplication :thinking:

administration/eslint.config.mjs
```diff
@@ -0,0 +1,219 @@
...
+  {
+    plugins: {
+      '@typescript-eslint': fixupPluginRules(typescriptEslint),
+      jest: fixupPluginRules(jest),
+      'jsx-a11y': fixupPluginRules(jsxA11Y),
+      'jsx-expressions': fixupPluginRules(jsxExpressions),
+      'prefer-arrow': preferArrow,
+      react: fixupPluginRules(react),
+      'react-hooks': fixupPluginRules(reactHooks),
```
steffenkleinle, line 47: :wrench: Same here, do we still need fixupPluginRules for every plugin? For example `typescript-eslint` already seems to support the new config. Please check others as well :)

administration/eslint.config.mjs
bahaaTuffaha: hmm 🤔 they are already from 'typescript-eslint' I didn't install them individually look at package.json

administration/eslint.config.mjs
bahaaTuffaha: I tried to move one by one but each one says `Cannot redefine plugin` except prettier showed me 7721 errors about Missing semicolons.
```
  ...compat.extends('prettier', 'plugin:jest/recommended', 'plugin:jest/style'),
  ...fixupConfigRules(
    compat.extends(
      'plugin:react/recommended',
      'plugin:react-hooks/recommended',
      'airbnb',
      'airbnb/hooks',
      'plugin:@typescript-eslint/recommended'
    )
  ),
```
It looks like the only way is to wrap it all with fixupConfigRules

steffenkleinle (approved):
LGTM, smoke tested on firefox.

f1sh1918:
is there sth missing @bahaaTuffaha or can it be merged?

PR #1962, opened 2025-03-05
1285: Fix launching apple maps

f1sh1918:
### Short Description

Currently we are facing an (handled) error if an address of a store was opened on ios

### Proposed Changes

- add launch mode external for ios

### Side Effects

- Apple maps is opened directly without safari. I guess that shouldn't be a problem though, since "Maps" can't be deleted from an ios device

### Testing (dev & ios only)
- Open a store in bayrische ehrenamtskarte f.e. search for "Brückstraße" then open on the detail view the address. There shouldn't be error in the console anymore

### Resolved Issues

Fixes: #1285


steffenkleinle (approved):
LGTM, not tested.

f1sh1918:
I could need a review from someone with ios simulator @andrew8er :)

andrew8er (approved):
👍 Works for me in iOS 18.3 Simulator.

PR #1964, opened 2025-03-06
1869: Finalize translations

f1sh1918:
### Short Description

There are some places in form where we still have hardcoded text

### Proposed Changes

- replace hardcoded text by i18n keys and move texts to translations
- avoid passing `TFunctions` where its not needed
- fix tests

### Side Effects

- application creation process labels may break
- error messages may not be displayed correctly

### Testing

1. Test the application creation process for bavaria
2. Check especially text form error messages for broken labels
3. Verein 360 now needs use `sports` instead of `Sport`, since we save the value instead of the (translatable label) 

### Resolved Issues

Fixes: #1869


administration/src/mui-modules/application/forms/OrganizationForm.tsx
```diff
@@ -24,22 +25,22 @@ const WarningContactPersonSamePerson = styled(Alert)`
...
-    'Tierschutz',
-    'Sport',
-    'Bildung',
-    'Umwelt-/Naturschutz',
-    'Kultur',
-    'Gesundheit',
-    'Katastrophenschutz/Feuerwehr/Rettungsdienst',
-    'Kirchen',
-    'Andere',
+    i18next.t('applicationForms:organizationCategories:social'),
```
f1sh1918, line 28: hm i realized that it may make sense to push the keys to the database and resolve them if we want to make them (for future issue) resolvable in different languages. Keys should then be moved to `application`

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
```diff
@@ -166,8 +166,8 @@ class ApplicationHandler(
             require(organizations.isNotEmpty()) {
                 "Work at organizations list cannot be empty if application is already verified"
             }
-            require(organizations.all { it.organization.category.shortText == "Sport" }) {
-                "All organizations must be of category Sport if application is already verified"
+            require(organizations.all { it.organization.category.shortText == "sports" }) {
```
f1sh1918, line 169: @connium if we store values instead of labels for dropdowns we would have also to inform verein 360 that the value they have to send us will slightly change with the next release from `Sport` to `sports`

connium: I don't think this would raise much of effort for them, as I'm not quite sure whether they even started with the implementation.
Is this this only change, that affects them?

I think, in the worst case, we could even add a mapping, as this affects exactly a single interface on our side.
Do you know the developer for the verein360 stuff? If so, can you please contact her/him? Otherwise I'd have to write this to our contact person at the BLSV.

f1sh1918: no unfortunately i don't have contact to a dev.
I think it will be the only change
I could also make a mapping for the data they send to us from `Sport` to `sports` but i think such a small change shouldn't be a problem for them to adjust

seluianova: I can see a small issue if they send as a label and not a value:
Technically they can send requests with isAlreadyVerified: false
In this case, checking it.organization.category.shortText == “sports” will not be performed, and if they submit something else, it will show up in the UI as an unresolved translation.

seluianova:
:question: do we want to use translations in defaultErrorMap as well?

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
seluianova: > I think, in the worst case, we could even add a mapping

Actually I don't see why we need to change the backend here?
We could make translatable labels in the UI form. But when the request is sent directly to the API, why not leave it as it is? It just won't be translated, but it will display the value from the database.

f1sh1918:
> defaultErrorMap

good point. i will add them

seluianova (commented):
Added some thoughts to the thread above, pls let me know if you'd prefer not to change that, I'll add my approval then.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
seluianova: Additional thought:
If we want to somehow apply localization to data from direct API requests, I think that should be a different task and concept if necessary.
Because we also have addresses (Deutschland vs Germany) and other fields. Why would we only translate category?

f1sh1918: I think dropdowns are the most important things since they are selections, also radio button selections.
Other values are directly typed values and can be stores as they were typed.
But maybe we also have to handle this in the future

seluianova: > Maybe i just don't get the point

1. User fills the form via /beantragen and selects category, for example, "Sport". We use localization to display the category name in the right language to the user.
2. We send the translated value in the request to the backend --> the “Sport” value is stored in the database
3. In the application overview we don't translate the category name, but use the one from the database.
For Verein360 requests it will be the value from the request.
For normal application flow, this will be the value translated according to the localization of the user who created the application.

seluianova: So my idea is to display the category in the application overview in the language in which it was submitted.
Because we basically do that with other fields as well - country, city names, other free input fields.

f1sh1918: So therefore we would have to check the translations json also in the backend.
Okay i'm okay that we just send the label for now and think about a general solution @seluianova 
Since we don't use another language right now its not super important to tackle this now

seluianova (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationHandler.kt
seluianova: alright! good to go then :)

Issue #1779, opened 2024-11-15
Unify form fields to one ui library


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Currently we are using different form fields especially for extensions which cause different stylings and lot of effort to style them equally. Since we moved the datepickers to mui since they provide better functionality, we should also move the form components to mui and unify the styling

**Describe the solution you'd like**
- implement mui form fields
- unify styling

**Describe alternatives you've considered**

**Additional context**

f1sh1918:
hey @hauf-toni ,
before we unify form fields the question is if we want to use the same style for mui form fields.
Currently we use `filled` and `outlined` which leads to a lot of styling overhead.
As @andrew8er also mentioned here the `outlined` seem to perform better and we also decided in the design system to use them.
[Design system](https://www.figma.com/design/o5p3ArtrrmOqG6lJggAQE5/%F0%9F%9F%A8-Designsystem-Integreat-(MUI)?node-id=14-744&p=f&t=RjdZyGPexiZ3E1HA-0)
So i guess it make sense to move from `filled` to `outlined` which means we have to adjust all form fields for the application process

hauf-toni:
I agree, thanks for sharing these insights @f1sh1918 . Let's do it.

f1sh1918:
@connium will create some small subtasks

connium:
I created meta task #2045 for the migration. So this can be closed.

PR #1973, opened 2025-03-10
1766: Unify date picker components 

andrew8er:
### Short Description

Replaces `<TextField>` components in `DateForm` with `<CustomDatePicker>`.

### Proposed Changes

- Updates the package `@mui/x-date-pickers` to 7.27.3.
- Replaces the `<TextField>` component in `DateForm` with `<CustomDatePicker>` components. 
- Uses `enableAccessibleFieldDOMStructure` for `<CustomDatePicker>`, which will be the [future default](enableAccessibleFieldDOMStructure) for MUI components.
- Uses the proper `format` for `<CustomDatePicker>`, fixing the text input. The format will also be displayed and having the month fields in lower case looks slightly misplaced, but I think this is tolerable. Could be fixed with some more CSS.
- Remove some box shadows that looked misplaced.
- Make all event handlers optional in `<CustomDatePicker>`.

### Side Effects

None.

### Testing

- Start backend and frontend
- Visit http://localhost:3000/beantragen and look at date form fields

### Resolved Issues

Fixes: #1766

### DIFF
--- a/administration/src/bp-modules/components/CustomDatePicker.tsx
+++ b/administration/src/bp-modules/components/CustomDatePicker.tsx
@@ -4,8 +4,6 @@ import type { DesktopDatePickerSlotProps } from '@mui/x-date-pickers/DesktopDate
 import { deDE } from '@mui/x-date-pickers/locales'
 import React, { ReactElement } from 'react'
 
-import useWindowDimensions from '../../hooks/useWindowDimensions'
-
 type CustomDatePickerProps = {
   value?: Date | null
   disabled?: boolean
@@ -36,60 +34,55 @@ const CustomDatePicker = ({
   disablePast,
   textFieldHelperText,
   textFieldSlotProps,
-}: CustomDatePickerProps): ReactElement => {
-  const { viewportSmall } = useWindowDimensions()
-
-  return (
-    <DesktopDatePicker
-      disabled={disabled}
-      label={label}
-      views={['year', 'month', 'day']}
-      openTo='year'
-      value={value}
-      format='dd.MM.yyyy'
-      sx={{ '& input[value=""]:not(:focus)': { color: 'transparent' } }}
-      slotProps={{
-        openPickerIcon: { fontSize: 'small' },
-        field: {
-          clearable: true,
-          onClear,
-        },
-        textField: deepmerge(
-          {
-            helperText: textFieldHelperText,
-            placeholder: 'TT.MM.JJJJ',
-            error: !isValid,
-            spellCheck: false,
-            onBlur,
-            size: 'small',
-            sx: {
-              width: '100%',
-              boxShadow: !isValid
-                ? `
+}: CustomDatePickerProps): ReactElement => (
+  <DesktopDatePicker
+    disabled={disabled}
+    label={label}
+    views={['year', 'month', 'day']}
+    openTo='year'
+    value={value}
+    format='dd.MM.yyyy'
+    sx={{ '& input[value=""]:not(:focus)': { color: 'transparent' } }}
+    slotProps={{
+      openPickerIcon: { fontSize: 'small' },
+      field: {
+        clearable: true,
+        onClear,
+      },
+      textField: deepmerge(
+        {
+          helperText: textFieldHelperText,
+          placeholder: 'TT.MM.JJJJ',
+          error: !isValid,
+          spellCheck: false,
+          onBlur,
+          size: 'small',
+          sx: {
+            width: '100%',
+            boxShadow: !isValid
+              ? `
                 0 0 0 0 rgba(205, 66, 70, 0),
                 inset 0 0 0 1px #cd4246,
                 inset 0 0 0 1px rgba(17, 20, 24, 0.2),
                 inset 0 1px 1px rgba(17, 20, 24, 0.3),
               `
-                : undefined,
-              '.MuiPickersInputBase-root': {
-                fontSize: viewportSmall ? '16px' : '14px',
-                borderRadius: '2px',
-              },
+              : undefined,
+            '.MuiPickersInputBase-root': {
+              borderRadius: '2px',
             },
           },
-          textFieldSlotProps
-        ),
-      }}
-      enableAccessibleFieldDOMStructure
-      localeText={deDE.components.MuiLocalizationProvider.defaultProps.localeText}
-      disableFuture={disableFuture}
-      disablePast={disablePast}
-      minDate={minDate}
-      maxDate={maxDate}
-      onChange={onChange}
-    />
-  )
-}
+        },
+        textFieldSlotProps
+      ),
+    }}
+    enableAccessibleFieldDOMStructure
+    localeText={deDE.components.MuiLocalizationProvider.defaultProps.localeText}
+    disableFuture={disableFuture}
+    disablePast={disablePast}
+    minDate={minDate}
+    maxDate={maxDate}
+    onChange={onChange}
+  />
+)
 
 export default CustomDatePicker
--- a/administration/src/bp-modules/components/CustomDatePicker.tsx
+++ b/administration/src/bp-modules/components/CustomDatePicker.tsx
@@ -5,7 +5,7 @@ import React, { CSSProperties, ReactElement } from 'react'
 import useWindowDimensions from '../../hooks/useWindowDimensions'
 
 type CustomDatePickerProps = {
-  date: Date | null
+  value?: Date | null
   disabled?: boolean
   label?: string
   onBlur?: () => void
@@ -21,7 +21,7 @@ type CustomDatePickerProps = {
 }
 
 const CustomDatePicker = ({
-  date,
+  value,
   disabled,
   label,
   onBlur,
@@ -44,7 +44,7 @@ const CustomDatePicker = ({
       disabled={disabled}
       label={label}
       views={['year', 'month', 'day']}
-      value={date}
+      value={value}
       format='dd.MM.yyyy'
       slotProps={{
         clearIcon: { fontSize: viewportSmall ? 'medium' : 'small' },
--- a/administration/src/bp-modules/components/CustomDatePicker.tsx
+++ b/administration/src/bp-modules/components/CustomDatePicker.tsx
@@ -1,22 +1,29 @@
 import { DesktopDatePicker } from '@mui/x-date-pickers'
 import { deDE } from '@mui/x-date-pickers/locales'
-import React, { ReactElement } from 'react'
+import React, { CSSProperties, ReactElement } from 'react'
 
 import useWindowDimensions from '../../hooks/useWindowDimensions'
 
 type CustomDatePickerProps = {
   date: Date | null
+  disabled?: boolean
+  label?: string
   onBlur?: () => void
   onChange?: (date: Date | null) => void
   onClear?: () => void
   isValid: boolean
   minDate?: Date
   maxDate?: Date
-  disableFuture: boolean
+  disableFuture?: boolean
+  disablePast?: boolean
+  textFieldHelperText?: string
+  textFieldStyle?: CSSProperties
 }
 
 const CustomDatePicker = ({
   date,
+  disabled,
+  label,
   onBlur,
   onChange,
   onClear,
@@ -24,6 +31,9 @@ const CustomDatePicker = ({
   minDate,
   maxDate,
   disableFuture,
+  disablePast,
+  textFieldHelperText,
+  textFieldStyle,
 }: CustomDatePickerProps): ReactElement => {
   const { viewportSmall } = useWindowDimensions()
   const formStyle = viewportSmall ? { fontSize: 16, padding: '9px 10px' } : { fontSize: 14, padding: '6px 10px' }
@@ -31,6 +41,8 @@ const CustomDatePicker = ({
     '0 0 0 0 rgba(205, 66, 70, 0), 0 0 0 0 rgba(205, 66, 70, 0), inset 0 0 0 1px #cd4246, inset 0 0 0 1px rgba(17, 20, 24, 0.2), inset 0 1px 1px rgba(17, 20, 24, 0.3)'
   return (
     <DesktopDatePicker
+      disabled={disabled}
+      label={label}
       views={['year', 'month', 'day']}
       value={date}
       format='dd.MM.yyyy'
@@ -39,10 +51,12 @@ const CustomDatePicker = ({
         openPickerIcon: { fontSize: 'small' },
         field: { clearable: true, onClear },
         textField: {
+          helperText: textFieldHelperText,
           placeholder: 'TT.MM.JJJJ',
           error: !isValid,
           spellCheck: false,
           onBlur,
+          style: textFieldStyle,
           sx: {
             width: '100%',
             input: formStyle,
@@ -52,6 +66,7 @@ const CustomDatePicker = ({
       }}
       localeText={deDE.components.MuiLocalizationProvider.defaultProps.localeText}
       disableFuture={disableFuture}
+      disablePast={disablePast}
       minDate={minDate}
       maxDate={maxDate}
       onChange={onChange}
--- a/administration/src/bp-modules/components/CustomDatePicker.tsx
+++ b/administration/src/bp-modules/components/CustomDatePicker.tsx
@@ -5,6 +5,8 @@ import type { DesktopDatePickerSlotProps } from '@mui/x-date-pickers/DesktopDate
 import { deDE } from '@mui/x-date-pickers/locales'
 import React, { ReactElement } from 'react'
 
+export type CustomDatePickerTextFieldProps = DesktopDatePickerSlotProps<Date, true>['textField']
+
 type CustomDatePickerProps = {
   value?: Date | null
   disabled?: boolean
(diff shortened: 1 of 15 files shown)


administration/src/mui-modules/application/primitive-inputs/DateForm.tsx
```diff
@@ -40,28 +43,29 @@ const DateForm: Form<State, ValidatedInput, AdditionalProps, Options> = {
...
-        style={{ margin: '4px 0', minWidth }}
-        type='date'
-        label={label}
-        required
-        inputProps={{
-          max: options.maximumDate ? formatISO(options.maximumDate, { representation: 'date' }) : undefined,
-          min: '1900-01-01',
-        }}
+      <CustomDatePicker
+        date={null}
```
f1sh1918, line 48: 🔧 it looks like the `DesktopDatePicker` from mui also accepts `undefined`
So you could remove this prop here and adjust the `customDatePickerProps`
`date?: Date`

f1sh1918 (changes requested):
Looks quite good so far except of the styling doesn't fit to the forms of the particular projects. Here are some examples

Koblenz & Nürnberg
https://staging.koblenz.sozialpass.app/erstellen

Bayern
https://staging.bayern.ehrenamtskarte.app/beantragen

michael-markl:
Please check whether the input form actually works on mobile devices (at least for Safari, Chrome & Firefox) (I am just adding this as the name `DesktopDatePicker` suggests otherwise :) )

f1sh1918:
> Please check whether the input form actually works on mobile devices (at least for Safari, Chrome & Firefox) (I am just adding this as the name `DesktopDatePicker` suggests otherwise :) )

We already use this component for koblenz and it was tested.
Don't panic @michael-markl :)

michael-markl:
Well, it doesn't work well on my Firefox for Android 😬

steffenkleinle:
> Well, it doesn't work well on my Firefox for Android 😬

Same for me actually.

f1sh1918:
Ok i expected that it was already tested on different Browsers before. For chrome it works like a charme.
That's now really bad before the koblenz launch :/

administration/src/mui-modules/application/primitive-inputs/DateForm.tsx
```diff
@@ -40,28 +41,27 @@ const DateForm: Form<State, ValidatedInput, AdditionalProps, Options> = {
...
```
michael-markl, line 48: We have to use controlled values here, as the state of the application form is persisted in local storage.

connium:
I have checked several websites that offer a date picker. On most of them, the user is shown a date picker and cannot enter the date value via a text field. The native `<input type="date">` element is also implemented this way.

Since Firefox seems to be the only browser where the text input does not work at the moment, I suggest to use the `DatePicker` (responsive variant) for Firefox and keep the `DesktopDatePicker` for all other browsers.

administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -1,51 +1,69 @@
...
+    0 0 0 0 rgba(205, 66, 70, 0),
+    inset 0 0 0 1px rgba(17, 20, 24, 0.2),
+  `
+
   return (
     <DesktopDatePicker
+      disabled={disabled}
+      label={label}
       views={['year', 'month', 'day']}
       value={date}
```
michael-markl, line 50: I think it would also be nice to add the prop `openTo="year"` as selecting a year is probably necessary for every of our date fields.

andrew8er:
@michael-markl 
 > Well, it doesn't work well on my Firefox for Android 😬

This should work now.

---

@connium 
>Since Firefox seems to be the only browser where the text input does not work at the moment, I suggest to use the DatePicker (responsive variant) for Firefox and keep the DesktopDatePicker for all other browsers. This should be fine for the users, as they are familiar with it from most sites.

The text input should work now as intended. We could however think about using a `DatePicker` here. The difference is explained [here](https://mui.com/x/react-date-pickers/base-concepts/#responsiveness)

administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -1,55 +1,75 @@
...
+
   return (
     <DesktopDatePicker
+      disabled={disabled}
+      label={label}
       views={['year', 'month', 'day']}
+      openTo='year'
       value={date}
-      format='dd.MM.yyyy'
+      format='dd.mm.yyyy'
```
f1sh1918, line 52: this breaks the date input for me. if i type a year. month will change to `00`

administration/src/mui-modules/application/primitive-inputs/DateForm.tsx
andrew8er: That was committed by accident, sorry.

administration/src/bp-modules/components/CustomDatePicker.tsx
andrew8er: Mh, works for me both in FF and Chrome, both mobile and desktop. Can you give me the exact steps to reproduce?

andrew8er:
#### Month issue on FF Mobile
So, currently I can reproduce the issue as follows:

- Click into a `<CustomDatePicker>` text field, select the days.
- Enter some days, focus will change to months
- Try to enter months

On Firefox Mobile, when entering months with a leading zero, the input fails, months with a leading `1` however will work. Also month input works again, when switching to year input and back to month input. Firefox Desktop, as well as Chrome Desktop/Mobile and Safari Desktop are unaffected.

- The corresponding issue in `mui-x` is most likely this: https://github.com/mui/mui-x/issues/13854
- The underlying issue is actually a Firefox bug, tracked here: https://bugzilla.mozilla.org/show_bug.cgi?id=1947143

Possible solutions:
1. Scrap the text input, only use a popover, i.e. what `<MobileDatePicker>` provides.
2. As @connium suggested, we could provide an extra version for FF users. I'm not a big fan of this, since this would make the behavior dependant on the user's browser and complicates user support and documentation quiet a bit.
3. Wait for the FF issue to be fixed and ignore the problem for now.
    - It is still possible to enter a date with the popover, the behavior is annoying though.
    - FF Mobile usage is probably minuscule.
    - The bug is already triaged and assigned to someone, but it will take some month for a fix to land in the release version.

#### Styling
I got the style pretty close to the current style, but the component height in the case for KoblenzPass is still slightly too high. Fixing this would require a lot of extra CSS however, because the component makes some assumptions about it's height when displaying the label (as seen on the Ehrenamtskarte Bayern form, not to be confused with the placeholder). The date picker's text field has its own prop `size` where the components size can be controlled in two steps.

The better solution IMHO would be to use MUI for all the input fields in KoblenzPass, since the current input components come from `blueprintjs`.

f1sh1918:
I'm okay with 3. since its rarely that people use firefox mobile and you can even use the datepicker in that case.

There are some small styling changes i would suggest f.e. the textfield `variant : filled` (for bavarian application process) and the `backgroundColor: white` is for that variant needed

I even couldn't find a way to reduce the paddings on quick testing but probably used the wrong class/prop for styling.

image

We already created a ticket for switching to mui also for koblenz so no worry in the slight height difference

administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -1,55 +1,81 @@
...
+          clearable: true,
+          onClear,
+        },
         textField: {
+          helperText: textFieldHelperText,
           placeholder: 'TT.MM.JJJJ',
           error: !isValid,
           spellCheck: false,
           onBlur,
+          style: textFieldStyle,
```
f1sh1918, line 59: pass here a variant.
` variant: 'filled',` for bavaria and `outlined` for koblenz

administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -1,55 +1,81 @@
...
+              ? `
+                0 0 0 0 rgba(205, 66, 70, 0),
+                inset 0 0 0 1px #cd4246,
+                inset 0 0 0 1px rgba(17, 20, 24, 0.2),
+                inset 0 1px 1px rgba(17, 20, 24, 0.3),
+              `
+              : undefined,
+            '.MuiPickersInputBase-root': {
+              fontSize: viewportSmall ? '16px' : '14px',
+              borderRadius: '2px',
```
f1sh1918, line 73: add here `backgroundColor: 'white'` that is required for the `filled` variant to align the background styling

andrew8er: I added some code to accommodate all kind of styles to match the requested style (see the last commit in this PR). However, MUI is clearly not designed for this. I think at some point we need to setup a way to have one theme for all MUI components per customer.

andrew8er:
The Material style EhrenamtsKarte Bayern is pretty dated now and not natively supported by MUI date pickers. I got it to work, but it is not pretty and we should think about updating the style, since its UX considered [inferior](https://medium.com/google-design/the-evolution-of-material-designs-text-fields-603688b3fe03) by the Google Design research team. Maybe as part of #1779.

f1sh1918 (approved):
Good work 👍 
Looks good

Just two small issues:
1. a11y issue that seem to be related to `enableAccessibleFieldDOMStructure`
The form field can not be escaped via tab if it has a value and the clear icon is shown
2. The second nit-picky one. Since you added this great deep merge for `textFieldSlotProps` could you please also pass the prop from the `BirthdayExtension` to adjust the height of the field.
```js
        textFieldSlotProps={{
          sx: {
            '.MuiPickersSectionList-root': {
              padding: '5px 0',
            },
          },
        }}
```
I think its a quite easy fix to adjust the height to look 99% equal :)

andrew8er:
> a11y issue that seem to be related to `enableAccessibleFieldDOMStructure` The form field can not be escaped via tab if it has a value and the clear icon is shown

I think this is a bug in the component itself, though I haven't found an issue. I tried a few options, but haven't found a good solution yet. We could use to disable keyboard access to both buttons (with `tabIndex: -1`) , but that is not making things better. The popover also has some keyboard navigation issues, for example, I cannot navigate beyond the first row of months.

seluianova:
:upside_down_face: nitpick, but the font in the birthday field is smaller then in other fields
image

seluianova:
we also have date in this form:
image
do we want to adjust it as well?

seluianova (commented):
The form allows me to submit an invalid year, such as 0011, if I enter it from the keyboard
image

andrew8er:
@seluianova 
> nitpick, but the font in the birthday field is smaller then in other fields

This was a style from the original component, but I agree, it looked out of place. Removed…

>  we also have date in this form:

I replaced the missing `<TextField>`s with `<CustomDatePicker>`, hope I caught everything now.

> The form allows me to submit an invalid year, such as 0011, if I enter it from the keyboard

We currently do not have any minimum date defined here, so this is technically correct. The original component provided a year picker starting from 1900 though. I think this would be a reasonable minimum date.

seluianova:
> We currently do not have any minimum date defined here

Yes, but in the current version of the form (main branch) it is not possible to submit a year such as 0011.
So I would treat this as a minor regression bug.
However if it's tricky to fix, I think we can create a separate issue for that, since it's not critical.

seluianova:
> We currently do not have any minimum date defined here

Yes, but in the current version of the form (main branch) it is not possible to submit a year such as 0011.
So I would treat this as a minor regression bug.
However if it's tricky to fix, I think we can create a separate issue for that, since it's not critical.

andrew8er:
The `<CustomDatePicker>` has a default `minDate` of 1900-01-01, but does not enforce it when entering the date via the text field, so I added another validation check.

seluianova:
:upside_down_face: When the birthday field is blank AND untouched, if I click “Nächster Schritt” without selecting a date, the error “Feld ist erforderlich.” appears in gray:
image
Expected: it should be red.
image

Sorry, I can't resist testing tricky cases when I review PRs 😀
But I think it's fine to move this "issue" to a separate ticket since it's not a blocking issue, up to you

seluianova:
In the card creation form (/cards/add) Ablaufdatum is outlined in red, even if the value is valid:
image

seluianova:
:upside_down_face: another nit-pick, probably might be ignored, but these fields look unnececarily long and the space on the right looks a bit odd:
image
This is how they looked like before:
image

seluianova:
:upside_down_face: Nürnberg project --> cards creation form (/cards/add):
As a tooltip, there was a validation error if the date was earlier than the min date.
image
Now this message is not displayed.

andrew8er:
#### birthday field
> When the birthday field is blank AND untouched, if I click “Nächster Schritt” without selecting a date, the error “Feld ist erforderlich.” appears in gray:

This does not happen for me with either FF or Chrome. Which browser are you using? Are the other fields filled? When clicking on the "Nächster Schritt" button, I get a notice to fill the form fields, nothing else. When I click on the birthday field, I get a red notice, as expected.

#### Misaligned fields
> Fields are missaligned here:

The misalignment stems from a lot of style customization for date picker's input field to match the current style. With #1779 we can ditch all those custom styles and this should be fixed automatically (assuming the MUI textfields are implemented coherently).

#### card creation date validation
> In the card creation form (/cards/add) Ablaufdatum is outlined in red, even if the value is valid:

Good catch! Fixed

#### date field icon padding
> another nit-pick, probably might be ignored for now, but these fields look unnececarily long and the space on the right looks a bit odd:

I noticed that too, and already tried to poke around the date picker's CSS but could not find a solution in a reasonable time. We already mess around the MUI date pickers CSS more than the the developers probably intended, but when implementing #1779, these should look normal again. If you find a quick'n'dirty solution, feel free to add a commit however!

#### missing tooltip
> 🙃 Nürnberg project --> cards creation form (/cards/add):
As a tooltip, there was a validation error if the date was earlier than the min date.

This tooltip here was automatically generated by the MUI `<TextField>`s. I think tooltips have a bad UX, especially for mobile. We generally use `helperText`s for displaying an error message beneath forms element. I could add this, but we would need appropriate translations.

administration/src/bp-modules/components/CustomDatePicker.tsx
```diff
@@ -1,62 +1,88 @@
...
-      minDate={minDate}
-      maxDate={maxDate}
-      onChange={onChange}
-    />
-  )
-}
+        textFieldSlotProps
+      ),
+    }}
+    enableAccessibleFieldDOMStructure
```
f1sh1918, line 78: can you please enable this only for viewport small @andrew8er 
Not be able to escape the field with keyboard is bad a11y

andrew8er: MUI switches between the desktop and mobile variant based on the `@media (pointer: fine)` query, as explained [here](https://mui.com/x/react-date-pickers/date-picker/#available-components). I think it would be more consistent to use this query here also then.

f1sh1918: I'm open for any solutions as long as I can leave the field with keyboard :)

seluianova:
> This does not happen for me with either FF or Chrome. Which browser are you using? Are the other fields filled?

@andrew8er yes, all mandatory fields before Birthday should be filled and valid, otherwise when the button is clicked, it validates only the first invalid field, and the rest are not validated.
I use Chrome.

seluianova:
> This does not happen for me with either FF or Chrome. Which browser are you using? Are the other fields filled?

@andrew8er yes, all mandatory fields before Birthday should be filled and valid, otherwise validation of the birthday field will not trigger.
I use Chrome.

administration/src/mui-modules/application/primitive-inputs/DateForm.tsx
```diff
@@ -41,28 +50,46 @@ const DateForm: Form<State, ValidatedInput, AdditionalProps, Options> = {
...
+            '.MuiInputLabel-root': {
+              transform: 'translate(0px, 13px) scale(1)',
+            },
+            '.MuiInputLabel-shrink': {
+              transform: 'translate(0px, -1.5px) scale(0.75)',
+            },
+            '.MuiFormHelperText-root': {
+              marginLeft: 0,
+            },
+          },
```
michael-markl, line 91:
```suggestion
          variant: 'standard',
          style: { margin: '4px 0', minWidth },
          sx: {
            'input': { paddingTop: '4px' },
            '.MuiInputAdornment-root': { transform: 'translateX(-8px)' }
          }
```

I tried locally to use the standard variant, and it seems to work better than filled, with only some minor adjustments.

Feel free to ignore this comment though.

andrew8er: Agreed, this is better.

f1sh1918:
Just found two small issue in the statistics button bar
1. There is a padding-right that shouldn't be there after the icon
2. Labels are not vertically center

image

The rest looks and works good afaics :)

andrew8er:
>    There is a padding-right that shouldn't be there after the icon

This came from a overly unpecific nested style in the `<StickyBottomBar>`, took me an hour to find 🤯 

>    Labels are not vertically center

I removed all blueprint components and replaced them with MUI components, look good IMHO now.

michael-markl (approved):
Nice work! Looks good to me. (I tested Bayern application form, statistics date form, cards creation date form, and Koblenz self-portal form)

Issue #1968, opened 2025-03-08
Map stores to region


f1sh1918:
### Describe the Problem

With #537 each store in the database has a foreign key reference to the region it belongs to. For now, all stores reference a mock region though and not the actual region that issued the store. 

**Describe the solution you'd like**
- Accepting stores should reference the correct `regionId`. 
- The stores from the freinet source have an `agencyId`, `freinetAgencies` can be used for reference the correct region
- The stores from LBE have an `anbieterName` field that can be mapped which region `name` and `prefix`. (There may be some spelling differences)
- The stores that are not from eak project should be referenced to the single region of the project (koblenz, nürnberg)

**Describe alternatives you've considered**
None

### Additional Information
Data source: https://www.ehrenamt.bayern.de/xml-json/app-daten.xml

f1sh1918:
This issue was originally included in #538 but i splitted it up

PR #1975, opened 2025-03-10
1968: Map stores to regions

f1sh1918:
### Short Description

As we want to add a reporting function at some point for accepting stores, we have to map them to a particular region

### Proposed Changes

- map eak stores with `freinetId` or `districtName` to region
- for projects with only one region (f.e. social passes) map the first matching region
- adjust region `München, Landeshauptstadt` to `München` (not sure if they will like it), we could also add the addition to a separate database field. But in this case we need exact prefix and name for proper mapping

### TODO
- cleanup
- fix tests
- some freinet agencies have duplicated region identifier, thats why 239 stores couldn't get mapped correctly. Waiting for response from freinet when they fix it

### Side Effects

- likely none

### Testing

- run the import job (backend) locally and check if regionIds were updated
- import stores for nürnberg and koblenz via project store manager and check if they have correct regionId in database

### Resolved Issues

Fixes: #1968

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
@@ -1,6 +1,7 @@
 package app.ehrenamtskarte.backend.regions.database.repos
 
 import app.ehrenamtskarte.backend.common.database.sortByKeys
+import app.ehrenamtskarte.backend.freinet.database.FreinetAgencies
 import app.ehrenamtskarte.backend.projects.database.Projects
 import app.ehrenamtskarte.backend.regions.database.RegionEntity
 import app.ehrenamtskarte.backend.regions.database.Regions
@@ -59,4 +60,15 @@ object RegionsRepository {
             .single().id
         return RegionEntity[regionId]
     }
+
+    fun findRegionByNameAndPrefix(name: String, prefix: String): RegionEntity? =
+        RegionEntity.find { (Regions.name eq name) and (Regions.prefix eq prefix) }.singleOrNull()
+
+    fun findRegionByFreinetId(freinetId: Int): RegionEntity? =
+        (FreinetAgencies innerJoin Regions)
+            .slice(Regions.columns)
+            .select { FreinetAgencies.agencyId eq freinetId }
+            .singleOrNull()?.let {
+                RegionEntity.wrapRow(it)
+            }
 }
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/database/repos/AcceptingStoresRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/database/repos/AcceptingStoresRepository.kt
@@ -1,7 +1,10 @@
 package app.ehrenamtskarte.backend.stores.database.repos
 
 import app.ehrenamtskarte.backend.common.database.sortByKeys
+import app.ehrenamtskarte.backend.projects.database.ProjectEntity
 import app.ehrenamtskarte.backend.projects.database.Projects
+import app.ehrenamtskarte.backend.regions.database.RegionEntity
+import app.ehrenamtskarte.backend.regions.database.repos.RegionsRepository
 import app.ehrenamtskarte.backend.stores.COUNTRY_CODE
 import app.ehrenamtskarte.backend.stores.database.AcceptingStoreEntity
 import app.ehrenamtskarte.backend.stores.database.AcceptingStores
@@ -88,8 +91,8 @@ object AcceptingStoresRepository {
 
     fun getIdIfExists(
         acceptingStore: AcceptingStore,
-        projectId:
-            EntityID<Int>
+        projectId: EntityID<Int>,
+        regionId: EntityID<Int>?
     ): Int? {
         return AcceptingStores.innerJoin(PhysicalStores).innerJoin(Addresses).innerJoin(Contacts)
             .slice(AcceptingStores.id).select {
@@ -103,7 +106,7 @@ object AcceptingStoresRepository {
                     (AcceptingStores.name eq acceptingStore.name) and
                     (AcceptingStores.description eq acceptingStore.discount) and
                     (AcceptingStores.categoryId eq acceptingStore.categoryId) and
-                    (AcceptingStores.regionId.isNull()) and // TODO #538: For now the region is always null
+                    (AcceptingStores.regionId eq regionId) and
                     (AcceptingStores.projectId eq projectId) and
                     (
                         PhysicalStores.coordinates eq Point(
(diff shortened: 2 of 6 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/database/repos/AcceptingStoresRepository.kt
```diff
@@ -114,17 +117,25 @@ object AcceptingStoresRepository {
             }.firstOrNull()?.let { it[AcceptingStores.id].value }
     }
 
-    fun importAcceptingStores(stores: List<CSVAcceptingStore>, projectId: EntityID<Int>, dryRun: Boolean): StoreImportReturnResultModel {
+    fun importAcceptingStores(stores: List<CSVAcceptingStore>, project: ProjectEntity, dryRun: Boolean): StoreImportReturnResultModel {
         var numStoresCreated = 0
         var numStoresUntouched = 0
+        val projectId = project.id
+        val region = RegionsRepository.findAllInProject(project.project).let {
```
f1sh1918, line 124: Not sure if there is a nicer solution for the social passes.
Another possibility would be to have a `Region_Store_Manager` Role and provide a region when uploading the csv

backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
```diff
@@ -59,4 +60,15 @@ object RegionsRepository {
...
+    fun findRegionByNameAndPrefix(name: String, prefix: String): RegionEntity? =
+        RegionEntity.find { (Regions.name eq name) and (Regions.prefix eq prefix) }.singleOrNull()
+
+    fun findRegionByFreinetId(freinetId: Int): RegionEntity? =
+        (FreinetAgencies innerJoin Regions)
+            .slice(Regions.columns)
+            .select { FreinetAgencies.agencyId eq freinetId }
+            .singleOrNull()?.let {
+                RegionEntity.wrapRow(it)
+            }
```
michael-markl, line 73: I think it would be good to also filter by project id, to make sure that we do not end up returning a region from a  different project.

f1sh1918: ah alright and also for both functions. thx for the hint

f1sh1918: added in last commit

seluianova:
:question: do we want to keep nullable regionId for some cases?

f1sh1918:
> ❓ do we want to keep nullable regionId for some cases?

it might happen that we can't get a region for bavaria if some freinet id is outdated or some `anbieter_name` does not 100% fit, so i wanted to avoid that they get filtered out

f1sh1918:
Result of the mapping

Total stores for eak: 5839
Stores with region for eak: 5797
= 99,3% of stores are mapped

There is just one region not correctly mapped by freinet (LK Ansbach) but i guess this will be fixed soon.

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/database/repos/AcceptingStoresRepository.kt
```diff
@@ -114,17 +117,25 @@ object AcceptingStoresRepository {
...
+            if (it.size == 1) {
+                it.first()
+            } else {
+                null
+            }
+        }
```
steffenkleinle, line 130: :upside_down_face: 
```suggestion
        val region = RegionsRepository.findAllInProject(project.project).singleOrNull()
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/importer/common/steps/Store.kt
```diff
@@ -30,17 +31,18 @@ class Store(config: ImportConfig, private val logger: Logger) :
                 var numStoresUntouched = 0
 
                 for (acceptingStore in input) {
+                    val region = getRegionIdFromFreinetIdOrDistrictName(project.id, acceptingStore.freinetId, acceptingStore.districtName)
```
steffenkleinle, line 34: :upside_down_face: 
```suggestion
                    val region = getRegionFromAcceptingStore(project.id, acceptingStore)
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/utils/MapStoreToRegion.kt
```diff
@@ -0,0 +1,15 @@
+package app.ehrenamtskarte.backend.stores.utils
+
+import app.ehrenamtskarte.backend.regions.database.RegionEntity
+import app.ehrenamtskarte.backend.regions.database.repos.RegionsRepository
+import org.jetbrains.exposed.dao.id.EntityID
+
+fun getRegionIdFromFreinetIdOrDistrictName(projectId: EntityID<Int>, freinetId: Int?, districtName: String?): RegionEntity? {
```
steffenkleinle, line 7: :wrench: See suggestion above. In any case, the naming is not correct as you return a whole region and not just an id.

steffenkleinle (approved):
Tested, looks good

seluianova:
❓ do we want to support csv store import for projects with more than 1 region? 
(I mean in general: probably we need a separate task to add regionId attribute to the csv import ?)

f1sh1918:
> ❓ do we want to support csv store import for projects with more than 1 region? (I mean in general: probably we need a separate task to add regionId attribute to the csv import ?)
> 
> or maybe we need an error handling for this case in the import acceprance store mutation? if project has more than 1 region --> return a meaningful error message

Good point.
I would say yes. I think it may make sense to create a separate role `REGION_STORE_MANAGER` or allow `RegionAdmins` to do that (there is no regionId needed)

For the `PROJECT_STORE_MANAGER` we should work with `ARS` so a region identifier in the csv instead of internal `regionIds`
So we can create a task for this and then remove this auto assigning region in the backend.
And then we also need a validation and a proper error message if the region is not available

I'm not sure if it makes sense to allow both roles in one project, because they may override stores

seluianova:
> Should i add the error handling directly here @seluianova ?

@f1sh1918 this one? "if project has more than 1 region --> return a meaningful error message"
for me it sounds like a small code change and could be added here, if I'm not missing anything. but up to you :)

seluianova:
> Update: Now all stores should be mapped :)

actually I currently get 2183 mapped stores and 3656 stores with regionId = null 🤔
not sure if I did something wrong

f1sh1918:
alright i throw an exception now which can only be tested if you pass a project with multiple regions 
 f.e. `bayern.ehrenamtskarte.app` manually
 
 For the other issue i created a ticket #2012

backend/src/main/kotlin/app/ehrenamtskarte/backend/stores/database/repos/AcceptingStoresRepository.kt
```diff
@@ -114,17 +118,20 @@ object AcceptingStoresRepository {
             }.firstOrNull()?.let { it[AcceptingStores.id].value }
     }
 
-    fun importAcceptingStores(stores: List<CSVAcceptingStore>, projectId: EntityID<Int>, dryRun: Boolean): StoreImportReturnResultModel {
+    fun importAcceptingStores(stores: List<CSVAcceptingStore>, project: ProjectEntity, dryRun: Boolean): StoreImportReturnResultModel {
         var numStoresCreated = 0
         var numStoresUntouched = 0
+        val projectId = project.id
+        // TODO 2012 provide region ars
+        val region = RegionsRepository.findAllInProject(project.project).singleOrNull() ?: throw RegionNotUniqueException()
```
seluianova, line 126: :upside_down_face: might be nice to add a test for this error

seluianova (approved):
lgtm, thanks :+1:

f1sh1918:
> > Update: Now all stores should be mapped :)
> 
> actually I currently get 2183 mapped stores and 3656 stores with regionId = null for bayern 🤔 not sure if I did something wrong. do you still get all mapped now?
> 
> upd. ah, it's because I don't have any freinet agencies loaded
> 
> upd-2. I loaded freinet agencies and got 81 stores with null region @f1sh1918 (not sure if we want to do anything about it)

Ah sorry i forgot to answer @seluianova 
I have just 3 stores from eak that can not be mapped because they have a wrong agencyId and shouldn't be included in the data stream. I will tell freinet (again)

The stores for other projects just have to be reuploaded via csv to get mapped correctly

seluianova:
@f1sh1918 I got 81 stores from eak that are not mapped.
78 of them have freinetId=1422
and 3 of them have freinetId=1732

I wonder why you have no issues with 1422 ?
do you get this agency from freinet api ?

Issue #1979, opened 2025-03-11
Server Error occurs when I discard entries


bahaaTuffaha:
### Describe the Bug

Server Error occurs when I discard entries at Bayern's application.

### Steps to Reproduce

1. Go to `/beantragen`
2. Fill the form choose blue +`Ich bin Inhaber:in einer JuLeiCa (Jugendleiter:in-Card). *` 
3. At stage 4 (Antrag Senden). just click discard (Alle Eingaben verwerfen)
4. Look at the console.

![Image](https://github.com/user-attachments/assets/5c6bc11d-922f-4a38-9470-0f71bdf3315f)

```
org.jetbrains.exposed.dao.exceptions.EntityNotFoundException: Entity RegionEntity, id=0 not found in the database
	at org.jetbrains.exposed.dao.EntityClass.get(EntityClass.kt:34)
	at org.jetbrains.exposed.dao.EntityClass.get(EntityClass.kt:36)
	at app.ehrenamtskarte.backend.regions.database.repos.RegionsRepository.findRegionById(RegionsRepository.kt:41)
	at app.ehrenamtskarte.backend.regions.webservice.schema.RegionsQueryService$regionByRegionId$1.invoke(RegionsQueryService.kt:41)
	at app.ehrenamtskarte.backend.regions.webservice.schema.RegionsQueryService$regionByRegionId$1.invoke(RegionsQueryService.kt:40)

```

### Related Issues

none

f1sh1918:
@bahaaTuffaha can you check which "Empfänger des Antrags"  you've chosen?

<img width="1031" alt="Image" src="https://github.com/user-attachments/assets/28d03da3-a8d7-4106-a575-f9d09d7cdce7" />

I guess this error happens because the region was not properly set and therefore the data privacy policy couldn't be loaded

Could be a duplicate of this https://github.com/digitalfabrik/entitlementcard/issues/1905
Never known how to reproduce it, so maybe we can now find out why this happens

PR #1980, opened 2025-03-11
1979: Fix region error discard entries

f1sh1918:
### Short Description

Currently we receiving some internal errors in our backend log, since the `getDataPrivacyPolicy` query is fetched without a valid region id, when the application data was discarded

### Proposed Changes

- add a skip for the query for empty string or invalid number

### Side Effects

- none

### Testing

1. Go to /beantragen for bavaria
2. Type in your personal data and requirements
3. Go to the last step and choose "Alle Eingaben verwerfen"
4. Check browser console log (no 500 error should be there) and check the backend log (local dev) that there are no errors anymore

### Resolved Issues

Fixes: #1979


steffenkleinle (approved):
Tested in firefox, works as expected

bahaaTuffaha (approved):
Nice, it works as expected 🚀

Issue #1978, opened 2025-03-11
Wrong title of websites


connium:
### Describe the Bug

When I visit the https://koblenz.sozialpass.app/ page, the browser displays "Showcase Berechtigungskarte Verwaltung" as the title.
The same title is also displayed on the [privacy page](https://koblenz.sozialpass.app/data-privacy-policy), the [create page](https://staging.koblenz.sozialpass.app/erstellen) and other pages. This also happens for our other projects.

### Steps to Reproduce

1. Go to https://koblenz.sozialpass.app/
2. See the browser title

### Expected Behavior

I'd expect that the management area shows something like "{ProjectTitle} Verwaltung" (e.g. "KoblenzPass Verwaltung") and the data privacy page has a title like "{ProjectTitle} Datenschutzerklärung".

### Additional Information

I also recognized, that in the html `header` an icon is specified, but this icon does not exist.
`<link rel="icon" href="/icons/showcase.entitlementcard.app.png">`

PR #1984, opened 2025-03-12
1978: Wrong title of website

f1sh1918:
### Short Description

Since the `ProjectConfigProvider` was refactored it looks like the meta tags hook does not properly work outside of the context.

### Proposed Changes

- move meta tags hook inside the project config provider (context), 
- add fallback logo for dev purposes

### Side Effects

- likely none

### Testing

1. Check the favicon, and page title of the different projects. It should be set now correctly even for the login page
2. On local dev you have to select the particular project

### Resolved Issues

Fixes: #1978


steffenkleinle (approved):
Tested locally on firefox, works as expected!

andrew8er (approved):

Issue #1986, opened 2025-03-12
Make location permission request compliant with apple guidelines


steffenkleinle:
### Describe the Bug

Make location permission request compliant with apple guidelines:
```
The app encourages or directs users to allow the app to access the location. Specifically, the app directs the user to grant permission in the following way(s):

- A custom message appears before the permission request, and to proceed users press a "Standort freigeben" button. Use words like "Continue" or "Next" on the button instead.

- A custom message appears before the permission request, and the user can close the message and delay the permission request with the Fertig button. The user should always proceed to the permission request after the message.

Permission requests give users control of their personal information. It is important to respect their decision about how their data is used.
```

### Steps to Reproduce

1. See email sent to us about koblenz

### Expected Behavior

The app passes the review.

### Related Issues

N/A

### Additional Information

N/A

PR #1987, opened 2025-03-12
1986: Location permission request compliant with apple app store guidelines

steffenkleinle:
### Short Description

Make location permission request compliant with apple app store guidelines.

### Proposed Changes

- Rename `Done` to `Next`
- Remove location request button on intro slide
- Request permissions on done button click

### Side Effects

N/A

### Testing

Delete app data, navigate through the intro slides and finish by seeing the location permission request.
After handling the permission request the location_button on the map should display correct status

Note: follow user location does not work without pushing on the button, but it also didn't before. Can be done in separate issue

### Resolved Issues

Fixes: #1986


f1sh1918 (approved):
Works good. Thx for the quick fix 👍 

💭 I was just thinking if it would be better if we implement the permissions request in a lifecycle method of the map page to appear when the map page was opened and no permissions were requested. Since then its directly connected to the map page and not to the intro slides. If for some reason the permission become undefined it can be rerequested easily (and intro slides will never show up)
But in any case also the gps button can be used for that.
So i think the implementation is okay as it is.
For me this request belongs more to the map page than to the intro slides :)

f1sh1918:
@steffenkleinle one issue with the request could be for eak.
1.user downloads the app
2.user activate card via deep link.
3. After clicking activation the card will be shown. If the user navigates to map the request won't be shown
4.the Intro screen will be shown on the next "regular" App start

frontend/lib/map/map_page.dart
```diff
@@ -39,6 +39,14 @@ abstract class MapPageController {
 class _MapPageState extends State<MapPage> implements MapPageController {
   MapController? _controller;
 
+  @override
+  void initState() {
+    super.initState();
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) async {
```
f1sh1918, line 45: ❓ Are you sure that we need the `addPostFrameCallback` here
Afaics its not required that the ui is ready when the permission will be requested.
Did you face any issues without that?

steffenkleinle: For me it actually does not show a permission request without this. Could you perhaps check and verify this?

f1sh1918: in theory it would be working but we have to wait for translation provider. So its okay
Another issue is now that the `followUserLocation` has to be retriggered.
At least on ios it will not be executed after granting permission

f1sh1918: i'm also getting an error
```js
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: A request for location permissions is already running, please wait for it to complete before doing another request.
#0      GeolocatorApple.requestPermission (package:geolocator_apple/src/geolocator_apple.dart:63:7)
<asynchronous suspension>
#1      checkAndRequestLocationPermission (package:ehrenamtskarte/location/determine_position.dart:129:29)
<asynchronous suspension>
#2      _MapPageState.initState.<anonymous closure> (package:ehrenamtskarte/map/map_page.dart:46:7)
<asynchronous suspension>

```

michael-markl: It's also a bit weird, that we would ask for permission every time the user opens the app, even though, the user might not *want* to give us permission; so he has to dismiss this dialog every time.
Imo, it would then be better to just wait for the user to click on the `location_button`.

f1sh1918: thats a good point @michael-markl but apple requested this explicitly.
So if we say we only request permissions if the user clicked on the location button of the map it might not be convenient for them.

I updated them pr so this is outdated.

f1sh1918:
@steffenkleinle i updated the pr to the only working solution i found.
So the user has to answer the permission request before the map page will be loaded.
That enables that the location button shows the correct permission status.

The general problem here is, that we can't properly update state in a `addPostFrameCallback` for android, since we are facing issues that the widget is not `mounted` anymore when we update the state. 

Since we have to find a solution for it quickly, i created this workaround. Feel free to propose a better solution.

I would suggest to create a follow up issue to improve the permission handling and request them on the `location_button` or `map_page` where they belong to.

andrew8er (approved):
Works for me.

When denying the permission, tapping the locator button opens a dialog which offers to navigate to the apps location settings. However on iOS, this sends me to the general app settings page, which is currently empty. [This](https://github.com/Baseflow/flutter-geolocator/issues/146) Github discussion might have some hints.

f1sh1918:
> Works for me.
> 
> When denying the permission, tapping the locator button opens a dialog which offers to navigate to the apps location settings. However on iOS, this sends me to the general app settings page, which is currently empty. [This](https://github.com/Baseflow/flutter-geolocator/issues/146) Github discussion might have some hints.

Hm for me it sends me to the correct settings page using ios 18.0 @andrew8er 
image

andrew8er:
> Hm for me it sends me to the correct settings page using ios 18.0

I used iOS 18.3.

f1sh1918:
> > Hm for me it sends me to the correct settings page using ios 18.0
> 
> I used iOS 18.3.

Okay then please open an issue for that with the particular ios version and the additional info you mentioned here.
I should update my Xcode to recent though :)

Issue #1967, opened 2025-03-06
Enforce kotlin code style


steffenkleinle:
### Describe the Problem

Currently, we don't have a specific code style enforced in the backend. Therefore, the code is inconsistent and does not follow code style best practices (e.g. trailing commas).

### Describe Your Preferred Solution

- Add a kotlin style following best practices
- Enforce the code style by adding a check in the CI
- If possible, add automatic formatting to IntelliJ save and pre-commit hooks

### Describe Possible Alternatives

Dont enforce a code style.

### Related Issues

N/A

### Additional Information

Discussed here: https://github.com/digitalfabrik/entitlementcard/pull/1902#discussion_r1971509057
Kotlin coding conventions: https://kotlinlang.org/docs/coding-conventions.html#trailing-commas
Ktlint gradle plugin: https://github.com/JLLeitschuh/ktlint-gradle

Comment by @andrew8er:
> We are using https://github.com/JLLeitschuh/ktlint-gradle as the Ktlint gradle plugin. I tested it and found the currently configured plugin not working as intended (well, not at all). Updated to the latest version, and it works. But before applying all the changes, I would like to check out all the configured rules first. We can also setup Ktlint to automatically create code styles for IntelliJ then.

PR #1996, opened 2025-03-19
1967: Enforce Kotlin code style

andrew8er:
### Short Description

Adds [Ktlint](https://pinterest.github.io/ktlint/latest/) and [Detekt](https://detekt.dev/) config and Gradle plugins.

### Proposed Changes

- This PR adds both Ktlint and detekt code check infrastructure. While both tools have some overlap (like checking line lengths), Ktlint cares more about the formatting, while detekt has some code smell and complexity checks. Detekt can apparently also [use Ktlint rules](https://detekt.dev/docs/intro#adding-more-rule-sets), but I haven't got this to work yet and since Ktlint – unless detekt – offers automatic formatting (with the IntelliJ plugin as well as a Gradle task) I kept it separate for now. Once all the code is formatted properly, this would be an option to think about however.

- The Ktlint Gradle plugin used to have a companion plugin that can automatically create an IntelliJ code style, but it is [unmaintained](https://github.com/JLLeitschuh/ktlint-gradle/issues/774) now. Instead I updated the existing code style to match. This needs to be kept in sync with the Ktlint/detekt settings manually.

- The Ktlint Gradle plugin is [seeking maintainers](https://github.com/JLLeitschuh/ktlint-gradle/issues/569) and might not be reliable in the longer run, but I could not find any good alternative out there. Maybe using Ktlint rules through detekt will resolve this.

- I have tried to configure the rules to not be overly draconian, some rules just do not make any sense to me, like [multiline expression wrapping](https://pinterest.github.io/ktlint/latest/rules/standard/#multiline-expression-wrapping). We can discuss the rules in this PR.

- I have not formatted the code in this repo accoprding to the rules, since I would like to discuss the rules first. To actually autoformat the code according to Ktlint use `./gradlew ktlintFormat`. This will fix _some_ issues Ktlint complains about, but not all, some still need to be resolved manually.

- ~~Once we're settled on the rules, I can also add the CI config to actually enforce the style on Github.~~ CircleCI will complain, as long as the formatting is not in place.

### Side Effects

You will probably curse Ktlint or detekt.

### Testing

- Run Ktlint diagnostic: `./gradlew ktlintCheck`
- Run Ktlint formatter: `./gradlew ktlintFormat`
- Run Detekt diagnostic: `./gradlew detekt`

### Resolved Issues

Fixes: #1967


f1sh1918:
I just googled some rules and they seem to be good to me.
Even i agree not be to strict at least in the first place, since it will be even lots of effort to align the current code to the rules introduced.
The plugins work fine for me. Good work 👍 
I would appreciate in end to add the command in the intellij `runConfig` maybe all three of them and name it `Format & Linting` or sth. similar

andrew8er:
> I even not sure if we could increase the line length a bit but it's reasonable to be display in one line even on smaller displays.

I used to have 120 characters but now tend to have 100. With a smaller screen, this width also fits two side bars from IDEs. I'm not really that opinionated about the line length though, I would also tolerate the occasional longer line.

andrew8er:
Both the `frontend` Dart code style and the `administration` prettier settings specify a line length of 120. I'd suggest we use this value to have a consistent line length then, otherwise we would need to reformat the whole code base.

andrew8er:
The CodeScene errors about `Global Conditionals` are completely bogus, `Large Method` might stem from reformatting and should probably be suppressed.

.editorconfig
```diff
@@ -0,0 +1,26 @@
...
+insert_final_newline = true
+
+[{*.kt,*.kts}]
+# Android Studio standard
+max_line_length = 120
+ij_kotlin_code_style_defaults = KOTLIN_OFFICIAL
+ij_kotlin_allow_trailing_comma = true
+ij_kotlin_allow_trailing_comma_on_call_site = true
+#  Disable wildcard imports entirely
+ij_kotlin_name_count_to_use_star_import = 2147483647
```
steffenkleinle, line 15: :thinking: ~~Could you quickly explain what these magic numbers mean?~~ Just saw that this is just taken from ktlin config, still looks weird. :D

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/database/repos/ApplicationRepository.kt
```diff
@@ -158,10 +154,12 @@ object ApplicationRepository {
...
             }
         }
-    }
 
-    fun delete(applicationId: Int, graphQLContext: GraphQLContext): Boolean {
-        return transaction {
+    fun delete(
+        applicationId: Int,
+        graphQLContext: GraphQLContext,
+    ): Boolean =
```
steffenkleinle, line 161: :wrench: Is there a way to configure this to not force the linebreaks for multiple parameters here? I think its easier to read without it.
```suggestion
    fun delete(applicationId: Int, graphQLContext: GraphQLContext): Boolean =
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -21,15 +21,19 @@ import org.jetbrains.exposed.sql.transactions.transaction
...
-        dfe: DataFetchingEnvironment
+        dfe: DataFetchingEnvironment,
     ): DataFetcherResult<Boolean> {
-        val applicationHandler = ApplicationHandler(dfe.getContext<GraphQLContext>(), application, regionId, project)
+        val applicationHandler = ApplicationHandler(
+            dfe.getContext<GraphQLContext>(),
+            application,
+            regionId,
+            project,
+        )
```
steffenkleinle, line 36: :wrench: With a line length of 120 chars this should still be fine? Same for quite a few other changes.

f1sh1918 (approved):
Looks good to me. 👍 
Thx for all the work.
I added a comment for one role that we may add, since i also think that not each parameter of a function should be wrapped in a separate line but respect the line length or a higher limit like `4` or sth similar since we should avoid passing too many functional parameters in general.

steffenkleinle (approved):
Approving so we can merge this ASAP to avoid further merge conflicts. Would be cool if you check the changes with the linebreaks again (and revert if applicable).

.editorconfig
andrew8er: Yes, apparently, there is no better way to disable this.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
andrew8er: I think the autoformatter is a bit overly enthusiastic. I corrected a few of these instances manually, but missed this one.

andrew8er:
I reconciled all conflicts with `theirs` and reformatted those files to not overwrite any changes.

Issue #2019, opened 2025-03-25
Add generate release github template


f1sh1918:
### Describe the Problem

As a tester i might want to identify the issue that has to be tested quickly, therefore some categorization for the prs would be great within the changelog

### Describe Your Preferred Solution

- add a `release.yml` template with particular categories

PR #2020, opened 2025-03-25
2019: Github release template

f1sh1918:
### Short Description

As it might take time for a tester to identify testable changes in a release, we may add a release template to distinguish between testable and not testable changes

### Proposed Changes

- add a category "testable features, fixes, enhancements" that includes everything except `not-testable` labeled prs
- create a category "not testable for end users" that only displays `not testable` labeled prs
- label `exclude-changelog` can be used for prs and shouldn't appear in the changelog at all

Note: I tried to keep it as simple as possible so no additional effort from us is needed. We could even create more categories if its needed but then we would have to assign all the particular labels in the pr. I also wanted to keep it `opt-in` so we ensure that no changes will be lost. The worst case is that we forget to mark a pr `not-testable` and it will appear in the first category

### Side Effects

- none

### Testing

You could generate a release based on this branch

### Resolved Issues

Fixes: #2019 

This is how it would look like

image


f1sh1918:
i talked to martin and he would be happy if we just add a label `Nuernberg` in the pr and show it separately in the changelog.
I will add this category. I still think this is way less effort than creating separate release notes files for end user which we may not need in the end since nobody will read it

.circleci/src/commands/install_app_toolbelt.yml
```diff
@@ -7,4 +7,4 @@ steps:
         echo 'export PATH=~/.npm-global/bin:"$PATH"' >> "$BASH_ENV"
   - run:
       name: Install app-toolbelt
-      command: npm install --unsafe-perm -g https://github.com/digitalfabrik/app-toolbelt/archive/refs/heads/main.tar.gz
\ No newline at end of file
+      command: npm install --unsafe-perm -g https://github.com/digitalfabrik/app-toolbelt/archive/refs/tags/v0.2.tar.gz
```
seluianova, line 10: special thanks :+1:

steffenkleinle, line 10:
```suggestion
      command: npm install --unsafe-perm -g https://github.com/digitalfabrik/app-toolbelt/archive/refs/tags/v0.2.tar.gz

```

.github/release.yml
```diff
@@ -0,0 +1,15 @@
...
+    - title: Testable features, fixes, enhancements 🎉
+      labels:
+        - '*'
+      exclude:
+        labels:
+        - not-testable
+        - Nuernberg
+    - title: Testable features for nuernberg 🎉
+      labels:
+        - Nuernberg
```
steffenkleinle, line 12: :thinking: Why do we only need this for Nürnberg but not for Bayern/Koblenz? And I think that we have to be careful when assigning labels to PRs this way that we base it on which projects the changes affect instead of who hired/pays us to implement the feature.

.github/release.yml
```diff
@@ -0,0 +1,15 @@
+changelog:
+  categories:
+    - title: Testable features, fixes, enhancements 🎉
```
steffenkleinle, line 3: :upside_down_face:
```suggestion
    - title: Features, fixes, enhancements 🎉
```

.github/release.yml
```diff
@@ -0,0 +1,15 @@
...
+      labels:
+        - '*'
+      exclude:
+        labels:
+        - not-testable
+        - Nuernberg
+    - title: Testable features for nuernberg 🎉
```
steffenkleinle, line 10: :upside_down_face: 
```suggestion
    - title: Nuernberg :tada: 
```
Maybe we even have a more distinct emoji on top? For example :orange_heart: ?

.github/release.yml
```diff
@@ -0,0 +1,15 @@
...
+      labels:
+        - Nuernberg
+    - title: Not testable for end users 🛠
```
steffenkleinle, line 13: :upside_down_face: 
```suggestion
    - title: Maintenance 🛠
```

steffenkleinle (approved):

.github/release.yml
f1sh1918: Because we only have to inform nuernberg about some  important changes. Currently no other customer requested that

.circleci/src/commands/install_app_toolbelt.yml
f1sh1918: Why we need an empty line?

.github/release.yml
f1sh1918: I think both are features that have to be tested. I don't we have to use different emojis tbh

.circleci/src/commands/install_app_toolbelt.yml
steffenkleinle: We use to have an empty line at the end of files. But optional since we don't enforce it for .yml files.

.github/release.yml
steffenkleinle: > I think both are features that have to be tested

I think thats clear without putting `Testable` in the category name.

.github/release.yml
steffenkleinle: I still think it would be helpful for both Martin, people from other teams and even us to see at a glance which changes are only related to a specific project. Also it feels inconsistent to only do this for Nuernberg.

f1sh1918: I don't think this will work, since changelog items won't be duplicated and some changes may not only be related to nuernberg but nuernberg may want to test them.
If one changelog item applies to a category it won't be shown in another again. 
We just highlight here changes that nuernberg may want to test but may be also related to other projects

.github/release.yml
f1sh1918: ok. md emoji usage won't work here.

seluianova:
🙃 Do we want to exclude “release”-PRs (like 1922) from release notes?

seluianova (approved):

.github/release.yml
steffenkleinle: For me thats even more reason to specify the other projects as well, such that we can tell Nuernberg that everything for all projects and Nuernberg is relevant? Or did you plan to label every single PR with Nuernberg, even if it affects the other projects as well? I am a little confused here, probably I am just not getting it though.

f1sh1918:
> 🙃 Do we want to exclude “release”-PRs (like 1922) from release notes?

This can only be done an exclude label set on the pr @seluianova

.github/release.yml
f1sh1918: Ok it looks like you didn't understand the mechanism how the release notes work, so i will try to explain it.
Lets say we have feature that affects nuernberg and koblenz.
Then we would add both project labels in the pr.
The release note template will take the first category listed in the template that fits to one label

So lets say we create a category for nuernberg that only shows prs with nuernberg label und do the same for koblenz. Then the pr that is marked with both labels will only appear for nuernberg.

So yes we tag every single pr with nuernberg that affects nuernberg where we think nuernberg should be informed about this change because sth. important might break

seluianova: > So yes we tag every single pr with nuernberg that affects nuernberg where we think nuernberg should be informed about this change because sth. important might break

So, we are making Nürnberg oriented release notes here.

The downside of this approach would be that if we later receive the same request from Koblenz, we would have to rework the approach again.

seluianova:
:thought_balloon: we may also need to agree on what we consider as not-testable maintenance.
for example, 1799 is technically testable for us because we have to test that card creation isn't broken.
but for end users it's rather a maintanance and "nothing changed", I think

f1sh1918:
> 💭 we may also need to agree on what we consider as not-testable / maintenance. for example, 1799 is technically testable for us (i.e. for our internal testing, e.g. by Martin) because we have to test that card creation isn't broken. but for end users it's rather a maintanance, because "nothing changed", I think
> 
> same for translations...
> 
> what should be the difference between enhancement and maintanance? visible VS not visible to users, or?

Maintenance was introduced by @steffenkleinle 
I really don't mind about the wording.
Only tickets that are not testable by martin should be marked as `not-testable`, f.e. CI, backend only issue (should be covered by tests, i don't see the point that martin should trigger queries and mutation manually, tbh., issue that will only be shown in logs and not to the end user, unit tests)

f1sh1918:
I just talked to @connium and we decided to only separate testable prs from non-testable for now, since the changelog structure will not stay clean.
So @Docmale you still have to identify within the testable issues, the ones that relate to nuernberg.

PR #2021, opened 2025-03-25
1802: Fix more actions button not properly displayed

f1sh1918:
### Short Description

It looks like the available height of the scroll view of the carousel wasn't calculated correctly.
So if a card extension is available for koblenz, there wasn't enough space to display the more actions button

### Proposed Changes

- Instead of calculating the available height by using BottomNavigationBar (height varies between ios and android) i added a LayoutBuilder that provides a `maxHeight` constraint.
- added some missing changes from the flutter update @michael-markl . They were created automatically for me

### Side Effects

- none

### Testing

- Check the steps in the issue description
- The Button should be still visible

### Resolved Issues

Fixes: #1802


seluianova:
Found a small issue on iOS:
when the button is pressed, it is still cropped
image

seluianova (commented):
Found a small issue on iOS:
the button looks cropped, when it is pressed

f1sh1918:
> Found a small issue on iOS: the button looks cropped, when it is pressed 

I would need some more info about what simulator and which ios version you used @seluianova

seluianova:
> Even how did you get this purple background?

it's a pressed button state. let me check the versions..

f1sh1918:
okay thx for the find @seluianova 
I realized that i was assuming the wrong height of the indicator.
I just checked it in the inspector and adjusted it to 28.

image

bahaaTuffaha (approved):
Looks good ✅ tested on Android with 5.0' screen and 6.71' screen emulator.
Why the card says its about to expire even if I changed the `endDate` to a future date?

f1sh1918:
> Looks good ✅ tested on Android with 5.0' screen and 6.71' screen emulator.
> Why the card says its about to expire even if I changed the `endDate` to a future date?

Because the date of the created card won't change. 
You can just create a new card with the extended end date. 
Thx for testing. I still need an ios Review :)

steffenkleinle (approved):
Tested on iOS simulator, works as expected 🎉

PR #2023, opened 2025-03-26
Use local app-toolbelt installation

steffenkleinle:
### Short Description

Use local app-toolbelt installation instead of installing it globally.
I frequently run into troubles because a path is not set, app-toolbelt is not detected or the wrong node version is used. By adding app-toolbelt locally, we can get rid of one step in t

### Proposed Changes

- Add app-toolbelt to package.json
- Use app-toolbelt from local node_modules instead of global installation

### Side Effects

1. An `npm install` is now needed in the CI to use the app-toolbelt
2. In order to install a newer app-toolbelt version (e.g. switch to `v0.3`), you need to either
  - `npm install app-toolbelt --force --no-save`
  - delete node_modules and run `npm install` again OR
  - remove `app-toolbelt` from package.json, run `npm install`, readd it with the correct version and run `npm install` again

However, this is only a local problem as in the CI dependencies are installed from scratch if the package-lock.json changes. Given how often we need to change something at the app-toolbelt, I think that is okay (and similar for installing app-toolbelt globally).

### Testing

Ensure that building the flutter app (on android) still works.

### Resolved Issues

Fixes: N/A

### DIFF
--- a/.circleci/src/commands/install_node_modules.yml
+++ b/.circleci/src/commands/install_node_modules.yml
@@ -6,7 +6,7 @@ steps:
       - v1-node-modules
   - run:
       name: Install node dependencies for npm workspaces
-      command: npm ci --ignore-scripts
+      command: npm ci
   - save_cache:
       key: v1-node-modules-{{ checksum "~/project/package-lock.json" }}
       paths:
--- a/.circleci/src/commands/install_node_modules.yml
+++ b/.circleci/src/commands/install_node_modules.yml
@@ -2,12 +2,13 @@ description: Restores and saves the node_modules directories of the npm workspac
 steps:
   - restore_cache:
       keys:
-      - v1-node-modules-{{checksum "~/project/package-lock.json" }}
+      - v1-node-modules-{{ checksum "~/project/package-lock.json" }}
+      - v1-node-modules
   - run:
       name: Install node dependencies for npm workspaces
-      command: "[ ! -d node_modules ] && npm ci --ignore-scripts --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache."
+      command: npm ci --ignore-scripts
   - save_cache:
-      key: v1-node-modules-{{checksum "~/project/package-lock.json" }}
+      key: v1-node-modules-{{ checksum "~/project/package-lock.json" }}
       paths:
         - ~/project/node_modules
         - ~/project/administration/node_modules
--- a/frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
+++ b/frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
@@ -43,7 +43,7 @@ private val json = Json { ignoreUnknownKeys = true }
 
 fun readBuildConfig(buildConfigName: String): BuildConfig {
     val jsonString = CommandLine.execute(
-        listOf("npx", "app-toolbelt", "v0", "build-config", "to-json", buildConfigName, "android"),
+        listOf("npx", "--no", "app-toolbelt", "v0", "build-config", "to-json", buildConfigName, "android"),
         File(System.getProperty("user.dir"))
     )
     return json.decodeFromString(jsonString)
@@ -53,7 +53,7 @@ data class ResValue(val type: String, val name: String, val value: String)
 
 fun readResValues(buildConfigName: String): List<ResValue> {
     val resString = CommandLine.execute(
-        listOf("npx", "app-toolbelt", "v0", "build-config", "to-properties", buildConfigName, "android"),
+        listOf("npx", "--no", "app-toolbelt", "v0", "build-config", "to-properties", buildConfigName, "android"),
         File(System.getProperty("user.dir"))
     )
     val props = Properties()
--- a/frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
+++ b/frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
@@ -43,7 +43,7 @@ private val json = Json { ignoreUnknownKeys = true }
 
 fun readBuildConfig(buildConfigName: String): BuildConfig {
     val jsonString = CommandLine.execute(
-        listOf("app-toolbelt", "v0", "build-config", "to-json", buildConfigName, "android"),
+        listOf("npx", "app-toolbelt", "v0", "build-config", "to-json", buildConfigName, "android"),
         File(System.getProperty("user.dir"))
     )
     return json.decodeFromString(jsonString)
@@ -53,7 +53,7 @@ data class ResValue(val type: String, val name: String, val value: String)
 
 fun readResValues(buildConfigName: String): List<ResValue> {
     val resString = CommandLine.execute(
-        listOf("app-toolbelt", "v0", "build-config", "to-properties", buildConfigName, "android"),
+        listOf("npx", "app-toolbelt", "v0", "build-config", "to-properties", buildConfigName, "android"),
         File(System.getProperty("user.dir"))
     )
     val props = Properties()
--- a/frontend/pubs/df_build_config/lib/builder.dart
+++ b/frontend/pubs/df_build_config/lib/builder.dart
@@ -78,8 +78,8 @@ class BuildConfigBuilder extends Builder {
 
   @override
   FutureOr<void> build(BuildStep buildStep) async {
-    final arguments = <String>["v0", "build-config", "to-json", name, "common"];
-    final process = await Process.run("app-toolbelt", arguments, runInShell: true);
+    final arguments = ["app-toolbelt", "v0", "build-config", "to-json", name, "common"];
+    final process = await Process.run("npx", arguments, runInShell: true);
     final exitCode = process.exitCode;
     final stdErr = process.stderr.toString();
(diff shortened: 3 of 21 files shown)


frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
```diff
@@ -43,7 +43,7 @@ private val json = Json { ignoreUnknownKeys = true }
 
 fun readBuildConfig(buildConfigName: String): BuildConfig {
     val jsonString = CommandLine.execute(
-        listOf("app-toolbelt", "v0", "build-config", "to-json", buildConfigName, "android"),
+        listOf("../../node_modules/app-toolbelt/bin/app-toolbelt", "v0", "build-config", "to-json", buildConfigName, "android"),
```
michael-markl, line 46: I think it would be better to use `npx --no app-toolbelt ...`  instead of manually selecting the path.

michael-markl (commented):
I believe iOS also needs to be adjusted (e.g. https://github.com/digitalfabrik/entitlementcard/blob/main/frontend/ios/Runner.xcodeproj/xcshareddata/xcschemes/bayern.xcscheme#L13)

michael-markl:
The only drawback that comes to my mind, are conflicting dependencies between our package.json and the app-toolbelt's dependencies. But I feel like, in this case, app-toolbelt was already behaving weirdly (sometimes confusing the typescript configs and similar weird stuff)^^
What are other drawbacks?

Another benefit is that all devs use a synchronized version of app-toolbelt for the project (assuming they npm install somewhat frequently - but some IDEs even notify if the dependencies are out of sync with package.lock).

steffenkleinle:
> The only drawback that comes to my mind, are conflicting dependencies between our package.json and the app-toolbelt's dependencies. But I feel like, in this case, app-toolbelt was already behaving weirdly (sometimes confusing the typescript configs and similar weird stuff)^^ What are other drawbacks?

Since its a direct install, npm does not pickup changes to the app-toolbelt dependency (i.e. changing `v0.2` to `v0.3` followed by a `npm install` does not update the dependency).
For example running `npm install app-toolbelt --force --no-save` seems to be needed.

michael-markl (commented):
> Furthermore, this makes it necessary to run npm install in the CI every time we want to use the app-toolbelt, which will take longer than just installing it globally.

That's a valid point 🤔 However, I think it's possible to only install the dependencies of the root package.json (using `npm ci --workspaces=false`)

steffenkleinle:
With the exception of building administration works like a charm, thanks @michael-markl for the help and really helpful tips!
For that I have to investigate a little more, no clue at the moment why it fails. But my best guess is conflicting typescript versions again (introduced by an outdated version in the now local app-toolbelt installation). Have to investigate more next week. I have seen the error during my CI refactoring as well, but I can't remember how I fixed it anymore :D
Interestingly enough, the error stems from https://github.com/digitalfabrik/entitlementcard/blob/f24ae8481b3000abe60d713ec0d527dd63b92fb2/frontend/build-configs/index.ts#L4.

https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/7751/workflows/6ba03887-4cab-409c-9712-bc752ceecd07/jobs/52913

frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
steffenkleinle: I was about to ask why it was necessary :D That makes a lot of sense, thanks :)

michael-markl: That would work too. I think you can even do
```json
  "app-toolbelt": "app-toolbelt"
```

michael-markl: But you would have to pass the arguments after a double dash:
```
npm run app-toolbelt -- v0 [...]
```

michael-markl:
It's due to the skipping of the install command if the cache is populated:
```
[ ! -d node_modules ] && npm ci --ignore-scripts --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache.
```
The build-configs npm package has a `prepare` script which compiles the TS to JS and stores it in the build-configs/dist folder. This folder is not restored from the cache, and the prepare script is not run as `npm ci` does not get executed (with the `--ignore-script` it would also not be executed).

I am not sure however, why this problem did not occur earlier 🤔 

I think it is [recommended](https://circleci.com/developer/orbs/orb/circleci/node#commands-install-packages) to rather cache the ~/.npm folder and just always run `npm ci` .

frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
steffenkleinle: Now I run into problems because the working directory is now the project root. Any tips to circumvent that? :D

steffenkleinle:
> I think It's due to the skipping of the install command if the cache is populated:
> 
> ```
> [ ! -d node_modules ] && npm ci --ignore-scripts --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache.
> ```
> 
> The build-configs npm package has a `prepare` script which compiles the TS to JS and stores it in the build-configs/dist folder. This folder is not restored from the cache, and the prepare script is not run as `npm ci` does not get executed (with the `--ignore-script` it would also not be executed).
> 
> I am not sure however, why this problem did not occur earlier 🤔
> 
> I think it is [recommended](https://circleci.com/developer/orbs/orb/circleci/node#commands-install-packages) to rather cache the ~/.npm folder and just always run `npm ci` .

Thanks, that did the trick! :rocket:

frontend/android/buildSrc/src/main/kotlin/BuildConfig.kt
michael-markl: I guess, that's better, then. 👍🏽

.circleci/src/commands/install_node_modules.yml
```diff
@@ -2,12 +2,13 @@ description: Restores and saves the node_modules directories of the npm workspac
 steps:
   - restore_cache:
       keys:
-      - v1-node-modules-{{checksum "~/project/package-lock.json" }}
+      - v1-node-modules-{{ checksum "~/project/package-lock.json" }}
+      - v1-node-modules
   - run:
       name: Install node dependencies for npm workspaces
-      command: "[ ! -d node_modules ] && npm ci --ignore-scripts --loglevel warn --yes || echo package.json and package-lock.json unchanged. Using cache."
+      command: npm ci --ignore-scripts
```
michael-markl, line 9: I still don't get why we use `--ignore-scripts`. Do you know why? 🤔

steffenkleinle: Otherwise it tries to run the following scripts automatically as they are configured as postinstall which fails if the current working directory is NOT `administration`:
``` json
    "generate-graphql": "graphql-codegen --config graphql-codegen.yml",
    "generate-protobuf": "protoc -I ../specs --es_out=src/generated --es_opt=target=ts --plugin=protoc-gen-es=../node_modules/.bin/protoc-gen-es ../specs/card.proto",
    "postinstall": "npm run generate-protobuf && npm run generate-graphql",
```
See https://github.com/digitalfabrik/entitlementcard/blob/main/administration/package.json#L121
I tried removing it [here](https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/7759/workflows/72a6204a-4ea1-4228-8d59-644a29c14d08/jobs/52952).

michael-markl: I don't think the problem there is that `administration` is not the working directory, but rather that the directory "administration/src/generated" does not exist when running "generate-protobuf", and protoc does not seem to automatically create the directory.
This could be fixed by adjusting the run config of  "generate-protobuf" to 
```json
    "generate-protobuf": "mkdir -p src/generated && protoc -I ../specs --es_out=src/generated --es_opt=target=ts --plugin=protoc-gen-es=../node_modules/.bin/protoc-gen-es ../specs/card.proto",
```

What I still don't get is: Why does the `prepare` script in "frontend/build-configs" get executed if the `--ignore-scripts` flag is used 🤔

andrew8er (approved):
LGTM.

f1sh1918 (approved):
Looks good to me.

One thing i recognized is that there is still an old flutter version in the pubspec-lock that should be updated (to `flutter: ">=3.24.5"`) @steffenkleinle

frontend/pubs/df_build_config/lib/builder.dart
```diff
@@ -78,8 +78,8 @@ class BuildConfigBuilder extends Builder {
 
   @override
   FutureOr<void> build(BuildStep buildStep) async {
-    final arguments = <String>["v0", "build-config", "to-json", name, "common"];
-    final process = await Process.run("app-toolbelt", arguments, runInShell: true);
+    final arguments = ["--no", "app-toolbelt", "v0", "build-config", "to-json", name, "common"];
+    final process = await Process.run("npx", arguments, runInShell: true);
```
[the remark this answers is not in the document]
steffenkleinle: Good question, seems fine without.

Issue #1994, opened 2025-03-18
CI should fail if Podfiles.lock is out of date


michael-markl:
### Describe the Problem

When updating dependencies, often times the Podfile needs to be updated as well.

Currently, the CI (commit_main) silently updates the `Podfiles.lock` and thus does not throw an error, if it needs to be updated.

### Describe Your Preferred Solution

Throw an error.

It would be great, if in this error case the CI would also log the new `Podfiles.lock` file such that it can be updated by devs that do not own a mac os device.

### Additional Information

I forgot to ask someone to update the podfiles here (#1956) resulting in an out-of-date Podfiles.lock on main, fixed by  #1993 .

PR #2034, opened 2025-04-01
1994: Use frozen lockfiles in CI

steffenkleinle:
### Short Description

Use frozen lockfiles in the CI for all installations.

### Proposed Changes

- Add `--deployment` for cocoapod installation
- Use `frozen` flag for ruby gems/fastlane installation
- Update cocoapod game to match Podfile.lock
- Remove unused gems in `frontend`

### Side Effects

N/A

### Testing

https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/7812/workflows/05963512-4841-4265-a461-8562d7128e50

### Resolved Issues

Fixes: #1994


andrew8er (approved):

f1sh1918:
@steffenkleinle can you explain a little bit more why this fixes the issue. It was mentioned that an error should be thrown if the podfile.lock is outdated. How do these ruby gems and cocoapods flags solve this issue

.circleci/src/commands/install_pods.yml
```diff
@@ -4,11 +4,12 @@ steps:
       name: Restore pods
       keys:
         - v1-pods-{{ checksum "ios/Podfile.lock" }}
+        - v1-pods
   - run:
       name: Install pods
       command: |
         bundle exec pod update
```
michael-markl, line 11: I feel like, we should remove the `pod update`.
If I understand correctly, `pod update` updates all dependencies and modifies `Podfile.lock` (I might be wrong, cannot test it :) )
Alternatively, we should abort if after `pod update`, the `Podfile.lock` changed (as compared to the git version), but I feel like this would be quite flaky, as new versions of pods may be released at any time (I suppose).
Probably, dependabot should take care of updating these dependencies "manually" (ie. in separate PRs).

f1sh1918: The announcement was just in january. I will create a ticket to not forget that but i think we may start in the beginning of next year. not sure if all packages will support that by default but this can be discussed in the particular issue

f1sh1918 (approved):

steffenkleinle:
> > @steffenkleinle can you explain a little bit more why this fixes the issue. It was mentioned that an error should be thrown if the podfile.lock is outdated. How do these ruby gems and cocoapods flags solve this issue
> 
> I don't get the question? That is exactly what the purpose of the flags is? If the corresponding lockfile is outdated, both `bundle install` and `pod install`/`bundle exec pod install` will fail now therefore failing the CI.
> 
> Edit: Oh, I see. Are you referring to the fact that both these steps are not executed in the regular `commit` workflow? I can add it to `check_frontend.yml`.

@f1sh1918 I chose to revert this as it would require additional runtime and is not that easy to implement in our current steps. If pod files are outdated, `commit_main` will fail and we'll get a notification in mattermost. Just opening a new PR to update the Podfile.lock seems like a reasonable workaround to me for now. Does that work for you?

.circleci/src/commands/install_pods.yml
steffenkleinle: Not sure if its necessary to create an issue for this now, I would guess that this will be addressed in a future flutter upgrade anyway?

Issue #1757, opened 2024-11-06
Add customizable text module to the confirmation email


dkehne:
**Is your feature request related to a problem? Please describe.**
Applicants quickly become impatient if no progress is made on their application. In some cities and districts, there is only one person who processes the applications. However, this person is sometimes sick or on vacation. Especially then, a notice should be sent to all applications received from then on.

**Describe the solution you'd like**
In the confirmations of receipt sent to applicants after the application has been sent, a different absence message for each region should be inserted as a text module in a position to be defined. 
The text message is maintained in the region settings for each region. If the input field is not filled in, the confirmation of receipt remains as before. If a text is inserted, e.g. “There is currently a processing time of 4-6 weeks due to the Christmas period”, this text is automatically added to all outgoing confirmation emails from this region.

Update:
- It may make more sense to have a textfield where the user can type the absence message and a checkbox that can be enabled and disabled. Often the same or similar message will be reused

Acceptance Criteria:
- create a section in the region settings with a textfield for the absence message and a checkbox that indicates whether its enabled or disabled
- update the application confirmation mail for the applicant and set this as a last sentence

f1sh1918:
@connium since some designs are not ready or fully reviewed we could also continue with that issue that may not require a design.

We should just define if this will be an absence notice or a general note that will be added after the mail text

connium:
I agree that we cannot tackle tasks for which there is no finished+reviewed design. So if you want to continue with tasks that are part of the Bavaria 2025 package, this is the right task.

Unfortunately I don't understand your proposed solution. I suggest that we have a short call about this.

f1sh1918:
There is no clear description how it should look like.
Since we do not support styling yet, the question is if we provide a fixed "headline" and then include the text part that the region provides

I think if you just add a normal text i may be overseen if there is no highlighting @connium

connium:
I checked how other services handle such messages, and this is how I would write it:

> Sehr geehrte/r {Vorname Nachname},
>
> wir haben Ihren Antrag auf die Bayerische Ehrenamtskarte erhalten.
>
> Den aktuellen Status Ihres Antrags können sie jederzeit unter folgendem Link einsehen. Dort haben Sie auch die Möglichkeit, Ihren Antrag bei Bedarf zurückzuziehen:
> {Link}
>
> Bitte beachten Sie, dass Antworten auf diese E-Mail werden nicht gelesen werden, da es sich um eine automatische Bestätigung handelt.
>
[...]

PR #2036, opened 2025-04-02
1757: Customizable text confirmation mail

f1sh1918:
### Short Description

As a region manager i want to add a note to the application confirmation mail to be able to add additional information like absence message or expected processing time.

### Proposed Changes

- extend region scheme by application confirmation note information
- add mutation to update application confirmation note
- add notes section to region settings
- add application confirmation note text area
- make character counter reusable
- fix z-index issue for appToaster when its used upon a modal
- add application confirmation note to mail templates
- handle new lines in mail text strings

### Side Effects

- likely none

### Testing
Migrate DB (dev only)
1. Login as region admin and go to region settings
2. Open "Eingangsbestätigung Antrag"
3. Add a text and save
4. Create an application for the region and check that your text will be included in the mail

Note: 
- Please feel free to add suggestion for better labels and texts
- I think we have to add tests for the region package in general so i didn't add them here, since its already a huge pr and would like to implement backend tests in a separate issue
### Resolved Issues

Fixes: #1757


backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -46,6 +46,14 @@ object Mailer {
         }
     }
 
+    private fun EmailBody.applicationConfirmationNoteParagraph(note: String?) {
+        if (!note.isNullOrEmpty()) {
+            p {
+                +"$note"
```
f1sh1918, line 52: 1. any smart idea how to render `\n` within a string as line break without splitting the string and rendering a `br()` instead of `\n`

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/EmailBody.kt
```diff
@@ -55,12 +55,25 @@ interface InlineRenderable {
 }
 
 private val multipleWhiteSpaces = Regex("\\s+")
+private val newLines = Regex("\\n+")
+
+private fun handleNewLines(text: String, children: ArrayList<InlineRenderable>) {
```
f1sh1918, line 60: ❓ anyone a better idea how to handle new lines `\n` in strings

michael-markl (commented):
The best approach would probably be to have a rich text editor in the administration that generates basic html, that we only need to sanitize in the backend. Then it would also be possible to embed links etc. But I think for the purposes of the issue, it's fine to keep it simple.

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/EmailBody.kt
```diff
@@ -55,12 +55,25 @@ interface InlineRenderable {
...
 class Paragraph {
     private val children: ArrayList<InlineRenderable> = ArrayList(0)
 
     operator fun String.unaryPlus() {
-        children.add(Text(this))
+        if (this.contains(newLine)) {
+            handleNewLines(this, children)
+        } else {
+            children.add(Text(this))
+        }
```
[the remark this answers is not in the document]
f1sh1918: yes i first also thought about to implement it like this. But i dunno how to pass text from a function within the email body. I can just add br() p and links and then everything is handle inside it @michael-markl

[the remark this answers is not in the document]
f1sh1918: i now implemented a plain function inside the paragraph that can be invoked. I also improved the new line handling with a proper regex to be able to display multiple new lines. Maybe this is good enough? :)

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/EmailBody.kt
```diff
@@ -63,6 +67,17 @@ class Paragraph {
...
 
+    fun plain(plainText: String) {
+        val sentences = plainText.split(keepNewLine)
+        sentences.forEach {
+            if (it.matches(newLine)) {
+                children.add(LineBreak())
+            } else {
+                children.add(Text(it))
+            }
+        }
```
michael-markl, line 78: I think, working with these complicated regexes is not necessary 🤔 

```suggestion
        plainText.trim().split("\n").forEachIndexed { index, text ->
            if (index != 0)
                children.add(LineBreak())
            children.add(Text(text))
        }
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -46,6 +46,14 @@ object Mailer {
         }
     }
 
+    private fun EmailBody.applicationConfirmationNoteParagraph(note: String?) {
+        if (!note.isNullOrEmpty()) {
```
michael-markl, line 50:
```suggestion
        if (!note.isNullOrBlank()) {
```

bahaaTuffaha:
I added a note then created a card but I can't find the note in the email.

f1sh1918:
> 
> 
> 
> - I added a note then created a card but I can't find the note in the email.
> Edit : It looks like newly created applications that have this email note included.
> 
> - And how can I migrate back the backend to the version that used at main ?

Delete the last entry in migration table. 
Delete the columns that were created by the migration
Restart backend

bahaaTuffaha (approved):
Nice work 👍️
As I mentioned previously that newly created applications will have this email note included... if thats expected then we are good to go ✅

administration/src/util/translations/de.json
```diff
@@ -335,7 +335,12 @@
     "freinetAgencyId": "AgenturId",
     "freinetAgencyAccessKey": "Zugangsschlüssel",
     "freinetActivateDataTransferCheckbox": "Personen- und Kartendaten an Freinet übermitteln",
-    "freinetActivateDataTransferSuccessful": "Ihre Einstellungen zur Datenübermittlung wurden erfolgreich gespeichert."
+    "freinetActivateDataTransferSuccessful": "Ihre Einstellungen zur Datenübermittlung wurden erfolgreich gespeichert.",
+    "applicationConfirmationMailNoteHeadline": "Eingangsbestätigung Antrag",
+    "applicationConfirmationMailNoteDialogPlaceholder": "Fügen Sie hier Ihre Notiz ein...",
+    "applicationConfirmationMailNoteExplanation": "Hier können Sie eine allgemeine Notiz hinterlegen, welche dem Antragssteller in der Eingangsbestätigungsmail des Antrags zusätzlich angezeigt wird. Sie können die Notiz jederzeit aktivieren oder deaktivieren.",
+    "applicationConfirmationMailNoteActivated": "Notiz bei Antragsbestätigung an Antragssteller:in senden",
```
seluianova, line 342:
```suggestion
    "applicationConfirmationMailNoteActivated": "Notiz aktivieren",
```
maybe this would be more straightforward.
otherwise I'm not sure, maybe it's not clear - if I don't check the box, does that mean the note won't be sent to Antragssteller or not sent at all?

administration/src/bp-modules/regions/application-confirmation-note/ApplicationConfirmationNote.tsx
```diff
@@ -0,0 +1,61 @@
...
+type ApplicationConfirmationNoteDialogProps = {
+  defaultConfirmationNote?: string | null
+  defaultConfirmationNoteActivated: boolean
+  isOpen: boolean
+  isSavingApplicationConfirmationNote: boolean
+  setIsOpen: (value: boolean) => void
+  saveApplicationConfirmationNote: (text: string, activated: boolean) => void
+}
+
+const MAX_NOTE_CHARS = 1000
```
seluianova, line 21: 🙃 as I understood it should be something really short, like 1-2 sentences, so 500 might be enough? 
(not so important actually)

seluianova (commented):
❓Note font size is larger than normal text on other pages. Is it intended?
image

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -32,6 +33,13 @@ class EakApplicationMutationService {
         val dataFetcherResultBuilder = DataFetcherResult.newResult<Boolean>()
 
         applicationHandler.validateRegion()
+        val region = transaction {
+            RegionsRepository.findRegionById(regionId)
+        }
+        val applicationConfirmationNote = takeIf {
+            region.applicationConfirmationMailNoteActivated &&
+                !region.applicationConfirmationMailNote.isNullOrEmpty()
+        }.let { region.applicationConfirmationMailNote }
```
seluianova, line 42:
```suggestion
        val applicationConfirmationNote = region.applicationConfirmationMailNote
            ?.takeIf { region.applicationConfirmationMailNoteActivated && it.isNotEmpty() }
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/EmailBody.kt
```diff
@@ -63,6 +63,15 @@ class Paragraph {
         children.add(Text(this))
     }
 
+    fun plain(plainText: String) {
+        plainText.trim().split("\n").forEachIndexed { index, text ->
```
seluianova, line 67:
```suggestion
        plainText.trim().lines().forEachIndexed { index, text ->
```

seluianova (approved):
I like the implementation! Just a few nit-picks and one improvement idea, feel free to ignore.
We could also disable the "Speichern" button if nothing has changed.

administration/src/util/translations/de.json
f1sh1918: this note is only for the "Antragssteller" so if not activated it won't be attached to the mail.

f1sh1918: i think your suggestion should be good

administration/src/bp-modules/regions/application-confirmation-note/ApplicationConfirmationNote.tsx
f1sh1918: hm there was no clear requirement. i think it does not harm to have 1000

administration/src/util/translations/de.json
seluianova: I understand that, I mean the current label might be ambitious (maybe just a personal feeling, up to you)

f1sh1918:
> ❓🙃 Note font size is larger than normal text on other pages. Is it intended? image

i removed the `large` flag

Issue #1901, opened 2025-02-13
Create demo region for freinet


f1sh1918:
**Is your feature request related to a problem? Please describe.**
As a developer i need a demo region to test the applicant and card data transfer to freinet

**Describe the solution you'd like**
- add a demo region with the `regionIdentifier` '00000' (staging and local env only) for bavaria. You can check `config.server.host` or create a config variable `environment: "Staging/Production/Development" and check for that
- this region will be mapped with the Demomandant 906 in freinet according to the `regionIdentifier`
- you can find the credentials for the demo mandant in passbolt (for later testing)

PR #2040, opened 2025-04-07
1901: Create demo region for Freinet

seluianova:
### Short Description
Create demo region for freinet (staging and local env only)

### Proposed Changes
Add demo region 00000 to the setup

### Side Effects
no?

### Testing

Dev only:

1. _Set freinet accessToken in the local config_ (if not set yet)
2. Run backend
3. Check regions and freinetagencies database tables --> new region 00000 should be created and mapped with the freinet agency 906

**Staging**:

1. Create region admin for `Stadt Freinet Demo`
2. Login with that region admin and check in `Region Settings` if Agenturname `EAK-Demo` and AgenturId `906`

### Resolved Issues
Fixes: #1901


seluianova:
@f1sh1918 I changed the config property "production" to "environment".
Corresponding salt PR:
https://git.tuerantuer.org/DF/salt/pulls/343
I guess, it should be merged just before release? Since it's not backward compatible.
If we merge it too early, there might a problem if we need to restart the server, right? (not sure how this works)

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
```diff
@@ -26,7 +27,8 @@ class Entry : CliktCommand() {
     private val config by option().file(canBeDir = false, mustBeReadable = true)
 
     // Options to overwrite single properties of the config
-    private val production by option().choice("true", "false").convert { it.toBoolean() }
+    private val environment by option().choice("dev", "staging", "prod")
```
f1sh1918, line 30: 🔧 i would prefer `development` and `production` , we also use this terms in web frontend

f1sh1918 (approved):
Looks good to me. 👍 

Indeed this pr is testable on staging. Please remove the flag and add testing instructions:
1. Create region admin for `Stadt Freinet Demo`
2. Login with that region admin and check in `Region Settings` if Agenturname `Eak- Demo` and AgenturId `906`

steffenkleinle (approved):
Nice :+1:

backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
seluianova: Okay I will update that. Just thought it might be more handy for terminal commands

seluianova:
> to avoid deploying this in sync i would recommend to keep the production variable for some time in the config (salt) and add a comment to remove it after we updated both environments

@f1sh1918 I've updated the PR https://git.tuerantuer.org/DF/salt/pulls/343: it now only adds a new environment property.
And I created a follow up PR to remove the production property after release: https://git.tuerantuer.org/DF/salt/pulls/344

f1sh1918:
Just rechecked your last commit. Still everything fine from my side!

connium:
🧪 Tested
- **Browser/Device**: Firefox 140.0, 137.0
- ✅ Create region administrator account for `Stadt Freinet Demo`
- ✅ Login as region administrator for `Stadt Freinet Demo`
- ✅ Check Freinet agency (agency name is `EAK-Demo`, agency ID is `906`)

Issue #1689, opened 2024-10-16
Add linting to `/frontend/build-configs`


ztefanie:
fronend/build-configs files should also be linted and checked by prettier

PR #2041, opened 2025-04-07
1689: Add linting to frontend build config

andrew8er:
### Short Description

Add linting to frontend build config.

### Proposed Changes

- Remove the `/frontend/build-configs/.prettierrc` file in favor of the workspace version.
- Move `/administration/eslint.config.mjs` to workspace root and use it as a base config.
- Add subproject specific `eslint.config.mjs` files to `/administration` and `/frontend/build-configs`.
- Format `/frontend/build-configs`.
- Add CircleCI config.
- Update some dependencies to fix weird errors around package management and imports.

I had to add some dependency overrides, since our currently installed `eslint-config-airbnb` package is nominally [not compatible](https://github.com/airbnb/javascript/issues/2961) with the currently used `eslint`. How this ever installed cleanly is beyond me. Hope this gets fixed soon though.

### Side Effects

None

### Testing

N/A

### Resolved Issues

Fixes: #1689


.circleci/src/jobs/check_frontend_build_configs.yml
```diff
@@ -0,0 +1,10 @@
+docker:
+  - image: cimg/node:20.18.0
+resource_class: small
+working_directory: ~/project/frontend/build-configs
+steps:
+  - prepare_workspace
+  - install_node_modules
+  - run:
+      name: Lint build config
+      command: npm run lint
```
steffenkleinle, line 10: 🔧 Is it actually necessary to create an own job for this? Can we just move the check (L8 - L10) to `check_frontend`?

Also, without adding this to a workflow, this is not run (moving to `check_frontend` would fix this).

steffenkleinle (commented):
Thanks!

andrew8er:
@f1sh1918 
>I just wonder if we may add the formatting command of frontend/buildConfig in a runConfig not sure if we need a separate runConfig for that but add it to Format administration and change the naming

I added two more commits to address this.

.circleci/src/jobs/check_frontend_build_configs.yml
steffenkleinle: I get that. The thing is just that it costs us runtime and credits to start yet another job in a separate image. I'd therefore prepare to have it in an existing job.

f1sh1918 (approved):

.circleci/src/jobs/check_frontend_build_configs.yml
andrew8er: I tried to merge both jobs, but I'd need to switch the `working_directory` between steps to run `install_node_modules`. Can I just `cd` in and out with two extra steps or can I somehow run a named step in a subshell?.

Do we pay by the job or by the runtime used to run a job? I think there are some potential improvements we could make to speed up tests.

steffenkleinle: Why do you need to switch the working directory to install node? We use npm workspaces so running `npm install` in any directory should install node modules for the root, administration and frontend/build-configs :thinking:

steffenkleinle: We pay by the runtime, therefore optimizing/speeding up the tests would definitely be nice. Could you create an issue with your ideas please?

andrew8er: > Why do you need to switch the working directory to install node?

When running `npm ci` in `/administration`, npm also runs the postinstall script which needs `protoc` installed. So we also need to install this… so including it in this job will make it more complicated and the runtime benefit gets smaller. 

It would be preferable, if we can either ditch `protoc` entirely and use an alternative npm package (such as [buf](https://github.com/bufbuild/buf), since we are already using [@bufbuild/protobuf](https://github.com/bufbuild/protobuf-es)) or use an npm package that pulls in `protoc` (like [this](https://www.npmjs.com/package/@protobuf-ts/protoc), this way we can avoid an extra step.

> We pay by the runtime, therefore optimizing/speeding up the tests would definitely be nice. Could you create an issue with your ideas please?

I'll add an issue tomorrow 👌

f1sh1918:
please add `not-testable`  label

.circleci/src/jobs/check_frontend_build_configs.yml
andrew8er: 👉 #2050

steffenkleinle: I just tested this in #2091 and it works perfectly well in one job. I chose to put it in `check_administration` now, even though it obviously does not really fit there. `check_frontend` would also be possible but then executing `install_node_modules` would be necessary first. I'd think of merging the check jobs (perhaps without `check_backend`) anyway though in a separate PR.

At least I would really prefer not to create another job just for linting frontend build configs.

steffenkleinle (approved):
LGTM

Issue #2002, opened 2025-03-20
Data could not be parsed with missing golden card option


bahaaTuffaha:
### Describe the Bug

At Bayerns form when you select the golden card and uncheck the checkbox on digital and physical options it will let you continue until the final stage where it will say `Daten konnten nicht geparsed werden.
`  without indication where is the missing field/box.

### Steps to Reproduce

1. Go to `http://localhost:3000/beantragen`
2. Fill the form normally but select the golden card and remove the checkbox on digital and physical options.
3. At last stage (4) it will show `Daten konnten nicht geparsed werden.`

### Expected Behavior

It least it shouldn't let you continue without one of the options is selected.

### Additional Information

![Image](https://github.com/user-attachments/assets/82f2d8e3-b6df-45d9-813f-39c93d57a8e3)

![Image](https://github.com/user-attachments/assets/5dd784c5-f119-4024-899c-172ddda3d221)

PR #2043, opened 2025-04-07
2002: Golden Card validation

bahaaTuffaha:
### Short Description

At Bayerns form when you select the golden card and uncheck the checkbox on digital and physical options it will let you continue until the final stage where it will say Daten konnten nicht geparsed werden.  without indication where is the missing field/box.

### Proposed Changes

- Just putting the partialValidationResult checking above the return for the golden card
`if (partialValidationResult.value.cardType !== BavariaCardType.Blue)`

### Side Effects

none.

### Testing

- Go to http://localhost:3000/beantragen
- Fill the form normally but select the golden card and remove the checkbox on digital and physical options.
- Test also the blue card's digital and physical options.

### Resolved Issues

Fixes: #2002

### DIFF
--- a/administration/src/mui-modules/application/forms/StepCardTypeForm.tsx
+++ b/administration/src/mui-modules/application/forms/StepCardTypeForm.tsx
@@ -72,12 +72,12 @@ const StepCardTypeForm: Form<State, ValidatedInput> = {
     if (partialValidationResult.type === 'error') {
       return { type: 'error' }
     }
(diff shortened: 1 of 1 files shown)


f1sh1918 (commented):
thx for the fix!
LGTM 👍

f1sh1918 (approved):

seluianova (approved):
Perfekt! Thx for the fix :)

Issue #2052, opened 2025-04-10
Translation overrides do not work on ios


f1sh1918:
### Describe the Bug

We are facing some issues on translation overrides on ios. So during initialization for all appLocales provided in the config the particular overrides.json of the project related translations should be applied.
On ios we face the following error and the fallback translation (bavaria) will be used

```js
Uncaught Platform Error
                       Null check operator used on a null value
                       #0      LocaleSettingsExt.overrideTranslations (package:slang/src/api/singleton.dart:504:51)
                       #1      LocaleSettings.overrideTranslations (package:ehrenamtskarte/l10n/translations.g.dart:175:137)
                       #2      overrideLocale (package:ehrenamtskarte/main.dart:74:25)
                       <asynchronous suspension>
                       #3      initializeTranslations.<anonymous closure> (package:ehrenamtskarte/main.dart:52:7)
                       <asynchronous suspension>
                       #4      Future._kTrue (dart:async/future.dart:670:3)
                       <asynchronous suspension>
                       #5      _RootZone.bindUnaryCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1633:12)
                       <asynchronous suspension>
```

### Steps to Reproduce

1. Install nuernberg sozialpass on ios (simulator with german locale)
2. Go to more -> Change the language to English and go to identify and check the label text f.e. intro slides and identity page
3. See wrong label texts of bavarian eak, since fallback was used. Also find an error in the dev console

### Expected Behavior

The language overrides work without issues and the texts of the overrides will be shown in the app

### Related Issues

https://github.com/slang-i18n/slang/issues/294

seluianova:
Addition to testing:
The behavior is different when the system language is German and when it is English.
So would be nice to test both.
When it's English - the English translation doesn't work on the intro slides, the rest of the pages look fine.
When it's German - the English translations throughout the app dont work.

PR #2053, opened 2025-04-10
2052: Workaround translation overrides ios

f1sh1918:
### Short Description

There is an issue that not all locales will be loaded initially with the current `slang` plugin version
https://github.com/slang-i18n/slang/issues/294
I added the workaround that they provided there

### Proposed Changes

- add load function to to add all locales on app start, since they were not properly initialized
- add comment to remove this workaround when new slang version was released (#2061)
- adjust runConfigs to use the flavor lowercase in the particular `buildFlavor`part. That seem to fix the issues tory mentioned

### Side Effects

- likely none

### Testing

See the issue instructions

### Resolved Issues

Fixes: #2052

### DIFF
--- a/.idea/runConfigurations/Run__env_local_buildConfig_bayern_.xml
+++ b/.idea/runConfigurations/Run__env_local_buildConfig_bayern_.xml
@@ -1,7 +1,8 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Run (env:local+buildConfig:bayern)" type="FlutterRunConfigurationType" factoryName="Flutter">
-    <option name="additionalArgs" value="--dart-define=environment=local --flavor Bayern" />
+    <option name="additionalArgs" value="--dart-define=environment=local" />
     <option name="attachArgs" value="--dart-define=environment=local" />
+    <option name="buildFlavor" value="bayern" />
     <option name="filePath" value="$PROJECT_DIR$/frontend/lib/main.dart" />
     <method v="2">
       <option name="RunConfigurationTask" enabled="true" run_configuration_name="Set build config bayern" run_configuration_type="ShConfigurationType" />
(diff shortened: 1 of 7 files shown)


f1sh1918:
Hm looks like it should be fixed with `slang 4.6.1`

andrew8er (approved):

seluianova (commented):
When the system language is German, switching the language to English in the app works now ✅
But when the system language is English, English translation still doesn't work on the intro slides (but it works on other pages, which is weird!) ❌
image
I also see a lot of errors like this in the log:
```
Failed to index parameter type: <WFHomeServicePickerParameter: 0x6000029f46e0> in is.workflow.actions.gethomeaccessorystate: (extension in ActionKit):__C.WFParameter.ToolKitConversionError.incompatibleParameterType(WFHomeServicePickerParameter) for locale: en
Failed to index parameter type: <WFSystemShortcutPickerParameter: 0x6000029f4b40> in com.apple.shortcuts.RunShortcutIntent: (extension in ActionKit):__C.WFParameter.ToolKitConversionError.incompatibleParameterType(WFSystemShortcutPickerParameter) for locale: en
```

Once the user closes the intro slides and restarts the application, there are no more errors and everything works smoothly.
So the problem seems to lie somewhere in the intro slides 🤔

seluianova:
I actually this we could treat this as two different issues.
One for the German language system and one for English.
Since the current PR fixes the first issue, I think we could merge this PR and open another issue for the English mode.
@f1sh1918 what do you think?

seluianova (approved):
Partially fixes the issue 👍

f1sh1918:
> I actually think we could treat this as two different issues.
> One for the German system language and one for English.
> Since the current PR fixes the first issue, I think we could merge this PR and open another issue for the English mode.
> @f1sh1918 what do you think?


Maybe I didnt get the whole issue. What is missing here to be fixed?

seluianova:
@f1sh1918 this:
```
But when the system language is English, English translation still doesn't work on the intro slides (but it works on other pages, which is weird!) ❌
```
Steps to reproduce:
1. Set system language to English in the emulator
2. Remove Nürnberg app (or clean the data to trigger the intro slides on start up)
3. Start the app --> intro slides will be displayed in English, but the fallback translation (Ehrenamtskarte) will be applied.

Issue #2049, opened 2025-04-09
Buttons on the scan and verification pages are partially covered by the system navigation bar


seluianova:
### Describe the Bug
"Blitz an" and "Frontkamera" buttons are a little bit covered by the system navigation bar.
I can only observe this problem on a real device. 
<img src="https://github.com/user-attachments/assets/60829a68-f115-4388-985e-f0066bdfae18" width="500"/>

### Steps to Reproduce
1. Go to 'Ausweisen' tab
2. Click on 'Karte aktivieren' or 'Gültigkeit prüfen'
3. Check the buttons 

### Expected Behavior
Buttons shouldn't be covered by the control bar.

### Related Issues
n/a

### Additional Information
Reproduced on device:
Pixel 8 Pro
Android 15

_Not sure if it's Anroid only issue, I don't have an iOS device to check._

**UPD**. same problem on this page and on the licenses page (last line of the text is covered)
<img src="https://github.com/user-attachments/assets/222b48af-9025-4f3d-ad8e-51709249f06a" width="500"/>

seluianova:
> Hm i can't reproduce this on Galaxy S10 Android 12

might be connected to screen resolution

f1sh1918:
Seem to be related to the navigation buttons

f1sh1918:
ok i checked several devices and android version and this issue seem to be introduced with android 15

<img width="552" alt="Image" src="https://github.com/user-attachments/assets/238b0ae2-0785-43a7-bca1-888849c5b88f" />

PR #2066, opened 2025-04-14
2049: Prevent screen view from overlap with system navigation

bahaaTuffaha:
### Short Description

"Blitz an" and "Frontkamera" buttons are a little bit covered by the system navigation bar.
same problem on this page and on the licenses page (last line of the text is covered)

### Proposed Changes

- Wrapped the `MaterialApp.router` with `SafeArea` at app.dart.
- I only enabled the bottom SafeArea.

Note: not tested on iOS

### Side Effects

- This will affect any view.
### Testing

- This test requires android 15 + Enabling the 3 Button Navigation in Settings -> System -> Navigation Mode
- Launch any Frontend app for ex. Bayern.
- Go to cards and add a card by QR code.
- Also check licenses page.
- check everything just in case because it wraps every view.

### Resolved Issues

Fixes: #2049


f1sh1918:
@bahaaTuffaha 
can you please add to the testing instruction that you have to enable 3 Button Navigation in Settings -> System -> Navigation Mode and its Android 15 only afaics

f1sh1918 (approved):
Thx for the quick fix.
I think there is a small issue on ios which is not super problematic


Normally the "ButtonSheet" should be white not black (area at the bottom).
I realized that this does not happen if you wrap the Scaffold (Home_Page) which is recommended in the official flutter docs:
https://docs.flutter.dev/ui/adaptive-responsive/safearea-mediaquery

But as you said the drawback is that we would have to wrap every single route and also the fullscreen modals like camera.
So i'm fine to stick to your suggestion since its more robust and less error prone

michael-markl (approved):
I think, as a hotfix, this works great!

We can think about removing this and properly handle the safe areas in the each route. It should not be too hard either.

Issue #2068, opened 2025-04-14
The app is loaded in the system language and then the language is automatically changed


seluianova:
### Describe the Bug
The problem occurs if the system language is different from the language selected by the user in the application settings. 
See the steps below for details.

### Steps to Reproduce

1. Set system language to English
2. Open Nürnberg app --> it's loaded in English
3. Change the app language to German (More --> Change language)
4. Stop the app (remove it from the background processes)
5. Lauch the app --> **first it will be loaded in English** (system language)! Then after a few seconds the application labels will be automatically changed to German (user preference).

### Expected Behavior
The application should load in the language saved by the user in the application settings

### Related Issues
n/a

### Additional Information
I can only observe this behavior in the iOS emulator.
Not sure if this is reproducible on a real device (I don't have one to check).

seluianova:
@f1sh1918 can you try to reproduce this? I think it has the same root cause as an issue with the intro slides...

seluianova:
The issue can be reproduced on the ios emulator, but cannot be reproduced on _real devices_ using browserstack.
We assume that real users won't encounter this....

PR #2070, opened 2025-04-15
2068: Fix locale loading on application startup

seluianova:
### Short Description
This PR fixes 2 issues that only show up on iOS.
First one is described in #2068.
Second issue is weird, it might depend on the environment because I can reproduce it stably on my machine, but it doesn't reproduce for @f1sh1918:
1. Set system language to English in the emulator
2. Remove Nürnberg app (or clean the user data to trigger the intro slides on start up)
3. Install / restart the app --> intro slides will be displayed in English, but in the the fallback translation (Ehrenamtskarte)
image

### Proposed Changes
- Convert App to StatefulWidget to add locale initialization to the AppState. This allows translations to be initialized before the app is initialized
- Move router creation inside AppState build function. This fixes the second issue for me. My assumption is that otherwise the IntroScreen widget is initialized before the translations are initialized. As a result I see intro slides in the fallback translation ❓ _But I'm not sure here, feel free to test and make other suggestions..._

### Side Effects
Hopefully no?
**But not tested on Android** (unfortunately I could not get the Android emulator to work on a MacBook yet)

### Testing
Follow steps from #2068

### Resolved Issues
Fixes: #2068


frontend/lib/app.dart
```diff
@@ -25,107 +26,127 @@ const homeRouteParamCardIndexName = 'cardIndex';
 const homeRouteName = '/home';
 const introRouteName = '/intro';
 
-final GoRouter router = GoRouter(
```
seluianova, line 28: moved as is into _createRouter() function

frontend/lib/app.dart
```diff
@@ -25,107 +26,127 @@ const homeRouteParamCardIndexName = 'cardIndex';
...
+            : isProduction()
+                ? buildConfig.mapStyleUrl.production
+                : isLocal()
+                    ? buildConfig.mapStyleUrl.local
+                    : buildConfig.mapStyleUrl.showcase;
 
-    // Load default language from settings
-    WidgetsBinding.instance.addPostFrameCallback((_) {
-      if (settings.language != null) {
-        LocaleSettings.setLocaleRaw(settings.language!);
```
seluianova, line 98: moved to locale_initializer so that the saved locale is initialized at application startup

f1sh1918:
@seluianova 
i tested your issue on my branch #2089 where the flavors were fixed with Browserstack
It seems like this issue does not happen on real devices with production builds

https://github.com/user-attachments/assets/7e6d2670-cb46-4b68-bce2-02267cd72844

seluianova:
okay, thanks for checking @f1sh1918 
it's so annoying to see and fix non-existing issues 😀

PR #2093, opened 2025-04-16
Add setup documentation

steffenkleinle:
### Short Description

Add more details to the setup documentation and recommend podman over docker.

### Side Effects

N/A

### Testing

N/A

### Resolved Issues

Fixes: N/A

### DIFF
--- a/docs/development-setup.md
+++ b/docs/development-setup.md
@@ -39,19 +39,39 @@
 
 ### Frontend Setup
 
-1. Install Android SDK via the [Android plugin](https://www.jetbrains.com/help/idea/create-your-first-android-application.html#754fd) or Android Studio
-2. Install [fvm](https://fvm.app/documentation/getting-started/installation) (flutter version manager)
-3. Install flutter dependencies
+1. Install [fvm](https://fvm.app/documentation/getting-started/installation) (flutter version manager)
+2. Install flutter dependencies
 ```shell
 cd frontend && fvm flutter pub get
 ```
 
-4. Open IntelliJ settings and
+3. Open IntelliJ settings and
    - Install the Android plugin and set the Android SDK path
    - Install the Dart plugin and set the Dart SDK path
    - Install the Flutter plugin and set the Flutter SDK path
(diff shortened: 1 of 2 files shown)


docs/development-setup.md
```diff
@@ -89,12 +109,11 @@ For IntelliJ, the following plugins are recomended:
...
-1. Install docker and docker-compose
-2. Setup docker-compose
+1. Install podman and podman-compose
+2. Setup podman-compose
 ```shell
-docker-compose rm
-docker-compose build
-docker-compose up --force-recreate
+podman machine init
+podman machine start && podman-compose up
```
seluianova, line 116: I would suggest to keep both options. I use docker for example. I can add a note how I made I'd work on mac

seluianova: Btw can you run backend tests with podman? we use TestContainers there, do they work fine with podman?

steffenkleinle: > yes same for me i would like to keep docker as an alternative

Hmm, I am not exactly sure why we would want to keep both. The setup docs are only necessary/intended for new installations in which case it does not matter whether you use docker or podman. Since docker is proprietary now and is more complicated to set up, I think it is valid to only recommend one solution (being podman in this case). This does not mean that you should or even have to switch away from it if your current setup works fine.

> Btw can you run backend tests with podman? we use TestContainers there, do they work fine?

How can I run those?

seluianova: Should work with a small additional setup
https://podman-desktop.io/tutorial/testcontainers-with-podman
Maybe we need to include this to the instruction, if you could test that

f1sh1918 (approved):

docs/development-setup.md
steffenkleinle: Did not have the motivation to get this up and running right now. I therefore opted to keep instructions for both docker and podman. This might be improved at a later point.

seluianova (approved):
lgtm thanks for the adjustments!

PR #2095, opened 2025-04-18
Fix querying for wrong physical store id

michael-markl:
### Short Description

We mix accepting store ids and physical store ids in the frontend.
On staging, these ids sometimes differ and thus we sometimes query physical stores using an invalid id.
Interestingly, on production, the ids always match, so we did not run into the bug yet.

### Proposed Changes

* Adjust naming of related queries and fields to explicitly specify accepting store id vs physical store id.
* Use physical store id instead of accepting store id when appropriate

### Side Effects

None

### Testing

* In the Bayern EAK staging app:
* Search for "Anna Apotheke"
* Clicking on it should successfully open the detail page of the Anna Apotheke

### DIFF
--- a/frontend/graphql_queries/stores/accepting_stores_search.graphql
+++ b/frontend/graphql_queries/stores/accepting_stores_search.graphql
@@ -4,6 +4,7 @@ query AcceptingStoresSearch($project: String!, $params: SearchParamsInput!) {
         name
         description
         physicalStore {
+            id
             address {
                 location
             }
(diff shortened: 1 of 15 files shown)


f1sh1918 (approved):
Works fine on ios.
I really faced this issue sometimes with local database also but never on production and didn't know why so great job and thx for the fix 👍

seluianova (approved):
lgtm!

PR #2098, opened 2025-04-21
2050: Speed up CI

steffenkleinle:
### Proposed Changes

-

### Side Effects

-

### Resolved Issues

Fixes: #


steffenkleinle:
> Hm I don't understand this approach with merging the checks. We may reduce the costs for resources but without parallel jobs I don't see any speedup in the check workflow @steffenkleinle 

Agreed, but there is a reason why this is a draft :D I am just trying some stuff here and I'll revert this again. No plans to make this a permanent change.

steffenkleinle:
Since we actually don't deliver/do anything with the things run on main, we might also just skip the `commit_main` altogether? Or what purpose does it serve exactly? Perhaps @f1sh1918?

f1sh1918:
We test the builds for all platforms. I would like to keep that

steffenkleinle:
> We test the builds for all platforms. I would like to keep that

Sure. How about all the `pack_xyz` commands? Does not seem like we test a lot of stuff with that/like that would easily break.

f1sh1918:
> > We test the builds for all platforms. I would like to keep that
> 
> Sure. How about all the `pack_xyz` commands? Does not seem like we test a lot of stuff with that/like that would easily break.

The packaging jobs can be removed afaics @steffenkleinle . We don't update there anything frequently so it should not likely break

andrew8er:
I think there's an error in the CircleCI config: https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/8222

f1sh1918 (approved):
LGTM

Issue #2003, opened 2025-03-21
Uncaught exception findRegionByRegionIdentifier


f1sh1918:
### Describe the Bug

We are facing some uncaught errors with empty collection for `findRegionByRegionIdentifier`

### Steps to Reproduce

has to be reproduced.

### Expected Behavior

region can be fetched without an error

### Related Issues

Might be related to #https://github.com/digitalfabrik/entitlementcard/issues/1979
And will be maybe fixed with the next release though

### Additional Information

```
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]: [JettyServerThreadPool-36332] WARN io.javalin.Javalin - Uncaught exception
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]: java.util.NoSuchElementException: Collection is empty.
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]:         at kotlin.collections.CollectionsKt___CollectionsKt.single(_Collections.kt:594)
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]:         at app.ehrenamtskarte.backend.regions.database.repos.RegionsRepository.findRegionByRegionIdentifier(RegionsRepository.kt:59)
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]:         at app.ehrenamtskarte.backend.regions.webservice.schema.RegionsQueryService$regionsByPostalCode$1.invoke(RegionsQueryService.kt:69)
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]:         at app.ehrenamtskarte.backend.regions.webservice.schema.RegionsQueryService$regionsByPostalCode$1.invoke(RegionsQueryService.kt:54)
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]:         at org.jetbrains.exposed.sql.transactions.ThreadLocalTransactionManagerKt.inTopLevelTransaction$run(ThreadLocalTransactionManager.kt:232)
Mar 21 07:54:58 entitlementcard.tuerantuer.org backend[2106365]:         at org.jetbrains.exposed.sql.transactions.ThreadLocalTransactionManagerKt.access$inTopLevelTransaction$run(ThreadLocalTransactionManager.kt:1)
[...]

PR #2101, opened 2025-04-23
2003: Fix uncaught exception regions by postal code

f1sh1918:
### Short Description

Currently a 500 error (check console) occurs if the user types a postal code in the application form that has no valid `regionIdentifier`(ars) in our `regions` table. You can find example postal codes in the issue comment.

Note: We use a csv file that consists all postalCodes and the corresponding regionIdentifiers to find the `ars` for a particular postal code, since a region can have multiple postal codes but has one single (4 digit) ars

### Proposed Changes

- adjust `findRegionByRegionIdentifier` to return `.singleOrNull)` before it was unsafely tried to get the region via id that couldn't be found in some cases and leading to internal error
- concatenate filtering regions by postalCode and finding the region by ars in the database

### Side Effects

- likely none

### Testing

- type in a postal code that has no corresponding ars in our database (check issue description)
- a)Check that the error mentioned in the issue description will not be thrown (dev only)
- b)Check that there is no 500 error in the browser console
- check that in both cases an info alert should be shown that the region couldn't be determined

### Resolved Issues

Fixes: #2003

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/regions/database/repos/RegionsRepository.kt
@@ -55,15 +55,13 @@ object RegionsRepository {
         region.activatedForCardConfirmationMail = activatedForConfirmationMail
     }
 
-    fun findRegionByRegionIdentifier(regionIdentifier: String, projectId: EntityID<Int>): RegionEntity {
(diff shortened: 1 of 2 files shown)


bahaaTuffaha (approved):
Tested with the list of postal codes that are provided in the issue and tried to enter weird compilation of letters and numbers it looks good. ✅
- The alert is working as intended.

seluianova (approved):
great investigation, changes look good!

Issue #791, opened 2023-02-07
Refactor qrcode.ts


maxammann:
**Is your feature request related to a problem? Please describe.**
The qrcode.ts currently contains a lot of code copied from [Encoder.ts](https://github.com/zxing-js/library/blob/5719e939f2fc513f71627c3f37c227e26efe06c1/src/core/qrcode/encoder/Encoder.ts#L503)

This is because some functions were private. Maybe we can simplify qrcode.ts by reusing code.

michael-markl:
I think we should add least add a disclaimer at the top of the file *now* that explains which functions are copy-pasted, which ones are adjusted, etc. Otherwise, we're building up quite some tech debt there (and noone knows how to maintain that code^^).

PR #2102, opened 2025-04-23
791: Add comments to functions

f1sh1918:
### Short Description

As a lot of functions were adapted and modified from https://github.com/zxing-js/library
To have a reference all functions, some comments should be added.

Note that the libraries is unmaintained and we should think about an alternative

### Proposed Changes

- add missing reference to functions

### Side Effects

- none

### Resolved Issues

Fixes: #791


bahaaTuffaha (approved):
Not tested

steffenkleinle (approved):

Issue #2100, opened 2025-04-23
Update backend dependencies


andrew8er:
### Describe the Problem

Backend dependencies are quiet old some have CVEs (as reported by Mend.io):

- Dependency maven:io.ktor:ktor-client-core-jvm:2.3.9 is vulnerable
  - CVE-2024-49580,  Score: 5.3
- Dependency maven:com.graphql-java:graphql-java:20.3 is vulnerable
  - CVE-2024-40094,  Score: 5.3
- Dependency maven:org.eclipse.jetty:jetty-server:11.0.23 is vulnerable
  - CVE-2024-8184,  Score: 5.9
  - CVE-2024-6763,  Score: 3.7
- Dependency maven:org.eclipse.jetty:jetty-http:11.0.23 is vulnerable
  - CVE-2024-6763,  Score: 3.7

### Describe Your Preferred Solution

Update all dependencies and Gradle plugins to their latest stable versions.

### Describe Possible Alternatives

None

### Related Issues

None

### Additional Information

None

PR #2104, opened 2025-04-23
2100: Update backend dependencies

andrew8er:
### Short Description

Updates all dependencies, build plugins and Gradle for the backend.

### Proposed Changes

- Update to Gradle 8.13.
- Update all Gradle build plugins to their latest stable versions.
- Update all dependencies to their latest stable versions.
- Remove unused dependencies `com.google.code.gson:gson` and `com.kohlschutter.junixsocket`.
- Use Gradle version catalog
- Add Gradle plugin `de.fayard.refreshVersions` to check for outdated versions.
- Remove deprecated API usage of the Gradle DSL.
- Use `assemble` instead of `build` in workflow `build_backend`, since `build` also runs tests which we already do in workflow `check_backend`.
- Add a run configuration for running backend tests.

I kept the GraphQL dependencies at version 6.10.2, since I get an error in a test, that only happens on CircleCi.

### Side Effects

None

### Testing

`./gradlew tests`

### Resolved Issues

Fixes: #2100


michael-markl (commented):
I like the changes (especially the new exposed query syntax... using `select` for a filter was definitely the wrong choice back then :D )

backend/src/main/kotlin/app/ehrenamtskarte/backend/auth/webservice/schema/types/Administrator.kt
```diff
@@ -26,6 +26,8 @@ class Administrator(
         if (regionId == null) {
             CompletableFuture.completedFuture(null)
         } else {
-            dataFetchingEnvironment.getDataLoader<Int, Region?>(REGION_LOADER_NAME).load(regionId)
+            dataFetchingEnvironment.getDataLoader<Int, Region?>(REGION_LOADER_NAME)
+                ?.load(regionId)
+                ?: CompletableFuture.completedFuture(null)
```
michael-markl, line 31:
```suggestion
            regionLoader.fromEnvironment(dataFetchingEnvironment).load(regionId)
```

andrew8er:
The tests in `Verein360ApplicationTest.kt` seem flaky. Is there any kind of limit in outbound connections for CircleCI or one on the freinet endpoint? Or is it just to slow?

andrew8er:
I added a commit to fix a test. Strangely, this test used to work with `com.expediagroup:graphql` versions below 7.0.0 but not from that version on. Took me hours to find this…

I also converted the Gradle dependency declaration to use version catalogs. This should result in simpler version specification (where different related packages use the same version) and safer dependabot PRs.

backend/gradle.properties
```diff
@@ -1,3 +1 @@
 # Properties to configure the gradle build
```
seluianova, line 1: do we need to keep this file?

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLContext.kt
```diff
@@ -16,10 +16,15 @@ data class GraphQLContext(
...
 }
+
+private const val ID = "context"
+
+fun GraphQLContext.Builder.put(context: app.ehrenamtskarte.backend.common.webservice.GraphQLContext) {
+    put(ID, context)
+}
+
+val GraphQLContext.context: app.ehrenamtskarte.backend.common.webservice.GraphQLContext
+    get() = get(ID)
```
seluianova, line 30:
```suggestion
fun GraphQLContext.Builder.put(context: app.ehrenamtskarte.backend.common.webservice.GraphQLContext) {
    put(GraphQLContext::class, context)
}

val GraphQLContext.context: app.ehrenamtskarte.backend.common.webservice.GraphQLContext
    get() = get(GraphQLContext::class)
```

:upside_down_face: type can be used as a key

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLContext.kt
```diff
@@ -16,10 +16,15 @@ data class GraphQLContext(
     val backendConfiguration: BackendConfiguration,
     val regionIdentifierByPostalCode: List<Pair<String, String>>,
     val request: HttpServletRequest,
-) : GraphQLContext {
```
seluianova, line 19: :thought_balloon: maybe we could also rename our GraphQLContext to avoid collisions with graphql.GraphQLContext

CustomGraphQLContext ?

seluianova (approved):
smoke tested, looks good to me!
just a few optional comments

backend/gradle/libs.versions.toml
```diff
@@ -0,0 +1,73 @@
...
+com-expediagroup-graphql = { id = "com.expediagroup.graphql", version.ref = "graphql" }
+com-google-protobuf = { id = "com.google.protobuf", version = "0.9.5" }
+io-gitlab-arturbosch-detekt = { id = "io.gitlab.arturbosch.detekt", version = "1.23.8" }
+org-jetbrains-kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version = "2.1.20" }
+org-jetbrains-kotlinx-kover = { id = "org.jetbrains.kotlinx.kover", version = "0.9.1" }
+org-jlleitschuh-gradle-ktlint = { id = "org.jlleitschuh.gradle.ktlint", version = "12.2.0" }
+
+[versions]
+
+# FIXME: Keeping this at 6.10.2 since all versions from 7.0.0 onwards produce an error on tests
```
seluianova, line 12: I have added a fix for that
https://github.com/digitalfabrik/entitlementcard/pull/2146

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/GraphQLContext.kt
andrew8er: I didn't want to let this PR grow even more, so I kept the name. But I think you're right, the name clashes with a library class and the `dfe.graphQlContext.context` accessor chain also does not look very good IMHO. 

Our `GraphQLContext` looks like a kitchen sink with all kinds of objects related to the request, and there is also the `fun GraphQLContext.getAdministrator()` extension function, which could be included directly in the class.
 
I would refactor this in a later PR, when I'm more familiar with the code base.

backend/gradle/libs.versions.toml
andrew8er: Thanks, I included it in this PR.

andrew8er: I cherry-picked your fix commit in this PR, thanks!

Issue #1679, opened 2024-10-14
Update GoRouter to recent


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Currently we can't update go router since we are facing issues with deeplinks.
Until version 13.2.1 a trailing slash was needed to properly resolve the activation code
With a dependency update a trailing slash is not needed but breaks the activation

Example (works until 13.2.1): 
```js
android % npx uri-scheme open berechtigungskarte://bayern.ehrenamtskarte.app/activation/code#CkkKHwoPQW5kcmVhcyBGaXNjaGVyEPGkARoICgIICSICCAASEDGjkYQBPaU2Dd6mI87eXtAaFHNsrKCgpGu20z2c06yyG7bU12ap/ --android
```
Example (works >13.2.1)
```js
android % npx uri-scheme open berechtigungskarte://bayern.ehrenamtskarte.app/activation/code#CkkKHwoPQW5kcmVhcyBGaXNjaGVyEPGkARoICgIICSICCAASEDGjkYQBPaU2Dd6mI87eXtAaFHNsrKCgpGu20z2c06yyG7bU12ap --android
```

Since we print this deeplinks on pdfs and send it via mail we can not simply adjust the deeplinks by removing the slash, since the old ones will still break

**Describe the solution you'd like**
- update go router to recent
- ensure deep links still work

**Describe alternatives you've considered**
- change the deeplinks since activation via qr code scanning still works

**Additional context**

bahaaTuffaha:
Is it just Checking hash code if it contains '/' before passing it to the Activation page?

PR #2105, opened 2025-04-23
1679: Update go router

f1sh1918:
### Short Description

The old `go_router` version didn't resolve url fragments properly without trailing slash, thats why we added some trailings slashes to the deeplinks.
This was not really a good practise since base64 strings could also include `/` 
This is quite error prone.

### Proposed Changes

- update got router to recent
- remove trailing slashes from deeplinks
- add legacy support for old deep links
- adjust tests

### Side Effects

- Likely none

### Testing

- Create cards for koblenz (/erstellen) and activate the card
- Create a card for bavaria and copy the custom scheme deeplink in the browser
- Use the command in `deeplinks.md`(devs only)
- `npx uri-scheme open https://staging.nuernberg.sozialpass.app/activation/code#<activationCode> --android`

Test legacy support:
- add to the deeplink a trailing slash using the command line (dev only)
- `npx uri-scheme open https://staging.nuernberg.sozialpass.app/activation/code#<activationCode>/ --android`

### Resolved Issues

Fixes: #1679

### DIFF
--- a/frontend/lib/activation/deeplink_activation.dart
+++ b/frontend/lib/activation/deeplink_activation.dart
@@ -202,8 +202,8 @@ class _WarningText extends StatelessWidget {
 
 DynamicActivationCode? getActivationCode(
     BuildContext context, String encodedBase64qrcode, Function(String message) updateErrorMessage) {
-  // GoRouter >= 13.x did not handle uri fragments properly without trailing slash. After the package update to recent version this was fixed.
-  // Though we need legacy support for old deep links that have a trailing slash.
+  // GoRouter 13.x did not handle uri fragments properly without trailing slash.
+  // Therefore we need legacy support for old deep links that have a trailing slash.
   final sanitizedBase64qrcode = removeTrailingSlash(encodedBase64qrcode);
   try {
     final activationCode = const ActivationCodeParser()
--- a/frontend/lib/activation/deeplink_activation.dart
+++ b/frontend/lib/activation/deeplink_activation.dart
@@ -11,6 +11,7 @@ import 'package:ehrenamtskarte/identification/util/activate_card.dart';
 import 'package:ehrenamtskarte/identification/verification_workflow/verification_qr_code_processor.dart';
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
 import 'package:ehrenamtskarte/proto/card.pb.dart';
+import 'package:ehrenamtskarte/util/string_utils.dart';
 import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
 import 'package:intl/intl.dart';
@@ -43,9 +44,9 @@ enum DeepLinkActivationStatus {
 }
 
 class DeepLinkActivation extends StatefulWidget {
-  final String base64qrcode;
+  final String encodedBase64qrcode;
 
-  const DeepLinkActivation({super.key, required this.base64qrcode});
+  const DeepLinkActivation({super.key, required this.encodedBase64qrcode});
 
   @override
   State<DeepLinkActivation> createState() => _DeepLinkActivationState();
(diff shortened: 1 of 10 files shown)


f1sh1918:
@michael-markl maybe you have some input/better idea how to resolve that without backward compatibility problems

michael-markl:
> [...] thats why we added some trailings slashes to the deeplinks.

😮 😮 😮 😮 😮 

Luckily, (most?) our base64 strings are padded with "=" (or "%3D") characters, so we can remove any slash occurring after the padding. That should likely solve the problem

f1sh1918:
Unfortunately not all (koblenz) deeplinks include paddings `=` in the base64 string
So i check the encoded deep links for trailing slash (that is not encoded) and remove them if exists.

frontend/lib/app.dart
```diff
@@ -37,7 +38,11 @@ final GoRouter router = GoRouter(
         GoRoute(
           path: '$activationRouteName/:$activationRouteCodeParamName',
           builder: (BuildContext context, GoRouterState state) {
-            return DeepLinkActivation(base64qrcode: Uri.decodeFull(state.uri.fragment));
+            // GoRouter >= 13.x did not handle uri fragments properly without trailing slash. After the package update to recent version this was fixed.
+            // Though we need legacy support for old deep links that have a trailing slash.
+            // Implemented in April 2025, can be removed around April 2027.
```
steffenkleinle, line 43: 🙃 
```suggestion
            // GoRouter 13.x did not handle uri fragments properly without trailing slash.
            // Therefore we need legacy support for old deep links that have a trailing slash.
            // Implemented in April 2025, can be removed around April 2027.
```

steffenkleinle: Just for clarification: Why does a trailing `/` "break the activation" now? Is it interpreted as part of the base64 encoded code?

f1sh1918: yes thats the point. before the go router update it wasn't interpreted but needed for whatever reason and now its interpreted as part of the base64 string

f1sh1918:
> I guess, the alternative is not to change the format of the paths, and instead just always remove the trailing slash. I think we have to support the old format (that includes the trailing slash) indefinitely, as some cards never expire?

Ah sure i forgot the golden cards. Thx for the hint. Ehm yes that would be also possible. I even think this is quite "dirty" so i wanted to avoid it

frontend/lib/app.dart
steffenkleinle: Perhaps it makes more sense then to strip the slash where the code is interpreted?

f1sh1918: Hm but we should strip it before decodingUri.
Why do you think that?

steffenkleinle: Since in general trailing slashes are not a problem, even with the new go router upgrade and could perhaps exist for different reasons than just deeplinks. By moving the stripping to where the code is interpreted (before decodeUri) we put the solution/workaround more closely to the problem and avoid problems possibly occurring somewhere in between. Furthermore, this reduces the risk that in the future its not clear anymore why this is needed/just removed imo.

f1sh1918: Ok so you think i should just send the `state.uri.fragment` without removing the slash and decoding it to the deep link activation and handle it there?

steffenkleinle: Exactly 👍

steffenkleinle (approved):
Tested on android

frontend/lib/activation/deeplink_activation.dart
```diff
@@ -200,9 +201,13 @@ class _WarningText extends StatelessWidget {
 }
 
 DynamicActivationCode? getActivationCode(
-    BuildContext context, String base64qrcode, Function(String message) updateErrorMessage) {
+    BuildContext context, String encodedBase64qrcode, Function(String message) updateErrorMessage) {
+  // GoRouter >= 13.x did not handle uri fragments properly without trailing slash. After the package update to recent version this was fixed.
+  // Though we need legacy support for old deep links that have a trailing slash.
+  final sanitizedBase64qrcode = removeTrailingSlash(encodedBase64qrcode);
```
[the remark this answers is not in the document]
f1sh1918: i use `QrCode` now

Issue #1999, opened 2025-03-19
Nürnberg-Pass-ID edge case


bahaaTuffaha:
### Describe the Bug

When creating a card at Nürnberg and entered 9999999999 for `Nürnberg-Pass-ID`  it will show (Ein Fehler ist aufgetreten) toast and the sentry catched it.

### Steps to Reproduce

1. Go to Nürnberg (region-admin)
2. Create card.
3. At Nürnberg-Pass-ID enter 9999999999.

### Expected Behavior

It should just show error message without closing the form.

### Additional Information

Also take care about dealing with uin64 in javascript, you may use bigint or string

![Image](https://github.com/user-attachments/assets/390fc9d7-4108-4856-9aa7-363c4ebd47c8)

f1sh1918:
@dkehne
can you ask nuernberg how big the numbers will grow?

dkehne:
@f1sh1918 can you tell me what the highest number would be that is supported by uint32?

f1sh1918:
4,294,967,295

f1sh1918:
any updates on this @dkehne?

connium:
I just wrote to Nürnberg and informed them, that we'd like to limit the number. I'll comment here, as soon as I receive an answer.

connium:
@f1sh1918 Nürnberg finally answered and they confirmed, that the number will be sufficient. At the moment the have 6-digit IDs.

f1sh1918:
Ok i would say we can limit it to 9 characters. That would be the easiest solution. This limit will never be reached @connium

PR #2109, opened 2025-04-28
1999: Limit the Nürnberg Pass-Id to 9 characters

seluianova:
### Short Description
In the card creation form, it was possible to enter Nürnberg-Pass-ID = 9999999999, but an error occurred when generating the card.
image

### Proposed Changes
Set the max length of the of nürnberg pass id to 9

### Side Effects
no

### Testing
1. Login to Nürnberg administration as region admin
2. Go to "Karten erstellen" --> "Einzelne Karten erstellen"
3. Check that it is not possible to enter Nürnberg-Pass-ID = 9999999999 (10 chars). Only 9 characters are allowed.
Additionally can be checked that a card with Nürnberg-Pass-ID = 999999999 (9 chars) can be generated without errors.

### Resolved Issues
Fixes: #1999


steffenkleinle (approved):
Not tested

f1sh1918 (approved):
LGTM.
Tested single card creation and bulk card creation

Issue #2094, opened 2025-04-17
Handle safe areas in the each route for android 15


bahaaTuffaha:
### Describe the Problem

We did a hotfix but this could be handled properly by adding safe areas in the each route.
"This is android 15 issue"

### Reproduction

One example where the current setup fails (as described first by @Docmale in #2313 ):

1. On android 15 goto Ehrenamtskarte - Über - Lizenz
2. Change to landscape (system bar left / right) 
3. view text near system bar

![Image](https://github.com/user-attachments/assets/08763523-0613-4b0c-ad98-5bdaaaaa0ecc)

### Expected Behavior

 text near system bar should not be cut off

### Describe Your Preferred Solution

- Current solution: The MaterialApp.router is wrapped with SafeArea at app.dart.
- Add safe areas in each route.

### Related Issues

Related to #2066

f1sh1918:
We should maybe do a research if this is really needed  @michael-markl
In the docs was just mentioned that the body of a scaffold is a good start.
Even the bottom bar color in ios changed after our change but i feel it could be more error prone if we forget to add it in every route and modal

f1sh1918:
hey @michael-markl,

i just checked some possible solutions here.
The biggest issue at least according to the mentioned issue in the screenshot is, that there is no easy way to detect if we have `Orientation.landscapeLeft` or `Orientation.landscapeRight` to set the particular safe area

I've even not sure if this issue is big enough to put a lot of additional logic into it, since its also not really possible to detect if this "Bottom Navigation Buttons" are used or the Gesture Navigation.
Afaics this issue is android only so we could at least only use the safe area wrapper for android to let it look better on ios.

I mean to avoid this issue we could also allow only portrait mode in the app at all.
I don't see any usage for landscape mode tbh.

PR #2127, opened 2025-04-29
2094: Improve system overlay and top unsafe area handling

michael-markl:
### Short Description

The system ui overlay was not properly configured:
 * on some screens in dark mode (e.g. Ausweisen, About), the system status bar was dark on a dark background (=> illegible)
 * on some screens (e.g. Search Page, Herausgeber, etc.), the system status bar was dark, but the title on our app bar was white (=> visually ugly, but all was legible)

Usually, the system ui overlay style is handled through the use of an AppBar.
I added a minimal appbar for status bar protection (using a Gradient) for the Ausweisen and About pages (the protections are especially relevant for landscape mode, where the content exceeds the viewport and can be scrolled).
For the maps page, we likely don't want to further protect the status bar (Google Maps does not do that either; it would look a bit weird), so we can use an AnnotatedRegion to set the system ui overlay style.

### Testing

* Check that system status bar (at the top) is legible on all pages/routes (also check landscape mode, dark mode, etc.)
* Also, especially check on iOS as I cannot test it there.

### Related Issues

Related: #2094

### DIFF
--- a/frontend/lib/about/about_page.dart
+++ b/frontend/lib/about/about_page.dart
@@ -14,6 +14,8 @@ import 'package:url_launcher/url_launcher_string.dart';
 
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
 
+import 'package:ehrenamtskarte/widgets/app_bars.dart';
+
 const String accessibilityPolicyUrl = 'https://berechtigungskarte.app/barrierefreiheit/';
 
 class AboutPage extends StatefulWidget {
@@ -41,7 +43,7 @@ class AboutPageState extends State<AboutPage> {
         final packageInfo = snapshot.data;
         if (snapshot.hasData && packageInfo != null) {
           children = [
-            Container(height: 20),
+            SizedBox(height: 20),
             Center(
               child: Padding(
                 padding: const EdgeInsets.all(10),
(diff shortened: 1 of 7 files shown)


bahaaTuffaha:
System status bar looks fine on all screens but I found these while using landscape (nav buttons covering content) Android with 3-button nav and 6.71' pixel 7 pro :

michael-markl:
Correct. This PR only resolves issues around the system status bar and/or top unsafe area. Otherwise, the PR would probably get too large.

bahaaTuffaha (approved):
Looks good ✅ just noticed this on android Api 35:
- The system nav buttons are in white after installing the app.

seluianova (approved):
Tested iOS in the simulator, looks good.
For Android, I'm a little concerned that the introductory slides have a white bottom navigation bar but all other pages have a black bottom navigation bar.
I would expect to see it white in the light mode and black in the dark mode.
However, it probably doesn't matter.
image
image

michael-markl:
>  For Android, I'm a little concerned that the introductory slides have a white bottom navigation bar but all other pages have a black bottom navigation bar.

This is likely a drawback of the current workaround of just adding a SafeArea (for bottom) around the app. I'll try to address this in an upcoming PR.

Issue #415, opened 2021-11-17
Map touch is not working after rotating device


maxammann:
**Describe the bug**
The map becomes partly unresponsive after rotating the device (at least on android).

**To Reproduce**
1. Go to Rotate device to landscape
2. Cold start app 
3. The map view is not responding on the bottom side.

**Expected behavior**
The app should be pinchable everywhere.

**Desktop (please complete the following information):**
 - OS: Android Emulator and real device

PR #2128, opened 2025-04-29
Fix usages of FutureBuilder

michael-markl:
### Short Description

Fix usages of FutureBuilder and remove the ScreenParentResizer workaround.

### Side Effects

I noticed, that we (probably mostly because of me :D ) used the FutureBuilder incorrectly.
We should supply it with a single future, that does not change on every call of build.

I realized that this also fixes problems with Flutter hot reloading, as the SettingsProvider was rebuilt, a new future was created, and the whole widget subtree was remounted.

This also revealed a bug involving screen_parent_resizer: The map always overflowed, without contracting itself again. We introduced screen_parent_resizer due to a bug in maplibre, namely that half of the screen was unresponsive after an orientation change (see #415). I could not reproduce this bug anymore, so it is likely safe to just remove the workaround altogether.

In fact, the parent resizer is already broken in the current production builds (on Android): If you change device orientation, the map will overflow without respecting the device's viewport which results in 
1. the map not being centered correctly
2. the licenses button (the small (i) at the bottom left) is not off the view.

### Testing

* Check that (on Android), an orientation change does not result in the earlier bug #415. Also, the map should be centered correctly, after an orientation change.


bahaaTuffaha (approved):
The map stays centered after changing orientation  (Android 34 and 35) ✅

PR #2147, opened 2025-05-05
2100: Update com.expediagroup.graphql version to 8.7.0

seluianova:
### Short Description
Follow up for https://github.com/digitalfabrik/entitlementcard/pull/2104

There was an issue with the com.expediagroup.graphql upgrade - one of the tests always failed in CI due to SocketTimeoutException.

I believe this update affected the initialization of the graphql endpoint, and the graphql endpoint was not yet initialized when the first test was run. As a result, the first graphql test failed.

### Proposed Changes
I changed the initialization of GraphQLHandler in the tests.
The GraphQLHandler object is now initialized before the graphql endpoint is configured.
This solved the issue in tests.

### Side Effects
Tests take more time in CI now! 😕
Appx 1 minute longer (3min 30sec VS 2min 30sec)

### Testing
Dev only: some graphql requests can be checked (smoke testing)

### Resolved Issues
n/a


andrew8er:
I picked your fix directly into my PR #2104 as f9fb73db739d3204570acd666e5e2f3fa65a037d so we can close this PR.

PR #2156, opened 2025-05-06
2268: Optimize backend tests performance

seluianova:
### Short Description
Our backend tests became slower after some dependency updates.
Suggested changes allow to speed up tests by half.

Main:
Image
Current branch:
image

### Proposed Changes
- Use singleton database container. This saves time because the container is not restarted and migrations are not reapplied between tests
- Move graphql handler creation for tests to the static context

### Side Effects
no

### Testing
Dev only:
Tests can be run locally using "Test backend" run configuration

### Resolved Issues
Fixes: #2268


seluianova:
I guess these 11 seconds exclude @BeforeAll hook where the main problem was (container setup etc).
For me the difference is more visible locally. 
After clean it took 6 minutes on main VS 2 mins on the current branch...
image

seluianova:
I also moved the graphql handler creation to a static context. This saved some more time.
image

Issue #2148, opened 2025-05-05
Uncaught runtime error when you log out


bahaaTuffaha:
### Describe the Bug

I can't log in and uncaught runtime error when you log out.

### Steps to Reproduce

1. Go to http://localhost:3000/
2. Fill credentials.
3. Click on the `Anmelden` button (nothing happens).
4. Click on Bayern button it will log you in.
5. Log out then click on try again button.
6. Uncaught runtime error will appear.

```
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'HTMLButtonElement'
    |     property '__reactFiber$6p9kja69brk' -> object with constructor 'FiberNode'
    --- property 'stateNode' closes the circle
    at JSON.stringify (<anonymous>)
    at canonicalStringify (canonicalStringify.ts:27:1)
    at QueryManager.getObservableFromLink (QueryManager.ts:1108:1)
    at QueryManager.getResultsFromLink (QueryManager.ts:1165:1)
    at resultsFromLink (QueryManager.ts:1597:1)
    at QueryManager.fetchQueryByPolicy (QueryManager.ts:1662:1)
    at fromVariables (QueryManager.ts:1275:1)
    at QueryManager.fetchConcastWithInfo (QueryManager.ts:1337:1)
    at ObservableQuery.fetch (ObservableQuery.ts:809:1)
    at ObservableQuery.reobserveAsConcast (ObservableQuery.ts:971:1)
```

### Expected Behavior

- It should log in as soon as I click on Anmelden (while selecting Bayern)
- We should handle this error.

### Additional Information

https://github.com/user-attachments/assets/2ffe1b5b-847a-467b-b6df-0cdfb388acea

f1sh1918:
Hm this a super weird issue. Not sure how it came up. But the login and logout do not redirect properly. We should fix that before a new release @bahaaTuffaha

f1sh1918:
this is the issue that causes the error
https://github.com/digitalfabrik/entitlementcard/pull/2118/files

PR #2160, opened 2025-05-06
Downgrade react-router-dom

f1sh1918:
### Short Description

Apparently the update to `react-router-dom 6.30.0` breaks the redirects for login and logout.
This is described here #2148

Note this is only a workaround to fix the issue. I also created an issue for updating react router to v7 (#2161)

### Proposed Changes

- downgrade react router dom

### Side Effects

- none

### Testing

1. Remove `node_modules` from administration and root folder(dev only)
2. run `npm i` and start administration (dev only)
3. Login should redirect to `/` and logout to the login form , check that this does not happen #2148

### Resolved Issues

I will keep #2148 open because its just a workaround.

### DIFF
--- a/administration/package.json
+++ b/administration/package.json
@@ -2,7 +2,8 @@
   "name": "administration",
   "private": true,
   "dependenciesComments": {
-    "typescript": "Keeping on 5.0.x because of current incompatibility of 5.1.x with @typescript-eslint"
+    "typescript": "Keeping on 5.0.x because of current incompatibility of 5.1.x with @typescript-eslint",
+    "react-router-dom": "Keep 6.26.2 since 6.30.0 breaks redirect for login and logout, update to v7 and fix the issue in #2161"
   },
   "dependencies": {
     "@apollo/client": "^3.7.16",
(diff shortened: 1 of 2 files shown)


bahaaTuffaha (approved):
Works as expected ✅
Nice we found this before a release 🙂

andrew8er (approved):
I also noticed this problem, hopefully this is resolved soon.

steffenkleinle (commented):
Just wondering if its could/should be tested this with a unit test?

f1sh1918:
> ~Just wondering if its could/should be tested this with a unit test?~ Edit: Nvm, does not make sense. This is not an issue on our side.

it was tested with e2e and failed @steffenkleinle but its a bit more difficult to include e2e in the ci, since we need a running backend

connium:
🧪 Tested
- **Browser/Device**: Firefox 140.0, Chromium 137.0
- ✅ Log in via https://staging.bayern.ehrenamtskarte.app works. Immediate logout leads to login page.
- ✅ Log in via https://staging.bayern.ehrenamtskarte.app/ works. Immediate logout leads to login page.
- ✅ Log in via https://staging.bayern.ehrenamtskarte.app/irgendwas works. Immediate logout leads to login page ("/").
  :warning: However, after login the URL still says https://staging.bayern.ehrenamtskarte.app/irgendwas . If I log in via https://staging.bayern.ehrenamtskarte.app/applications, I am redirected to the list of applications.

PR #2162, opened 2025-05-07
Docs backend first steps

f1sh1918:
### Short Description

To get a better onboarding for newcomers, i provided some first steps for backend development.

### Proposed Changes

- added some documentation about how to create a simple graphql query and use it in the frontend

### Side Effects

- none

### Testing

N/A


seluianova (approved):
I agree regarding the name of the document.
Looks good anyway

Issue #2165, opened 2025-05-08
Add ability to load test data into DB


andrew8er:
### Describe the Problem

While working on various parts of the project, I often need to have some kind of test data. We have no way to load test data yet.

### Describe Your Preferred Solution

Add a backend command to load test data

### Describe Possible Alternatives

None

### Related Issues

None

### Additional Information

None

PR #2166, opened 2025-05-08
2165: Load test data

andrew8er:
### Short Description

This PR enables developers to load test data into the DB.

#### Main changes:

- Move discovery of config file into `EntryPoint.kt` and make it log the chosen file.
- Add command `db-clear` (and IntelliJ run config) to clear the DB content.
- Add command `db-import-dev` (and IntelliJ run config) to load developer data.
- Rename existing IntelliJ run configs to have a "DB" prefix and move to "Backend commands" folder, so they visually group together.
- Add a small set of test data in the form of SQL files in `/backend/src/dev/sql`.

#### Smaller changes:

- Add an explicit name to the DB volume in `docker-compose.yaml`.
- `BackendConfiguration.toImportConfig()` is now an extension function to aid decoupling.
- `Importer.findProject()` is now an immutable property `project`, since it never changes.
- `Database.executeScript()` is now named `Database.executeSqlResource()` to better reflect what it does.
- Class `Database` is now an object, since it only contains utility functions.
- Various smaller refactorings
- Slightly out of scope of this issue but fixes an annoying problem: Exclude all files in `/.idea` from version control and whitelist some, that are useful across developers and do not contain user/machine specific paths. Note that this does not affect files already under version control.

### Side Effects

None

### Testing

In `/backend`: `/gradlew test`

### Resolved Issues

Fixes: #2165


bahaaTuffaha (changes requested):
Thank you for this great idea it will save time .
 `DB import dev data` :
```
Caused by: org.postgresql.util.PSQLException: ERROR: relation "public.applications" does not exist
```
so I migrated the db then this error occurred after importing dev data:
```
Caused by: org.postgresql.util.PSQLException: ERROR: insert or update on table "applications" violates foreign key constraint "fk_applications_regionid__id"
```
`/.gradlew test` is SUCCESSFUL

andrew8er:
- Which DB version are you running on?
- Is there any data previously stored (apart from the data that is loaded through `import`)?
- You could try (in this order): `clean`, `migrate`, `import`, `db-import-dev`

bahaaTuffaha:
> * Which DB version are you running on?
> * Is there any data previously stored (apart from the data that is loaded through `import`)?
> * You could try (in this order): `clean`, `migrate`, `import`, `db-import-dev`

PostgreSQL (ver. 17.4)
I did compose down then up I so no data.. even cleaned the db and migrated.

bahaaTuffaha (approved):
LGTM.. Thank you ✅

andrew8er:
> LGTM.. Thank you ✅

No worry. I'm thinking of adding another Gradle task that runs all those in sequence, so we can easily get a clean and developer-ready DB. What do you think?

andrew8er:
I added a command `db-recreate` to completely wipe and rebuild the DB to the latest version an load sample data.

steffenkleinle:
> ./gradlew migrate
./gradlew import

I guess these should be
- `./gradlew db-migrate`
- `./gradlew db-import-online`
right?

docs/development-setup.md
```diff
@@ -196,22 +190,31 @@ projects:
...
-```shell
-docker exec -i <container_id> psql ehrenamtskarte postgres < dump-$(date +%F).sql
-```
+### DB tasks
+
+- Clear the DB's contents: `./gradlew db-clear`
+- Run all migrations: `./gradlew migrate`
+- Import data from online stores: `./gradlew import`
+- Load developer sample data: `./gradlew db-import-dev`
+- Create a clean DB with sample data (all of the above steps): `./gradlew db-recreate`
```
steffenkleinle, line 205: Could you replace steps 4 and 5 above with this command please? I think this is the way better way to set everything up now.

steffenkleinle (approved):
Very nice, love it! Works flawlessly, I adjusted the commands in the PR description.

Issue #963, opened 2023-05-08
Professionalize Emails


sarahsporck:
**Is your feature request related to a problem? Please describe.**
We get reports that the mails look phishy. Therefore we should use some mail testing services to identify the pain points to get some proper suggestions what to improve. Additionally we should check the particular wordings of each mail.

**Describe the solution you'd like**
- This is more a research ticket to identify issue.
- Collect here the issues we have and create separate issues for solving them

**Additional context**
We should do some research
There are some mail testing services we could test to identify issues
f.e. https://www.mail-tester.com/

michael-markl:
I believe "Emails are looking kind of unprofessional" and "Emails often land in spam." might be two different issues.
I would think that adding an icon does not decrease spam false positives.

sarahsporck:
Hmm you might be right. I still unprofessional and spam are related. Although for now I'd priorize mails not landing in spam. 

We could check out https://www.mail-tester.com/ and https://svn.apache.org/repos/asf/spamassassin to see how unbiased spam filters might classify our e-mails. And adjust wording, structure annd more based on that.

sarahsporck:
See #1002 .

f1sh1918:
In the confirmation mail we use "Antrags" it should be replaced by "Antrages"

```js
Sie wurden gebeten, die Angaben eines Antrags auf eine Ehrenamtskarte zu bestätigen. ... die Angaben eines Antrages ... 
```

michael-markl:
According to the Duden, both "Antrags" and "Antrages" are valid for Genitiv Singular :D :
https://www.duden.de/deklination/substantive/Antrag

f1sh1918:
if there are any issues the customer will reach us with that

f1sh1918:
i reopened this ticket, since we got reports from customer that the application mails look phishy.

f1sh1918:
@connium i think from a technical side there is not much to improve, since we have a top score in mail-checker.com after adding a DKIM signature and pass the DMARC test. 
I think the only improvement would be the texts. Not sure who would be a good person to improve wordings etc.

PR #2170, opened 2025-05-12
963: Adjust email templates

seluianova:
### Short Description
Adjust email templates according to the task #963

### Proposed Changes
- adjust email templates

### Side Effects
no

### Testing
Check emails:
- confirmation email for the applicant after the application has been submitted
- email to the contact person - request to approve/reject the application
- confirmation email for the application when the card has been issued (should be enabled in the region settings: Regionsspezifische Einstellungen --> Nach der Erstellung einer Karte verschickt das System eine E-Mail-Bestätigung an den Antragsstellenden mit einem Link zur Vorab-Aktivierung)

### Resolved Issues
Fixes: #963


backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -310,7 +318,7 @@ object Mailer {
                 )
             }
             p { +"Dieser Link ist 24 Stunden gültig." }
-            finalInformationParagraph(projectConfig)
+            finalInformationParagraph()
```
michael-markl, line 321: For this mail and for the mail below, we have to use a more generic way, as these are also necessary for Nürnberg & Koblenz.

connium: I'm pretty sure we don't send any emails for the KoblenzPass. The entitled person doesn't have to provide an email address and there is also none contained in the provided data.
For Nürnberg we don't manage the applications at all, so there shouldn't be any email communication as well.

seluianova: ah, indeed! 🤦‍♀️
thanks @michael-markl 
@connium is it fine if we keep it as Ehrenamtskarte-Bayern-Verwaltung for Bayern?

f1sh1918: we could also introduce another string variable for mail templates for each project

f1sh1918 (commented):
Looks good to me except of the sender name in the `finalInformationParagraph` which we may set as a variable in the backend config

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
seluianova: @connium any suggestions for signature for Koblenz and Nürnberg?
Ihr Team von SocialPass Nürnberg/Koblenz ?

connium: Ihr Team des Nürnberg-Passes/KoblenzPasses

PR #2173, opened 2025-05-12
2038: Fix PermissionRequestInProgressException

seluianova:
### Short Description
Fix exception from sentry:
```
PermissionRequestInProgressException: A request for location permissions is already running, please wait for it to complete before doing another request.
  File "geolocator_apple.dart", line 63, in GeolocatorApple.requestPermission
  File "determine_position.dart", line 129, in checkAndRequestLocationPermission
  File "intro_screen.dart", line 21, in _IntroScreenState._onDonePress
```
This happens when the user clicks "Weiter" on the last intro screen multiple times.

### Proposed Changes
- add _isProcessing variable to the _IntroScreenState
- check it before executing _onDonePress() and return if the button has already been clicked

### Side Effects
no

### Testing
Dev only:
- perform fresh app installation (intro slides should be displayed)
- open the app, go to the last intro slide
- click "Weiter" or "Next" button multiple times and very fast
- see the exception in the log

### Resolved Issues
Fixes: #2038

### DIFF
--- a/frontend/lib/intro_slides/intro_screen.dart
+++ b/frontend/lib/intro_slides/intro_screen.dart
@@ -17,11 +17,22 @@ class IntroScreen extends StatefulWidget {
 }
 
 class _IntroScreenState extends State<IntroScreen> {
+  bool _isProcessing = false;
+
   Future<void> _onDonePress(SettingsModel settings) async {
-    await checkAndRequestLocationPermission(context, requestIfNotGranted: true);
-    settings.setFirstStart(enabled: false);
-    if (!mounted) return;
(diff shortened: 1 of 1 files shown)


bahaaTuffaha (approved):
Tested only on Android ✅ 
I noticed the following sentence gone after this fix:
```
The grantResults array is empty. This can happen when the user cancels the permission request
```

f1sh1918 (approved):
Thx for investigation and fixing it 👍 
I hoped to find sth easier like a Debounce Button/Gesture widget or sth.
But we may create something reusable in the future if we need, since we may face similar issue somewhere else

Tested on ios

PR #2183, opened 2025-05-14
Fix postgis image on ARM macs

steffenkleinle:
### Short Description

The recent update of the postgis image (#2151) breaks starting the database docker compose service on ARM macs:
```
Resolving "postgis/postgis" using unqualified-search registries (/etc/containers/registries.conf.d/999-podman-machine.conf)
Trying to pull docker.io/postgis/postgis:17-3.5-alpine...
Error: choosing an image from manifest list docker://postgis/postgis:17-3.5-alpine: no image found in image index for architecture "arm64", variant "v8", OS "linux"
```

### Proposed Changes

- Use `linux/amd64` instead

### Side Effects

N/A

### Testing

Check that starting docker-compose still works (or works again on macs).

### Resolved Issues

Fixes: N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,7 @@ services:
       - network
   db-postgis:
     image:  postgis/postgis:17-3.5-alpine
+    platform: linux/amd64
     command: [ "postgres", "-c", "log_statement=all", "-c", "log_destination=stderr" ]
     environment:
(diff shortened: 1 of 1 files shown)


seluianova (approved):
thanks! I saw this error, but didn't figure out how to fix it in the config for everyone (fixed  locally)

andrew8er (approved):

Issue #2141, opened 2025-05-05
Check if person exists in Freinet when a card was created


connium:
### Describe the Problem

As a region administrator or a region manager,
I want to add personal information for volunteers to Freinet-Online if they are not known yet
so that I can manage volunteers and issued cards there.

### Describe Your Preferred Solution

Given that I am logged in as a region administrator or a region manager
and data transfer to Freinet-Online is enabled,
when I approve a pending application,
then it is checked whether the person is already present in Freinet-Online by sending the applicants first name, family name and birth date.

Given that I am logged in as a region administrator or a region manager
and data transfer to Freinet-Online is enabled,
when I approve a pending application
and there are multiple persons with the same first name and family name and birth date in Freinet-Online,
then I am shown a message, that we have found multiple candidates and the update must be performed manually
and no data is sent to Freinet-Online.

Given that I am logged in as a region administrator or a region manager
and data transfer to Freinet-Online is disabled,
when I approve a pending application,
then no communication with Freinet-Online takes place.

### Acceptance Criteria
- create a new mutation f.e. `sendApplicationAndCardDataToFreinet` that receives the `applicationId, cardId` from the frontend.
- ensure that the user is logged in and has `RegionManager` or `RegionAdmin` role
- within this mutation a get request to the freinet api should be trigger using `vorname`, `nachname` and `geburtsdatum` (see examples)
- before sending the call check if the region is activated for data transfer to freinet and an `accessKey` and `agencyId` is provided
- the call will receive a list of users or emptyList
- a) If there is more than one user found, send an error to the user that data can not be send to freinet because the person can't be identified explicitly
- b) No user is found, #2142 will be triggered when implemented
- c) One user is found, #2143 will be triggered when implemented
- trigger the mutation after a card was successfully created
- ensure that the mutation will only be triggered for staging/development stage, since its just a part of the service (#2142, #2143, #2144)
- find login information for freinet in passbolt. There you can check which users exist

Example (Unique user): 
```
https://freinet-online.de/api/input/v3/personen/suche?vorname=Rudi&nachname=Dutschke&geburtstag=1987-02-01&agencyID=<agencyId>&accessKey=<accessKey>
```
Example (User not unique):
```
https://freinet-online.de/api/input/v3/personen/suche?vorname=Peter&nachname=Pan&geburtstag=1992-10-01&agencyID=<agencyId>&accessKey=<accessKey>
```

f1sh1918:
[...]

PR #2184, opened 2025-05-14
2141: Check if person exists before sending Application data to freinet 

bahaaTuffaha:
### Short Description

Given that I am logged in as a region administrator or a region manager
and data transfer to Freinet-Online is enabled,
when I approve a pending application,
then it is checked whether the person is already present in Freinet-Online by sending the applicants first name, family name and birth date.

### Proposed Changes

- I create a backend function called `FreinetSendApplicationData` that receives applicationId and project and send back a response `SendToFreinetResponse` that has `success` as boolean and `errorMessage` as string.
- This function will get the application from `ApplicationRepository` by id then extract "forenames" ,"surname" and dateOfBirth.
- Then it will send this data as a request to freinet `/api/input/v3/personen/suche` 
- I checked for size of the response if its more that 1 it should send a success false and an error message.
- If the response have 1 or empty this needs to be implemented at 2142 and 2143 tickets.
- At `useCardGenerator.ts` I used the mutation and checked for 'disabled for region' in case if the region is not supported to not show any Toaster.

### Side Effects

- None.

### Testing

- Create a card at Bayern from a form.
- keep in mind that you have to:
  - Add the accessToken at the config file.
  - enable a region at frienetagencies in the backend for ex. `EAK-Demo` and switch the region admin's region number to same as regionID that is 94 for frienet demo.
- Try a name like Peter Pan at the issue description.
- Try a new name to see if it successful.
- See if the toaster is showing up after creating the card.
### Resolved Issues

Fixes: #2141


administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -59,6 +65,20 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
   const [createCardsMutation] = useCreateCardsMutation()
   const [deleteCardsMutation] = useDeleteCardsMutation()
   const appToaster = useAppToaster()
+  const [sendToFreinet] = useSendApplicationDataToFreinetMutation({
+    onCompleted: data => {
```
f1sh1918, line 69: ❌ separate errors with `onError` fkt

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -59,6 +65,20 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
...
+      if (appToaster && data.result.errorMessage !== 'disabled for region') {
+        appToaster.show({
+          intent: data.result.success === true ? Intent.SUCCESS : Intent.DANGER,
+          message:
+            data.result.success === true
```
f1sh1918, line 74: 🔧 i would just send a success message if `(data.result.success === true)` 
If its false (like `disabled for region`) nothing will be shown
In error cases just use `onError` function as mentioned above

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -59,6 +65,20 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
...
+              ? 'Bewerbung erfolgreich an Freinet gesendet'
```
f1sh1918, line 75: 🔧 please use i18n

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetSendApplicationData.kt
```diff
@@ -0,0 +1,160 @@
...
+import io.ktor.http.HttpStatusCode
+import io.ktor.http.URLProtocol
+import io.ktor.http.path
+import kotlinx.coroutines.runBlocking
+import kotlinx.serialization.json.Json
+import org.jetbrains.exposed.sql.transactions.transaction
+import org.slf4j.LoggerFactory
+
+@Suppress("unused")
+class FreinetSendApplicationData {
```
f1sh1918, line 31: 🔧 Please separate external api call into a separate class and move it to util folder f.e.
Then just call it in your mutation transaction.

f1sh1918 (commented):
Hey @bahaaTuffaha nicely done for a first touch with the backend 👍 

There are some patterns and improvements i added to be convention with out structure and some best practises.

I will explicitly test everything after these things were adjusted. If you need any help or you have questions just let me know

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetSendApplicationData.kt
```diff
@@ -0,0 +1,160 @@
...
+            if (!Authorizer.mayViewApplicationsInRegion(admin, regionId)) {
+                logger.warn("unauthorized access")
+                return@transaction SendToFreinetResponse(false, userErrorMessage)
+            }
+
+            val freinetAgency = FreinetAgencyRepository.getFreinetAgencyByRegionId(regionId)
+                ?: throw NotFoundException("Freinet agency not found for region")
+
+            if (!freinetAgency.dataTransferActivated) {
+                logger.info("Freinet is disabled for region $regionId")
```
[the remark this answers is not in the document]
bahaaTuffaha: I can create an exception for it and we can ignore that exception at administration so that if the syncing is successful it will show success if it fails it will show failure and if region disabled it should show nothing.

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -59,6 +66,23 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
...
   const [deleteCardsMutation] = useDeleteCardsMutation()
   const appToaster = useAppToaster()
+  const { t } = useTranslation('cards')
+
+  const [sendToFreinet] = useSendApplicationDataToFreinetMutation({
+    onCompleted: data => {
+      if (data.sendApplicationDataToFreinet === true) {
+        appToaster?.show({ intent: 'success', message: t('freinetDataSyncSuccessMessage') })
+      } else {
+        appToaster?.show({ intent: 'danger', message: t('errors:freinetDataSyncError') })
```
f1sh1918, line 76: ❓ in which case is this needed?
I think if there is no error but the data was not send because its not activated for the region, there shouldn't be any toast

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -59,6 +66,23 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
...
+      }
+    },
+    onError: error => {
+      if (error.message.includes('FREINET_DATA_TRANSFER_NOT_ACTIVATED') === false) {
```
f1sh1918, line 80: 🔧 I wouldn't throw an error if its not activated in the backend. Then this check is not needed

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
```diff
@@ -0,0 +1,145 @@
...
+            }
+
+            val applicationData = application.jsonValue
+            val objectMapper = jacksonObjectMapper()
+            val jsonNode = objectMapper.readTree(applicationData)
+
+            val personalDataNode = jsonNode
+                .path("value").firstOrNull { it["name"].asText() == "personalData" }
+
+            fun JsonNode.findValueByName(fieldName: String): String? =
```
f1sh1918, line 90: ❓ what about this function? you just execute it but you don't assign any return value

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
```diff
@@ -0,0 +1,145 @@
...
+                logger.warn("unauthorized access")
+                throw UnauthorizedException()
+            }
+
+            val freinetAgency = FreinetAgencyRepository.getFreinetAgencyByRegionId(regionId)
+                ?: throw FreinetAgencyNotFoundException()
+
+            if (!freinetAgency.dataTransferActivated) {
+                // Freinet is disabled for this region
+                throw FreinetDataTransferNotActivatedException()
```
f1sh1918, line 80: 🔧 i wouldn't throw an error here, because every region that hasn't activated this in region settings (lots of them not using freinet) will recieve an error. just return false

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
```diff
@@ -0,0 +1,145 @@
...
+
+            val personalDataNode = jsonNode
+                .path("value").firstOrNull { it["name"].asText() == "personalData" }
+
+            fun JsonNode.findValueByName(fieldName: String): String? =
+                this.firstOrNull { it["name"].asText() == fieldName }
+                    ?.get("value")
+                    ?.asText()
+
+            val firstName = personalDataNode?.get("value")?.findValueByName("forenames").orEmpty()
```
f1sh1918, line 95: 🔧 Please move the logic of external api calls in a separate class as i requested already before

administration/src/util/translations/de.json
```diff
@@ -561,7 +562,10 @@
     "regionNotUnique": "Den Akzeptanzstellen konnte keine eindeutige Region zugewiesen werden. Bitte wenden Sie sich an den IT-Support",
     "pdfCreationError": "Etwas ist schiefgegangen beim Erstellen der PDF.",
     "csvCreationError": "Etwas ist schiefgegangen beim Erstellen der CSV.",
-    "freinetAgencyNotFound": "Diese Freinetagentur existiert nicht."
+    "freinetAgencyNotFound": "Diese Freinetagentur existiert nicht.",
```
f1sh1918, line 565:
```suggestion
    "freinetAgencyNotFound": "Diese Agentur konnte in Freinet nicht gefunden werden.",
```

administration/src/util/translations/de.json
```diff
@@ -561,7 +562,10 @@
...
+    "freinetDataSyncError": "Die Daten konnten nicht mit Freinet-Online synchronisiert werden. Bitte aktualisieren Sie die Daten in Freinet-Online manuell.",
+    "freinetMultiplePersonError": "Es wurden mehrere Personen mit diesen Daten gefunden. Bitte aktualisieren Sie die Daten in Freinet-Online manuell.",
+    "freinetApiNotReachableError": "Die Freinet-Online-Schnittstelle konnte nicht erreicht werden."
```
f1sh1918, line 568:
```suggestion
    "freinetApiNotReachableError": "Es ist keine Verbindung zu Freinet-Online möglich. Bitte versuchen Sie es zu einem späteren Zeitpunkt erneut. Bitte aktualisieren Sie die Daten in Freinet-Online manuell."
```

administration/src/util/translations/de.json
```diff
@@ -561,7 +562,10 @@
     "regionNotUnique": "Den Akzeptanzstellen konnte keine eindeutige Region zugewiesen werden. Bitte wenden Sie sich an den IT-Support",
     "pdfCreationError": "Etwas ist schiefgegangen beim Erstellen der PDF.",
     "csvCreationError": "Etwas ist schiefgegangen beim Erstellen der CSV.",
-    "freinetAgencyNotFound": "Diese Freinetagentur existiert nicht."
+    "freinetAgencyNotFound": "Diese Freinetagentur existiert nicht.",
+    "freinetDataSyncError": "Die Daten konnten nicht mit Freinet-Online synchronisiert werden. Bitte aktualisieren Sie die Daten in Freinet-Online manuell.",
```
f1sh1918, line 566: 🔧 No exception needed afaics

f1sh1918 (commented):
Very good improvement. Just some small things i added @bahaaTuffaha

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
bahaaTuffaha: false = fail , true = success , if exception = 'FREINET_DATA_TRANSFER_NOT_ACTIVATED' it means ignore.

https://github.com/digitalfabrik/entitlementcard/blob/b86470a501258261394c2f84c85d936f8dcf150b/administration/src/bp-modules/cards/hooks/useCardGenerator.ts#L79-L84

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
bahaaTuffaha: Aha I get what you mean... just error for exceptions and ignore false return.. right?

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
bahaaTuffaha: How?  findValueByName returns a string.

administration/src/util/translations/de.json
bahaaTuffaha: Removed `freinetDataSyncError` .

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
f1sh1918: ah now i got it, my mistake. you just declared this function in the middle of the code, so i was confused. Please move it to util. There is already a `JsonNodeSubstitute` file. So you can create a separate one for this function

administration/src/util/translations/de.json
```diff
@@ -561,7 +562,9 @@
     "regionNotUnique": "Den Akzeptanzstellen konnte keine eindeutige Region zugewiesen werden. Bitte wenden Sie sich an den IT-Support",
     "pdfCreationError": "Etwas ist schiefgegangen beim Erstellen der PDF.",
     "csvCreationError": "Etwas ist schiefgegangen beim Erstellen der CSV.",
-    "freinetAgencyNotFound": "Diese Freinetagentur existiert nicht."
+    "freinetAgencyNotFound": "Diese Agentur konnte in Freinet nicht gefunden werden.",
+    "freinetMultiplePersonError": "Es wurden mehrere Personen mit diesen Daten gefunden. Bitte aktualisieren Sie die Daten in Freinet-Online manuell.",
```
f1sh1918, line 566: @connium i think we have to find here a better description like:
"Der Antragsteller konnte nicht eindeutig einer Person in Freinet-Online zugewiesen werden. Bitte aktualisieren Sie die Daten in Freinet-Online manuell."

Maybe you got even a better idea

administration/src/util/translations/de.json
```diff
@@ -561,7 +562,9 @@
...
+    "freinetApiNotReachableError": "Es ist keine Verbindung zu Freinet-Online möglich. Bitte versuchen Sie es zu einem späteren Zeitpunkt erneut. Bitte aktualisieren Sie die Daten in Freinet-Online manuell."
```
f1sh1918, line 567: @connium 
I think here we even need a better description

andrew8er:
@f1sh1918 Shouldn't all methods in `FreinetSearchPersonApi` be `suspend` functions? In my opinion, we should never call blocking functions in a request handler code path.

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -82,6 +102,16 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
         downloadDataUri(dataUri, filename)
         cards.forEach(saveActivityLog)
 
+        if (!isProductionEnvironment() && applicationId != null) {
```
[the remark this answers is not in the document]
bahaaTuffaha: image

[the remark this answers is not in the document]
seluianova: If this is a temporary condition, I would suggest adding a comment with the task number when the condition should be removed

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -82,6 +102,16 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
...
+          const { projectId } = projectConfig
```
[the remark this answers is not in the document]
andrew8er: It is a suggestion, you can keep it the way it is. But I see zero benefit in it.

administration/src/util/translations/de.json
connium: @f1sh1918 @hauf-toni how about "Der antragstellenden Person konnte keine Person in Freinet-Online zugeordnet werden. Bitte aktualisieren Sie die Daten in Freinet-Online manuell."

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetSearchPersonApi.kt
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+
+        if (responseBody == null) {
+            throw FreinetDataTransfernApiNotReachableException()
+        }
+
+        return objectMapper.readTree(responseBody)
+    }
+}
```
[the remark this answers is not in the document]
andrew8er: What do you mean, the expression form? As for this, I think it is very idiomatic Kotlin and easy to read. It reads strictly from top to bottom with the default expression (after the `?:`) at the very bottom. 

I just noticed there are some other improvements that could be made:

- The body is read into a `String` and then fed to the parser. For large files, this can become a performance issue, since we need the whole file stored completely in memory. It is better to feed the data via an `InputStream` to the parser and ensure that it gets closed properly. This can be achieved with `request.bodyAsChannel().toInputStream().use { objectMapper.readTree(it) }`
- Currently, the `ClientRequestException` gets swallowed and replaced by a custom exception, but that is unnecessary. I would just rethrow it after logging and consumer code can catch the standard Ktor exceptions.

```kt
        runBlocking {
            try {
                httpClient.request {
                    url {
                        protocol = URLProtocol.HTTP
                        this.host = host
                        path("/api/input/v3/personen/suche")
                        parameters.append("vorname", firstName)
                        parameters.append("nachname", lastName)
                        parameters.append("geburtstag", dateOfBirth)
                        parameters.append("accessKey", accessKey)
                        parameters.append("agencyID", agencyId.toString())
                    }
                    method = HttpMethod.Get
                }.bodyAsChannel().toInputStream().use { objectMapper.readTree(it) }
            } catch (e: ClientRequestException) {
                logger.error("Freinet error: ${e.response.status} – body:\n${e.response.bodyAsText()}")
                throw e
            }
```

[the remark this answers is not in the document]
bahaaTuffaha: I think I can't put runBlocking in side a suspended function.. so I will remove runBlocking from here and wrap it around the calling of this method.

administration/src/util/translations/de.json
bahaaTuffaha: It wont send any request to freinet If the card creation got aborted. so I will go with second one...

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
```diff
@@ -44,37 +46,28 @@ class FreinetApplicationMutationService {
...
 
-            val persons = FreinetSearchPersonApi.searchPerson(
-                host = projectConfig.freinet.host,
-                firstName = firstName,
-                lastName = lastName,
-                dateOfBirth = dateOfBirth,
-                accessKey = freinetAgency.apiAccessKey,
-                agencyId = freinetAgency.agencyId,
-            )
+            val persons = runBlocking {
```
f1sh1918, line 62: @andrew8er i guess if we use `runBlocking` here it will not solve the issue to unblock the api call
I would suggestion to wrap the api call with
```
CoroutineScope(Dispatchers.IO).launch {
// api call
}
```
Even it would be great to get some clarification why this makes sense and what happens behind the scenes

administration/package.json
andrew8er, line 1: ❌ This downgrades a lot of packages. You accidentally resolved the wrong parent while merging this file.

administration/src/App.tsx
andrew8er, line 1: ❌ Merge error.

andrew8er (changes requested):
I think this branch needs some cleanup, before we can proceed.

@bahaaTuffaha Could it be that you use merge pulls to sync remote branches? Because there are several merge commits in this PR that merge the remote tracking branch (338a1c7, 584a5cd, 5fb4763, a620c5d, 11a30bf) into this PR. This introduced a lot of old code, and I must admit, I have no idea what's going on by now in this PR.

Please run `git config get pull.rebase` and check if it says true. If not, run these two commands:

```
git config --global pull.rebase true
git config --global pull.autosetuprebase always
```

Did you do an interactive rebase before? Now it would be a good time, you can drop the merge commits, the erroneous commits and their reverts.

Also this might be easier, if you use a gui. I use https://git-fork.com/ which is free (except for the occasional nag screen, that can easily be dismissed with Esc)

andrew8er:
@bahaaTuffaha I created a new branch `2141-send-application-data-to-freinet-2` with all your commits (minus the merge commits). You can reset your branch on `main` and cherry-pick all those on your branch if you want.

seluianova:
:question: I assume that when we implement application approval, there will be no need to send the sendApplicationDataToFreinet() request from the administration to the backend as it can be executed in the backend as part of the approval mutation?

seluianova:
> No the data will not be send while approval

@f1sh1918 but in the issue description the freinet request is connected to the approval:

> when I approve a pending application, then it is checked whether the person is already present in Freinet-Online by sending the applicants first name, family name and birth date.

bahaaTuffaha:
> > But then we would have to adjust the CardCreationResultModel and add there a boolean if data was send to freinet to show some additional onCompleted toast
> 
> right, but this sounds simpler to me than a separate mutation 🤔
> 
> > I have no strong opinion what is the better way
> 
> me too actually, since it's already been done that way. up to you @bahaaTuffaha sorry for flooding here. I remember there was a meeting regarding this task but I couldn't attend. so I had to clarify my questions and concerns somewhere 😀

On paper it sounds perfect but what happened is when there is an error thrown it will interrupt the card creation and if I returned them all false I can't now know which error has occurred.
So I will stick with the current implantation unless there is another idea .

f1sh1918:
> > > But then we would have to adjust the CardCreationResultModel and add there a boolean if data was send to freinet to show some additional onCompleted toast
> > 
> > 
> > right, but this sounds simpler to me than a separate mutation 🤔
> > > I have no strong opinion what is the better way
> > 
> > 
> > me too actually, since it's already been done that way. up to you @bahaaTuffaha sorry for flooding here. I remember there was a meeting regarding this task but I couldn't attend. so I had to clarify my questions and concerns somewhere 😀
> 
> On paper it sounds perfect but what happened is when there is an error thrown it will interrupt the card creation and if I returned them all false I can't now know which error has occurred. So I will stick with the current implementation unless there is another idea .

Yes thats kind of true, we have a fallback mechanism if there is an error in the card creation mutation. It can be little tricky to figure out which error will result in rollback and which not. Even freinet should never block any card creation

f1sh1918 (approved):

f1sh1918:
@bahaaTuffaha please also add tests for your new mutation. Since it may be bit more complicated due to external api call you can also create a separate task for it. But it shouldn't be forgotten

seluianova (approved):
> Yes thats kind of true, we have a fallback mechanism if there is an error in the card creation mutation

I think if we catch FreinetApiNotReachableException() in mutation, use it to set sendApplicationDataToFreinet = false in the response and don't throw any further, it will work.
And if we want to pass a specific error message, we could add an extra field for the message...
Or we return an enum with a status instead of boolean, like PERSON_CREATED, DISABLED_FOR_REGION, PERSON_ALREADY_EXISTS, etc.
But it's more complicated than I initially thought, right.
I don't mind the current implementation, it has some advantages :)
Thanks @bahaaTuffaha

f1sh1918:
@andrew8er can you please unblock this pr

connium:
> ❓ is it enough to show the toaster for success / failure? maybe we want to save a boolean in the db / display some note in the application overview if the application has been sent to freinet? the concern is: the toaster is displayed for a few seconds. if the user does not read it, the information about whether the data was sent successfully or not will be lost.

IMO we don't need to add additional logic for this error message. I'd say it's enough to show the toast. If we receive feedback that the people miss data in Freinet because they didn't recognize the toast, we might change the behaviour.
=> Good enough for now.

Issue #2190, opened 2025-05-19
Koblenz card creation does not work anymore


f1sh1918:
### Describe the Bug

When i want to create a card using the form
http://localhost:3000/erstellen

i get an error that there was no data found for this user

<img width="581" alt="Image" src="https://github.com/user-attachments/assets/bce32b89-6893-424b-980a-57bce97f39a8" />

### Steps to Reproduce

1. Go to http://localhost:3000/erstellen
2. Type in user data. Ensure that the userentitlements are correct in the database
3. Click on KoblenzPass erstellen.

### Expected Behavior

The card will be created in the typed in data is correct

### Additional Information

Some quick investigation showed that the `userHash` is still correct but if its converted to `userHash.toByteArray()` it results to a different value than in the database.
In the last release before some dependency updates it worked properly
It looks like this library may cause the issue `kotlin-stdlib` (updated from v.2.0.0 to v.2.1.20)

Issue #2055, opened 2025-04-11
Migrate input fields of the KoblenzPass application form to MUI


connium:
### Describe Your Preferred Solution

Replace the blueprint.js input fields of the KoblenzPass application form with MUI controls.
💡 Hint: there is a separate task for migrating the buttons.

### Additional Information

Maybe it makes sense to document the necessary steps and learnings as we will have to migrate a lot more screens.

f1sh1918:
hey @bahaaTuffaha,

we discussed before and it was mentioned in the meta that we use mui outlined. Looks like you've overseen that
https://github.com/digitalfabrik/entitlementcard/issues/2045

Probably this will lead to that the CardCreation Form has to be updated in the same issue, since the birthday field will be used there.
Otherwise you could duplicate for the meantime the birthday extension, one with mui for koblenz and one keeping the old style for nuernberg. Later we can remove the deprecated one

PR #2191, opened 2025-05-19
2190: Fix KOBLENZ_PEPPER in gradle config

seluianova:
### Short Description
Fix KOBLENZ_PEPPER in gradle config

### Proposed Changes
Fix KOBLENZ_PEPPER in gradle config

### Side Effects
These variables wont be used on production, right? (not sure how do we run backend there)

### Testing
Card creation for Koblenz should work locally 

### Resolved Issues
Fixes: #2190

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -130,7 +130,7 @@ kover {
 tasks.withType<JavaExec>().configureEach {
     systemProperties = properties
     environment("JWT_SECRET", "HelloWorld")
-    environment("KOBLENZ_PEPPER", "testasdf")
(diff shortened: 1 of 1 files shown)


steffenkleinle (approved):
My bad, sorry! I planned to revert this but apparently missed it...

f1sh1918:
Ok i've overseen this and I'm glad it does not rely on dependency updates. Thx for the quick :+1

f1sh1918:
Hm i tested the pr for issue #2055 and there is the correct pepper set and it doesn't work for me @seluianova

PR #2193, opened 2025-05-20
1933: Fix null check on null value

f1sh1918:
### Short Description

This issue occurs if a setState action is executed on a unmounted component and most cases only for ios

### Proposed Changes

- create a separate async function for `locationStatusInitialization`
- after getting the location status, check if component is mounted before setting the state

### Side Effects

- should be none

### Testing

This can only be tested for devs, since this error is thrown silently

1. Make a fresh install on ios! 
2. Click finish in the intro slides, give permission and wait until the map and location Button are initialized
3. There shouldn't be an error in the dev console anymore. You can check for the error by testing this with main branch
4. Ensure that the location button in the right corner of the map shows the correct location status

### Resolved Issues

Fixes: #1933


f1sh1918:
> I do not get any error when using `main`. Could you be more specific on what to look for?


https://github.com/user-attachments/assets/d93ef454-9e3f-47ef-aca5-d0d8d3964699

@andrew8er

f1sh1918:
> After a few tries, I could reproduce the error. However, on your branch, I now get the following, slightly different error:
> 
> ```
> [sentry.platformError] [error] Uncaught Platform Error
>                        This widget has been unmounted, so the State no longer has a context (and should be considered defunct).
> Consider canceling any ...
>                        #0      State.context.<anonymous closure> (package:flutter/src/widgets/framework.dart:951:9)
>                        #1      State.context (package:flutter/src/widgets/framework.dart:957:6)
>                        #2      _LocationButtonState.initializeLocationStatus (package:ehrenamtskarte/map/location_button.dart:60:9)
>                        <asynchronous suspension>
> ```

Hm not sure why this is the case here but looks like it works here with `mounted` instead of `context.mounted` according to the post here https://yawarosman.medium.com/this-widget-has-been-unmounted-so-the-state-no-longer-has-a-context-and-should-be-considered-7592ab8afe36

can you please recheck @andrew8er

andrew8er (approved):
> Hm not sure why mounted instead of context.mounted has to be used here to avoid errors
Just found that: https://yawarosman.medium.com/this-widget-has-been-unmounted-so-the-state-no-longer-has-a-context-and-should-be-considered-7592ab8afe36
>
> Even in stackoverflow its mentioned that it should be the same...
https://stackoverflow.com/questions/79341433/difference-between-context-mounted-and-mounted-in-flutter

Cannot answer your questions, but it works now. Maybe add a comment on the need to use `context` here.

seluianova (approved):
I reproduced the error on main, fix looks good!

PR #2194, opened 2025-05-20
Fix card activation

seluianova:
### Short Description
Card activation is currently broken
Exception in the backend:
```
[JettyServerThreadPool-287] WARN io.javalin.Javalin - Uncaught exception
java.lang.IllegalArgumentException: maxAttempts must be set to perform at least 1 attempt.
	at org.jetbrains.exposed.sql.Transaction.setMaxAttempts(Transaction.kt:77)
	at app.ehrenamtskarte.backend.cards.webservice.schema.CardMutationService.activateCard$lambda$4(CardMutationService.kt:312)
```

### Proposed Changes
Change the transaction's max attempts property to 1 so that the transaction is executed once instead of never

### Side Effects
no

### Testing
1. create card (for any project)
2. activate/add card in the app

### Resolved Issues
n/a


andrew8er (approved):

steffenkleinle (approved):
LGTM, not tested

Issue #1981, opened 2025-03-12
Approve application for Ehrenamtskarte Bayern


connium:
### Describe the Problem

As a region manager or a region administrator,
I want to approve an application
so that I can filter duplicated applications and I don't issue cards twice.

### Describe Your Preferred Solution

Given that I am logged in as a region manager or a region administrator,
when I navigate to the page for incoming applications
and there is a pending application
and I expand the card of the application to show its details,
then I can approve the application via a button.

Given that there is a pending application that can be approved,
when I approve the application,
then I am redirected to the screen for card creation.

Given that there is a pending application that needs to be approved,
when I approve the application
and I issue the card,
then the application is marked as being approved.

Given that there is a pending application that needs to be approved,
when I approve the application
and I cancel the issuing of the card,
then the application is not marked as being approved but is kept in its (pending) state.

Given that there is an approved application,
when I expand the card of the application to show its details,
then I see a hint "Für diesen Antrag wurde am dd. Month yyyy um HH:mm eine Karte erstellt." above the action buttons
and there are two actions available: "Karte erneut erstellen" and "Antrag löschen".

### Additional Information

Design proposal can be found [📌 here.](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=993-587&t=dNJn8iXuBM4mDzCv-1)

connium:
I had a chat with @hauf-toni exactly about that this morning. Maybe I misunderstood her. Let's wait for her response.

hauf-toni:
My personal best case scenario from a uiux perspective would be to include another cta "create card"[ in this screen](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=1124-774&t=zupIRsI13Dd3d2cC-1). However, a less complex version (code-wise) would probably [be this one](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=1136-3665&t=zupIRsI13Dd3d2cC-1). let me know which one you prefer. I'm okay with both @f1sh1918

f1sh1918:
@hauf-toni no i absolutely with you. I think the first step should be "approve" or "reject"
Then we have an application process status.
Depending on that we can either create a card and delete the application or just delete the application
Export functions will be available in all the cases

f1sh1918:
@andrew8er i updated the user story by two more sections.
[...]

PR #2196, opened 2025-05-20
1981.1: Refactor application overview in administration

andrew8er:
### Short Description

In preparation of #1981 I refactored the application overview to use MUI components and adhere to the [design](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=993-587&t=dNJn8iXuBM4mDzCv-1).

Almost all components are now using MUI components rather than Blueprint, except for some components used in the application filter bar and the application detail tree.

### Proposed Changes

- Update `eslint` dependencies.
- Update `@mui` dependencies.
- Update `date-fns` dependeny.
- Add a preliminary MUI theme for use throughout the application.
- Use MUI components and implement design in `/applications`

### Side Effects

None

### Testing

- Load test data with `backend: $ ./gradlew db-recreate`.
- Run backend and adminstration dev server.
- Visit http://localhost:3000/applications.
- Run tests.

### Resolved Issues

Partially addresses #1981


administration/src/bp-modules/applications/VerificationsQuickIndicator.tsx
```diff
@@ -1,51 +1,57 @@
...
-const VerificationQuickIndicator = memo(({ verifications }: { verifications: Application['verifications'] }) => {
-  const verificationStati = verifications.map(getVerificationStatus)
+const ToolTipContent = forwardRef<HTMLDivElement, unknown>((p, ref) => {
   const { t } = useTranslation('applicationsOverview')
+  const theme = useTheme()
+
+  return (
+    <Container
+      ref={ref}
+      sx={{
```
f1sh1918, line 17: ❓Just a general question. Why do we not use the mui `styled` api here?
I mean if we want to style just a particular part of a complex component we have to need the particular props but i general i prefer the styled components style, since the readability is way better

administration/src/i18n.ts
```diff
@@ -32,6 +32,7 @@ export const loadTranslations = (): Translations => ({
     stores: deTranslations.stores,
     userSettings: deTranslations.userSettings,
     users: deTranslations.users,
+    component_Accordion: deTranslations.component_Accordion,
```
f1sh1918, line 35: 💭 hm currently we set namespaces for pages and we already, so i would tend to say this can be in `applicationOverview` 
If we really want to have a namespace for "base component" that we reuse in several places, we can even create a "baseComponents" namespace or put it in misc. 
I don't see an advantage in creating a namespace for each component.
What was your intend?

administration/src/App.tsx
```diff
@@ -20,7 +20,9 @@ const App = (): ReactElement => (
       <AuthProvider>
         <AppApolloProvider>
           <LocalizationProvider dateAdapter={AdapterDateFns}>
-            <Router />
+            <ThemeProvider theme={theme}>
```
f1sh1918, line 23: ❌ The imports are missing for `ThemeProvider` and `theme` itself, so the app breaks

administration/src/bp-modules/applications/NoteDialogController.tsx
```diff
@@ -67,7 +68,12 @@ const NoteDialogController = ({
   return (
     <>
       <Tooltip content={toolTipContent}>
-        <NoteButton onClick={() => onOpenNoteDialog(true)} intent='none' icon='annotation'>
+        <NoteButton
```
f1sh1918, line 71: 🙃 I'm not super happy with the styling of the button. Its gray and onHover white.
Not sure if this is intended in the design but seems weird to me since a hover should be highlighting.
But maybe thats just a personal preference

f1sh1918 (changes requested):
Looks really nice already. Great work 👍 
Theme provider imports are missing (see comments and ci pipeline), so the app is broken currently.

One small issue:
PDF Export the status bar is broken. I would tend to just hide it for print
image
And i added another minor button styling comment and namespaces

In general if you add TODOs please create an issue for it and refer to it :)

administration/src/bp-modules/applications/VerificationsQuickIndicator.tsx
andrew8er: When making a one-off customization of a component, I generally use the `sx` prop. For components that get reused or I want to externalize, I would probably use the `styled()` approach. I tend to follow these guidelines:

https://mui.com/material-ui/customization/how-to-customize/#overriding-nested-component-styles

Though I'm a bit annoyed that there seems to be no API to combine both ways (make a component with a styled top-level element and provide e.g. custom `slotProps`) and that the `styled()` API are slightly different in functionality: https://mui.com/system/styled/#difference-with-the-sx-prop

administration/src/i18n.ts
andrew8er: Yes, I had the idea of one namespace per component for reusable components. I would like to refactor the file layout later after the MUI migration and wanted to introduce a `/shared` dir for components and other functionality at some point. 

We could use one namespace for all modules in (future) `/shared`, but I still would prefix component specific strings with e.g. `Accordion-` or something. Would you agree?

administration/src/App.tsx
andrew8er: I don't get it. This worked locally up until some point a few minutes ago. Should be already fixed.

administration/src/bp-modules/applications/NoteDialogController.tsx
andrew8er: I agree, did not test this thoroughly enough.

The MUI palette system is a bit lacking in view, too many common cases are not covered, for example, there is no configurable way for inverted panels, for example light text on dark background in a light theme. It can be fixed with module augmentation as shown in `src/types/mui-theme.d.ts` though.

I would like an official style guide here.

administration/src/App.tsx
f1sh1918: Just checked the example code of mui and the imports are defenitively needed :)

administration/src/bp-modules/applications/VerificationsQuickIndicator.tsx
f1sh1918: Hm my personal preference would be to use `styled()` if its possible even if the components are not reused because i find it better to read and it's clearer when to use what.
In all other cases where particular parts via props have to be styled for sure this way has to be used.
Thats also how we did it in other projects so far afaics.

But maybe others of the team have another opinion

administration/src/App.tsx
andrew8er: I had the imports in there, but somehow they got deleted, probably in a rebase…

connium:
@andrew8er @f1sh1918 From my experience the best way is to tweak the theme settings and if something cannot be handled by theme overrides, we should ask ourselves if it is worth it to differ from the default design/behaviour (in most cases it is not).

administration/src/bp-modules/applications/VerificationsQuickIndicator.tsx
f1sh1918: ok i'm fine to keep that for now.
yes sure we shouldn't mix styled apis

administration/src/util/translations/de.json
```diff
@@ -606,5 +606,9 @@
     "referenceNrInvalidLengthError": "Das Aktenzeichen muss eine Länge zwischen {{KoblenzReferenceNumberMinLength}} und {{KoblenzReferenceNumberMaxLength}} haben.",
     "referenceNrLabel": "Aktenzeichen",
     "startDayLabel": "Startdatum"
+  },
+  "shared": {
+    "Accordion-showLess": "Weniger anzeigen",
```
f1sh1918, line 611: 🙃 i would like to keep the camel case conventions for namespaces
```suggestion
    "accordionShowLess": "Weniger anzeigen",
```

Same for showMore

bahaaTuffaha (approved):
Nice work ✅

🙃 nitpicking:
it looks different from the design



Maybe for a future ticket: making application view responsive on mobile.🤔

administration/src/mui-modules/NonIdealState.tsx
```diff
@@ -0,0 +1,20 @@
+import { Container, Stack, Typography, useTheme } from '@mui/material'
+import React, { ReactElement } from 'react'
+
+/* eslint-disable react/destructuring-assignment */
```
bahaaTuffaha, line 4: You can deconstruct it without adding an eslint rule:
`const NonIdealState = ({ title, icon, description }:{ title: string; icon: ReactElement; description: string }`

andrew8er:
> 🙃 nitpicking:
> it looks different from the design

You are right, I didn't see this dialogue in the design. But I think, it is more than nitpicking here. It looks like pretty out of place from all other design elements, and is inconsistent with other dialogues. I will check back with Toni and ask for clarification on this.

andrew8er:
@f1sh1918 @bahaaTuffaha I addressed all of your concerns and rebased the branch on top of `main` again.

As for the dialog styles, I kept it as it is for now, as I am not sure in what general direction the design will go. The biggest inconsistency is the position of the primary button, in one instance it is on the left, in the other on the right. The [Material2 design guide](https://m2.material.io/components/buttons#hierarchy-and-placement) states that the primary button goes to the right.

Screenshot 2025-05-27 at 12 24 55
Screenshot 2025-05-27 at 12 25 06

f1sh1918:
> @f1sh1918 @bahaaTuffaha I addressed all of your concerns and rebased the branch on top of `main` again.
> 
> As for the dialog styles, I kept it as it is for now, as I am not sure in what general direction the design will go. The biggest inconsistency is the position of the primary button, in one instance it is on the left, in the other on the right. The [Material2 design guide](https://m2.material.io/components/buttons#hierarchy-and-placement) states that the primary button goes to the right.
> 
> Screenshot 2025-05-27 at 12 24 55
> Screenshot 2025-05-27 at 12 25 06
> 

@hauf-toni i think this should be reflected in the designs that primary buttons are on the right side

PR #2200, opened 2025-05-21
Update gradle related xmls

f1sh1918:
### Short Description

Not sure if i'm alone but i'm facing the issue "Invalid Gradle JDK configuration found" for intelliJ, probably due to the gradle update

### Proposed Changes

- update gradle related intellij xml config files

### Side Effects

- none

### Testing

- starting the backend should work

### Resolved Issues

Fixes: #


.idea/modules.xml
```diff
@@ -3,6 +3,9 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/administration/administration.iml" filepath="$PROJECT_DIR$/administration/administration.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.main.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.main.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.test.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.test.iml" />
```
andrew8er, line 8: ❌ As far as I know, `.iml` files for Gradle modules are redundant, since IntelliJ will infer the module definition from the Gradle object model (this is what */Build/Sync Project with Gradle files* does). 

The `.iml` files are an interoperability feature for projects that also have non-Gradle users, and are automatically created from the Gradle model, when enabled (Disable _"Generate *.iml files for modules imported from Gradle"_ in *Settings:/Build, Execution, Deployment/Build tools/Gradle*).

Also these files do not exist.

andrew8er (changes requested):
Some module references are redundant, probably created because of an IntelliJ setting. Disable "Generate *.iml files for modules imported from Gradle" in "Settings:/Build, Execution, Deployment/Build tools/Gradle".

Otherwise that cleanup is really necessary, thanks!

bahaaTuffaha (approved):
image

Can we also ignore these 2 files
- backend/versions.properties
- backend/.kotlin/sessions/kotlin-compiler-7550422918674850352.salive

andrew8er (approved):

PR #2202, opened 2025-05-23
Group dependabot updates and reduce frequency

michael-markl:
### Short Description

Groups dependabot updates and reduces frequency to reduce the number of PRs.

### Testing

n/a

### Comment

Could someone with sufficient permissions check in the repository settings whether we have to enable dependabot security updates (and if so enable them)? I have no access to the repo settings...

Here's a guide:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#enabling-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,60 +1,89 @@
+# yaml-language-server: $schema=https://json.schemastore.org/dependabot-2.0.json
+
 version: 2
 updates:
-
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: monthly
     ignore:
       - dependency-name: "*"
         # Ignore major updates for all dependencies
-        update-types: [ "version-update:semver-major" ]
-    open-pull-requests-limit: 2
+        update-types: ["version-update:semver-major"]
+    groups:
+      minor:
+        patterns: ["*"]
+        update-types: [minor]
+    open-pull-requests-limit: 1
 
   - package-ecosystem: gradle
     directory: "/backend"
     schedule:
-      interval: daily
+      interval: monthly
     labels:
       - "not-testable"
-    open-pull-requests-limit: 2
+      - "dependencies"
+    ignore:
+      - dependency-name: "*"
+        # Ignore major updates for all dependencies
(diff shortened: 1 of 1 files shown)


steffenkleinle:
> Could someone with sufficient permissions check in the repository settings whether we have to enable dependabot security updates (and if so enable them)? I have no access to the repo settings...

I would say it looks good:
image

.github/dependabot.yaml
```diff
@@ -1,60 +1,89 @@
...
-      interval: daily
+      interval: monthly
     labels:
       - "not-testable"
-    open-pull-requests-limit: 2
+      - "dependencies"
+    open-pull-requests-limit: 1
+    groups:
+      all:
+        patterns: ["*"]
```
steffenkleinle, line 89: 🤔 Is there `update-types` missing here?
```suggestion
        patterns: ["*"]
        update-types: [minor]
```

steffenkleinle (approved):
LGTM

michael-markl:
Thanks @steffenkleinle :)  Out of curiousity, could you also sned a screenshot of the dependabot rules (ie. click on the button next to "1 rule enabled")?

.github/dependabot.yaml
michael-markl: I would say, for the docker containers, we can also let dependabot do major upgrades, that's why I didn't put it there so far :) I can definitely add it, if desired. Let me know

steffenkleinle:
> Thanks @steffenkleinle :) Out of curiousity, could you also send a screenshot of the dependabot rules (ie. click on the button next to "1 rule enabled")?

Good question:
image

f1sh1918 (approved):
LGTM
I'm not sure how well grouped the updates are.
It may make sense to start with `weekly` interval until we are better up-to-date and then move to monthly. But we can also test if `monthly` is sufficient

michael-markl:
@f1sh1918 let's try :)

.github/dependabot.yaml
steffenkleinle: No, was just a question if this was on purpose or you just forgot it. But that answers the question, thanks.

Issue #2056, opened 2025-04-11
Migrate buttons of the KoblenzPass application form to MUI


connium:
### Describe Your Preferred Solution

Replace the blueprint.js buttons of the KoblenzPass application form with MUI controls.
Also make sure to include the help modal.

### Additional Information

Maybe it makes sense to document the necessary steps and learnings as we will have to migrate a lot more screens.

PR #2207, opened 2025-05-24
2056: Migrate checkbox to mui

f1sh1918:
### Short Description

Migrated missing checkbox to MUI.
I think in the `primitive-inputs` is still too much application form related code, thats why i even create a more basic and reuseable component

### Proposed Changes

- create reusable checkbox base component (without form logic)
- use it in the `CheckboxForm` and in `CardSelfServiceForm` 
- move `FormAlert` to mui/base
- create unit tests for base components
- replace spinner

### Side Effects

Some error messages look different in the application form. We should decide on a particular format (with or without the icon)

image

### Testing

1. Open card creation self service portal for koblenz. Test the checkbox works and the error message will be shown
2. Open application form and test the checkboxes there

### Resolved Issues

Fixes: #2056


administration/src/mui-modules/base/__tests__/FormAlert.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+describe('FormAlert', () => {
+  it('should not render when errorMessage is null', () => {
+    const { container } = render(<FormAlert errorMessage={null} />)
+    expect(container.firstChild).toBeNull()
+  })
+
+  it('should not render when errorMessage is undefined', () => {
+    const { container } = render(<FormAlert errorMessage={undefined} />)
+    expect(container.firstChild).toBeNull()
+  })
```
bahaaTuffaha, line 15: Do we need to test both (null and undefined)? both are falsy anyway. 🤔

bahaaTuffaha (approved):
Tested on Firefox and chromium based browser ✅

Does these other check-boxes needs replacement or it's just for card creation?

image

administration/src/mui-modules/base/__tests__/FormAlert.test.tsx
f1sh1918: you are right. i will change that

f1sh1918:
> #2056

Well the root of this ticket was to migrate koblenz related stuff. i reused then the checkbox for application process to have all end user facing checkbox same style. I would adjust the other checkboxes in a separate ticket

seluianova (approved):
looks good to me. nit-pick fix is up to you

f1sh1918:
Updated the missing alert old styled alert (with background)
I kept the small glitch in the alignment under the radio buttons, since we may have to override the radio buttons stylingand its really minor

Issue #2000, opened 2025-03-19
Setup sentry backend


f1sh1918:
### Describe the Problem

Since we received some errors in the backend recently and needed couple of time to recognize it would be nice to also get some alerts from sentry.

### Describe Your Preferred Solution

- setup a kotlin project in sentry
- add an alert that creates a github bug issue if a particular threshold is reached (check the alerts for the other projects as a baseline)
- add the requirement implementation in the `backend`
- provide a `reportErrorToSentry` function that sends errors in particular cases
- general exception and errors should be reported automatically

PR #2209, opened 2025-05-27
2000: Setup Sentry backend

f1sh1918:
### Short Description

I added sentry for backend jvm to better collect errors in the backend. Sentry will automatically capture errors.

### Proposed Changes

- initialize sentry on webservice start for production environment
- add reporting function that can be used in particular cases where its needed. Currently only for testing
- add tracing to all applications

https://docs.sentry.io/concepts/key-terms/tracing/

### Side Effects

- none

### Testing (local dev only)
1. Get the auth token from here:
https://passbolt.tuerantuer.org/app/passwords/view/9aa6c68e-3d03-4dc4-9152-bdc4bdc24f0e
2. Add in the intellij runConfig `run backend` the environment variable (for local testing)
`SENTRY_AUTH_TOKEN=<Token>`
3. Adjust in `Webservice.kt` the environment to enable sentry for development
```java
  if (config.environment == Environment.DEVELOPMENT) {
            initializeSentry()
            println("Init Sentry for production environment")
        }
```
4. Throw an example error f.e. in `SignInMutationService` and login in the administration

```java
    try {
            throw Exception("This is a test.")
        } catch (e: Exception) {
            reportErrorToSentry(e)
        }
```

5. Check that the error was transmitted to sentry
https://sentry.tuerantuer.org/organizations/digitalfabrik/projects/entitlementcard-backend/?issuesType=all&project=7

Note: Please don't commit any environment variables :)

Here is particular salt pr to set the environment variable for the services on the server
https://git.tuerantuer.org/DF/salt/pulls/358
### Resolved Issues

Fixes: #2000

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
@@ -85,7 +85,7 @@ class WebService {
 
         app.start(host, port)
 
-        if (config.environment == Environment.DEVELOPMENT) {
+        if (config.environment == Environment.PRODUCTION) {
             initializeSentry()
             println("Init Sentry for production environment")
         }


bahaaTuffaha (approved):
Nice ✨ ..Tested ✅

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/webservice/WebService.kt
```diff
@@ -82,6 +84,11 @@ class WebService {
         app.post("/users/import") { context -> userImportHandler.handle(context) }
 
         app.start(host, port)
+
+        if (config.environment == Environment.DEVELOPMENT) {
```
andrew8er, line 88: Should this be `config.environment == Environment.PRODUCTION`?

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/utils/Sentry.kt
```diff
@@ -0,0 +1,17 @@
...
+        // https://docs.sentry.io/concepts/key-terms/tracing/
+        options.tracesSampleRate = 0.05
+    }
+}
+
+fun reportErrorToSentry(e: Exception) {
+    if (Sentry.isEnabled()) {
+        Sentry.captureException(e)
+    }
+}
```
andrew8er, line 17: 🔧  This function is unnecessary, we can just use `Sentry.captureException(e)`

andrew8er (approved):
Tested web, backend.

backend/src/main/kotlin/app/ehrenamtskarte/backend/common/utils/Sentry.kt
f1sh1918: @andrew8er but is this function checking if `Sentry.isEnabled()` ? Couldn't find anything in the docs about that

PR #2212, opened 2025-05-27
2211: Add info data transfer error message

f1sh1918:
### Short Description

Koblenz wants an additional info if user entitlement can't be found

### Proposed Changes

- adjust text in translation
- add white-space preline to respect new lines in FormAlerts

### Side Effects

- none
### Testing
- Open koblenz pass self service `/erstellen`
- Input form with data that can't be found
- Check for the correct error message and that the message respect new lines

### Resolved Issues

Fixes: #2211


seluianova:
This PR should have been created one PR ago 🙃
image

PR #2214, opened 2025-05-27
Post install restart backend after migration

f1sh1918:
### Short Description

Since we had some issue that after a package update, the backend service was not restarted successful, i added a service restart in the debian package post install step.

### Proposed Changes

- backend service restart in the backend debian package post install step

### Side Effects

- should be none, if migration runs first

### Testing

Only testable when new package will be installed in the server


f1sh1918:
@svenseeberg after merging this and doing the next release we should remove the service restart from salt I guess to avoid duplication

seluianova (approved):
looks good

Issue #2215, opened 2025-05-27
MUISvgIcons are not clearly visible


f1sh1918:
### Steps to Reproduce

1. localhost:3000/erstellen (koblenz)
2. Check the icons on the right corner of the birthday field (CustomDatePicker)
3. See error

<img width="344" alt="Image" src="https://github.com/user-attachments/assets/3d8a8af3-bb00-40ec-90ee-0f24a0a09332" />

### Expected Behavior

Icons should be visible like before

<img width="496" alt="Image" src="https://github.com/user-attachments/assets/668f3c89-4f40-470a-b07f-0afcdd42d92b" />

PR #2216, opened 2025-05-27
2215: MUI SVG Icons not visible

f1sh1918:
### Short Description

Due to the creation of the mui theme and declaring default style palette for Buttons, even the MuiIconButton inherited the default color from the Button.
MuiIconButtons are currently only used for datepickers and can be set to dark gray

### Proposed Changes

- override color for MuiIconButton

### Side Effects

- can have effect on other MuiIconButtons

### Testing

1. Please check for icon buttons in all extensions (mostly datepickers)
2. Also check for the application card (extended) that the icons there still have the text color
3. Check the buttons with icons in the application form to still be correct

### Resolved Issues

Fixes: #2215


bahaaTuffaha (approved):
Tested .. looks good✅

seluianova:
Not related to the current changes, but just in case you want to add a fix for that too. Otherwise I can create a separate issue.
The calendar icon is slightly shifted to the left:
image

seluianova (approved):
lgtm, thanks!

f1sh1918:
> Not related to the current changes, but just in case you want to add a fix for that too. Otherwise I can create a separate issue. The calendar icon is slightly shifted to the left: image

@seluianova yes you can create an issue for that. I even think this will be fixed when we switch to outlined forms also for the application. So i think this is really minor and just temporary

PR #2238, opened 2025-06-02
Fix typos in the application form

seluianova:
### Short Description
We had labels of the buttons in uppercase before and it didn't show that verbs started with uppercase letters.
image
image
image

Alternatively this should be changed to uppercase again, not sure what is more desired?

### Proposed Changes
- fix translations

### Side Effects
no

### Testing
Check labels in the application form

### Resolved Issues
n/a


f1sh1918 (commented):
Looks good. Just one small issue

bahaaTuffaha (approved):

f1sh1918:
> Added translation for attach file button.
> And also fixed description of the attachment. I have just noticed that we use wrong label there - attachment title instead of attachment description.
> image

Yes that may happened during refactoring. 
We just have to ensure that we don't replace a key that will be stored in the database and can't be resolved in the application card anymore @seluianova 

I will have a deeper look tommorow :)

Issue #2239, opened 2025-06-02
Runtime error when entering Startdatum from the keyboard in the card creation form


seluianova:
### Describe the Bug
Runtime error on the card creation form
`Parameters do not form a valid PlainDate.`

### Steps to Reproduce
1. Go to Nürnberg administration
2. Click on 'Karten erstellen' /cards/add
3. Click on Startdatum field and enter any key from the keyboard

<img width="892" alt="Image" src="https://github.com/user-attachments/assets/b6ce9ba2-6292-4761-90f2-557371bfb953" />

### Expected Behavior
No runtime errors

seluianova:
@connium only Nürnberg, because we have Startdatum extension only for Nürnberg

f1sh1918:
prefilled fields are likely less tested :/
nice find 👍 
I created a pr

PR #2242, opened 2025-06-03
2239: Runtime error start date

f1sh1918:
### Short Description

The plainDate of the startDay was not constructed safely

### Proposed Changes

- use SafeFromLocalDate to safely construct plain date 
- add null for initial value as fallback to avoid controlled/uncontrolled error in the console

### Side Effects

- none

### Testing

1. Login nuernberg project
2. Go to Karte erstellen
3. Click in startDay form and try to change the value
4. No error should be shown

### Resolved Issues

Fixes: #2239


bahaaTuffaha (approved):
Nice.. tested on Firefox and Chrome ✅

PR #2249, opened 2025-06-04
1981.2: Approve application for Ehrenamtskarte Bayern

andrew8er:
### Short Description

This PR enables the explicit approval of an entitlement card application. Region managers can now approve (later also reject) an application. An approved application is eligible for the issuance of a card, but can also be deleted if necessary. 

#### Major changes
- backend: Add a DB column `application.status` with the allowed states `Pending`, `Approved`, `ApprovedCardCreated`, `Rejected`. State transitions are checked for plausibility.
- backend: Add a DB column `application.statusResolvedDate`.
- backend: The DB column `application.cardCreated` is folded into `application.status` via the migration.
- backend: Add GraphQL mutation `approveApplicationStatus()`
- backend: Use one single transaction in `CardMutationService.createCardsByCardInfos()`. There are several DB operations in this mutation, and they all can fail, so they should be run in a single transaction.
- API: Deprecate `ApplicationView.cardCreated`.
- API: Add fields `status` and `statusResolvedDate` to type `ApplicationView`.

#### Minor changes

- administration: Remove some unused files.
- administration: Improve the GraphQL code generator config.
- administration: Refactor some areas around application handling.
- administration: Remove a redundant memoization in `<ApplicationsOverview>`.
- backend: Let `ApplicationRepository`'s methods only return entities.
- backend: Remove redundant transactions in `ApplicationRepository`' methods.
- backend: Move the function `isAlreadyVerified()` to class `ApplicationVerificationEntity` as a property.
- backend: Add more test data.

#### Open questions

- administration: Should code accessing the deprecated field `ApplicationView.cardCreated` be refactored?
- backend: There should be some unit tests for the new mutation.

### Side Effects

None

### Testing

- Recreate the DB from scratch and run migrations: `./gradlew db-recreate` (in `/backend`)
- Start backend and administration
- Navigate to http://localhost:3000/applications

### Resolved Issues

Fixes: #1981


backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0028_Add_Application_Status.kt
```diff
@@ -0,0 +1,23 @@
...
+            """
+            -- Add column "status"
+            alter table "applications" add column "status" varchar(32) not null default 'Pending'
+                 check ("status" in ('Pending', 'Approved', 'ApprovedCardCreated', 'Rejected'));
+            -- Set "status" to 'AcceptedCardCreated' for rows that have "cardCreated" set to true
+            update "applications" set "status" = 'AcceptedCardCreated' where "cardCreated";
+            -- Remove ”cardCreated" column
+            alter table "applications" drop column "cardCreated";
+            -- Add column "statusResolvedDate"
+            alter table "applications" add column "statusResolvedDate" timestamptz default null;
```
f1sh1918, line 19: ❓ why is `timestamptz` needed here. I think we only store UTC in the database

f1sh1918: hm okay i understand. as long as we don't have issues with mixing the types when we display it for the users, i'm fine with that. Even it looks like we can not convert the current `timestamp` to `timestampz`

f1sh1918:
Just some questions to the testing instructions.
Why should i recreate the whole db from scratch? That shouldn't in any case necessary.
It should work in a normal migration.
To test this pr some proper testing instruction would be good like create an application. reject an application, check this or check that and so on.

f1sh1918:
> @f1sh1918
> 
> > i canot not do a migration from current main to this branch
> 
> The error seems to stem from an insertion into the DB rather than a migration. So is this maybe from the `db-import-dev` phase? In any case, this works for me.
> 
> Steps to reproduce:
> 
> * On `main` get a clean state of the DB with `./gradlew db-recreate`
> * Switch to branch `1981.2…`: run `./gradlew db-import-dev`

Well i'm on migration 27. before i do the current one.
This is also the current production state. So if i do a normal migration with "run migration" it should work or it also won't work if we make a new deployment @andrew8er

connium:
Sounds like a good candidate for a quick pairing session @andrew8er @f1sh1918

f1sh1918:
Some general things:
1. When the application was approved, there should be a toast "Antrag genehmigt" 
2. Since the application rejection is not impliment yet, i would just hide the button for now
3. Maybe this will be fixed by updating from main
image

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/schema/view/ApplicationView.kt
```diff
@@ -14,33 +15,42 @@ data class ApplicationView(
...
-                false,
+        fun fromDbEntity(entity: ApplicationEntity, includePrivateInformation: Boolean = false): ApplicationView =
+            ApplicationView(
+                id = entity.id.value,
+                regionId = entity.regionId.value,
+                createdDate = entity.createdDate.toString(),
+                jsonValue = entity.jsonValue,
+                withdrawalDate = entity.withdrawalDate?.toString(),
+                note = entity.note.takeIf { includePrivateInformation },
+                cardCreated = (entity.status == ApplicationEntity.Status.ApprovedCardCreated)
```
f1sh1918, line 32: ❓ Can you explain me what this `includePrivateInformation` is for?

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -125,27 +129,104 @@ const AccordionExpandButton = (p: { expanded: boolean }) => {
...
+}: {
+  cardAlreadyCreated: boolean
+  primaryButtonHref: string
+  onSecondaryButtonClick: () => void
+}) => {
+  const { t } = useTranslation('applicationsOverview')
+
+  return (
+    <>
+      <Tooltip title={t('incompleteMappingTooltip')}>
```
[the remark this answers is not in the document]
andrew8er: I tend to remove the tooltip, because no user (including me) will understand that message.

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -125,27 +129,104 @@ const AccordionExpandButton = (p: { expanded: boolean }) => {
   )
 }
 
+const ButtonsCardPending = ({
```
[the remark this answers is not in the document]
andrew8er: I removed some other components that _might_ be useful in other places from this file, but kept the `<ButtonsCardPending>`, `<ButtonsCardApproved>` etc. components, since they are exclusively used here. The file is reasonably sized now for my taste, and I have some ideas how component definitions might get terser even further. I will present those at some later point.

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -125,27 +129,104 @@ const AccordionExpandButton = (p: { expanded: boolean }) => {
...
+  onSecondaryButtonClick: () => void
+}) => {
+  const { t } = useTranslation('applicationsOverview')
+
+  return (
+    <>
+      <PrimaryButton variant='contained' startIcon={<CreditScore />} disabled={disabled} onClick={onPrimaryButtonClick}>
+        {t('applicationApprove')}
+      </PrimaryButton>
+      <Button
```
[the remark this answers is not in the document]
andrew8er: Unfortunately, we cannot build a unified secondary button here, the color as well as the icon are changing for every use case, and the `variant` is not settable in the style in a satisfactory way.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/schema/view/ApplicationView.kt
andrew8er: From my understanding, this view (should actually be named type IMHO) is reused on several queries, which may or may not include sensitive fields.

I would have created two separate types here, but kept the logic as is for now.

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0028_Add_Application_Status.kt
michael-markl: In fact, exposed [only added support for this two years ago](https://github.com/JetBrains/Exposed/pull/1787). That's why most of our timestamps do not use this type.
While the name ("timestamp with timezone") is more than confusing to me, it would be the better to have all our timestamps use that column type.
I think the old timestamps are the reason why we have this weird workaround where we set the system time zone to UTC on startup.

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -164,43 +192,44 @@ const ApplicationCard = ({
...
+    onCompleted: result => {
+      onChange({ ...application, status: result.status })
+    },
+  })
+
   const createCardQuery = useMemo(
     () =>
-      `${config.applicationFeature?.applicationJsonToCardQuery(jsonValueParsed)}&applicationIdToMarkAsProcessed=${id}`,
-    [config.applicationFeature, jsonValueParsed, id]
+      `./cards/add${config.applicationFeature?.applicationJsonToCardQuery(jsonValueParsed)}` +
```
michael-markl, line 207: `applicationJsonToCardQuery` may return null. In which case the popup below (or maybe a simplified error à la "Karte kann nicht automatisch erstellt werden." should be shown. It may happen if the structure of the application form changes (older applications can not be migrated to a new application form structure).

f1sh1918: Alright thx for the hint. I would then tend to use a less cryptic message and better Show a toast if it can not be resolved

andrew8er: What options do administrators have in such a case beside deleting the application?

michael-markl: I'm not sure what you mean by "options", but they can still view the application, and if desired, manually create a card. (I don't know what is possible wrt. Freinet-related functionality)

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -106,14 +110,38 @@ class EakApplicationMutationService {
...
+     *
+     * @return The updated ApplicationView.
+     * @throws NotFoundException If an application with the specified accessKey is not found.
+     */
+    @GraphQLDescription("Approve an application")
+    fun approveApplicationStatus(applicationId: Int, dfe: DataFetchingEnvironment): ApplicationView {
+        dfe.graphQlContext.context.enforceSignedIn()
+
+        return transaction {
+            if (mayUpdateApplicationsInRegion(dfe.graphQlContext.context.getAdministrator(), applicationId)) {
```
f1sh1918, line 132: ❌ You have to pass the regionId of the application not the `applicationId` @andrew8er 
So you check if the current user is in the same region as the application and has the correct role

andrew8er: I was just debugging, why I can't approve applications other than the first…

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -6,15 +6,37 @@ import FlipMove from 'react-flip-move'
...
+  {
+    title: 'accepted',
+    status: ApplicationVerificationStatus.Approved,
+  },
+  {
+    title: 'rejected',
+    status: ApplicationVerificationStatus.Rejected,
+  },
+  {
+    title: 'withdrawed',
```
seluianova, line 32:
```suggestion
    title: 'withdrawn',
```
or revoked as an option

andrew8er: There were a lot of this specific typo in the code base, fixed them.

andrew8er:
@f1sh1918 

>    When the application was approved, there should be a toast "Antrag genehmigt". Actually we could also use a state in the application card to indicate the application state directly, since the note about approval only appears if the page was reloaded

Toast was added.

>    Since the application rejection is not impliment yet, i would just hide the button for now, to not confuse the user
    Maybe this will be fixed by updating from main

Button was hidden

>    Not sure about that. It actually does not help any end user since this message is not really understable for them and the tooltip should be deleted

I tried to improve the tooltip, it now reads: "Aus den Antragsdaten kann keine Karte erzeugt werden." 

The error is still pretty vague and should convey a more actionable hint to users IMHO. Any ideas?

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0028_Add_Application_Status.kt
```diff
@@ -0,0 +1,23 @@
...
+        exec(
+            """
+            -- Add column "status"
+            alter table "applications" add column "status" varchar(32) not null default 'Pending'
+                 check ("status" in ('Pending', 'Approved', 'ApprovedCardCreated', 'Rejected'));
+            -- Set "status" to 'ApprovedCardCreated' for rows that have "cardCreated" set to true
+            update "applications" set "status" = 'ApprovedCardCreated' where "cardCreated";
+            -- Remove ”cardCreated" column
+            alter table "applications" drop column "cardCreated";
+            -- Add column "statusResolvedDate"
```
seluianova, line 18: 🙃 personal preference, feel free to ignore:
Smthg like statusUpdatedAt sounds more precise to me, considering that it can be updated more than once

specs/backend-api.graphql
```diff
@@ -73,12 +73,14 @@ type ApplicationVerificationView {
 }
 
 type ApplicationView {
-  cardCreated: Boolean
+  cardCreated: Boolean @deprecated(reason : "Use 'status' instead")
```
seluianova, line 76: should be keep this field ?

andrew8er: The field will be removed in the follow-up PR #2276

backend/src/main/kotlin/app/ehrenamtskarte/backend/migration/migrations/V0028_Add_Application_Status.kt
seluianova: > The field captures the time an application changes from Pending to either Approved or Rejected

this field is also changed when the card is issued. is it intended?

andrew8er: Ah the condition guarding the change to this field was wrong, I updated it, should work now as intended.

seluianova (approved):
Works like a charm 👍

kalairubachandrakumaran-oss:
Tested:
- Two issues created #2485, #2486
- Rest of requirement works fine

Issue #1096, opened 2023-08-22
Improve validity checks for card


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Since we had some issue with checking validity or cards due to timezone issues, we should improve the validity checks like `isCardNotYetValid` and `isCardExpired`.

**Describe the solution you'd like**
- We could use https://pub.dev/packages/clock to test how `isCardNotYetValid` behaves when the time is very close to the startDay.
- The timezone logic in the app should be properly solved. For the expiration day, we did some weird tolerance stuff. This should be removed
- Some proper tests should be added.

seluianova:
> Since we had some issue with checking validity or cards due to timezone issues

@f1sh1918 could you please elaborate more on what exactly the time zone issue was and what behavior should be expected?
to make sure we're on the same page

f1sh1918:
I have no real clue.
I think @michael-markl requested at some point this could be improved and i created a ticket probably out of a Pull request comment

PR #2254, opened 2025-06-07
1096: Improve validity checks for card

seluianova:
### Short Description
Currently we don't consider timezone when checking `isCardNotYetValid` and `isCardExpired`.
This results in 2 issues:
1. if the card’s start date is 20.05.25, but the current time in Germany is 20.05.25 at 00:01, the function isCardNotYetValid() returns false, because the check is performed against UTC time (19.05.25 22:01).
2. for the expiration day, _we did some weird tolerance stuff_ (c)

### Proposed Changes
- add timezone package to the project, refactor validity checks to consider german timezone
- extend tests

### Side Effects
no?

### Testing
I suppose the issues mentioned in the description can't be easily tested unless testing is performed around 00:01, but I've covered these corner cases in the autotests.

We can do regression checks here:
1. Create cards with different start dates (already active, not active yet), check activation and verification 
(_start date is only used in the Nuremberg project_).
3. Create cards with different expiry dates (still active, already expired), check activation and verification 
(_I believe an expired card can only be created on dev env - the corresponding validation in administration module should be commented out_)

### Resolved Issues
Fixes: #1096


frontend/test/card_info_util_test.dart
```diff
@@ -120,5 +154,113 @@ void main() {
...
+        ..fullName = 'Max Mustermann'
+        ..expirationDay = getEpochDaysFromUtcOffset(1)
+        ..extensions = (CardExtensions()
+          ..extensionRegion = (RegionExtension()..regionId = 93)
+          ..extensionBirthday = (BirthdayExtension()..birthday = -365 * 10)
+          ..extensionNuernbergPassId = (NuernbergPassIdExtension()..passId = 99999999));
+      expect(isCardExpired(cardInfo), false);
+    });
+
+    test('should return true if expirationDay is today', () {
```
f1sh1918, line 192: ❓ i guess the expired check in this pr didn't change, but i would expect if the expirationDay is today that my card is still valid and will be expired the day after expirationDay. Just a general idea

f1sh1918 (approved):
Tested on android and ios. works fine 👍

frontend/test/card_info_util_test.dart
seluianova: I think you are right, I will adjust that 👌🏻

seluianova:
Adjusted isCardExpired check: if expiration date is "today" - card is still valid.

frontend/lib/identification/util/card_info_utils.dart
```diff
@@ -22,19 +26,35 @@ extension Hashing on CardInfo {
 }
 
 bool isCardExpired(CardInfo cardInfo) {
-  final expirationDay = _getExpirationDayWithTolerance(cardInfo);
-  return expirationDay != null && expirationDay.isBefore(DateTime.now());
+  final expirationDay = _getExpirationDay(cardInfo);
+  if (expirationDay == null) return false;
+  final now = TZDateTime.from(clock.now(), currentTimezone);
+  return expirationDay.isBefore(TZDateTime(currentTimezone, now.year, now.month, now.day));
```
andrew8er, line 32: 💭 With conditional early returns (i. e. returns in an `if` in the middle of the function), I always put the remaining code in the `else` block. This makes it clear from the indentation, when this code gets executed (and particularly when not). While it is quiet easy to see here, in larger functions it gets hard to spot the `if (...) return ...` in the middle of the code. Sadly, this is actively discouraged by a lint rule in the JS code, and I will propose to drop that rule.

```suggestion
  final expirationDay = _getExpirationDay(cardInfo);
  if (expirationDay == null) {
    return false
  } else {
    final now = TZDateTime.from(clock.now(), currentTimezone);
    return expirationDay.isBefore(TZDateTime(currentTimezone, now.year, now.month, now.day));
  }
```

andrew8er (approved):
Formatting suggestion, otherwise LGTM.

seluianova:
> Adjusted isCardExpired check: if expiration date is "today" - card is still valid till the end of the day.

I forgot to adjust this in the card verification.
Currently, when we perform card verification and the expiration date is today -- > the verification result is card expired

f1sh1918:
Okay then we have to create an issue for that

kalairubachandrakumaran-oss:
> Okay then we have to create an issue for that

> I had the same error message, but only when I was not in the staging environment

seluianova:
> To create a card that is expired, I need help from developers

I think the only option for staging is to create a card with an expiration date = today and test the activation tomorrow.

Issue #2262, opened 2025-06-12
[dev]Print deep links for nuernberg in the console


f1sh1918:
### Describe the Problem

Currently card activation can only be tested with deeplinking for ios simulators.
Therefore we should also get proper deeplinks for the nuernberg project even its not really supported for the project

### Describe Your Preferred Solution

- print the deep links in the console after card creation only for non production environment

PR #2263, opened 2025-06-12
2262: Print deep links nuernberg

f1sh1918:
### Short Description

Since some refactoring deep links will only be printed for bavarian entitlementcard and if confirmation mail is activated in region settings. Since testing cards is way easier with deep links and for ios simulators it can only be tested using them, the deep links should be printed in dev environments in any case to simplify testing. 

### Proposed Changes

- print deep links to the console even for nuernberg project
- add custom scheme deep link example to the docs, so we don't have to add any trusted domain for android (needed for https scheme deep links)

### Side Effects

-

### Testing

- create a card for nuernberg
- activate a card via deep link

### Resolved Issues

Fixes: #2262


andrew8er (approved):

bahaaTuffaha (commented):
I copied the deep-link from the console but it shows the following:

Issue #2267, opened 2025-06-12
index.html uses wrong source paths


f1sh1918:
### Describe the Bug

Currently our administration build is broken because in the `index.html` file full paths were used instead of relative ones

f.e.
```
href="http://localhost:3000/static/css/main.6d27b579.css"
```

### Steps to Reproduce

1. run `npm run build` and check `index.html
2. See wrong paths

### Expected Behavior

paths will be relative f.e. 
```
href="/static/css/main.6d27b579.css"
```
Ensure this applies to css and js files

PR #2219, opened 2025-05-28
Remove cra deps

michael-markl:
### Short Description

This removes the remaining create-react-app dependencies (CRA is dead).
Also updates npm versions according to our semver rules.

### Proposed Changes

This is a continuation of #1026 . A few left-over dependencies of CRA were still there (namely react-dev-utils and babel-preset-react-app). I removed these dependencies (most of their functionality was not necessary, and / or outdated). I mirrored the babel-preset-react-app in our own config code and updated it (mostly removed plugins that are now part of `@babel/preset-env` ).
This way, we no longer depend on deprecated or vulnerable packages (at least as reported by npm :D things like the airbnb eslint package are also quasi dead^^ ).

### Side Effects

Hopefully none.

### Testing

Test all the things in administration :D

### Resolved Issues

Fixes: #810

PR #2269, opened 2025-06-12
2267: Use '/' as public URL instead of localhost

michael-markl:
### Short Description

Use '/' as the root path for referencing any (JS) assets.

Fixes a bug introduced in #2219 .

### Testing

Build the production app and verify that the index.html references the JS asset with "/" instead of "http://localhost:3000/".

### Resolved Issues

Fixes: #2267


michael-markl:
@f1sh1918 Let me know if I should rather base this on the release branch.

Issue #2273, opened 2025-06-16
E-Mail service does not work [staging]


f1sh1918:
### Describe the Bug

As a user i want to be able to get my password reset mail, to reset my password

### Steps to Reproduce

1. [pw reset](https://staging.bayern.ehrenamtskarte.app/forgot-password)
2. Type in your account mail address
3. Click "Passwort zurücksetzen

### Expected Behavior

The mail will be send

### Additional Information

<img width="670" alt="Image" src="https://github.com/user-attachments/assets/48fa4d84-a7da-48cb-89b6-9d8ac79c0928" />

Backend-Log:
```
Jun 16 07:50:34 entitlementcard-test.tuerantuer.org backend[3422338]: [JettyServerThreadPool-59] ERROR app.ehrenamtskarte.backend.mail.Mailer - Failed to send email [Subject: 'Passwort Zurücksetzen'], reason: Third party error
```

PR #2274, opened 2025-06-16
2273: upgrade-java-mail dependency

f1sh1918:
### Short Description

We had some issues with smtp auth with the recent java simple mail version.
It has turned out that it was related to deprecated smtp host and will be fixed with the next server update including this pr
https://git.tuerantuer.org/DF/salt/pulls/361

Mind that this salt pr has to be merged before we release that

### Proposed Changes

- update version
- use correct smtp host

### Side Effects

- none

### Testing

Please adjust your local smtp config (local backend config.yml) for the projects and use
    smtp:
      host: mail.your-server.de
      port: 587
      username: noreply@ehrenamtskarte.app
      password: <passbolt>
 Please also test mailing service for other projects like nuernberg and koblenz (pw reset f.e.)
      
Find the password here: https://passbolt.tuerantuer.org/app/passwords/view/63908d6c-4325-4180-aa26-a0274c8ffcbe
### Resolved Issues

Fixes: #2273

### DIFF
--- a/backend/src/main/resources/config/config.yml
+++ b/backend/src/main/resources/config/config.yml
@@ -28,7 +28,7 @@ projects:
     timezone: "Europe/Berlin"
     selfServiceEnabled: false
     smtp:
-      host: mail.ehrenamtskarte.app
+      host: mail.your-server.de
       port: 587
       username: OVERRIDE_IN_LOCAL_CONFIG
       password: OVERRIDE_IN_LOCAL_CONFIG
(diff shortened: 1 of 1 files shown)


seluianova:
@f1sh1918 Is it important to use this particular server for dev testing? mail.your-server.de
or it can be any server ?

f1sh1918:
> @f1sh1918 Is it important to use this particular server for dev testing? mail.your-server.de or it can be any server ?

Well it was just the issue that we didn't use the recommended hetzner smtp config which can be checked in their docs.
If you use the correct smtp config from your mail provider, it will not fail. So i think it was just a config issue on our side and i don't think we need additional docs tbh

andrew8er (approved):
Sending mail works.

When creating a new user with my email address, I get two to three invitation mails however. Can anyone confirm this?

andrew8er:
@seluianova 

>  Is it important to use this particular server for dev testing? mail.your-server.de
or it can be any server ?

You could also use a different server, but most mail servers are configured to only route mails with specific senders (otherwise you'd have a public spam relay), so you would need to adjust your sender address accordingly.

Issue #2176, opened 2025-05-12
Replace NPM dep `react-flip-move `


andrew8er:
### Describe the Problem

The dependency [react-flip-move](https://github.com/joshwcomeau/react-flip-move) is [not maintained anymore](https://github.com/joshwcomeau/react-flip-move/issues/233), and apparently [does not work with function components](https://github.com/digitalfabrik/entitlementcard/blob/ed497bb2a6e7beeb91c12996b0422f66ed6c02ca/administration/src/bp-modules/applications/ApplicationsOverview.tsx#L33), it should be replaced.

### Describe Your Preferred Solution

Use another list animation library, possibly [react-flip-toolkit](https://github.com/aholachek/react-flip-toolkit).

PR #2280, opened 2025-06-17
2176 Replace `react-flip-move` with `motion`

andrew8er:
### Short Description

Replace package `react-flip-move` with [motion](https://github.com/motiondivision/motion) (formerly known as `framer-motion`) for animating lists.

### Side Effects

None

### Testing

- Visit http://localhost:3000/cards/add to add and remove card items
- Visit http://localhost:3000/applications and delete applications

### Resolved Issues

Fixes: #2176

### DIFF
--- a/administration/package.json
+++ b/administration/package.json
@@ -33,7 +33,6 @@
     "notistack": "^3.0.2",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
-    "react-flip-move": "^3.0.5",
     "react-i18next": "^15.5.2",
     "react-router": "^7.6.1",
     "styled-components": "^5.3.11",
--- a/administration/package.json
+++ b/administration/package.json
@@ -28,6 +28,7 @@
     "graphql": "^16.11.0",
     "i18next": "^23.16.8",
     "localforage": "^1.10.0",
+    "motion": "^12.18.1",
     "normalize.css": "^8.0.1",
     "notistack": "^3.0.2",
     "react": "^18.3.1",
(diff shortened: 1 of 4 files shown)


bahaaTuffaha (approved):
Nice work... tested on Firefox and Chrome ✅
I have two points:

1) How about `react-transition-group` package its size is way smaller
2)  This is a nitpick 🙃: I noticed that the button not centered after closing the card creation window.

https://github.com/user-attachments/assets/07ee8297-4c4b-4cf6-a251-a961e4e6892c

andrew8er:
@bahaaTuffaha 

>    How about react-transition-group package its size is way smaller

To be honest, the reason I choose this library is because [the one I initially intended to use](https://github.com/aholachek/react-flip-toolkit) had a crappy API and I used this one before and knew it worked and is also very popular. I would need to measure how much code the usage really adds, since tree-shaking removes everything unused. It is not possible to gauge that from the package size alone.

>    This is a nitpick 🙃: I noticed that the button not centered after closing the card creation window.

Do you mean the "Karte hinzufügen" button? Can you give me a screenshot?

bahaaTuffaha:
> Do you mean the "Karte hinzufügen" button? Can you give me a screenshot?

I left a video... it looks like GitHub not always playing posted videos 😐️... thanks Andy for the comparison.

andrew8er:
I aligned the cards from the card creation form at the top of the grid rows now, since the items change their height after interaction (because of the appearing error message). This made it hard to hit the close button, since the whole item jumps upwards a little.

bahaaTuffaha (approved):
Tested again looks good ✅
You can do the same flipMove animation I looked at the source-code and I came up with this :
```
const animationProperties: MotionNodeAnimationOptions = {
  initial: { opacity: 0, scale: 0 },
  animate: { opacity: 1, scale: 1 },
  exit: { opacity: 0, scale: 0 },
  transition: {
    duration: 0.3,
  },
}
```

kalairubachandrakumaran-oss:
Tested ✅
image
image

Issue #1738, opened 2024-11-04
Improve csv import column descriptions


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
The CSV card import description is ambiguous. The mentioned column format does not inform the user whether the column is actually required or optional.

**Describe the solution you'd like**
It should be clear what columns are required and which are optional.
This can be done by either:
- marking required columns with an asterisk `(*)`
- marking optional columns with an `(optional)`
- adding a separate bullet point `Optionale Spalten` or similar

Refactoring the card config to avoid having three separate config arrays (i.e.   extensionColumnNames, extensions, extensionRequirement). Instead, we could do something like:
``` ts
  extensions: [{
    name: 'Kartentyp',
    extension: BavariaCardTypeExtension,
    required: true,
  },
  ...
]
```

**Describe alternatives you've considered**
Leave it as it is.

**Additional context**
![image](https://github.com/user-attachments/assets/d76bba8d-3ee9-4a48-bc1e-1237a351c54f)

This was brought up here: https://github.com/digitalfabrik/entitlementcard/pull/1715#pullrequestreview-2412788115

seluianova:
> Describe alternatives you've considered

As an alternative, maybe we should just expect "MailNotification" column to be presented in the CSV
It can be empty if the value is not required

PR #2286, opened 2025-06-23
1738:  Checked for isMandatory for each extension and added an asterisk

bahaaTuffaha:
### Short Description

The CSV card import description is ambiguous. The mentioned column format does not inform the user whether the column is actually required or optional.

### Proposed Changes

- I went for the asterisk option.
- From cardConfig I took `nameColumnName` and `expiryColumnName` these two are mandatory headers then added to them any isMandetory extentions from true from `cardConfig.extensions`.
- Added asterisk to required headers.
- I did the same for stores at `StoresRequirementsText.tsx`.

### Side Effects

None.

### Testing

go to:
- http://localhost:3000/cards/import (bayern), (nuernberg).
- http://localhost:3000/stores/import (koblenz), (nuernberg)

### Resolved Issues

Fixes: #1738

### DIFF
--- a/administration/src/bp-modules/cards/ImportCardsRequirementsText.tsx
+++ b/administration/src/bp-modules/cards/ImportCardsRequirementsText.tsx
@@ -1,7 +1,8 @@
-import React, { ReactElement } from 'react'
+import React, { ReactElement, useContext, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
+import { ProjectConfigContext } from '../../project-configs/ProjectConfigContext'
 import { ENTRY_LIMIT, FILE_SIZE_LIMIT_MEGA_BYTES } from './constants'
 
 const RequirementsList = styled.ul`
@@ -18,7 +19,23 @@ const ImportCardsRequirementsText = ({
   csvHeaders,
   isFreinetFormat = false,
 }: ImportCardsRequirementsProps): ReactElement => {
+  const projectConfig = useContext(ProjectConfigContext)
+  const { card: cardConfig } = projectConfig
   const { t } = useTranslation('cards')
+
+  const requiredHeaders = useMemo(() => {
+    const headers = [cardConfig.nameColumnName, cardConfig.expiryColumnName]
+    cardConfig.extensions.forEach((extension, index) => {
+      const header = cardConfig.extensionColumnNames[index]
+      if (extension.isMandatory && header) {
+        headers.push(header)
+      }
+    })
+    return headers
+  }, [cardConfig])
+
+  const decoratedHeaders = csvHeaders.map(header => (requiredHeaders.includes(header) ? `${header}*` : header))
+
   return (
     <RequirementsList>
       <li>{t('maxFileSize', { maxFileSize: FILE_SIZE_LIMIT_MEGA_BYTES })}</li>
@@ -26,7 +43,7 @@ const ImportCardsRequirementsText = ({
       <li>
         {t('maxNumberOfEntries')}: {ENTRY_LIMIT}
       </li>
-      <li>{isFreinetFormat ? t('minColumnsForFreinet') : `${t('columnFormat')}:  ${csvHeaders.join(', ')}`}</li>
+      <li>{isFreinetFormat ? t('minColumnsForFreinet') : `${t('columnFormat')}:  ${decoratedHeaders.join(', ')}`}</li>
       <li>{t('dateFormatHint')}</li>
     </RequirementsList>
   )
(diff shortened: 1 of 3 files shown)


administration/src/bp-modules/cards/ImportCardsRequirementsText.tsx
```diff
@@ -18,15 +19,31 @@ const ImportCardsRequirementsText = ({
...
+  const requiredHeaders = useMemo(() => {
+    const headers = [cardConfig.nameColumnName, cardConfig.expiryColumnName]
+    cardConfig.extensions.forEach((extension, index) => {
+      const header = cardConfig.extensionColumnNames[index]
+      if (extension.isMandatory && header) {
+        headers.push(header)
+      }
+    })
+    return headers
+  }, [cardConfig])
```
andrew8er, line 35: A functional approach (which I find easier to understand) would be this:
```suggestion
  const requiredHeaders = useMemo(
    () => [
      projectConfig.card.nameColumnName,
      projectConfig.card.expiryColumnName,
      ...projectConfig.card.extensions.flatMap((extension, index) => {
        const header = projectConfig.card.extensionColumnNames[index]
        return extension.isMandatory && header ? [header] : []
      }),
    ],
    [projectConfig.card]
  )
```

Also I think we re overusing `useMemo()` quiet a bit in our code base. For example, here it is quiet unnecessary, since `cardConfig` changes almost never in practical use. We could use a plain function here.

administration/src/bp-modules/stores/StoresRequirementsText.tsx
```diff
@@ -1,7 +1,8 @@
-import React, { ReactElement } from 'react'
+import React, { ReactElement, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
+import { StoresFieldConfig } from '../../project-configs/getProjectConfig'
```
andrew8er, line 5: This could be
```suggestion
import type { StoresFieldConfig } from '../../project-configs/getProjectConfig'
```

administration/src/bp-modules/stores/StoresRequirementsText.tsx
```diff
@@ -10,17 +11,18 @@ const RequirementsList = styled.ul`
 `
 
 type ImportCardsRequirementsProps = {
-  header: string[]
+  header: StoresFieldConfig[]
 }
 
 const StoresRequirementsText = ({ header }: ImportCardsRequirementsProps): ReactElement => {
+  const headers = useMemo(() => header.map(field => (field.isMandatory ? `${field.name}*` : `${field.name}`)), [header])
```
andrew8er, line 18: Again `useMemo()` is complete overkill IMHO.

andrew8er (changes requested):
Some unnecessary `useMemo()` usages.

I noticed that this is a larger problem in this code base, way too many small, easily computable things are memoized.

A good explainer why this is bad is this article: https://blog.logrocket.com/when-not-to-use-usememo-react-hook/

administration/src/bp-modules/cards/ImportCardsRequirementsText.tsx
```diff
@@ -18,15 +19,31 @@ const ImportCardsRequirementsText = ({
...
 }: ImportCardsRequirementsProps): ReactElement => {
+  const projectConfig = useContext(ProjectConfigContext)
+  const { card: cardConfig } = projectConfig
   const { t } = useTranslation('cards')
+
+  const requiredHeaders = useMemo(() => {
+    const headers = [cardConfig.nameColumnName, cardConfig.expiryColumnName]
+    cardConfig.extensions.forEach((extension, index) => {
+      const header = cardConfig.extensionColumnNames[index]
+      if (extension.isMandatory && header) {
```
f1sh1918, line 30: 🔧 i would prefer explicit null check here
```suggestion
      if (extension.isMandatory && header !== null) {
```

f1sh1918 (commented):
Looks good already! 👍 
A general discussion could be if we rename the `MailNotification` column header, since its not really self explaining. I think `E-Mail-Adresse` would be better:
But i have to check if this queryParam is used somewhere for card creation from an external system (Freinet).
So this can be done in a separate ticket.

I would love to see a unit test here :)

administration/src/bp-modules/cards/ImportCardsRequirementsText.tsx
f1sh1918: i agree. Even there is not state in this function it could also be outside of the component to avoid unneeded rerendering. But just removing `useMemo` should be sufficient

andrew8er (approved):

michael-markl:
> I noticed that this is a larger problem in this code base, way too many small, easily computable things are memoized.

I don't think it's overused; there seem to be only 12 uses of `useMemo` and maybe two or three of them do not make any sense.

Issue #2294, opened 2025-06-24
Create "Weitere Optionen" button in application detail view


connium:
### Describe the Problem

With the buttons to approve and reject applications and the PDF export button, there are already 3 buttons. With the [CSV export feature](#1752), we will add another button. This makes it hard for users to find the appropriate action.

### Describe Your Preferred Solution

The primary actions for an application are:
- approve / create card
- reject
- delete

The less important actions, which also aren't used by all agents, are:
- print as PDF
- export as CSV

Thus, we should hide these less important buttons behind a "Weitere Optionen" button/dropdown, as can be seen in this 📌[Figma](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=1003-2292&t=k2FkP5A4nDxBQkg3-0).

### Describe Possible Alternatives

Show all buttons directly below the application.

### Related Issues

#1752

PR #2297, opened 2025-06-25
2294: Other options button

f1sh1918:
### Short Description

Since we are reaching limits how many buttons/actions we can display next to each other, we introduce are more actions menu button

### Proposed Changes

- add base component `BaseMenu`, 
- add typography for button, 
- replace button labels by Typography element for ApplicationCard
- add required translations
- add unit test for `BaseMenu`
- move pdf print to other options menu

### Side Effects

- none

### Testing

1. Start bavaria project
2. Go to "Anträge"
3. Open the application accordion "Mehr Anzeigen" and click on "Weitere Option"
4. PDF Print should be shown as an option
5. If you click the button pdf print should be shown
6. If you print/close the print menu, the "Weitere Option" menu should be closed

### Resolved Issues

Fixes: #2294

### DIFF
--- a/administration/src/bp-modules/applications/ApplicationCard.tsx
+++ b/administration/src/bp-modules/applications/ApplicationCard.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/destructuring-assignment */
 import { MutationResult } from '@apollo/client'
-import { CreditScore, Delete, ExpandMore, InfoOutline, Print, Warning } from '@mui/icons-material'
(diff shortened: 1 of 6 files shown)


f1sh1918:
We may extend this component a bit if we want to use it for our main user menu

connium:
> We may extend this component a bit if we want to use it for our main user menu

I'm pretty sure the navigation bar will receive major changes. So I wouldn't do anything about that at the moment.

f1sh1918:
opened #2301 instead, since i took the wrong issue number on branch

Issue #2144, opened 2025-05-05
Store issued card in Freinet-Online when a card was created


connium:
### Describe the Problem

As a region administrator or a region manager,
I want to add issued volunteer cards to Freinet-Online
so that I can manage volunteers and issued cards there.

### Describe Your Preferred Solution

Given that I am logged in as a region administrator or a region manager
and data transfer to Freinet-Online is enabled,
when I approve a pending application,
then information about the issued volunteer card is stored at the corresponding person in Freinet-Online
and a protocol entry will added which includes the mail address of the current user.

Given that I am logged in as a region administrator or a region manager
and data transfer to Freinet-Online is disabled,
when I approve a pending application,
then no communication with Freinet-Online takes place.

### Acceptance Criteria
- within #2141 or #2142 you received a `userId` from freinet
- send a post request to create a card in freinet
- send a success message to the user (toast)
- send error message to the user if not successful

Example:
- EndpointUrl: https://freinet-online.de/api/input/v3/ehrenamtskarte
- Json Body:
```
{
    "data": {
        "init": {
            "agencyID":  <agencyId>,
            "accessKey": "<accessKey>",
            "author": "TuerAnTuer",
            "author_mail": "berechtigungskarte@tuerantuer.org"
        },
        "karten_farbe": 1, // 1= Blue, 2= Gold
        "gueltig_bis": "2026-02-01", // This has only to be set for blue card
        "user_id": 1276621,
        "digital_card": true // Always true
    }
}
```

#blocked by #2142 , #2143

PR #2299, opened 2025-06-26
2144: Store card information freinet

f1sh1918:
### Short Description

When a service user creates a new card, the card information should be transferred to freinet

### Proposed Changes

- pass required card information to the `sendApplicationAndCardDataToFreinetMutation`
- connect new endpoint `ehrenamtskarte` (FreinetApi)
- map the relevant data to the required format
- send card data to freinet for existing users and created users
- add error handling

### Side Effects

- likely none

### Testing
1. Create an application for Freinet Demo region
2. Create an region admin for Freinet Demo region
3. Login and go to "Region Verwalten" and enable "Freinet Datenübermittlung"
4. Go to "Anträge verwalten" and create a card from application
5. Login to Freinet and check that applicant and card data were transferred
6. Test "Standard" and "Golden Cards"
7. Check that card data transfer works for existing users and newly created users
Freinet Login information: https://passbolt.tuerantuer.org/app/passwords/view/692b0bc6-7974-4dab-a69e-1979949758fa
Note that: "Protocol" does not exists for this endpoint so we can not create a separate protocol entry
### Resolved Issues

Fixes: #2144


backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/schema/types/FreinetCard.kt
```diff
@@ -0,0 +1,19 @@
+package app.ehrenamtskarte.backend.freinet.webservice.schema.types
+
+import app.ehrenamtskarte.backend.exception.webservice.exceptions.FreinetCardDataInvalidException
+
+const val CARD_TYPE_STANDARD = "Standard"
+const val CARD_TYPE_GOLD = "Goldkarte"
+
+data class FreinetCard(
+    val id: Int,
```
seluianova, line 9: ❓ id field is not used ?

f1sh1918: Removed it

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
```diff
@@ -23,8 +25,13 @@ import org.slf4j.LoggerFactory
 class FreinetApplicationMutationService {
     private val logger = LoggerFactory.getLogger(FreinetApplicationMutationService::class.java)
 
-    @GraphQLDescription("Send application info to Freinet")
-    fun sendApplicationDataToFreinet(applicationId: Int, project: String, dfe: DataFetchingEnvironment): Boolean {
+    @GraphQLDescription("Send application and card information to Freinet")
+    fun sendApplicationAndCardDataToFreinet(
+        applicationId: Int,
+        project: String,
+        freinetCards: List<FreinetCard>,
```
seluianova, line 32: ❓why do we have list as param here? 
we can issue the card multiple times for one application, but we can't (and shouldn't) do it in one request, right?

seluianova:
General question to the feature:
When I click "Karte erstellen" in the application, card creation form is opened.
The card creation form contains pre-filled card data from the application and a button to add one more card.
When I add another card, **both cards are sent to freinet and bound to the same user.**
But is this expected and obvious to administrators? It seems to me that there is no clue that all cards on a form should be associated with the same user...
Because we use the same form to create cards without application, and there you can create cards for different users...
image

connium:
I'd expect that only a single card is created from an application. I'd be happy to hear the opinion of others because I haven't talked to many users so far.

If there is no scenario where 1 application leads to multiple cards, I think your suggested solution fits perfect.

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationService.kt
seluianova: > It would actually work and there is no harm in doing it

the problem I see is that it's not obvious. we literally allow to enter different names for different cards on the form. how to the user should understand that all of them will be linked to the same user in freinet?

f1sh1918: Well but then we should not delete the query params immediately after forwarding from application. Because if there is any reload of the page the related applicant data will be lost and restriction to one card will not work

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetApi.kt
```diff
@@ -135,13 +133,52 @@ class FreinetApi(private val host: String, private val accessKey: String, privat
...
+                logger.devInfo("Successfully created person in freinet: ${response.bodyAsText()}")
+                FreinetPersonCreationResultModel(
+                    true,
+                    response.bodyAsChannel().toInputStream().use {
+                        objectMapper.readTree(it)
+                    },
+                )
             } catch (e: Exception) {
                 logger.devWarn("Error creating person in freinet $e")
                 throw FreinetApiNotReachableException()
```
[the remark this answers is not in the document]
andrew8er: Then throwing a simple `Exception` (or whatever Javalin has as a standard for generic 500 codes) should do as well. On top of that, we should have a generic exception handler in place that maps exceptions to status codes and hides/shows error messages based on the run environment. Maybe for another issue…

administration/src/bp-modules/cards/hooks/useCardGenerator.ts
```diff
@@ -70,9 +71,10 @@ const useCardGenerator = ({ region, initializeCards = true }: UseCardGeneratorPr
   const appToaster = useAppToaster()
   const { t } = useTranslation('cards')
 
-  const [sendToFreinet] = useSendApplicationDataToFreinetMutation({
+  const [sendToFreinet] = useSendApplicationAndCardDataToFreinetMutation({
     onCompleted: data => {
-      if (data.sendApplicationDataToFreinet === true) {
+      if (data.sendApplicationAndCardDataToFreinet === true) {
+        console.log('test')
```
seluianova, line 77:
```suggestion
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetApi.kt
```diff
@@ -135,13 +132,63 @@ class FreinetApi(private val host: String, private val accessKey: String, privat
                     contentType(ContentType.Application.Json)
                     setBody(requestBody)
                 }
-                val body = response.bodyAsText()
-                logger.devInfo("Successfully created person in freinet: $body")
+                logger.error("Successfully created person in freinet: ${response.bodyAsText()}")
```
seluianova, line 135:
```suggestion
                logger. devInfo("Successfully created person in freinet: ${response.bodyAsText()}")
```
why changed to error?

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetApi.kt
```diff
@@ -135,13 +132,63 @@ class FreinetApi(private val host: String, private val accessKey: String, privat
...
                 true
             } catch (e: Exception) {
-                logger.devWarn("Error creating person in freinet $e")
+                logger.error("Error creating card in freinet $e")
                 throw FreinetApiNotReachableException()
             }
         }
     }
+
+    private fun ObjectNode.addFreinetInitInformation(agencyId: Int, accessKey: String, moduleName: String) {
```
seluianova, line 184:
```suggestion
    private fun addFreinetInitInformation(agencyId: Int, accessKey: String, moduleName: String) {
```

seluianova (approved):
lgtm, just minor suggestions 👍

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetApi.kt
f1sh1918: yes. thx :)

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetApi.kt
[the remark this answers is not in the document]
f1sh1918: Okay thx for this detailed information. I think i now got the idea behind that.

Hm the only thing i didn't get was the issue with `ApplicationDataIncompleteException` i mean should that also result in a `500` ?  I think the reason why this might be thrown is very specific. So why not throw an specific error?

kalairubachandrakumaran-oss:
Currently not testable (from the Testing Team) ℹ️

Issue #1754, opened 2024-11-06
Trigger approval mail for organizations manually


dkehne:
### Describe the Problem

Sometimes approval processes take time and email get "organizational" lost on the organizations site. So there should be a possibility for the administrative processors to trigger the email again. 

As a region manager,
I want to trigger the approval email to an organization
so that they receive a reminder in case the original message was overlooked or lost.

### Describe Your Preferred Solution

Given that I am logged in as a region manager or a region administrator,
when I navigate to the page for incoming applications
and there is a pending application
and I expand the card of the application to show its details
and the application has a pending approval by an organization,
then I can resend the approval mail to the organization.

Given that there is a pending application,
when an organization has approved or rejected the volunteer work of the applicant,
then there is no option to resend the approval mail to the organization.

Given that there is a pending application,
when there are multiple organizations listed for approval,
then I can only resend the approval mail to organizations, that have neither confirmed nor rejected the volunteer work of the applicant.

Given that there is a pending application
and the approval by an organization is pending,
when I resend the approval mail,
then the approval mail is send again to the email address of the organization that was entered in the application
and I get a hint, that the email has been sent
and the option to resend is disabled unless I reload the page to avoid multiple emails being sent.

### Describe Possible Alternatives

We could also add the possibility to manually set up a "new/other" mail in that sending process.

### Additional Information

- A design will be provided by @hauf-toni

f1sh1918:
@hauf-toni how is the progress with the design. I think we only need an additional button

f1sh1918:
@connium i think to say the button can not be triggered for one minute makes it unnecessary complicated. I would just say the button will be disabled after successful sending but if you refresh the page it will be enabled again

hauf-toni:
Design can be found 📌 [here](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=1790-509&t=6zdSyykmZzxd4Fa0-1).

connium:
Discussed with @hauf-toni : Deviating from the design, the timestamp of the last e-mail is not displayed, and the button is only locked until the page is reloaded.

seluianova:
@connium should the email be any different or just the same as the first one? Maybe with "Erinnerung" in the title or something like that?

connium:
@seluianova for now I'd just send the same email as the first one without any modifications. I don't think that the mention of "Erinnerung" would change anything for the recipients.

PR #2300, opened 2025-06-26
1754: Trigger approval mail for organizations manually

seluianova:
### Short Description
Add button to resend approval request to the organization.
📌 [Design](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=1790-509&t=6zdSyykmZzxd4Fa0-1)

### Proposed Changes
- add sendApprovalMailToOrganisation mutation
- add resent approval request button to the verification view
- the button is only shown when the verification status is pending approval
- after clicking the button is disabled until the page is reloaded
- _not requested in design_, but I also change the button text to "Anfrage erneut gesendet" after the click (feel free to object)

### Side Effects
no?

### Testing
- Create an application that requires verification
- Log in as a region admin, check the new button in the application view
- Check that an approval request is sent when the button is clicked

image

### Resolved Issues
Fixes: #1754


administration/src/bp-modules/applications/components/VerificationListItem.tsx
```diff
@@ -59,6 +90,17 @@ const VerificationListItem = ({ verification }: VerificationListItemProps): Reac
           </tr>
         </tbody>
       </table>
+      {authData !== null && status === VerificationStatus.Pending && (
```
seluianova, line 93: I wanted to hide the button on the applicant view (/antrag-einsehen) but not sure what was the optimal way

f1sh1918 (commented):
It already looks really good 👍 
I found an issue that when the application was withdrawed by the applicant. The retrigger button will still be shown.
Not sure if that was defined in the requirements but i think it shouldn't be possible to trigger that in that case, i think.

Rest looks good. Also good refactoring in the backend!

image

Update: Since this is a general issue that the withdrawn status will probably not set as application status. This can also be done in a separate issue @seluianova, since the check for status "Pending" should then be correct.

administration/src/bp-modules/applications/components/VerificationListItem.tsx
seluianova: sounds good. thanks for the idea, applied

bahaaTuffaha (approved):
Great work!
Tested on Chrome and Firefox ✅

kalairubachandrakumaran-oss:
For open requests, can the request be resent (/)
The request can be resent every time the page is refreshed. Is that expected?
image

PR #2301, opened 2025-06-26
2294: Other options

f1sh1918:
### Short Description

Since we are reaching limits how many buttons/actions we can display next to each other, we introduce are more actions menu button

### Proposed Changes

- add base component `BaseMenu`, 
- add typography for button, 
- replace button labels by Typography element for ApplicationCard
- add required translations
- add unit test for `BaseMenu`
- move pdf print to other options menu

### Side Effects

- none

### Testing

1. Start bavaria project
2. Go to "Anträge"
3. Open the application accordion "Mehr Anzeigen" and click on "Weitere Option"
4. PDF Print should be shown as an option
5. If you click the button pdf print should be shown
6. If you print/close the print menu, the "Weitere Option" menu should be closed

### Resolved Issues

Fixes: #2294

### DIFF
--- a/administration/src/bp-modules/applications/ApplicationCard.tsx
+++ b/administration/src/bp-modules/applications/ApplicationCard.tsx
@@ -337,7 +337,7 @@ const ApplicationCard = ({
   )
 
   const otherOptionsContainerWidth = 170
-  const otherOptionsItemHeight = 40
+  const otherOptionsItemHeight = 36.5
   const menuItems: MenuItemType[] = [
     {
       name: t('exportPdf'),


seluianova:
@f1sh1918 could you please resolve conflicts, I would test this with the merged rejection button

f1sh1918:
@seluianova done

administration/src/util/translations/de.json
```diff
@@ -203,7 +203,6 @@
     "applicationApprove": "Antrag bestätigen",
     "applicationReject": "Antrag ablehnen",
     "applicationApprovedToastMessage": "Der Antrag wurde genehmigt.",
-    "applicationRejectedToastMessage": "Der Antrag wurde abgelehnt.",
```
bahaaTuffaha, line 206: I think this is deleted by mistake.

administration/src/mui-modules/base/__tests__/BaseMenu.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const menuButton = getByRole('button', { name: 'Test Menu' })
+
+    fireEvent.click(menuButton)
+    const menuItem = getByText('Print PDF')
+    fireEvent.click(menuItem)
+
+    const menu = queryByRole('menu')
+    expect(menu).toBeNull()
+  })
+})
```
bahaaTuffaha, line 77: If I put the menu expect statement before clicking on menuItem it will get a pass 🫣.
something wrong here: ex.
```
  it('should close menu after clicking menu item', () => {
    const { getByRole, getByText, queryByRole } = render(<BaseMenu {...defaultProps} />)
    const menu = queryByRole('menu')

    expect(menu).toBeNull()
    const menuButton = getByRole('button', { name: 'Test Menu' })

    fireEvent.click(menuButton)
    expect(menu).toBeNull()
    const menuItem = getByText('Print PDF')
    fireEvent.click(menuItem)
  })
```

bahaaTuffaha (approved):
Nice work... tested on Chrome and Firefox ✅
Left some comments...

administration/src/mui-modules/base/__tests__/BaseMenu.test.tsx
f1sh1918: Well i dunno why this is the case but if you move the menu item before the click menu button, it will not be found (which is correct). Not sure what happens in mui under the hood. Maybe the menu is in the dom but not visible.
The important thing is that the items are not visible before clicking the "menuButton"

f1sh1918: Yes i guess this is due to this is inside a modal and its always difficult to check that.
I spent some time in investigating and even with the suggestion here:
https://stackoverflow.com/questions/69889380/how-to-test-material-ui-popover-close-implementation
it doesn't work.
If you want i can remove the test. The component actually works and i won't spend more time in unit testing here

seluianova:
the height of the buttons is slightly unbalanced (not reproduced on main)
image

seluianova:
🙃 and one more nit-pick level issue:
the title and the icon of new button are misaligned.
on the other buttons we have the same space on the top / bottom / left / right.
but here there is more space on the left and right:
image

seluianova:
🙃 I believe the font of “Weitere Optionen” and “PDF exportieren” is different. I can see the different space between the letters, it's bigger in the drop down item:
image

P.S. feel free to object to this level of nitpicking 😀
when I see such inconsistencies, they bother me. but I don't mind to leave it as is.

administration/src/mui-modules/base/__tests__/BaseMenu.test.tsx
bahaaTuffaha: I will give it a look again if I can find a solution

f1sh1918:
> the height of the buttons is slightly unbalanced (not reproduced on main - all buttons have the same height there) image

This is not an issue of my pr. The Create card button (used as anchor link) is too small (not 40px) feel free to create a bug ticket for that :)

In general i'm super fine with adding nit-picks but i also think it would be ok to approve the pr if there are only nit-picks.
What do you think? @seluianova

administration/src/mui-modules/base/__tests__/BaseMenu.test.tsx
bahaaTuffaha: I think you need to apply the same for "should call onClick handler when menu item is clicked" test.

andrew8er:
I'm probably a bit late, but would'n it have been way more effective to just adjust the MUI `MuiButton` style's typography instead of defining a new typography style and applying it manually everywhere?

Something like this:

```ts
export const theme = createTheme({
  /* ... */
  components: {
    MuiButton: {
      styleOverrides: {
        root: {
          textTransform: 'none',
          fontSize: 14,
          fontWeight: 600,
          variants: [
            {
              props: { variant: 'contained', color: 'primary' },
              style: {
                '&:hover': {
                  color: '#ffffff',
                },
              },
            },
          ],
        },
      },
    },
  /* ... */
})
```

kalairubachandrakumaran-oss:
Test result:
- PDF Print should be shown as an option (/)
- If you click the button pdf print should be shown (/)
- If you print/close the print menu, the "Weitere Option" menu should be closed (/)
https://github.com/user-attachments/assets/b55bf2e6-6158-4324-9035-d64a392447d2

Issue #2298, opened 2025-06-25
Hovering Karte erstellen button makes button text invisible


f1sh1918:
### Steps to Reproduce

1. Go to applications, accept application
2. Hover over "Karte erstellen"
3. See error

<img width="509" alt="Image" src="https://github.com/user-attachments/assets/3d92c2a5-5304-44c0-9c8d-e9a3fd64f9df" />

### Expected Behavior

text color remains visible

kalairubachandrakumaran-oss:
Fixed :
https://github.com/user-attachments/assets/11a8d97b-1e61-48fe-b70d-86ff640d7083

PR #2307, opened 2025-06-30
2298: Creat button hover fix

f1sh1918:
### Short Description

If our mui button is used as an anchor tag, the button text gets invisible if the user hovers on it

### Proposed Changes

- override the hover color for the particular color palette and variant

### Side Effects

- should be none

### Testing

1. Accept an application in the application list
2. Hover over the card creation button
3. Text should remain visible

### Resolved Issues

Fixes: #2298


seluianova (approved):
lgtm!

connium:
> [...] (the hover style comes from an unscoped blueprint CSS).

Does this mean we should a comment here to remove this overwrite as soon as blueprint is gone?

Issue #1752, opened 2024-11-06
export application data to CSV


dkehne:
**Is your feature request related to a problem? Please describe.**
It will still take some time before the Freinet interface #1751 is ready, and at the same time there are cities and districts that do not rely on Freinet for master data management but use a different system.
It should also be possible to export applications in a machine-readable format. Currently this is only possible as a PDF.

**Describe the solution you'd like**
It should also be possible to download each application in csv format at the touch of a button.

**Additional Context**
We already have a "Download PDF"-Button, so maybe it is useful to ask @hauf-toni for a design to merge both functions into one button/menu.

hauf-toni:
Design proposal can be found [📌 here](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=1003-2292&t=dNJn8iXuBM4mDzCv-1).

f1sh1918:
@hauf-toni design is good to go. i will remove ui/ux flag.
The only issue is that we currently didn't find anyone who really wants this feature even it was highly ranked and the community meeting last year. @connium  wanted to discuss which fields should be exported since it will be probably a mess if we export all fields

connium:
We will export more or less the same data, that we send to Freinet Online.

- first name
- family name
- street
- house number
- address suffix
- zip code
- city
- email address
- phone number
- birth date
- card type (blue/gold)
- application date
- card expiry date

PR #2314, opened 2025-07-01
1752: Export application csv

f1sh1918:
### Short Description

This feature enables the user to export application data to csv

### Proposed Changes

- add `applicationDataHelper` for getting the particular application data for the csv
- add `exportApplicationToCsv` that export the data to csv and triggers download
- add error handling to avoid errors if application data is missing that is required for csv export
- add error handling if the feature is not enabled via project config
- add unit tests and application test data

### Side Effects

- please check if card can still created from application and values will be passed correctly to the card form, since i refactored there sth. add also added error handling

### Testing

- Create an application for bavarian entitlementcard
- login as region admin or region manager
- go to the application details
- use "Weitere Optionen" and "CSV exportieren"
- check that the particular data and headers fit to the application data

### Resolved Issues

Fixes: #1752


administration/src/bp-modules/applications/utils/exportApplicationToCsv.ts
```diff
@@ -0,0 +1,63 @@
...
+      ...{ creationDate: formatDateWithTimezone(application.createdDate, config.timezone) },
+    }
+
+    const csvHeader = Object.keys(csvData).map(key => i18next.t(`application:${key}`))
+    const csvContent = Object.values(csvData)
+    const csv = stringify([csvHeader]) + stringify([csvContent])
+
+    const blob = new Blob([csv], { type: CSV_MIME_TYPE_UTF8 })
+    const { surname, forenames, creationDate } = csvData
+    downloadDataUri(blob, `${surname}_${forenames}_${creationDate}.csv`)
```
bahaaTuffaha, line 53: 🤔 I was trying to add a special char to surname like ? or / and that got handle by the OS the file name adjusted and the char became `__` not sure if other operating systems could do the same.
Do we need to sanitize or no need for that?

bahaaTuffaha (approved):
Tested on Chrome and Firefox ✅

:question:  Do we need to add the city/region too?

administration/src/bp-modules/applications/utils/exportApplicationToCsv.ts
f1sh1918: Hm we also use this naming format for PDFs somewhere I think. If there is no error I don't think this is super important

andrew8er (approved):
LGTM.

kalairubachandrakumaran-oss:
Test result:
- use "Weitere Optionen" and "CSV exportieren" (/)
- check that the particular data and headers fit to the application data (/)
image

Issue #2195, opened 2025-05-20
Add backend tests for activate card mutation


seluianova:
### Describe the Problem
Follow up for https://github.com/digitalfabrik/entitlementcard/pull/2194

### Describe Your Preferred Solution
Add tests for card activation functionality

PR #2315, opened 2025-07-01
2195: Add tests for activate card mutation

seluianova:
### Short Description
Add tests for activate card mutation

### Proposed Changes
- refactoring of ExampleCardInfo helper class to improve readability
- new tests

### Side Effects
no, only tests

### Testing
n/a

### Resolved Issues
Fixes: #2195


andrew8er (commented):
The usage of attributes with getters in `SampleCards.kt` hides the fact that there is a new object returned for every call (which is itself desirable here). I would suggest to just use methods here to make this transparent.

seluianova:
> The usage of attributes with getters in SampleCards.kt hides the fact that there is a new object returned for every call

I think you are right. Changed to functions

andrew8er (approved):

Issue #2305, opened 2025-06-30
Fix spacing above and below application list


connium:
### Describe the Problem

On the list of incoming applications, there is very little white space between the filter bar and the first application.

![Image](https://github.com/user-attachments/assets/7bec0d7e-35ca-4a9c-9343-b6c7a264be94)

Additionally, the last element sits flush at the bottom of the screen.

@andrew8er also [mentioned](https://github.com/digitalfabrik/entitlementcard/issues/2235#issuecomment-2983038356), that there is no scrollbar visible, if there are only few items in the list. If any of the elements is expanded, the scrollbar appears and causes a layout shift.
Hint: I cannot reproduce this with Fedora/latest Firefox.

Finally, the whole page scrolls, instead of just the list area.

### Describe Your Preferred Solution

- Add a vertical spacing of 24 px (1,5 rem) between the filter bar and the first application.
- Add a vertical spacing of 32 px (2 rem) between the last application and the bottom of the screen.
- Make sure there is no layout shift, when the scrollbar needs to be shown.
- Only scroll the list of applications (without menu bar and filter bar) instead of the whole page.

### Related Issues

- #2235

andrew8er:
> @andrew8er also https://github.com/digitalfabrik/entitlementcard/issues/2235#issuecomment-2983038356, that there is no scrollbar visible, if there are only few items in the list. If any of the elements is expanded, the scrollbar appears and causes a layout shift.
> Hint: I cannot reproduce this with Fedora/latest Firefox.

This is apparently macOS specific. This for me when I have configured scrollbars to show "Automatically based on mouse or trackpad" in macOS System Settings/Appearance. When using this setting, the scrollbars are rendered next to the scroll container, whereas with "When scrolling", the scrollbars are overlayed on top of the scroll container.

![Image](https://github.com/user-attachments/assets/2fc23982-21c3-471f-9c3c-e58947818854)

PR #2321, opened 2025-07-03
2305: Spacing around application list

andrew8er:
### Short Description

Refactors the nesting of elements in all authenticated routes to provide flexible layout possibilities for all types of content.

### Proposed Changes

- Removes the `<Main>` component that encapsulated the main content and forced a flex layout on it.
  The layout was:
  ```tsx
  <div id="root">
    <Nav/>
    <Main display=flex>
      <container><content/></container>
      <Footer/>
    </Main>
  </div>
  ```
  now:
  ```tsx
  <div id="root">
    <Nav/>
    <container><content/></container>
    <Footer/>
  </div>
  ```
  - The layout has one element less nesting.
  - The main content can (must) chose its own layout (ie, can have a sticky footer, scrolling content etc).
  - Sticky headers and footers are now possible without resorting to old-school hacks like `position: fixed`.

- Adds or adjusts main content containers.
- The navbar is now sticky at the top.
- Some blueprint.js components where replaced by their MUI counterparts.

### Side Effects

None.

### Testing

Test all authenticated routes.

### Resolved Issues

Fixes: #2305


seluianova (changes requested):
scrolling down on the application form does not work - not possible to fill the form
http://localhost:3000/beantragen

seluianova:
this change was probably not intended:
image

seluianova:
🙃 might be subjective, but I think it was better when the menu buttons were centered vertically
image

seluianova:
scrolling down the list of the users looks like this:
image

bahaaTuffaha (changes requested):
- I can't scroll at:
http://localhost:3000/beantragen
before:
Screenshot from 2025-07-14 17-36-24

After:
Screenshot from 2025-07-14 17-39-19

seluianova:
>> I can't scroll at: http://localhost:3000/beantragen

> Fixed 👍

I still can't scroll actually

andrew8er:
> might be subjective, but I think it was better when the menu buttons were centered vertically

Buttons are now centered

> scrolling down the list of the users looks like this:

This should be fixed now. The `sticky` position attribute is really tricky…

> No space on top, no bottom line

Fixed

> I still can't scroll actually

This works for me an Firefox and Chrome. Could you reload the page and try again? I noticed that at least Firefox sometimes has layout problems after hot-reloading flex-box styled components.

administration/src/bp-modules/cards/AddCardsForm.tsx
```diff
@@ -77,29 +54,43 @@ const AddCardsForm = ({
...
   return (
-    <Scrollable>
-      <FormsWrapper>
+    <Stack sx={{ flexGrow: 1, overflow: 'auto', justifyContent: 'safe center', alignItems: 'center' }}>
+      <Stack
+        sx={{
+          flexDirection: 'row',
+          flexWrap: 'wrap',
+          justifyContent: 'center',
+          alignItems: 'flex-start',
```
bahaaTuffaha, line 63: :thinking: I think it looks better when it's centerd.
:question:  And why do you prefer inline styling?
```suggestion
          alignItems: 'center',
```

bahaaTuffaha (commented):
When you print the page this is how it looks like this (This is one page):
Screenshot from 2025-07-18 00-35-36

And the interesting thing is I can print it with the snackbar if I was quick 😅
Screenshot from 2025-07-18 00-34-54

administration/src/bp-modules/cards/AddCardsForm.tsx
andrew8er: > 🤔 I think it looks better when it's centerd.

I'll change it, I have no strong preference on this one

> ❓ And why do you prefer inline styling?

Several reasons:
- When using styled components or similar, you always end up with stupid names like `StyledFoo` or `CustomBar` that carry no meaning, but you need to have a one, otherwise it does not work.
- This form is actually shorter for many occasions.
- No jumping around when editing a layout.
- This is one of the preferred styling methods for one-off MUI components.
- The styles get compiled into a css class module, so it's not actually "inline" styling in the html/css sense.

seluianova (changes requested):
Statistics page is corrupted (login as project admin)
image
image

seluianova:
Some pages are vertically aligned at the top and some are centered. I think we should stick to the same alignment
image
image

seluianova:
> I cannot reproduce this, for me, the statistics page looks like I would expect it, both in FF and Chrome.

@andrew8er can you try to create some cards and re-check?

f1sh1918:
The whole svg is messed up
image

f1sh1918:
Some things that changed between `main`

1. Spacing between elements was not there before (minor)
image

2. Wrong height here, should be like items above, since there will be more cards coming soon

image

seluianova:
Statistics for region admin looks different from main, I think the main-version is neater
Current branch:
image
Main:
image

seluianova:
> And I still see the issue with pdf export reported by @bahaaTuffaha
>> This should be fixed now.

Perhaps something went wrong with the commits but the issue is still there. 
I even checked it in another browser that I had never used for testing before to make sure it wasn't cached or something

image

f1sh1918:
@andrew8er since this fix from #2375 is not present here (`ApplicationList` container was removed), i have the same issue as there if a scroll down the list and then using export function
I cannot see the status bar and headlines currently

andrew8er:
> since this fix from https://github.com/digitalfabrik/entitlementcard/issues/2375 is not present here (ApplicationList container was removed), i have the same issue as there if i scroll down the list and then using export function.

I have ported the fix to this version.

> As i read from the description that the menu should be sticky. Was this removed in the meantime? Then it would be good to update the description

This was an error, fixed.

f1sh1918:
Okay then i only found one thing.
On the statistics side there is the gap too small for statistics. Before we had some gap.

image

Looks like the styles are not working on the `Box` if responsive bar is rendered.
The `margin-right` destroys the grid layout and thats why the gaps don't work anymore

seluianova:
> I cannot see the status bar and headlines currently, so this seem to be fine.

ok that's getting odd 😀 I still see it.
I cleared the cache and updated Chrome to the latest version 138.0.7204.184 (official build) (arm64).
also tested in Safari
image

seluianova:
btw I spotted one more issue:
the “Fehlende Berechtigung” error is now aligned to the top edge (previously it was centered).
image

seluianova:
> I think the problem is rather webpack, which can get confused when changing branches. Sometimes it is necessary to restart it.

not in this case. I always restart it when I change branches.

andrew8er:
> Looks like the styles are not working on the Box if responsive bar is rendered.
The margin-right destroys the grid layout and thats why the gaps don't work anymore

Looking at your referenced code, I think you got an outdated version, because there are two nested grid layouts, which are not present in the current code.

I currently have a grid with `<Box sx={{ disply: 'grid', gap: 4 }}>` to layout the region statistic components, which should translate to 16px of gap horizontally and vertically:

Screenshot 2025-07-30 at 12 59 19

f1sh1918:
> > Looks like the styles are not working on the Box if responsive bar is rendered.
> > The margin-right destroys the grid layout and thats why the gaps don't work anymore
> 
> Looking at your referenced code, I think you got an outdated version, because there are two nested grid layouts, which are not present in the current code.
> 
> I currently have a grid with `<Box sx={{ disply: 'grid', gap: 4 }}>` to layout the region statistic components, which should translate to 16px of gap horizontally and vertically:
> 
> Screenshot 2025-07-30 at 12 59 19

There is nothing to fetch or update. i'm on recent version @andrew8er 
If you check your code snippet in current changed file here you won't find it

seluianova:
> Can @bahaaTuffaha or @seluianova just return null instead of the alert Box if statistic values are 0 and check if the gaps between charts and legend work correct

```
  if (statistic.cardsCreated === 0) {
    return null
  }
```

image

the gaps look different, it's also visible with alert boxes

main:
image

current branch:
image

andrew8er:
@f1sh1918 I included your changes.

seluianova:
❓ Is the increased distance between the buttons intentional?
image

seluianova:
This message was centered on main:
image

Main:
image

connium:
@andrew8er Maybe the [screenflow](https://nextcloud.tuerantuer.org/index.php/f/6620212) helps to see, which pages are available for which users/projects.

connium:
> I'll see if I can come up with a simple solution, I have some other smaller quality of (developer-)life improvements in mind already.

Please create separate tasks for this and do not include this in a feature task.

andrew8er:
So I think this is ready to merge, do you all agree?

PR #2323, opened 2025-07-04
2244: Refactoring deepLink functions

f1sh1918:
### Short Description

I refactored some deeplink functions a bit to simplify testing.

### Proposed Changes

- pass buildConfig and isProduction to getDeepLinkFromQrCode,
- pass buildConfig to getCustomSchemeDeepLinkFromCode,
- adjust tests

### Side Effects

- deep links could be broken

### Testing

1. Test deep links when a card was created for bavaria (console or pdf)
2. Test the link for koblenz in the self service portal (3rd step Card activation) 

### Resolved Issues

Fixes: #

### DIFF
--- a/administration/src/util/getDeepLinkFromQrCode.ts
+++ b/administration/src/util/getDeepLinkFromQrCode.ts
@@ -1,18 +1,18 @@
+import { BuildConfigType } from 'build-configs'
 import { ACTIVATION_FRAGMENT, ACTIVATION_PATH, HTTPS_SCHEME } from 'build-configs/constants'
 
 import { PdfQrCode } from '../cards/pdf/PdfQrCodeElement'
 import { QrCode } from '../generated/card_pb'
 import { uint8ArrayToBase64 } from './base64'
-import { getBuildConfig } from './getBuildConfig'
 import { isProductionEnvironment } from './helper'
 
-const getDeepLinkFromQrCode = (qrCode: PdfQrCode): string => {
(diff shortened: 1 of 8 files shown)


administration/src/util/getDeepLinkFromQrCode.ts
```diff
@@ -1,18 +1,18 @@
...
 import { ACTIVATION_FRAGMENT, ACTIVATION_PATH, HTTPS_SCHEME } from 'build-configs/constants'
 
 import { PdfQrCode } from '../cards/pdf/PdfQrCodeElement'
 import { QrCode } from '../generated/card_pb'
 import { uint8ArrayToBase64 } from './base64'
-import { getBuildConfig } from './getBuildConfig'
 import { isProductionEnvironment } from './helper'
 
-const getDeepLinkFromQrCode = (qrCode: PdfQrCode): string => {
+const getDeepLinkFromQrCode = (qrCode: PdfQrCode, buildConfig: BuildConfigType, isProduction: boolean): string => {
```
andrew8er, line 9: Instead of conditional logging, we could also remove log statements altogether when building for production:

https://medium.com/@shaangontia/remove-console-statements-webpack-5-38455e4b471a

f1sh1918: Hm we also have some logs for testing on staging environment to get the deep links so not sure if entirely removing them is the solution, but we could distinguish between staging and production

PR #2325, opened 2025-07-07
Fix android builds ninja cmake

f1sh1918:
### Short Description

New kotlin version requires ninja.

### Proposed Changes

- add ninja that are required for fastlane android builds
- added some gradle properties to speedup gradle build a bit
- adjusted macos ressource class to fix deprecation warning

### Side Effects

-

### Testing

https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/8843/workflows/c59259eb-089b-4761-9cd6-1a10978b271f

### Resolved Issues

Fixes: #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,6 +278,10 @@ jobs:
                 precache: android
             - install_fastlane:
                 directory: android
+            - run:
+                command: |
+                    sudo apt-get update && sudo apt-get install -y ninja-build
+                name: Install Ninja
             - restore_cache:
                 keys:
                     - v1-gradle-{{ checksum "android/build.gradle.kts" }}-{{ checksum "android/app/build.gradle.kts" }}-{{ checksum "android/settings.gradle.kts" }}
@@ -355,6 +359,7 @@ jobs:
                     - nuernberg
                     - koblenz
                 type: enum
+        resource_class: m4pro.medium
         steps:
             - prepare_workspace
             - install_app_toolbelt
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -278,6 +278,20 @@ jobs:
...
+                name: Install Ninja
+            - run:
+                command: |
+                    wget https://github.com/Kitware/CMake/releases/download/v3.22.1/cmake-3.22.1-Linux-x86_64.tar.gz
+                    tar -xzvf cmake-3.22.1-Linux-x86_64.tar.gz -C $HOME
+                    export PATH=$HOME/ninja:$HOME/cmake-3.22.1-Linux-x86_64/bin:$PATH
+                name: Install CMake
+            - run:
+                command: |
+                    export PATH=$HOME/ninja:$HOME/cmake-3.22.1-Linux-x86_64/bin:$PATH
```
seluianova, line 293: I actually thought each "run" step has its own shell. so if you do "export PATH" here it won't be applied in other steps

f1sh1918: i will test that. i just got that from steffen

andrew8er (approved):

.circleci/config.yml
```diff
@@ -278,6 +278,10 @@ jobs:
                 precache: android
             - install_fastlane:
                 directory: android
+            - run:
+                command: |
```
seluianova, line 282: I think „|“ is needed when command contains multiple lines

PR #2341, opened 2025-07-14
1983: Send application rejected mail

seluianova:
### Short Description
Send an email to applicant for an Ehrenamtskarte Bayern if their application is rejected

### Proposed Changes
- add sending email to rejectApplicationStatus mutation
- bugfix: if the application has already been approved - the rejectApplicationStatus mutation returned a successful response and a rejectionMessage was added to the application. I added an exception for the case when the request has already been processed.
- add ApplicationJsonExtensions util and clean up FreinetApplicationMutationService a bit

### Side Effects
no?

### Testing
- create an application
- login as region admin, reject the created application
- check that the corresponding email has been sent to the applicant

### Resolved Issues
Fixes: #1983

### DIFF
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -3,6 +3,9 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/administration/administration.iml" filepath="$PROJECT_DIR$/administration/administration.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.main.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.main.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.test.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.test.iml" />
       <module fileurl="file://$PROJECT_DIR$/.idea/entitlementcard.iml" filepath="$PROJECT_DIR$/.idea/entitlementcard.iml" />
       <module fileurl="file://$PROJECT_DIR$/frontend/.idea/frontend.iml" filepath="$PROJECT_DIR$/frontend/.idea/frontend.iml" />
     </modules>
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -3,9 +3,6 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/administration/administration.iml" filepath="$PROJECT_DIR$/administration/administration.iml" />
-      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.iml" />
-      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.main.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.main.iml" />
-      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.test.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.test.iml" />
       <module fileurl="file://$PROJECT_DIR$/.idea/entitlementcard.iml" filepath="$PROJECT_DIR$/.idea/entitlementcard.iml" />
       <module fileurl="file://$PROJECT_DIR$/frontend/.idea/frontend.iml" filepath="$PROJECT_DIR$/frontend/.idea/frontend.iml" />
     </modules>
--- a/administration/src/graphql/applications/rejectApplicationStatus.graphql
+++ b/administration/src/graphql/applications/rejectApplicationStatus.graphql
@@ -1,5 +1,9 @@
-mutation rejectApplicationStatus($applicationId: Int!, $rejectionMessage: String!) {
-  application: rejectApplicationStatus(applicationId: $applicationId, rejectionMessage: $rejectionMessage) {
+mutation rejectApplicationStatus($applicationId: Int!, $project: String!, $rejectionMessage: String!) {
+  application: rejectApplicationStatus(
+    applicationId: $applicationId
+    project: $project
+    rejectionMessage: $rejectionMessage
+  ) {
     status
     statusResolvedDate
     rejectionMessage
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
@@ -18,7 +18,7 @@ import app.ehrenamtskarte.backend.common.webservice.context
 import app.ehrenamtskarte.backend.exception.service.ForbiddenException
 import app.ehrenamtskarte.backend.exception.service.NotFoundException
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.ApplicationDataIncompleteException
-import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidJsonException
+import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidInputException
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidLinkException
 import app.ehrenamtskarte.backend.exception.webservice.exceptions.InvalidNoteSizeException
 import app.ehrenamtskarte.backend.mail.Mailer
@@ -129,7 +129,7 @@ class EakApplicationMutationService {
      * Approves an application if it is in the Pending status.
      *
      * @return The updated [ApplicationView].
-     * @throws NotFoundException If an application with the specified accessKey is not found.
+     * @throws InvalidInputException If an application with the specified id is not found
      * @throws ForbiddenException If the user is not allowed to modify this application
      */
     @GraphQLDescription("Approve an application")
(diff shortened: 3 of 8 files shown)


.idea/modules.xml
```diff
@@ -3,6 +3,9 @@
   <component name="ProjectModuleManager">
     <modules>
       <module fileurl="file://$PROJECT_DIR$/administration/administration.iml" filepath="$PROJECT_DIR$/administration/administration.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.main.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.main.iml" />
+      <module fileurl="file://$PROJECT_DIR$/.idea/modules/backend.test.iml" filepath="$PROJECT_DIR$/.idea/modules/backend.test.iml" />
```
seluianova, line 8: maybe something is wrong with my local setup, but the backend module desperately wants to be added

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -142,7 +140,7 @@ class EakApplicationMutationService {
                     it.tryChangeStatus(Status.Approved)
                     ApplicationView.fromDbEntity(it, true)
                 }
-                ?: throw NotFoundException("Application not found")
```
seluianova, line 145: for NotFoundException we show "Seite nicht gefunden" error message in administration, I believe it's not relevant for this case

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationJsonExtensions.kt
```diff
@@ -0,0 +1,36 @@
...
+        return listOfNotNull(forenames, surname).filter { it.isNotBlank() }.joinToString(" ")
+    }
+
+    fun ApplicationEntity.getApplicantEmail(): String =
+        this.getPersonalDataNode().findValueByName("emailAddress")
+            ?: throw ApplicationDataIncompleteException()
+
+    fun ApplicationEntity.getPersonalDataNode(): JsonNode =
+        this.parseJsonValue().findValueByPath("application", "personalData")
+            ?: throw ApplicationDataIncompleteException()
```
andrew8er, line 35: 🔧  Since we are using these methods in a variety of packages, they could just as well be defined in the `ApplicationEntity` themselves, preferably as attribute getters, instead of Java-style `get…()` methods.

Also, putting extension methods in an enclosing scope only makes sense, if you want to force the usage sites to have a specific `this` scope, but this would be better solved with [context parameters](https://kotlinlang.org/docs/whatsnew22.html#preview-of-context-parameters), and is not necessary here.

administration/src/graphql/applications/rejectApplicationStatus.graphql
```diff
@@ -1,5 +1,9 @@
-mutation rejectApplicationStatus($applicationId: Int!, $rejectionMessage: String!) {
-  application: rejectApplicationStatus(applicationId: $applicationId, rejectionMessage: $rejectionMessage) {
+mutation rejectApplicationStatus($applicationId: Int!, $project: String!, $rejectionMessage: String!) {
+  application: rejectApplicationStatus(
+    applicationId: $applicationId
+    project: $project
```
andrew8er, line 4: ❓ I know this is used in a lot of GraphQL functions, but I cannot imagine this is a good idea to have this as a parameter. Shouln't this come from a build time constant or configuration somewhere along the deployment process? Wouldn't this allow any logged in user to issue any project name and mess with an unrelated project here?

seluianova: Perhaps I misunderstand your point, but we have one backend instance across all projects. So we can‘t have project as a constant in the backend.

andrew8er: I don't have any insights into the deployment process yet, so I'm just speculating here. Are users then supposed to manipulate data across projects?

andrew8er: Then shouldn't the project name be derived from the current sessions administrator `projectId` property instead of blindly trusting user input here?

This seems like a medium security issue here. It's not catastrophic, since users must be authenticated, but we have real possibility for unauthorized access across projects here.

Or did I miss something here?

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/utils/ApplicationJsonExtensions.kt
seluianova: > Since we are using these methods in a variety of packages, they could just as well be defined in the ApplicationEntity themselves

I see these functions as utilities to parse the application json.
I don't think applicantName, applicantEmail, etc are ApplicationEntity properties.
I think putting it all in ApplicationEntity will make a clutter there, so I'd rather keep it separate, considering that we might want to parse even more json fields in the future.

andrew8er: I can see your point, the might as well be extension methods and we keep the entity classes small.

But the surrounding `object ApplicationJsonExtensions` is unnecessary. In Kotlin freestanding functions (and objects) can exist and are idiomatic. I would even say that `object` is rarely necessary, and is in most cases a Java-ism.

seluianova: Removed ApplicationJsonExtensions object

administration/src/graphql/applications/rejectApplicationStatus.graphql
seluianova: I think it's a cool observation.
If we take the project from the admin - it would mean an additional database request to get the project name by its id, but it's not a big deal, I guess.

f1sh1918: Yes we may fix that. The only thing that can happen in this case is that the mail signature is from a different project but this would be prevented by the `Authroizer` so no panic :)

But we will have cases where we have to pass the project because the request doesn't enforce authorization

f1sh1918 (commented):
Looks good so far but if during mail sending an error occurs nothing will be shown to the user @seluianova 
I think we then need a rollback and an error message

seluianova:
> Looks good so far but if during mail sending an error occurs nothing will be shown to the user @seluianova
> I think we then need a rollback and an error message

@f1sh1918 I actually assumed we don't.
The user will be able to see the status of the request on their /antrag-einsehen page, so I thought a failed email shouldn't be a blocker for an admin who wants to process an application.

.idea/modules.xml
seluianova: reverted

andrew8er (approved):

backend/src/main/kotlin/app/ehrenamtskarte/backend/mail/Mailer.kt
```diff
@@ -405,4 +405,36 @@ object Mailer {
...
+            sendMail(
+                backendConfig,
+                projectConfig.smtp,
+                projectConfig.administrationName,
+                applicantAddress,
+                subject,
+                message,
+            )
+        } catch (_: MailNotSentException) {
+        }
```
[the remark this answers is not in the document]
seluianova: It's logged in sendMail() function

f1sh1918:
> > > > Looks good so far but if during mail sending an error occurs nothing will be shown to the user @seluianova
> > > > I think we then need a rollback and an error message
> > > 
> > > 
> > > @f1sh1918 I actually assumed we don't. The user will be able to see the status of the request on their /antrag-einsehen page, so I thought a failed email shouldn't be a blocker for an admin who wants to process an application.
> > > @connium do you have an opinion on that?
> > > Can we have a case where an email can never be delivered? So that the application can never be rejected?
> > 
> > 
> > Well its likely that this issue only occurs if the mail service does not work and not if the mail address is not correct or existing. But i'm also okay if we don't check this. I actually think if you received a mail 2 or 3 weeks again when you submitted your application, you will likely not recheck the status. So this additional mail will be more obvious that sth. happened I'm also fine to keep it as it is
> 
> What would happen, if a user misspelled their mail address? Then the application will be sent, but the applicant won't receive an email, right?
> 
> I think, that there are multiple reasons why an email cannot be delivered and some of these we will never recognize (mailbox full, invalid redirect, ...). So I'd show a note to the user but not roll back the transaction.

We can just catch issues with the mail service itself. Can't be reached, wrong credentials whatever.
As we currently show an error message for the application submit, i would do the same. It will even likely not happen

kalairubachandrakumaran-oss:
Test result:
-  an email is sent to the applicant ✅ 
FDE83BEC-DA94-4096-985A-C4571EAAA625
0A4BD571-C95A-45BF-B528-528B4D48803E

- and the email contains the subject "Ihr Antrag wurde abgelehnt" ✅ 
- and the email contains a clear description of the reason for rejection in German language ✅ (The email is being sent, but the last sentences show slight deviations from the requirement.)
- and I get a notification that the email has been sent. ❔ 
- >[connium](https://github.com/connium) Question: what exactly is meant by notification? As far as I can see, there is currently no notification (apart from a short note shown when rejecting — see screenshot). 
image

connium:
>     * and I get a notification that the email has been sent. ❔
>
>     * > [connium](https://github.com/connium) Question: what exactly is meant by notification? As far as I can see, there is currently no notification (apart from a short note shown when rejecting — see screenshot).

I try to avoid describing UI or technical details, to allow UI/UX and the developers find the best solution (in terms of usability and technical effort).

In this case, I'd see the requirement to be fulfilled, if for instance a toast message appears saying "Email was sent successfully".

Issue #2251, opened 2025-06-04
Prevent submitting the application if error occurred while sending an email


bahaaTuffaha:
### Describe the Bug

If the email failed to be sent for any reason the application shouldn't be submitted as this could create duplicates if the service got back online again.

### Steps to Reproduce

1. Make sure you did the setup for local email server using [MailHog](https://github.com/mailhog/MailHog) i talked about before on matter-most.
2. keep the hog server container shut/stopped.
3. Go to http://localhost:3000/beantragen and fill the form.
4. and submit. You should see and error message saying `email konnte nicht an test@outlook.com gesendet werden.`
5. go to region manager and look for the application being there.

### Additional Information

![Image](https://github.com/user-attachments/assets/7b54a918-1e4f-4350-bc35-ed08841ef3c3)

f1sh1918:
well that shouldn't be too difficult to implement. We could just set everything inside of a `transaction{}` and if an error occurs everything will be rollbacked

PR #2342, opened 2025-07-14
2251: Prevent submitting with error

f1sh1918:
### Short Description

As a user i might not get any mail if during my application submit if the mail service has issues and i'm even not aware that the application was submitted. That can also result in duplicates if the application was submitted twice

### Proposed Changes

- wrap the `addEakApplication` function in a transaction
- adjust error message
- adjust style of error message

### Side Effects

- none

### Testing (local)

1. Adjust your local mail config to include an error
2. Restart backend service
3. Create an application that now should show an error when submitting
4. Check that the particular application will not be listed in "Eingehende Aufträge"

### Resolved Issues

Fixes: #2251

### DIFF
--- a/administration/src/util/translations/de.json
+++ b/administration/src/util/translations/de.json
@@ -599,7 +599,7 @@
     "invalidApplicationConfirmationNoteSize": "Unzulässige Zeichenlänge der Notiz erreicht. Maximal sind {{maxSize}} Zeichen erlaubt.",
     "entitlementNotFound": "Wir konnten Ihre Angaben nicht im System finden.\nAus technischen Gründen stehen die Daten des Jobcenters Koblenz ab Mitte des folgenden Monats zur Verfügung.\nBitte überprüfen Sie Ihre Angaben und versuchen Sie es erneut.",
     "entitlementExpired": "Sie sind nicht länger berechtigt, einen KoblenzPass zu erstellen. Bitte kontaktieren Sie koblenzpass@stadt.koblenz.de für weitere Informationen.",
-    "mailNotSend": "Email konnte nicht an {{recipient}} gesendet werden.",
+    "mailNotSend": "E-Mail konnte nicht an {{recipient}} gesendet werden.\nBitte probieren Sie es zu einem späteren Zeitpunkt erneut.\nIhre Antragsdaten bleiben erhalten.",
     "passwordResetKeyExpired": "Die Gültigkeit ihres Links ist abgelaufen",
     "regionNotFound": "Diese Region existiert nicht.",
     "regionNotActivatedForApplication": "Für diese Region kann der zentrale Beantragungsprozess noch nicht genutzt werden, kontaktieren Sie bitte Ihre zuständige Behörde direkt.",
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
@@ -37,39 +37,39 @@ class EakApplicationMutationService {
         application: Application,
         project: String,
         dfe: DataFetchingEnvironment,
-    ): DataFetcherResult<Boolean> {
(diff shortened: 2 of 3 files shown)


administration/src/util/translations/de.json
```diff
@@ -599,7 +599,7 @@
     "invalidApplicationConfirmationNoteSize": "Unzulässige Zeichenlänge der Notiz erreicht. Maximal sind {{maxSize}} Zeichen erlaubt.",
     "entitlementNotFound": "Wir konnten Ihre Angaben nicht im System finden.\nAus technischen Gründen stehen die Daten des Jobcenters Koblenz ab Mitte des folgenden Monats zur Verfügung.\nBitte überprüfen Sie Ihre Angaben und versuchen Sie es erneut.",
     "entitlementExpired": "Sie sind nicht länger berechtigt, einen KoblenzPass zu erstellen. Bitte kontaktieren Sie koblenzpass@stadt.koblenz.de für weitere Informationen.",
-    "mailNotSend": "Email konnte nicht an {{recipient}} gesendet werden.",
+    "mailNotSend": "E-Mail konnte nicht an {{recipient}} gesendet werden.\nBitte probieren Sie es zu einem späteren Zeitpunkt erneut.\nIhre Antragsdaten bleiben erhalten.",
```
f1sh1918, line 602: @connium maybe you got some better ideas here what to show. This error only occurs if our mail service is broken, just for some context

bahaaTuffaha (approved):
Tested. It works as expected ✅

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -37,39 +37,39 @@ class EakApplicationMutationService {
...
+                    verificationEntities,
+                    dataFetcherResultBuilder,
+                    applicationConfirmationNote,
+                )
+            } else {
+                applicationHandler.sendApplicationMails(
+                    applicationEntity,
+                    verificationEntities,
+                    dataFetcherResultBuilder,
+                    applicationConfirmationNote,
```
andrew8er, line 68: ❌ Since `ApplicationHandler.sendPreVerifiedApplicationMails()` entails connecting to a mail service, I would strongly suggest to move this out of the transaction, as those should be as short as possible.

Unrelated to this PR, but the `ApplicationHandler` class is a weird kitchen sink of functionality, that should be split up into different (extension) functions.

andrew8er: This is actually a very classic problem with distributed systems, and all solutions have some pros and cons.

Even with one big transaction like we currently have, there is the possibility of an inconsistency: sending them email works, but the transaction somehow fails and everything is rolled back. You now have sent a verification email but the application is not in the correct state.

Usually, this is solved with a task queue system like Celery or – for emails only – could also be a local mail forwarder that is very fast, possibly running on the same machine as the backend, or a message queuing system like NATS or Kafka (just to name some well known example, absolutely not suggesting it).

A simple solution would be to put all mails into their own DB table and have a service (or a job within the backend), dequeue them after successfully sending them.

f1sh1918: So you prefer to not rollback if mail error occurs but to just shown an error message and forward to the application submitted screen?
I mean then we can never resend the mail to the applicant except we provide a separate functionality that can be triggered by service users

seluianova (approved):
If you decide to keep the current solution: to me it looks good.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
seluianova: > Even with one big transaction like we currently have, there is the possibility of an inconsistency: sending them email works, but the transaction somehow fails and everything is rolled back. You now have sent a verification email but the application is not in the correct state.

The same situation will happen if the admin rejects and deletes the application before it is verified by the organizations.
We return a nice fallback page for such cases.

Issue #2336, opened 2025-07-10
Misaligned invalid card activation screen


bahaaTuffaha:
### Describe the Bug

The error message `Der Aktivierungscode ist ungiiltig.` is not in the center of the screen.

### Steps to Reproduce

1. Create a card or you have a deeplink of an existent card.
2. Change the first char of the hash after `code#` to any other letter.
3. Try to activate the deeplink.

### Expected Behavior

- Centered error message.

### Related Issues

None.

### Additional Information
<img src="https://github.com/user-attachments/assets/c46ee422-483f-44dc-b108-204970c3138f" width="300px" />

PR #2343, opened 2025-07-14
2336: Misaligned card activation error

f1sh1918:
### Short Description

When the user opens an invalid deeplink, the error message container should be center even if the text length does not need the entire screen width

### Proposed Changes

- wrap error container with `Center` widget

### Side Effects

- none

### Testing (local)

1. `npx uri-scheme open ehrenamtbayern://bayern.ehrenamtskarte.app/activation/code# --ios`
2. Test same for `--android`

### Resolved Issues

Fixes: #2336


bahaaTuffaha (approved):
Nice!
Tested on Android ✅

BTW the Resolved Issue number is wrong.

Issue #2356, opened 2025-07-15
Improve route protection and unify error boxes


f1sh1918:
### Describe the Problem

Currently our route protection with roles and conditions is very repetitive uses lots of redundant code

### Describe Your Preferred Solution

- It would be great to use a reuseable component for that
-  also remove blueprint components

Issue #2245, opened 2025-06-03
Unhanded error at Nürnberg's activity logs


bahaaTuffaha:
### Describe the Bug
Sometimes while I'm testing a weird error comes when I open the activity logs at the store manger (Nürnberg) and was hard to reproduce.

```bash
React Router caught the following error during render TypeError: Cannot read properties of undefined (reading 'fromSerialized')
    at Card.ts:95:1
    at Array.reduce (<anonymous>)
    at deserializeCard (Card.ts:92:1)
    at ActivityLog.ts:20:1
    at Array.map (<anonymous>)
    at loadActivityLog (ActivityLog.ts:18:1)
    at ActivityLogController (ActivityLogController.tsx:29:1)
    at renderWithHooks (react-dom.development.js:15486:1)
    at mountIndeterminateComponent (react-dom.development.js:20103:1)
[...]

PR #2358, opened 2025-07-16
2356: Route protection and error boxes

f1sh1918:
### Short Description

When i was investing some time in reproducing #2245 i realized that our route and component render protection is quite repetitive and has lots of redundant code.  Therefore i created a separate component to avoid all the if/else conditions with particials and return all the error boxes.
All these turned in a bigger refactoring though.

### Proposed Changes

- remove redundant `getQueryResult.tsx` and replace imports
- replace `NonIdealState` (BP) by `AlertBox`(MUI)
- replace `Spinner` (BP) by `CircularProgress`(MUI), 
- create `RenderGuard` component to restrict routes and component renderings by roles or/and condition
- add unit tests for `RenderGuard`, `HomeController`, `NavigationBar`

### Side Effects

- i refactored a lot so permission to see particular routes or components can be broken, but should be hopefully catched by correct unit tests ;)

### Testing

1. Login with particular role of projects and test if the roles can still do and see the same than before

### Resolved Issues

Fixes: #2356 
Fixes: #2245


administration/src/bp-modules/UserMenu.tsx
```diff
@@ -58,11 +60,13 @@ const UserMenu = ({ onSignOut }: UserMenuProps): ReactElement => {
       <NavLink to='/user-settings'>
         <MenuItem minimal icon='settings' text={t('userSettings')} />
       </NavLink>
-      {projectConfig.activityLogConfig && (
+      <RenderGuard
```
f1sh1918, line 63: This was actually a change. The menu item should be role restricted

administration/src/mui-modules/base/CenteredCircularProgress.tsx
```diff
@@ -0,0 +1,7 @@
+import { CircularProgress } from '@mui/material'
+import { styled } from '@mui/system'
+
+const CenteredCircularProgress = styled(CircularProgress)`
+  margin: 'auto';
```
bahaaTuffaha, line 5:
```suggestion
  margin: auto;
```

administration/src/bp-modules/regions/RegionController.tsx
```diff
@@ -20,19 +20,24 @@ const RegionSettingsContainer = styled.div`
...
-      <DataPrivacyCard />
-      <RegionSettingsController regionId={region.id} />
-      <ApplicationConfirmationNoteController regionId={region.id} />
-      {freinetDataTransferEnabled && <FreinetSettingsController regionId={region.id} project={projectId} />}
-    </RegionSettingsContainer>
+    <RenderGuard
+      allowedRoles={[Role.RegionAdmin]}
+      condition={region !== undefined}
+      error={{ description: t('notAuthorizedForRegionSettings') }}>
+      {region && (
```
bahaaTuffaha, line 32: Why are we checking `region` twice at condition and here? does condition is sufficient?

bahaaTuffaha (changes requested):
I like the `RenderGuard` already 👍️

administration/src/bp-modules/regions/RegionController.tsx
f1sh1918: Because i wanted to avoid `!` you can check what happens without.
Thats the drawback of this component instead of an `if else` statement

f1sh1918:
> There is no actions at `admin@bayern.ehrenamtskarte.app` image
> 
> Region-admin is having 2 extra buttons now (Benutzer verwalten und Statistiken) both says I'm not authorized. image

I used `RegionManager` instead of `ProjectAdmin`. Should be fine now

f1sh1918:
> > Fixes: https://github.com/digitalfabrik/entitlementcard/issues/2245
> 
> I can still reproduce this error
> image
> 

How @seluianova ? This route shouldn't be accesible with project store manager

seluianova:
login as externer API Nutzer to Bayern, open /applications route:
image

f1sh1918:
> > Fixed it here: [0a7b55f](https://github.com/digitalfabrik/entitlementcard/commit/0a7b55f08082c58f8c9bcbf0636e2715c8748a6a)
> 
> well I know it is still not possible on production, but there is a following "bug" now:
> 
> 1. login as region admin to bayern, create card, logout
> 2. login as region admin to nürnberg, open activity log --> it shows the card created on the previous step :)
> 
> image
> not sure if we want to fix that

We would have to clean session storage on logout. You can create an issue for that if you want. but it will not happen on production

bahaaTuffaha (approved):
Great work...
Tested on chrome tried to visit everything.✅
I think it solved the `fromSerialized` error. ✅

seluianova (approved):
lgtm

Docmale:
various roles tested

PR #2367, opened 2025-07-22
2094: Handle safe areas

f1sh1918:
### Short Description

Wrapping the whole MaterialApp with `SafeArea` doesn't lead to really good results, since the theme background will not be considered and we should only use `SafeArea` where its really needed to avoid blank parts even there are no clickable items or texts that have to be readable

Note that the scope of this was that all ui element are clickable and texts are readable.
Its not very likely that this app will be used in landscape mode, so i did not optimize that.

### Proposed Changes

- remove `SafeArea` from MaterialApp to avoid black areas even with light theme, 
- create `LandscapeSafeArea` that only sets SafeArea for landscape mode, since in most cases this its only needed there
- use this widget where needed, 
- wrap `SafeArea` around IntroPage, 
- add padding for location button in landscapeMode for the map, since its the only ui component that can not be touched

### Side Effects

- none

### Testing
Mind to use Android 15 Device with BottomNavigationButtons. Test also ios device
1. Test the MapPage in portrait and landscape (Location button should be clickable
2. Test the LicensePage (About) in portrait and landscape. Everything should be readable.
3. Test the SearchPage in portrait and Landscape. Everything should be readable and all buttons clickable

### Resolved Issues

Fixes: #2094


andrew8er:
The safe area on the store detail view is missing:

Screenshot 2025-07-28 at 14 07 52

seluianova:
perhaps the bottom navigation bar also needs a safe area
image

f1sh1918:
> The safe area on the store detail view is missing:
> 
> Screenshot 2025-07-28 at 14 07 52

Fixed with last commit
Ios
image
Android
image

seluianova:
extra space is added on top of the description
image
as discussed, not a blocker but still a remark...

andrew8er (approved):

seluianova:
the top app bar look different on different pages...
image
image

Issue #2319, opened 2025-07-02
Handle withdrawn applications correctly


f1sh1918:
### Describe the Problem

As a service user i want to have to correct application actions if an application was withdrawn by the applicant

### Describe Your Preferred Solution

If an application was withdrawn by the user, the following restrictions should be applied:
- not allow triggering approval mail
- not allow confirming the application
- not allow rejecting the application
- not allow creating a card
- only export and delete application should be possible

### Related Issues

#1754

PR #2373, opened 2025-07-28
2319: handle withdrawn applications

andrew8er:
### Short Description

Enables correct handling of withdrawn applications. This includes administration frontend and the backend. This was a little bit more involved than initially estimated, since to correctly handle the filtering and sorting, we need to take the application's pre-verification status into account, which is hidden in the raw `jsonValue` string data. There were four different places, where this property was parsed and passed around separately from the application object.

The `jsonValue` property is now parsed (in `parseApplication()` immediately after receiving the application objects from the different queries, which simplifies the passing of data a bit.

Please have a close look at the backend part of the PR, I might have missed a spot where a withdrawn application should have been handled.

#### Major changes

- Remove DB column `application.withdrawalDate` and replace it with `statusResolvedDate`. The withdrawal is already captured with the `status` column.
- Removed the `withdrawalDate` from the GraphQL type, replace with `statusResolvedDate`.
- Show the withdrawal notice in the `<ApplicationCard>` component.
- Preparse the `application.jsonValue` immediately after querying and pass it down to components.
- Respect the pre-verification status when sorting and filtering applications in `<ApplicationOverview>`.
- In the backend, prohibit some mutations on applications when an application is withdrawn.

#### Minor changes

- Improve the IntelliJ SQL code style, it is now a little less horrible.
- Add Freinet demo administrator and application.
- Add demo data for an undecided application

### Testing

- Inspect the filtering and sorting in http://localhost:3000/applications

### Resolved Issues

Fixes: #2319


administration/src/mui-modules/application-verification/ApplicationApplicantController.tsx
```diff
@@ -23,9 +24,9 @@ const ApplicationApplicantController = ({ providedKey }: { providedKey: string }
   if (!applicationQueryHandler.successful) {
     return applicationQueryHandler.component
   }
-  const application = applicationQueryHandler.data.application
+  const application = parseApplication(applicationQueryHandler.data.application)
```
f1sh1918, line 27: ❌ parsing the application here, leads to null values for some properties

image

This also leads to `ApplicantApplicantView` will not receive proper status

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/schema/view/ApplicationView.kt
```diff
@@ -25,7 +24,6 @@ data class ApplicationView(
                 regionId = entity.regionId.value,
                 createdDate = entity.createdDate.toString(),
                 jsonValue = entity.jsonValue,
-                withdrawalDate = entity.withdrawalDate?.toString(),
```
f1sh1918, line 28: ❌ Related to LN 29- LN 30
Even this issue came with an pr before, i just realized it while testing the `ApplicationApplicantView`

This information will not be passed for `getApplicationByApplicant` 
We need this information in the `ApplicationApplicantController`

f1sh1918 (changes requested):
Looks already really good but i have to request changes here, since the `ApplicationApplicantView` breaks or has been broken (before)
1. Create an application
2. Open the applicant status link
http://localhost:3000/antrag-einsehen/<accesskey>/
3. There should be a withdraw button

I added a comment why this happens, check `ApplicantView`
The return model has some privacy settings that prevent `status` and `statusResolveDate` to be exposed for non logged in users.

It would be also okay to fix it in a separate issue. But it just affects two lines of code

bahaaTuffaha (commented):
image

This is pending and I can't see the withdraw button.

f1sh1918:
> image
> This is pending and I can't see the withdraw button.

see why i requested changes the comment before

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -229,43 +228,69 @@ const ButtonsApplicationResolved = (props: {
...
+// Return type should actually be ComponentProps<typeof Typography>['color'] | undefined, but this type is too generic
+const statusColor = (applicationStatus: ApplicationStatus): string | undefined => {
+  switch (applicationStatus) {
+    case ApplicationStatus.Approved:
+    case ApplicationStatus.ApprovedCardCreated:
+      return 'success'
+    case ApplicationStatus.Rejected:
+      return 'error'
+    case ApplicationStatus.Withdrawn:
+      return 'warning '
```
seluianova, line 254:
```suggestion
      return 'warning'
```

seluianova:
Broken translations:
image

For other statuses as well:
image

seluianova:
Something messed up with the counts
64 =! 6 + 2 + 2 + 56
image

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -1,39 +1,65 @@
...
-import { ApplicationStatusBarItemType, ApplicationVerificationStatus, GetApplicationsType } from './types'
-import { applicationEffectiveStatus } from './utils'
+import { getPreVerifiedEntitlementType } from './preVerifiedEntitlements'
+import type { Application, ApplicationStatusBarItemType } from './types'
 
-export const barItems: ApplicationStatusBarItemType[] = [
-  {
-    title: 'allApplications',
-    status: undefined,
+const countApprovingVerifications = (application: Application): number =>
```
seluianova, line 15: countApprovedVerifications ?
countRejectedVerifications

administration/src/bp-modules/applications/ApplicationsOverview.tsx
```diff
@@ -1,39 +1,65 @@
...
+    i18nKey: 'allApplications',
+    filter: (_: Application): boolean => true,
   },
-  {
-    title: 'accepted',
-    status: ApplicationVerificationStatus.Approved,
+  accepted: {
+    i18nKey: 'accepted',
+    filter: (application: Application): boolean =>
+      countApprovingVerifications(application) > 0 ||
```
seluianova, line 29: the application is approved when every verification is approved, not just > 0

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/schema/view/ApplicationAdminGql.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import app.ehrenamtskarte.backend.application.database.ApplicationEntity
+import app.ehrenamtskarte.backend.application.webservice.dataloader.verificationsByApplicationLoader
+import app.ehrenamtskarte.backend.common.webservice.fromEnvironment
+import com.expediagroup.graphql.generator.annotations.GraphQLName
+import graphql.schema.DataFetchingEnvironment
+import java.util.concurrent.CompletableFuture
+
+@GraphQLName("ApplicationAdmin")
+data class ApplicationAdminGql(
```
seluianova, line 11: 🙃 I think the previous name was clearer tbh

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/EakApplicationMutationService.kt
```diff
@@ -249,3 +259,11 @@ class EakApplicationMutationService {
         return true
     }
 }
+
+private fun ApplicationEntity.changeStatusOrThrow(status: Status): ApplicationEntity =
+    try {
+        this.status = status
+        this
+    } catch (e: IllegalArgumentException) {
+        throw InvalidInputException("Cannot approve application, status is '${this.status}'")
```
seluianova, line 268: this function is used not only for approval but also for rejection

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/schema/view/ApplicationView.kt
```diff
@@ -1,53 +0,0 @@
-package app.ehrenamtskarte.backend.application.webservice.schema.view
-
-import app.ehrenamtskarte.backend.application.database.ApplicationEntity
-import app.ehrenamtskarte.backend.application.webservice.dataloader.verificationsByApplicationLoader
-import app.ehrenamtskarte.backend.common.webservice.fromEnvironment
-import graphql.schema.DataFetchingEnvironment
-import java.util.concurrent.CompletableFuture
-
-@Suppress("unused")
-data class ApplicationView(
```
seluianova, line 10: I didn't get why this data class was split up tbh.
I guess the idea was to hide some fields from the applicant. but this can be done at the graphql query level.
we can remove unnecessary fields from the query, and they will not be returned
```
query GetApplicationByApplicant {
    getApplicationByApplicant(accessKey: "ghjg") {
        createdDate
        id
        jsonValue
        regionId
        status
        statusResolvedDate
        withdrawalDate
    }
}
```

administration/src/shared/verifications.ts
```diff
@@ -0,0 +1,21 @@
+import { ApplicationVerificationView } from '../generated/graphql'
+
+export const isEmailValid = (email: string): boolean => /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(email)
```
seluianova, line 3: 🙃 this function is used not only for verifications but also to check applicant email (and can be potentially used to verify other emails, e.g. when creating a new admin), not sure if this function should be moved to "verifications.ts"

seluianova:
The button to re-send an approval request is not available if the application has multiple verifications and one of them has been approved.
image

seluianova (changes requested):
Adding “request changes” as I already mentioned some breaking issues, but I have not checked everything yet as it is really a lot.

backend/src/main/kotlin/app/ehrenamtskarte/backend/application/webservice/schema/view/ApplicationView.kt
andrew8er: Several reasons:
- The most important one was that it is conceptually clearer to have several "views" for different audiences and thus to avoid accidentally revealing data. As @f1sh1918 [pointed out here](https://github.com/digitalfabrik/entitlementcard/pull/2373#pullrequestreview-3063029355), there might even be a third view necessary. I tend to agree here. Having different view types at the lowest level (server controller layer) removes the possibility for accidents in any of the upper levels (e.g. UI).
- The other one is that having separate types removes some unnecessary "nullable" properties, which were only nullable, because they might need to be omitted for more restricted views. The nullability also trickles down to the client side and having properties non-nullable removes some redundant null-checks.

> I guess the idea was to hide some fields from the applicant. but this can be done at the graphql query level.

This would open an information disclosure vulnerability. The attributes a GraphQL query retrieves of an object is simply a client side construct meant to avoid transmitting unnecessary data and nothing would stop an attacker from requesting the "hidden" attributes if you do not prohibit this at the server.

administration/src/shared/verifications.ts
andrew8er: To me this is a very generic utility function, I would place that in a file `/src/misc/emails.ts` or so.

What worries me more is the possibility that this regex will reject valid email addresses because it is hilariously simplified for what it claims to do. 

For the curious, a regex that correctly validates real world RFC822 email addresses looks like [this](https://pdw.ex-parrot.com/Mail-RFC822-Address.html), but we should use a specialized package, since it is really slow.

andrew8er:
> Broken translations:
> For other statuses as well:

Good catch. I expanded the i18n a little bit and separated the status bar texts from the "no applications" texts for grammatical correctness.

> Something messed up with the counts / filters by status
> 64 =! 6 + 2 + 2 + 56
> 2 applications are lost somewhere in "all applications" and cannot be filtered

Not necessarily, since the different statuses are not strictly mutually exclusive. An application can be both "withdrawn" (from the `status` field) and be either approved/rejected/open (from the application verifications _or_ the preverification status).

We had a discussion [here](https://github.com/digitalfabrik/entitlementcard/issues/2319#issuecomment-3102446784), but it could be argued that withdrawn applications should be removed from all other filters.

> I think it’d be better to open separate PRs for large refactorings.
> It’s a pretty big change (1000+ lines), and seems to require an extra testing.

Generally, you're totally right. The thing here is, I did not plan to make this huge refactoring, I only noticed that it would be necessary pretty late in the process, when fine-tuning the sorting of the application list. I think this PR is better review on a commit by commit basis.

> btw the testing instructions don't include the key changes requested in the task (handling of the “withdrawn” status)

This is because I do not have a firm grasp of the ways the backend processes applications (especially with respect to sending emails). I tried to wade through the code, but it's pretty convoluted imho. So I added just some very general instructions and a note for the backend in the PR introductory text.

> The button to re-send an approval request is not available if the application has multiple verifications and one of them has been approved.

- Original logic was `application.status !== ApplicationStatus.Rejected`. 
- Current logic is `application.status === ApplicationStatus.Pending`, 
- Now changed to `verificationStatus(verification) === VerificationStatus.Pending`. Is this correct?

seluianova:
> Original logic was application.status !== ApplicationStatus.Rejected.
> Current logic is application.status === ApplicationStatus.Pending,
> Now changed to verificationStatus(verification) === VerificationStatus.Pending. Is this correct?

No 🙈
Original logic was:
Show button if **verification is pending** AND **application is not rejected**

Expected change within the [2319](https://github.com/digitalfabrik/entitlementcard/pull/2373/files#top):
Show button if **verification is pending** AND **application is not rejected** AND **application is not withdrawn**

andrew8er:
> Show button if verification is pending AND application is not rejected AND application is not withdrawn

👍  Updated and rebased.

Docmale:
found:
Withdrawn data are not deleted #2510

Issue #2375, opened 2025-07-28
Fix print dialog


f1sh1918:
### Describe the Bug

We had lots of layout change since the last release and our print (window print) layout is messed up

### Steps to Reproduce

Reproducible only on GoogleChrome
1. Go to "Eigehende Anträge
2. Scroll the list a bit down
3. Weitere Optionen -> PDF exportieren
4. Error: There is lots of additional space before and after application content
### Expected Behavior

There is no additional space before and after the application

f1sh1918:
Motion is causing the issue :/
This commit: https://github.com/digitalfabrik/entitlementcard/commit/f9fcc198d706bd5b35e81e16bc31f4b5dbc4fda9

seluianova:
can you attach a screenshot?
I don't see the issue on current main....
<img width="1512" height="753" alt="Image" src="https://github.com/user-attachments/assets/5e294743-f8e7-4b70-ab2d-cb76d6749dba" />
<img width="1382" height="775" alt="Image" src="https://github.com/user-attachments/assets/2fa638cf-af61-450b-b2ca-e5b3e108273a" />

andrew8er:
I don't have such huge margins:
<img width="646" height="133" alt="Image" src="https://github.com/user-attachments/assets/95770c95-d3c1-45f4-b0c5-581c4eef65cf" />

I would be fine with removing the animation library here, I don't think it adds much to the usability anyway. But our current approach to print one single application is completely bonkers, and needs to be rearchitected:

Currently, we have a route to display all applications (`/applications`) and hide all unwanted elements for printing. Instead we should have another route (e.g. `/applications/:id/print`) that only renders the bare application in a simplified form, maybe with additional headers and immediately opens a print dialogue.

andrew8er:
Have have not examined the issue exactly, but my hunch is this: the margins are actually gaps from the enclosing flex-box layout, that are rendered even if the `<ApplicationCard>`  components of all "not-printed" cards are hidden. 

Maybe a `@print { display: none }` style on the `<motion.div>` component will do the trick.

PR #2379, opened 2025-07-29
2375: Fix print dialog

f1sh1918:
### Short Description

Print dialog currently adds lots of whitespace before and after the application, depending on the scroll position of the list

### Proposed Changes

- remove gap for printing view

### Side Effects

- none

### Testing

1. Go to "Eingehende Anträge"
2. Scroll down the application list
3. Click export pdf
4. Application shouldn't have whitespace before and after the application details

### Resolved Issues

Fixes: #2375

### DIFF
--- a/administration/src/bp-modules/applications/ApplicationsOverview.tsx
+++ b/administration/src/bp-modules/applications/ApplicationsOverview.tsx
@@ -43,6 +43,10 @@ const ApplicationList = styled.div`
   flex-direction: column;
   height: 100%;
   gap: 16px;
+  // This fixes a whitespace issue in print dialog when using motion #2375
+  @media print {
+    gap: 0;
+  }
 `
 
 const getEmptyApplicationsListStatusDescription = (activeBarItem: ApplicationStatusBarItemType, t: TFunction): string =>


f1sh1918:
> ```diff
>  @media print {
> +    gap: 0;
> +  }
> ```

good find thx. I tried a lot yesterday and was assuming that it relates to `gap` but couldn't fix it

andrew8er (approved):
LGTM.

We could also move the print handling out of the `<ApplicationCard>` (removing the `sx={{ displayPrint: isSelectedForPrint ? 'block' : 'none' }}` of the outermost `<Accordion>` and move it to the `<ApplicationOverview>` using:

```ts
const MotionDiv = styled(motion.div)`
  @print {
    display: none;
  }
`
```
instead of `<motion.div>`.

But maybe better move this to another PR.

bahaaTuffaha (approved):
Tested on brave and Firefox ✅
For the Snackbar also you can add this at `StyledOverlayToaster` AppToaster.tsx :
```js
@media print {
    display: none !important;
  }
```

Issue #1948, opened 2025-02-27
Error toast in online form for Koblenz doesn't disappear automatically


connium:
### Describe the Bug

In the application form for the KoblenzPass error toasts are shown. These messages don't disappear automatically.

When I fill in the application form for the KoblenzPass, and I don't fill a mandatory field and click the "KoblenzPass erstellen" button, a toast message appears.

### Steps to Reproduce

1. Open the KoblenzPass app.
2. Go to the 'Ausweisen' tab.
3. Click on 'Aktivieren'.
4. Fill in the form, but leave at least one mandatory field empty.
5. Click on 'KoblenzPass erstellen'
6. See the error toast.
7. Wait for several seconds. The toast stays visible.

Additionally, if I don't dismiss the toast and click the button again, a second toast is added.

### Expected Behavior

I expect the toast to disappear after several seconds.

See https://mui.com/material-ui/react-snackbar/ for example.

### Additional Information

- OS: Android 9
- Browser: Stock (Chrome)

![Image](https://github.com/user-attachments/assets/0a71f7eb-cdd7-4ebb-8fc5-bbfedaf46a17)

f1sh1918:
It's not a bug @connium its timeout 0 set and can be closed manually

connium:
Then I was misled by learned behavior. I will clarify with @hauf-toni  what corresponds to the expected behavior for users and what WCAG says about it.

Would it be possible to close previous toasts if a new toast is shown?

f1sh1918:
> Then I was misled by learned behavior. I will clarify with [@hauf-toni](https://github.com/hauf-toni) what corresponds to the expected behavior for users and what WCAG says about it.
> 
> Would it be possible to close previous toasts if a new toast is shown?

No, i don't think so

seluianova:
> for example to about 5-6 seconds

should this also apply to “long” messages like this?

<img width="556" height="222" alt="Image" src="https://github.com/user-attachments/assets/2a437108-c4a7-4cab-b9f9-abb8e3b310e4" />

connium:
> > > for example to about 5-6 seconds
> > 
> > 
> > should this also apply to “long” messages like this?
> 
> yes, possibly with an extended display duration of 8–10 seconds if the message is as long as this one. Is it possible to implement a logic where the display duration is longer if the message exceeds x characters?

[...]

PR #2391, opened 2025-08-04
1948: Use default timeout for error toast in the self-service portal

seluianova:
### Short Description
In the application form for the KoblenzPass error toasts are shown. These messages don't disappear automatically.

As discussed in https://github.com/digitalfabrik/entitlementcard/issues/1948 we would like to make them disappear automatically.

### Proposed Changes
- I removed `timeout: 0` for the "Mindestens eine Ihrer Angaben ist ungültig." toaster, the default timeout (5 secs) is now applied;
- for apollo errors I set `timeout: 8000`, as discussed in the task.

### Side Effects
no

### Testing
1. Open Koblenz card creation form: /erstellen
2. Submit form with errors
--> for "field validation" error (Mindestens eine Ihrer Angaben ist ungültig), the toaster should disappear after 5 seconds
--> for other errors (e.g. user entitlements revoked or expired), the toaster should disappear in 8 seconds

### Resolved Issues
Fixes: #1948

### DIFF
--- a/administration/src/bp-modules/util/cardGenerationError.tsx
+++ b/administration/src/bp-modules/util/cardGenerationError.tsx
@@ -20,6 +20,7 @@ export const showCardGenerationError = (appToaster: Toaster, error: unknown): vo
     const { title } = getMessageFromApolloError(error)
     appToaster.show({
       message: <FormAlert isToast errorMessage={title} />,
+      timeout: 8000,
       intent: 'danger',
     })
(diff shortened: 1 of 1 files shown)


bahaaTuffaha (approved):

Docmale:
If I don't fill in a field after opening the activation, only the name hint is displayed, not a toast. However, for other fields and later, it is. That's OK for me.

Issue #2250, opened 2025-06-04
Add tests for the new FreinetApplicationMutationService


bahaaTuffaha:
### Describe the Problem

FreinetApplicationMutationService needs tests at the backend.

### Describe Possible Alternatives

N/A

### Related Issues

https://github.com/digitalfabrik/entitlementcard/issues/2141

PR #2402, opened 2025-08-10
2250: Added tests for FreinetApplicationMutationService

bahaaTuffaha:
### Short Description

FreinetApplicationMutationService needs tests at the backend.

### Proposed Changes

- Created a mock for FreinetApi that runs before each test.
- Test for the following:
   - Unauthorized access.
   - Freinet is not configured.
   - Forbidden error for authorized.
   - Data transfer is not activated.
   - Data transfer is activated and creates new person.
   - Returns true when person exists.
   - Returns error when multiple persons are found in Freinet.

### Side Effects

- None.

### Testing

Run the Test Backend configuration. 

### Resolved Issues

Fixes: #2250

### DIFF
--- a/backend/src/test/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationServiceTest.kt
+++ b/backend/src/test/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationServiceTest.kt
@@ -1,0 +1,263 @@
+package app.ehrenamtskarte.backend.freinet.webservice
+
+import app.ehrenamtskarte.backend.GraphqlApiTest
+import app.ehrenamtskarte.backend.application.database.Applications
+import app.ehrenamtskarte.backend.application.webservice.schema.view.JsonField
+import app.ehrenamtskarte.backend.application.webservice.schema.view.Type
+import app.ehrenamtskarte.backend.common.webservice.EAK_BAYERN_PROJECT
+import app.ehrenamtskarte.backend.freinet.database.FreinetAgencies
+import app.ehrenamtskarte.backend.freinet.database.FreinetAgenciesEntity
+import app.ehrenamtskarte.backend.freinet.util.FreinetApi
+import app.ehrenamtskarte.backend.freinet.webservice.schema.types.FreinetPersonCreationResultModel
+import app.ehrenamtskarte.backend.generated.SendApplicationAndCardDataToFreinet
+import app.ehrenamtskarte.backend.generated.inputs.FreinetCardInput
+import app.ehrenamtskarte.backend.helper.TestAdministrators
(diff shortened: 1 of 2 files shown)


bahaaTuffaha:
> Afaics we fire here requests on every test run against the freinet api which does not have any testing instance but only test "mandant". I'm not really sure if its okay to do that on `commit` workflow tbh. But i'm also aware that we shouldn't mock code that we don't own and in this scenario we cannot do this at least for an integration test.

https://mockk.io/#constructor-mocks:~:text=Sometimes%2C%20especially%20in%20code%20you%20don%E2%80%99t%20own

bahaaTuffaha:
And also we didn't provide a real accessKey to this test

backend/src/test/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    private fun createMutation(applicationId: Int, project: String): SendApplicationAndCardDataToFreinet {
+        val freinetCard = FreinetCardInput(
+            cardType = "Standard",
+            expirationDate = "2026-12-31",
+        )
+        val variables = SendApplicationAndCardDataToFreinet.Variables(
+            applicationId = applicationId,
+            project = project,
```
seluianova, line 258: project param was removed, the branch should be actualized

backend/src/test/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetApplicationMutationServiceTest.kt
```diff
@@ -0,0 +1,263 @@
...
+                agency.dataTransferActivated = false
+            }
+
+            val applicationId = createTestApplicationForFreinet()
+            val mutation = createMutation(applicationId, projectId)
+            val response = post(client, mutation, regionAdminFreinet.getJwtToken())
+            assertEquals(200, response.code)
+
+            val jsonResponse = response.json()
+            assertEquals(false, jsonResponse.path("data").path("sendApplicationAndCardDataToFreinet").asBoolean())
```
seluianova, line 95: we can also check that no request is sent to freinet, e.g.

```
            verify(exactly = 0) {
                anyConstructed<FreinetApi>().createPerson(any(), any(), any(), any(), any())
            }
```

seluianova:
> Well still think that we shouldn't add tests that fire to freinet. Probably we need to mock the freinet api service and call it in the graphql test

@f1sh1918 as I see, the Freinet API is mocked in these tests, so requests are not sent to the real Freinet

seluianova (approved):
lgtm!

bahaaTuffaha:
> Can this PR be merged?

It needs adjustment after #2441

Issue #2396, opened 2025-08-06
Invalid application link view is misaligned


seluianova:
### Describe the Bug
Invalid application link view is misaligned

Mobile browser:
<img width="380" height="790" alt="Image" src="https://github.com/user-attachments/assets/3bd78940-823d-488f-b05d-ee5358d5b4cf" />

Desktop browser:
<img width="1512" height="825" alt="Image" src="https://github.com/user-attachments/assets/69092c10-6316-4472-a7d8-d4cb53b53ff8" />

Current view on staging:
<img width="1512" height="354" alt="Image" src="https://github.com/user-attachments/assets/234eaf83-39e1-4696-bda5-8ec86557f90c" />

### Steps to Reproduce
Open /antrag-einsehen link with an invalid access code, e.g. /antrag-einsehen/hjh

### Expected Behavior

_I guess the view might just be reverted to the previous version (?)_

- the view should be vertically aligned to the top (remove empty space at the top)
- the view should be better adapted for mobile browser

PR #2407, opened 2025-08-18
2396: Fix invalid application link view's alignment

bahaaTuffaha:
### Short Description

Invalid application link view is misaligned.

### Proposed Changes

- Changed the `flex-direction` on small screens (xs and sm) to column.
- Adjusted margin to have 5px instead of auto to stay at top and keep a small margin.

### Side Effects

other places that uses `AlertBox` like activation page and `StatisticsController` ...

### Testing

- Test http://localhost:3000/antrag-einsehen/hjh
- check other places where `AlertBox` is used.

### Resolved Issues

Fixes: #2396


administration/src/mui-modules/base/AlertBox.tsx
```diff
@@ -3,10 +3,18 @@ import React, { ReactElement } from 'react'
...
   justify-content: center;
   align-items: center;
+
+  @media (max-width: 899px) {
+    flex-direction: column;
+    & .MuiAlert-action {
+      align-self: start;
+      margin-left: 0;
+    }
+  }
```
andrew8er, line 17: 💭  Here a simple `sx` prop could be used, along with the MUI defined breakpoints and grid system:

```tsx
    <Alert
      sx={theme => ({
        margin: 1,
        justifyContent: 'center',
        alignItems: 'center',
        [theme.breakpoints.down('md')]: {
          flexDirection: 'column',
          alignSelf: 'flex-start',
          marginLeft: 0,
        },
      })}
      ...
      >
    </Alert>

f1sh1918:
> > Update: As an improvent we could also think about providing a custom error component for these few cases where we render large lists in the error description
> 
> * How it differs from the `Alert` component in terms of design?
> * And is invalid link at (antrag-einsehen) is considered a large list?

There is no design. The question is if we want to request one.
I mean its at least that large that the content exceeds the space that fits properly in a general alert design from mui.

But you can also just keep the `AlertBox` for now if it looks okay. It was just an idea

bahaaTuffaha:
> If possible, this should be done in the theme.

Outlined is a type of variation... unless we want to adjust the outlined variation to have a specific styling.

bahaaTuffaha:
> Currently, we show the content and the button side-by-side. This PR changes this to a vertical stack for smaller viewports. I would propose to always display it as a vertical stack, with buttons aligned right. Also we should have a max width of about 600px (you could use a [breakpoint](https://mui.com/material-ui/customization/breakpoints/#default-breakpoints)), because overly long paragraphs are hard to read.

Always vertically it will look like this 
image

instead of

image

f1sh1918:
> Maybe we just don't need an icon for invalid link view? And for other alerts it could stay as before
> 
> (_I personally don't like the vertical alignment_)

Yes same for me. Even as @andrew8er i these cases here we don't need a retry button so i could fit better even in mobile

andrew8er:
> Always vertically it will look like this

With vertical alignment, I meant the retry button goes under the main content, not the icon on top.

andrew8er (approved):

Issue #1758, opened 2024-11-06
Make notes more visible, if available


dkehne:
**Is your feature request related to a problem? Please describe.**
Local authorities would like to be able to see at a glance whether there are notes for an application. These usually only exist if there is something wrong with the application or if there is a deviation from the standard process.

**Describe the solution you'd like**
We were immediately provided with a solution as a suggestion for improvement:
"Notes should be highlighted in color if they are available and displayed immediately when opened"

However, I think that @hauf-toni should work out a design suggestion again based on the request.

hauf-toni:
Design can be found 📍 [here](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=2519-1725&t=mzp7xTHZsyKeZfzX-1). Please let me know your thoughts @f1sh1918 @andrew8er

andrew8er:
@hauf-toni 
In the design you mentioned "new content". I have a pretty clear understanding of what constitutes new content, but when is a note content _not_ new, i.e. when would we hide the indicator again? The current logic is to show the indicator, when there is any note content

An adjacent question: do these applications get reviewed by multiple persons? I had the idea that those are usually reviewed by only one person, but after thinking about, I realized there might be a few reasons for multiple reviewers (vacations, illness).

PR #2415, opened 2025-08-19
1758 Notes more visible

f1sh1918:
### Short Description

Its important to know for the service user, if there are notes to the application. So we want to highlight it a bit more

### Proposed Changes

- change icon color
- change button color and text if there was a note created

### Side Effects

- none

### Testing

1. Go to applications
2. Create a note and delete a note. Check the effect

### Resolved Issues

Fixes: #1758


seluianova:
❓ I thought we temporary moved the button here because we waited to the new design.
image

Can it be placed in the corner? For example, as an overlay? I think we discussed this option before but decided to wait for the design
image

seluianova:
❓ Can the button be placed in the corner, for example, as an overlay?
image

seluianova:
Suggestion regarding button placement:
https://github.com/digitalfabrik/entitlementcard/pull/2425

f1sh1918:
> Suggestion regarding button placement: #2425

merged it. thx

Issue #2390, opened 2025-08-04
QR code scanning does not work in landscape mode


seluianova:
### Describe the Bug
Not possible to activate or verify the card in the landscape mode

### Steps to Reproduce
Try to activate or verify the card in the landscape mode using qr code scanner
--> the image in the camera is rotated to the side, as the result the code is not recognized

Tested on real android device:

<img width="2244" height="1008" alt="Image" src="https://github.com/user-attachments/assets/db537c91-4bb4-4ad2-80c5-5dda011e175b" />

### Expected Behavior
Scanning should work in landscape mode (or maybe landscape mode should be disabled?)

### Additional Information

I guess it's an upstream issue

PR #2416, opened 2025-08-19
2390: Fix qr code scanning prevent landscape

f1sh1918:
### Short Description

We are facing some issue with qr code recognition in landscape mode and decided to restrict the app to portrait, since there is no real scenario where landscape mode might be used or improve user experience

### Proposed Changes

- restrict app to portrait mode
- remove landscape only safe area widget and its uses

### Side Effects

- app doesn't work in landscape mode

### Testing

1. Start the app and switch orientation to landscape by turning device or emulator

### Resolved Issues

Fixes: #2390

### DIFF
--- a/frontend/ios/Runner/Info.plist
+++ b/frontend/ios/Runner/Info.plist
@@ -65,15 +65,12 @@
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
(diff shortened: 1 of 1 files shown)


f1sh1918:
> We also have a landscape related code in CardDetailView: image
> 
> And in LocationButton: image
> 
> And there is also UISupportedInterfaceOrientations in Info.plist (I guess this should also be updated?)

updated @seluianova

seluianova (approved):

Issue #2397, opened 2025-08-06
"Erneut versuchen" button in AlertBox doesn't work


seluianova:
### Describe the Bug
Click on the "Erneut versuchen" button produces an exception
<img width="1512" height="549" alt="Image" src="https://github.com/user-attachments/assets/fb6c5564-acde-466f-b646-3f4cfd5b887f" />
<img width="1509" height="829" alt="Image" src="https://github.com/user-attachments/assets/fee17907-1107-4299-af29-2a514848efb0" />

I also think this alert should be adjusted to look more error-like (red color, error icon)

### Steps to Reproduce
Dev only:
1. Start administration & backend
2. Login as region admin to administration
3. Stop backend, open /applications
--> see "Server nicht erreichbar." error

### Expected Behavior
- the "Erneut versuchen" button should work
- the color should be red for errors and the icon should look like an error
- the button translation should be adjusted "Erneut Versuchen" --> "Erneut versuchen"

PR #2417, opened 2025-08-19
2397: Fix retry

f1sh1918:
### Short Description

Currently we are facing errors if retry was executed for an request using `AlertBox`

### Proposed Changes

- execute `onAction` with arrow function
- pass `severity: 'error'`

### Side Effects

- none

### Testing

- open this route to get an erroneous request : http://localhost:3000/antrag-einsehen/hjh
- check the network tab for requests or backend service (devs)

### Resolved Issues

Fixes: #2397

### DIFF
--- a/administration/src/mui-modules/base/AlertBox.tsx
+++ b/administration/src/mui-modules/base/AlertBox.tsx
@@ -33,7 +33,7 @@ const AlertBox = ({
       variant='outlined'
       action={
         onAction ? (
-          <Button color='inherit' size='small' onClick={onAction}>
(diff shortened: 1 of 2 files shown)


andrew8er (approved):
Works.

Any idea for the reason this did not work? Usually, it should, however without the function wrapper, I got this error:
```
Uncaught TypeError: cyclic object value
```

Another thing is that the retry here is completely pointless from a user perspective. We should differentiate between retryable errors (5xx, network errors, timeouts) and those that are not.

Docmale:
Simulated this by using an invalid servername  in URL

Issue #2062, opened 2025-04-14
Show information about application status on user-facing pages


connium:
Describe the Problem

As an applicant,
I want to see the status of my application
so that I know what is going on.

### Describe Your Preferred Solution

Given that I have applied for a volunteer card
and my application has been approved,
when I click on the status link from the initial confirmation mail,
then I see, that my application has been approved
and I see when the my application was approved
and i can no longer withdraw my application.
Here is some example view, since design is missing:
https://staging.bayern.ehrenamtskarte.app/antrag-einsehen/GDYCypIg7YcTR7ruNzeukchBK4hZE8S9zEnR4dPW1dIN_8HOibjaLsYAoc5ifhMClJgPTmXCErIAgunW0gncXQ==
The area after the application details can be replaced by the hint. ❓

Given that I have applied for a volunteer card
and my application has been rejected,
when I click on the status link from the initial confirmation mail,
then I see, that my application has been rejected
and I see when the my application was rejected
and i can no longer withdraw my application.

Given that I am a contact person of an organization to confirm the voluntary work of an applicant
and the application of that person has been approved
or the application of that person has been rejected,
when I click on the link from the request-for-confirmation mail,
then I see, that the application has been processed ("Der Antrag wurde bereits bearbeitet.")
and i can no longer confirm or reject the voluntary work of the applicant.
Here is some example view, since design is missing:
https://staging.bayern.ehrenamtskarte.app/antrag-verifizieren/MzsH2He7kpka7gK2f61u3Dk0e0wjfIgnMVgsydJJi2WVpV2Za9k3Oo3WuijNzpFxRPm_TFJCYq1NM56TgZJ4Uw==
The area after the application details can be replaced by the hint. ❓

f1sh1918:
it might get conflichts with #2319 so i will just wait until this task is finished

PR #2419, opened 2025-08-19
2062: Status information user facing pages

f1sh1918:
### Short Description

In #2319 the user facing information for the applicant were already implemented, when an application was processed, so only the information for verifying organizations were missing

### Proposed Changes

- add helper function,
- add translations for the application already processed, 
- show `AlertBox` if the application was already processed
- adjust `AlertBox` to render it `borderless`

### Side Effects

- none

### Testing

1. Create an application with organizations to verify
2. Login as region manager/region admin and reject or accept the application
3. Check the verification url that you got via the e-mail
Example: http://localhost:3000/antrag-verifizieren/<"verificationAccessKey">

### Resolved Issues

Fixes: #2062

Note: Since the design is very close to our `AlertBox` component, i decided to reuse it (borderless). So there are slight differences that shouldn't harm.

image


f1sh1918:
ok thx. it was not attached to the ticket. Will adjust it

seluianova (approved):
lgtm!

bahaaTuffaha (commented):
1) Rejection an application will show a blue AlertBox but in the design it doesn't look blue:

Image


2) If an organization has rejected the voluntary work do we need to do something for the initial mail status or the application it self  like hiding the withdraw @connium :

(this image shows the initail mail status and the application page)
Image

f1sh1918:
> Generally works fine.
> 
> What I find a little irritating is that we are dealing with `Withdrawn` applications in `<ApplicationApplicantController>`, while all other states are handled in `<ApplicationApplicantView>`. Is there a specific reason for this?
> 
> I would have expected, that withdrawn and accepted or rejected applications are generally handled the same. Since the information we show in an already rejected application are pointless to the reviewer, why do we not hide them like in the withdrawn case?

Good point. I talked to @connium and it makes no sense to show the whole application card to the verifier if the application was already processed by the service user. I will adjust that

f1sh1918:
> Generally works fine.
> 
> What I find a little irritating is that we are dealing with `Withdrawn` applications in `<ApplicationApplicantController>`, while all other states are handled in `<ApplicationApplicantView>`. Is there a specific reason for this?
> 
> I would have expected, that withdrawn and accepted or rejected applications are generally handled the same. Since the information we show in an already rejected application are pointless to the reviewer, why do we not hide them like in the withdrawn case?

Hm well i think we have two different cases:

1. Verifier: We should show the application information as short as possible, so if an application was processed we just show an AlertBox that it was processed and no further steps are needed from the verifier
2. Applicant: If an application was processed especially rejected, the applicant get the rejection reason via mail but also want to recheck the application. So in this case we may show the application to him. And thats also the reason why the states were handled in different places

f1sh1918:
Well I would tend to not show the application Information. After an application was rejected it should be delete quite soon. Then the application Information will not be available anymore so the recheck can only be done in a very short period

f1sh1918:
Alright i adjusted Applicant and Verifier view if it was already processed

Verifier
image

Applicant
image

So no further application information will be shown

seluianova:
🙃 Now I think these alert boxes should have the same style
image
image

seluianova (approved):
Re-tested

connium:
> > 🙃 The last sentence probably no longer makes sense, because it is not possible to “leave a feedback” and it is not even clear on the page itself what feedback is being referred to. _(perhaps the entire "hint" should be rephrased or removed)_ image
> 
> @connium what do you think. I don't think that removing the application information impacts the content of this note

Well, if I cannot verify or reject the volunteer work because the application has already been processed, then it doesn't make much sense to show this hint. It might even confuse people more, “what feedback?”/"welche Rückmeldung"?

connium:
> > Well, if I cannot verify or reject the volunteer work because the application has already been processed, then it doesn't make much sense to show this hint. It might even confuse people more, “what feedback?”/"welche Rückmeldung"?
> 
> Perhaps we should show this hint when the user **is** able to approve/reject the application. (Currently we don't show it in such case)

I think it's better to not show it at all, than to show it in a scenario where the user cannot do anything. IRC we include already a similar note in the mail to the contact person, so I'm totally fine with removing this hint completely (will make it easier for all of us :slightly_smiling_face: ).

f1sh1918:
Alright I can create a ticket for that. If it should be removed urgently I could also retrigger a release if needed

PR #2420, opened 2025-08-19
2361: update mui library to recent

f1sh1918:
### Short Description

See issue description. This was reported by sentry. According to mui issue it should be fixed with recent mui version

### Proposed Changes

- update mui to recent

### Side Effects

- none

### Testing

I couldn't reproduce, but it should happen on clicking on the "Chip" that appears when a file was uploaded and the user wants to remove it

### Resolved Issues

Fixes: #2361


seluianova:
I could reproduce the issue!

> it should happen on clicking on the "Chip" that appears when a file was uploaded and the user wants to remove it

It's important to click on the chip but not the “delete” icon.

image

Library update fixes the issue ✅

Issue #2322, opened 2025-07-04
Update geolocator and geolocator_android


f1sh1918:
### Describe the Problem

Currently we have issues with `geolocator_android: > 4.6.1 causing this issue. It came up with this ticket #2316
https://stackoverflow.com/questions/79528379/could-not-get-unknown-property-flutter-for-extension-android

### Describe Your Preferred Solution

- update `geolocator and `geolocator_android` to recent versions

### Related Issues

#2316

PR #2421, opened 2025-08-19
2322: Update geolocator

f1sh1918:
### Short Description

With older version of geolocator there were some issue, that blocked us from updating. Seem to be fixed now

### Proposed Changes

- update geolocator and geolocator_android to recent
- remove unneeded `forceAndroidLocationManager` flag, since it was only used for FLOSS builds

### Side Effects

- likely none

### Testing

- Test gps location is still working
- Test if app can be build with android (dev only): 
```
fvm flutter build appbundle --flavor bayern --release --dart-define=environment=staging
```
### Resolved Issues

Fixes: #2322


seluianova (approved):
tested android & ios (virtual devices)

Issue #2382, opened 2025-07-30
Toasters are printed on the PDF form


seluianova:
### Describe the Bug
If the user clicks “PDF exportieren” when some toaster is displayed, it is printed in a PDF file.
<img width="1170" height="398" alt="Image" src="https://github.com/user-attachments/assets/3e653332-dcce-4dff-b244-15e784afb83d" />

### Steps to Reproduce
1. Login as region admin, go to '/applications'
2. Select some pending application, click "Antrag bestätigen", "Antrag ablehnen" or "Anfrage erneut senden" button
3. When toaster appears, click "Weitere Optionen" --> "PDF exportieren" button very fast
4. See the toaster on the PDF

### Expected Behavior
Toasters should not be printed

bahaaTuffaha:
> Instead of hiding a bunch of elements and add lots of extra code for that, we could add another route (e.g. `/application/:id/print`) that displays a single application in a simplified print layout with all unnecessary elements omitted.
> 
> There are a few caveats, opening a new tab is only possible within a click event and some browsers (or some settings) prohibit opening tabs.
> 
> What we could try instead is:
> 
> * Navigating to the print page programmatically on print button click
> * Open the print dialogue automatically here
> * Navigate back after the print dialogue was closed (using the [`afterprint` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/afterprint_event)).

You can just do this: https://github.com/digitalfabrik/entitlementcard/pull/2379#pullrequestreview-3067172900

andrew8er:
I think solving this with CSS is pretty futile, we will always have some elements that somebody forgot to hide. The main problem is, that we can only print a full page, and our "current" full page happens to show a lot of components that are not relevant to the thing we want to actually print. This necessitates a lot of tricky code, that I would rather get rid of.

I [tried the extra print route approach](https://github.com/digitalfabrik/entitlementcard/tree/2382-create-print-application-route) in a proof of concept and it sort of works, but I get the print dialog twice.

Another option would be [react-to-print](https://www.npmjs.com/package/react-to-print), seems like a pretty solid library.

PR #2422, opened 2025-08-19
2382: Application print view

andrew8er:
### Short Description

Adds a specialized, very simple print view for printing a single application.

### Proposed Changes

- Removes some code that was used to hide components while printing.
- Add dependency `react-to-print`.
- Add an `<ApplicationPrintView>` component.

### Side Effects

None

### Testing

- Visit http://localhost:3000/applications.
- Click "Weitere Optionen/Als PDF exportieren".

### Resolved Issues

Fixes: #2382


f1sh1918 (commented):
Looks really good already.

In the previous print view, we also had the timestamp on it, which might be also an important information for the service user i think

image

f1sh1918 (approved):
❓ I'm not really sure if there is a scenario where this withdrawn note might be needed
image
Rest LGTM

f1sh1918:
> > ❓ I'm not really sure if there is a scenario where this withdrawn note might be needed
> 
> It was in the design once, although I cannot find it anymore… I will remove it.

No sorry for the misunderstanding, it is an Information which we may also consider for print view not for removing it in general

administration/src/bp-modules/applications/ApplicationPrintView.tsx
bahaaTuffaha, line 1: on firefox there is a weird path on the top`about:srcdoc`
Image

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -240,22 +241,23 @@ const headerTypography: SxProps = {
...
   const { t } = useTranslation('applicationsOverview')
   const theme = useTheme()
   const config = useContext(ProjectConfigContext)
   const baseUrl = `${getApiBaseUrl()}/application/${config.projectId}/${application.id}`
   const appToaster = useAppToaster()
+  const printContentRef = useRef<HTMLDivElement>(null)
+  const printApplication = useReactToPrint({
+    contentRef: printContentRef,
+    pageStyle: applicationPrintViewPageStyle.styles,
+  })
```
bahaaTuffaha, line 260: [Firefox issue] You can add `documentTitle: document.title,` so when you save the print it should have the name of the page instead of srcdoc.

bahaaTuffaha (approved):
Nice work.
I found couple of small problems:

1) The outline/box-shadow is missing .
2) There is no break lines .
3) The approval text is missing.
Image

PR #2428, opened 2025-08-20
2396: Invalid application link alignment

f1sh1918:
### Short Description

Currently we show a retry button even a retry could not solve the issue (for all graphql errors f.e.)

### Proposed Changes

- add retryable option in `GraphQLErrorMessage`
- only show retry button if it's needed
- remove AlertBox vertical style for mobile

### Side Effects

- none

### Resolved Issues

Fixes: #

### DIFF
--- a/administration/src/errors/DefaultErrorMap.ts
+++ b/administration/src/errors/DefaultErrorMap.ts
@@ -14,10 +14,10 @@ const defaultErrorMap = (error: ApolloError): GraphQLErrorMessage => {
     return { title: i18next.t('errors:pageNotFound') }
   }
   if (error.message.includes('500')) {
-    return { title: i18next.t('errors:internalError') }
+    return { title: i18next.t('errors:internalError'), retryable: true }
   }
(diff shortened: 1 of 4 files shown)


f1sh1918:
@andrew8er since you raised this issue, i'm not sure if there are any other error codes where it makes sense to provide a retry

bahaaTuffaha:
what about hiding the icon?
I was going with this:
```js
        [theme.breakpoints.down('md')]: {
          margin: '5px',
          '& .MuiAlert-icon': {
            display: 'none',
          },
        },
```

f1sh1918:
> what about hiding the icon? I was going with this:
> 
> ```js
>         [theme.breakpoints.down('md')]: {
>           margin: '5px',
>           '& .MuiAlert-icon': {
>             display: 'none',
>           },
>         },
> ```

I would tend to not hide it, since its the icon and the colors are the indicator for the severity of the Alert

andrew8er (approved):

PR #2442, opened 2025-08-28
fix-jest-test-coverage: adjust output file

f1sh1918:
### Proposed Changes

-

### Side Effects

-

### Resolved Issues

Fixes: #


f1sh1918:
I think we should just run the tests either in actions or in circleCI but not both

bahaaTuffaha (approved):

PR #2455, opened 2025-08-30
Gh action check administration

f1sh1918:
### Proposed Changes

- added missing protobuf
- adjusted test-script to produce `report.json` and run silent
- move run commands from circle ci `check-administration` to gh action
- comment out gh job `check-administration`
- run workflow only if administration files changed

### Side Effects

- none

### Testing
I reverted the changed file that triggers that workflow (administration), you can find the triggered workflow here: https://github.com/digitalfabrik/entitlementcard/actions/runs/17342927351/job/49239198048

### Resolved Issues

Fixes: #


f1sh1918:
closed in favor of #2455

.github/workflows/check-administration.yaml
```diff
@@ -0,0 +1,38 @@
+name: 'Check Administration'
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'administration/**'
+jobs:
+  check-administration:
```
andrew8er, line 9: 💭  I would reorder the naming here: `administration-check`

This way, related jobs group together visually when we have more than one job per high level module.

The same principle would also apply to the file name.

andrew8er (approved):

seluianova (approved):

Issue #2163, opened 2025-05-07
Improve Playwright e2e tests


bahaaTuffaha:
### Describe the Problem
Playwright tests are a bit flaky on Firefox and mac OS.

This part is always an issue:
```
      await page
        .getByRole('radio', { name: applicationType == 'initial' ? 'Erstantrag' : 'Verlängerungsantrag' })
        .check()
    }
```
- Since its working with `headless:false` i guess there are some racing conditions with Firefox.
- Note: on Linux (for my case) it works good all the time even with headless mode except date picker I did a workaround for Firefox.

### Describe Your Preferred Solution

Investigate a workaround or solution for radio buttons or anything could be flaky.

### Describe Possible Alternatives

Using `headless:false` in that cases.

### Related Issues

- You can take a look at the previous discussions at : https://github.com/digitalfabrik/entitlementcard/pull/2013 for more info.

bahaaTuffaha:
After this PR https://github.com/digitalfabrik/entitlementcard/pull/2129  the application icon for juleica needs fixing.
```
Error: Timed out 5000ms waiting for expect(locator).toBeVisible()

Locator: getByRole('listitem').filter({ hasText: 'Antrag vom' }).nth(3).locator('img[src*="juleica"]')
Expected: visible
Received: <element(s) not found>
Call log:
  - expect.toBeVisible with timeout 5000ms
  - waiting for getByRole('listitem').filter({ hasText: 'Antrag vom' }).nth(3).locator('img[src*="juleica"]')
```

PR #2456, opened 2025-08-30
2163: Improve e2e tests

seluianova:
### Short Description
Several improvements

### Proposed Changes
- actualize locators in check_applications test as they were changed;
- one test failed due to the real bug: wrong translation label (fixed);
- do not open browser when e2e tests are run; it was started by "npm run start" command, so I added "start:ci" for tests
- setup baseUrl in the config
- add screenshot to the report for failed tests to simplify analysis
image
- attempt to fix flaky tests: I replaced “check” with “click” for radio buttons, and it seems to be working more stably now; this simulates the actual user behavior - they click on the elements.

<hr>

_I tried to re-run tests multiple times and they are always green for me. Probably on different environment something else may come up_ ❓

### Side Effects
no

### Testing
n/a

### Resolved Issues
Fixes: #2163


administration/playwright.config.ts
```diff
@@ -25,76 +25,22 @@ export default defineConfig({
...
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
+    screenshot: 'only-on-failure'
   },
 
   /* Configure projects for major browsers */
   projects: [
     {
       name: 'chromium',
-      testMatch: 'create_applications.spec.ts',
```
seluianova, line 39: I guess the idea behind this was to setup the order of the test execution?
The same can be achieved if we add numbers to the file names - they are executed in the alphabetical order.
But:
Actually tests should be independent from each other.
If "create application" fails due to a potential bug - this should not cause all other tests to fail.
In order to be able to test cases that require test data to be created, we must configure the loading of test data.
I will create a task for that.

bahaaTuffaha: Yes this should keep them in order.

bahaaTuffaha: I can see delete applications is running in parallel 🫣

bahaaTuffaha: I had to revert the changes to keep it in order.

f1sh1918:
ok i still have issues (mostly timeout) on firefox but one less

image

image

image

f1sh1918 (approved):
I approved since chromium works and there shouldn't be "formal" bugs in the tests
Even for pipeline integration (in a next step) i would tend to say one browser engine should be okay and reduce costs

bahaaTuffaha:
Note: Let `npm run test:e2e` run its own backend that has the flag disableMailService set to true.

Issue #2441, opened 2025-08-28
Restrict freinet data transfer for staging


f1sh1918:
### Describe the Problem

Since we don't have a freinet staging system for testing, but only a demo region, we have to ensure that during testing on staging system, we do not send user and card data by accident

### Describe Your Preferred Solution

- add a check in the backend that for environment staging and development user and card data can only be send to the Freinet Demo Region

PR #2459, opened 2025-08-30
2441: Restrict freinet data transfer

f1sh1918:
### Short Description

Currently it is possible to send user and card data to freinet from staging environment, which should not happen. Since freinet has no staging environment the data transfer has to be tested with the `Freinet Demo` region.

### Proposed Changes

- add validation function that only allows sending data to freinet if `production` environment or `Freinet Demo` region
- send custom error message if data transfer is not permitted
- adjust freinet settings controller to reset state on update error
- set initial state onComplete for the freinet data query

### Side Effects

- none

### Testing

1. Go to "Region verwalten" enable "Freinet Data transfer"
2. Process an open application, create a card and check that you get an "Fehlende Berechtigung" error message
3. Switch to "Freinet Demo" region. Repeat step 1. and 2. and check if you get a success message and data will be send to freinet (credentials for freinet demo region can be found in passbolt)

### Resolved Issues

Fixes: #2441


f1sh1918:
@bahaaTuffaha if your test pr is merged, please add a test for this case (probably some tests of you should already fail)

seluianova:
perhaps it's better to restrict this in updateDataTransferToFreinet() ?

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/util/FreinetDataTransferPermission.kt
```diff
@@ -0,0 +1,16 @@
...
+import app.ehrenamtskarte.backend.exception.service.ForbiddenException
+
+/**
+ * Validates if Freinet data transfer is permitted for the given environment and region.
+ * - Production environment: allowed for all regions.
+ * - Non-production environments: only allowed for demo region, because freinet has no staging environment.
+ */
+fun validateFreinetDataTransferPermission(environment: Environment, regionName: String) {
+    if (environment != Environment.PRODUCTION && regionName != FREINET_DEMO_REGION_NAME) {
+        throw ForbiddenException()
```
f1sh1918, line 14: ❓ Not sure if we should create or throw a separate exception, but since this will not happen on production, i think it should be good enough

andrew8er (approved):
LGTM.

backend/src/main/kotlin/app/ehrenamtskarte/backend/freinet/webservice/FreinetAgencyMutationService.kt
```diff
@@ -27,6 +29,12 @@ class FreinetAgencyMutationService {
             throw NotImplementedException()
         }
 
+        val region = RegionsRepository.findRegionById(regionId)
```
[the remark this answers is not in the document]
seluianova: pls ping me when ready

seluianova:
FreinetAgencyMutationServiceTest should be adjusted

administration/src/util/translations/de.json
```diff
@@ -625,7 +625,8 @@
     "applicationToCsvDisabledError": "Die Antragsexportfunktion ist nicht verfügbar.",
     "applicationToCsvError": "Es ist ein Fehler beim Export des Antrags aufgetreten.",
     "freinetPersonDataInvalid": "Erforderliche Personendaten von Freinet sind nicht verfügbar. Die Kartendaten konnten deshalb nicht übertragen werden.",
-    "freinetCardDataInvalid": "Die Kartendaten können nicht an Freinet übertragen werden, weil diese ungültig sind."
+    "freinetCardDataInvalid": "Die Kartendaten können nicht an Freinet übertragen werden, weil diese ungültig sind.",
+    "freinetDataTransferNotPermitted": "Für diese Region ist kein Datentransfer erlaubt."
```
seluianova, line 629:
```suggestion
    "freinetDataTransferNotPermitted": "Für diese Region ist keine Datenübermittlung erlaubt."
```

seluianova (approved):
Tested 👍
Thanks for the quick adjustment :)

seluianova:
> Should i keep the environment and adjust the regionId or wouldn't it make more sense to use production?

I would go for the region adjustment

f1sh1918:
> > do you want to re-check @seluianova ?
> 
> I am away till the late evening (for testing) But the commit looks good, I think it shouldn't break anything

okay you can just check it if you want or directly merge it. Then another beta release for backend/administration could be triggered i guess

Issue #2461, opened 2025-09-01
Remove applicationAlreadyProcessedHint


f1sh1918:
### Describe the Problem

Since we do not show any application information to the verifier when an application was already processed, we do not need to show the hint. As we discussed in #2419

### Describe Your Preferred Solution

- remove the hint text
- remove the translation entry

kalairubachandrakumaran-oss:
Tested ✅ 

<img width="100" height="200" alt="Image" src="https://github.com/user-attachments/assets/9af591ae-848a-43b4-8c0a-d49193d148cc" />

<img width="100" height="200" alt="Image" src="https://github.com/user-attachments/assets/647e7b6d-2646-47f4-8999-78c95067de8d" />

<img width="100" height="200" alt="Image" src="https://github.com/user-attachments/assets/bd68a6ac-b188-4c68-b694-cd8c3e034762" />

PR #2463, opened 2025-09-01
2461: Remove hint

f1sh1918:
### Proposed Changes

- remove hint for verifiers, when application was already processed

### Side Effects

- none

### Testing

- Create an application
- Process an application (Reject/approve)
- Check the verification link

### Resolved Issues

Fixes: #2461


f1sh1918:
> I think it shouldn't be bold (and bigger than the normal text) image
> 
> image

okay then i will pass it as description

PR #2490, opened 2025-09-08
Add fallback to install step to avoid failures from stale npm cache in CI

seluianova:
### Short Description
Due to frequent dependency updates, we encounter npm cache-related failures in CI from time to time.
Previous time it was discussed here: https://github.com/digitalfabrik/entitlementcard/pull/2386

### Proposed Changes
- add a fallback step that clears the npm cache and retries npm ci if the initial install fails
- this should help to avoid manual adjustments of the cache keys every time

### Side Effects
no?

### Testing
Successful workflow: 
https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/9432/workflows/966bc99a-6da7-46a7-ab21-32e7965f6581/jobs/59819

image

### Resolved Issues

Fixes: #


f1sh1918 (approved):
LGTM thx

seluianova:
@f1sh1918 or we can merge this with one review perhaps ? I believe this change is quite safe

Issue #2480, opened 2025-09-04
Implement custom "callout" component


connium:
We should implement a custom component for this, could be lifted from the "withdrawn" notice in `<ApplicationCard>`.

- `src/bp-modules/project-settings/PepperSettings.tsx`
- `src/bp-modules/user-settings/ChangePasswordForm.tsx`
- `src/bp-modules/users/DeleteUserDialog.tsx`
- `src/bp-modules/users/EditUserDialog.tsx`

For details, ask @andrew8er

PR #2491, opened 2025-09-09
2480 implement custom callout component

Elham-EN:
### Short Description

Implemented custom MUI Callout component to replace Blueprint.js Callout in EditUserDialog as part of MUI migration.

### Proposed Changes

- Created reusable Callout component in src/shared/components/Callout.tsx
- Replaced Blueprint.js Callout with custom component in EditUserDialog.tsx
- Updated props: intent='danger' → color='error', intent='primary' → color='info'

### Side Effects

- Blueprint.js Callout dependency can eventually be removed once all instances are migrated
- Future callout implementations across the app should use this new component for consistency

### Testing

- Verify callout styling matches original appearance
- Test EditUserDialog info and error callouts display correctly
- Confirm links and checkboxes work as expected

### Resolved Issues

 Fixes: #2480

### DIFF
--- a/administration/src/bp-modules/applications/ApplicationCard.tsx
+++ b/administration/src/bp-modules/applications/ApplicationCard.tsx
@@ -60,6 +60,7 @@ import { ApplicationStatusNote } from './components/ApplicationStatusNote'
 import { ApplicationIndicators } from './components/VerificationsIndicator'
 import type { Application } from './types'
 import { ApplicationToCsvError, exportApplicationToCsv } from './utils/exportApplicationToCsv'
+import { Callout } from '../../shared/components/Callout'
 
 const DeleteDialog = (props: {
   isOpen: boolean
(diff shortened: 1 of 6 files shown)


andrew8er (changes requested):
First of all, thanks for the PR. There are a few issues, that need to be resolved, before we can merge it though.

The type of issues also indicate that you did not actually run the code, because some of the changed files would not compile. Is there a reason for this? Do you need help in setting up the project? We have some documentation [here](https://github.com/digitalfabrik/entitlementcard/blob/main/docs/development-setup.md), but as always, it could certainly be improved.

We currently have settings for IntelliJ included in the project, but I want to add some for VS Code in the future, since this is what most people will use for frontend code.

If you do use VS Code, here is a little snipped for your local `.vscode/settings.json`:

```json
{
    "[typescriptreact]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode",
        "editor.formatOnSave": true
    },
}
```

When having the prettier plugin installed, this will automatically set up the IDE format the files with our code style. TypeScript should be already set up, since VS Code will pick up `tsconfig.json` by default.

f1sh1918:
hey @Elham-EN ,

he pr seems to be stale unfortunately. If there won't be any reaction from your side until next monday, we will close this pull requests, since it doesn't meet our conventions.

thx for contribution anyways. Feel free to reopen it and fix the requested issues any time.

Issue #2488, opened 2025-09-08
Hide setting for Freinet data exchange


connium:
### Describe the Problem

Freinet does not seem to be happy with the current process. After consulting with Ms. Weniger, the option for Freinet data exchange will therefore not be available for the time being.

Since the administrators are not yet familiar with the function, the easiest way is to hide the corresponding configuration in the Region Settings.

### Describe Your Preferred Solution

Hide the configuration for Freinet data exchange in the administration.

### Describe Possible Alternatives

- Disable the configuration settings and show an explanation. The administrators don't know the functionality yet, so we decided to hide it.

seluianova:
Should be adjusted in the salt config

PR #2492, opened 2025-09-09
2488: Disable freinet data transfer

seluianova:
### Short Description
Freinet data transfer should be temporarily disabled

### Proposed Changes
Set freinetDataTransferEnabled to false

### Side Effects
no

### Testing
Login as region admin to administration, check region settings /region
- "Freinet Datenübermittlung" should not be available
image

### Resolved Issues
Fixes: #2488


kalairubachandrakumaran-oss:
Test result:
"Freinet Datenübermittlung" section should not be available ✅ 
image

PR #2499, opened 2025-09-15
2447: Restructure package layout for backend

andrew8er:
### Short Description

Restructures the backend's package layout. Primary goals:

- Shallower nesting
- Uniformity
- Layout should follow code usage

### Proposed Changes

- Have more sensible package layout, that follows code usage:

| package        | purpose |
|-                      |-     |
|`config`  | configuration related code |
|`db.columns` | custom column types |
|`db.entitites` | table objects and entity classes, one file each |
|`db.migration` | db migrations |
|`db.repositories` | repositories for entities (most of the code here could be inlined, moved to their entity classes or expressed as extension functions at their usage site). |
|`db.setup` | setup code for DB data.  |
|`graphql.*` | API controllers and types for the GraphQL API, one package per topic. |
|`import` | Data import functionality |
|`routes` | HTTP API routes |
|`shared` | A kitchen sink of code that is used in various places. Some of it should be refactored later and moved into more specific places. |

- `getPaths()` methods from route handlers where inlined into the Javalin request handler configuration.
- Some types/file contents where moved to more appropriate places (`AuthContext`, `EakBayernRegions.k`, `Distance.kt`, `StoreFieldCleaner.kt`, `MapStoreRegion.kt`, `MapCsvToStore.kt`). This is code that is effectively used in only one place, and most those could be private to the file as long as test code is not using it.
- A few Kotlin objects (`PasswordValidator`, `Authorizer`, `PostalCodesLoader`) that were used as function namespaces where dissolved and the functions were moved to more appropriate locations (e.g. at their sole usage site). Methods that have a clear _main_ parameter where converted to extension functions of the main object's type.
- Assertions in `AuthorizerTest.kt` are now using more specialized versions.
- Inline `Hmac.hash()`
- Some classes were never instantiated and only used as a function namespace via a `companion object` (`CanonicalJson`, `ValidityPeriodUtil`). These were refactored into objects or as freestanding or extension functions.

### Side Effects

None

### Testing

- Compile
- Test

### Resolved Issues

Fixes: #2447


backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/shared/exceptions/ApplicationDataIncompleteException.kt
```diff
@@ -0,0 +1,7 @@
+package app.ehrenamtskarte.backend.graphql.shared.exceptions
```
seluianova, line 1: I would put exceptions direct in graphql package (just because they are "self-sufficient" enough), but this might be too personal

seluianova:
> I regenerated the graphql for front and back and still showing Apollo errors is this related?

@bahaaTuffaha these errors also occur in the main branch, so they should not be related

bahaaTuffaha:
Tested most of the things we usually test like card creation and application creation and approval ... Nice work ✅

Issue #2475, opened 2025-09-04
Migrate login screen to MUI


connium:
Replace the blueprint.js parts of the login form with MUI controls. Remember the forgot-password functionality.
💡 Please raise your hands if you recognize that this will create a massive amount of work.

PR #2515, opened 2025-09-22
2475: Login screen mui

f1sh1918:
### Short Description

Remove blueprint components and use mui for login and pw forgot screen

### Proposed Changes

- adjust login form and password reset to use mui components
- fix endIcon direction in the userMenu popover
- adjusted typography "manually" since the issue is still open and the sizes have to be defined there for each typo

Note: 
- i decided not to use the mui labels and helpers that we use for card input fields but keep separated labels, since there are some issues with the `OutlineInput`. We could adjust that if we have a real design for that

### Side Effects

- none

### Testing

1. Open Login screen, type in data and login
2. Logout
3. Click "Passwort vergessen", type in users mail address and send submit. 

All texts and buttons should still be properly aligned and sized with the mui components

### Resolved Issues

Fixes: #2475


connium:
@f1sh1918 what do you mean with the following?
> i decided not to use the mui labels and helpers that we use for card input fields but keep separated labels, since there are some issues with the OutlineInput. We could adjust that if we have a real design for that

What kind of design do you need?

seluianova (approved):
looks nice to me!
just one point: 
_I think the font size in the fields is currently larger than our standard font and it would be better to decrease it to the normal size._ (might be a personal impression❔)

image

bahaaTuffaha (approved):
LGTM ✅️

Issue #2464, opened 2025-09-01
Misleading error message when scanning a card that is not yet valid


f1sh1918:
### Describe the Bug

For nuernberg sozialpass, we have a digital and physical card.
Additionally we have a startDay that can be in the future

### Steps to Reproduce

1. Open administration and create a nuernberg card with a startDay that is in the future
2. Open the pdf and scan the physicial code in the app (verification)
3. See the error message: Karte konnte nicht verifiziert werden

### Expected Behavior

The card will be scanned and the preview window will be open with the card is not valid

Or show at least an error message that is clear understandable

seluianova:
<img width="382" height="792" alt="Image" src="https://github.com/user-attachments/assets/1f740b2a-09d4-4d0d-80aa-104e55eb5ef6" />

PR #2518, opened 2025-09-23
2464: Misleading error card not yet valid

f1sh1918:
### Short Description

We had no particular check for card info consistency according to a card that is not yet valid and just throw a random invalid message. For the verifier of the card this is very confusing

### Proposed Changes

- add `cardIsNotYetValid` check in `assertConsistentCardInfo` and throw an exception, 
- add the particular error messages

### Side Effects

- none

### Testing

Follow the issue description

And you should get this error
image

### Resolved Issues

Fixes: #2464

### DIFF
--- a/frontend/lib/identification/verification_workflow/verification_qr_code_processor.dart
+++ b/frontend/lib/identification/verification_workflow/verification_qr_code_processor.dart
@@ -49,8 +49,10 @@ Future<CardInfo?> verifyStaticVerificationCode(
   String projectId,
   StaticVerificationCode code,
 ) async {
-  assertConsistentCardInfo(code.info);
+  CardInfo cardInfo = code.info;
+  assertConsistentCardInfo(cardInfo);
   _assertConsistentStaticVerificationCode(code);
+  _assertCardIsAlreadyValid(cardInfo);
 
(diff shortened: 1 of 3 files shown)


seluianova (changes requested):
Just found a side effect in the activation:
Create a card with start date = tomorrow, try to activate it.
The error in the UI is **incorrect**:
image
While the error in the log is:
`I/flutter ( 8142): QrCodeParseException: card is not yet valid`

On the main branch we allow the activation of cards that are not yet valid.
It looks like this:
image

frontend/lib/identification/verification_workflow/verification_qr_code_processor.dart
```diff
@@ -48,24 +49,21 @@ Future<CardInfo?> verifyStaticVerificationCode(
...
   _assertConsistentStaticVerificationCode(code);
+  _assertCardIsAlreadyValid(cardInfo);
+
   if (!(await queryStaticServerVerification(client, projectId, code)).valid) {
     return null;
   }
   return code.info;
 }
 
-void assertConsistentCardInfo(CardInfo cardInfo) {
```
seluianova, line 59: 🙃 I think the location was fine here.
other assertions are here & the function is only used here

Issue #2477, opened 2025-09-04
Migrate ApplicationStatusBarHelpButton to MUI


connium:
The actual help content should be split into a popover component that is controlled by a standard MUI button.
- `src/bp-modules/applications/ApplicationStatusBarHelpButton.tsx`

For details, ask @andrew8er

Issue #2474, opened 2025-09-04
Migrate menu bar to MUI


connium:
Replace the blueprint.js button bar and user menu with MUI controls (e.g. [App Bar](https://mui.com/material-ui/react-app-bar/)).
💡 Please raise your hands if you recognize that this will create a massive amount of work.

connium:
I discussed this with @hauf-toni.

We like the mui app bar. The mail address of the user should move into the user menu, above the role.
For the visualization of the menu in the app bar, we would go with the rounded user icon https://mui.com/material-ui/react-app-bar/#app-bar-with-responsive-menu.
As logo, you can use the outlined icon of the entitlement card. https://nextcloud.tuerantuer.org/index.php/f/6588964

f1sh1918:
@hauf-toni could you please provide a quick design how it should look like with the username above the role

PR #2520, opened 2025-09-23
2477: Migrate popover

f1sh1918:
### Short Description

Migrate the popover for `application status` and `user role explanation` from blueprint to mui

### Proposed Changes

- create `PopoverWrapper.tsx` with a basic popover config using mui popover
- the popover uses an icon button but could be extended with a normal button if needed
- use the wrapper for application status and role explanation

Note: I didn't migrate the `Popover` that is used in the `UserMenu` since there is a separate task to replace the whole app bar #2474 

### Side Effects

- there is no arrow anymore. Not sure if this is mandatory. It could be done via lots of custom styles. Other components like `Popper` (no backdrop) or `Tooltip` (lots of additional styling, disabling hover, focus etc.) have other drawbacks
- So i decided that is the most suitable component for the purpose

### Testing

1. Login to bavarian entitlementcard (region manager/admin)
2. Go to "Eingehende Aufträge" click on "?" icon
3. Go to Benutzer verwalten (Region admin or project admin) click also on "?" icon
4. Check if Popover opens and content is okay

### Resolved Issues

Fixes: #2477


bahaaTuffaha (approved):
Tested on Chrome and Firefox ✅️

Issue #2476, opened 2025-09-04
Migrate user creation/management to MUI


connium:
Replace the blueprint.js parts of the user creation and management with MUI controls.
💡 Please raise your hands if you recognize that this will create a massive amount of work.

Issue #2335, opened 2025-07-10
Unify style of the dialogs in administration


seluianova:
### Describe the Problem
We have very different dialogs in the administration UI.
I think they should be unified.
<img width="451" height="184" alt="Image" src="https://github.com/user-attachments/assets/70858ccc-ae39-4277-bd06-b971c572f8de" />
<img width="460" height="176" alt="Image" src="https://github.com/user-attachments/assets/034b2c49-40e5-4c2d-abae-aec4acdaa19c" />
<img width="525" height="151" alt="Image" src="https://github.com/user-attachments/assets/af0be339-10ad-4cef-ad2e-a7682394e38a" />

f1sh1918:
@connium and @hauf-toni 
I'm about to unify our dialogs to the 4th option but we have some legacy dialogs for koblenz

<img width="458" height="349" alt="Image" src="https://github.com/user-attachments/assets/dcefe7c3-eb86-4cd2-862e-7db4f904c86b" />

After unifying it would be

<img width="479" height="223" alt="Image" src="https://github.com/user-attachments/assets/4e38a50d-7937-495b-bf16-fe2c07af1705" />

hauf-toni:
> That's perfectly fine with me.

same for me. you could remove the line break to make it look even better, but thats more on the icing on the cake side

Issue #2328, opened 2025-07-07
Migrate dialogs to MUI


connium:
Replace the blueprint.js dialogs with MUI dialogs.
Make sure, the primary action button is the right most button in the dialog.
💡 Please raise your hands if you recognize that this will create a massive amount of work.

PR #2524, opened 2025-09-25
2335: Unify dialogs

f1sh1918:
### Proposed Changes

- use mui dialogs thoughout the application
- remove `BaseDialog` as not needed
- adjust `ConfirmDialog` to use mui dialog with default onCancel action 
- use `ConfirmDialog`  where we don't have custom `onCancel` action (sufficient in 95% of all cases)
- use `react-router` `Link` for internal routing (prevents complete rerender)
- migrate user creation/edit to mui
- rename `CardTextField` property `showRequired` to `forceError` and add comment
- move `touch` handling to `CardTextField` to simplify usage of component

### Side Effects

- none

### Testing

- Test all occurrences of dialogs. If they look equal and work. Also check if the labels are okay. In some cases like `RejectDialog` i'm not really sure why `Abbrechen` is not good enough but i kept custom implementation
- Test internal links (Karte erneut erstellen, etc.)
- Test user creation / user edit
- Test field errors for card creation form fields (Nürnberg/Bayern/Koblenz)

### Resolved Issues

Fixes: #2335
Fixes: #2328
Fixes: #2476


f1sh1918:
### Open questions
- i removed the button tooltip for the text dialog but show a `AlertBox` if maxCharacters are exceeded.

image

Maybe the design could be improved a bit but i think this highlighting is more consistent and more visible to the user than just a tooltip

seluianova:
🙃 I would change the dialog title to “Benutzer löschen”, because we have such titles for other dialogs
image

seluianova (changes requested):
Required check doesn't work here:
`<Checkbox required>{t('ownAccountWarningConfirmation')}</Checkbox>`
I can edit (and delete) my own account without setting the checkbox.

Also the translation is missing 
image

seluianova:
Popover with the roles description is opened behind the dialog
image

administration/src/bp-modules/users/EditUserDialog.tsx
```diff
@@ -71,88 +73,83 @@ const EditUserDialog = ({
...
-          {regionIdOverride !== null || role === null || !rolesWithRegion.includes(role) ? null : (
-            <FormGroup label={t('region')}>
-              <RegionSelector onSelect={region => setRegionId(region.id)} selectedId={regionId} />
-            </FormGroup>
+        )}
+        <Callout intent='primary'>
+          {selectedUser?.id === me?.id ? (
+            <>
+              {t('youCanChangeYourOwnPassword')}{' '}
+              <Link to='/user-settings`' target='_blank' rel='noreferrer'>
```
seluianova, line 129: this link opens `/user-settings%60` for some reason

seluianova, line 129:
```suggestion
              <Link to='/user-settings' target='_blank' rel='noreferrer'>
```

f1sh1918:
> Required check doesn't work here:
> `<Checkbox required>{t('ownAccountWarningConfirmation')}</Checkbox>`
> I can edit (and delete) my own account without setting the checkbox.
> 
> ❓ is it intended that the fields in this dialog were not migrated?
> 
> Also the translation is missing 
> image

Yes there are separate issues for that. But translations have to be fixed

seluianova:
🙃 "Add user" dialog:
When a required field is not set, the button is disabled, but there are no errors or prompts indicating what the problem is.image
Before:
image

administration/src/util/translations/de.json
```diff
@@ -531,14 +533,18 @@
     "deleteUser": "Benutzer löschen",
     "delete": "Löschen",
     "deleteUserSuccess": "Benutzer erfolgreich gelöscht.",
-    "deleteUserConfirmPrompt": "Benutzer '{{mail}}' löschen?",
-    "deleteUserIrrevocableConfirmPrompt": "Möchten Sie den Benutzer '{{mail}}' unwiderruflich löschen?",
+    "deleteUserConfirmPrompt": "Benutzer löschen?",
```
seluianova, line 536:
```suggestion
    "deleteUserConfirmPrompt": "Benutzer löschen",
```

connium:
> image 

Is it possible to vertical align the checkbox at the top? IMO it's related to the bold text. Maybe we could also put the explanation in a `FormHelperText`.

f1sh1918:
> > > 🙃 I think it would be ideal to align the text in the alert boxes to the left 
> >
> > Well we could do that by removing the left padding but then the padding on both side won't be equal. So in any case it will look odd
> > 
> > Not sure if thats better. Removing the overall padding of the component is not an option i think
> 
> I agree with @seluianova that the padding oft the alert boxes is too large. I wouldn't remove the x-padding completely but try to match it with the padding of the text inputs (according to m3, text inputs and containers have a padding of 16dp).

Well the technical issue is that we can either set a fixed  padding or set it fullwidth not both because the Container width depends on the content by default but i did a override here because two different Container width next to each other would Look odd

f1sh1918:
@connium and @seluianova 
we can do a `space-around` (instead of center)for the flex content which would result in this
image

administration/src/bp-modules/applications/ApplicationCard.tsx
```diff
@@ -191,29 +168,29 @@ const ButtonsApplicationPending = (props: {
...
-              variant='contained'
-              disabled={props.primaryButtonHref === undefined}
-              href={props.primaryButtonHref}
-              startIcon={<CreditScore />}>
-              {' '}
-              {props.applicationStatus === ApplicationStatus.ApprovedCardCreated
-                ? t('createCardAgain')
-                : t('createCard')}
-            </Button>
+            <Link to={props.primaryButtonHref}>
```
bahaaTuffaha, line 182: Instead of Wrapping the button with Link we can use component:

```js
  component={Link}
  to={props.primaryButtonHref}
```

administration/src/mui-modules/application/DiscardAllInputsButton.tsx
```diff
@@ -14,12 +14,14 @@ const DiscardAllInputsButton = ({ discardAll }: { discardAll: () => void }): Rea
         Alle Eingaben verwerfen
```
bahaaTuffaha, line 14: 🔧  Need to be moved to the translation file.

bahaaTuffaha (approved):
1)  At Edit user dialog: despite the extra padding, there is a `<div>` as descendant of `<p>`.

Image

Image

2) Found unrelated issue: when you create a user it let pass with any text at the email input:

Image

administration/src/bp-modules/applications/ApplicationCard.tsx
f1sh1918: good find but it will require some custom hover styling if i apply the provided changes.
I think we have lots of such occurrences.
I added it to #2535

Issue #2507, opened 2025-09-17
Migrate GraphqlHandler


seluianova:
### Describe the Problem
- migrate GraphQLHandler
- migrate mutation and queries
- fix graphql-export command

### Additional Information
DataLoaders: https://docs.spring.io/spring-graphql/reference/request-execution.html#execution.batching.dataloader

https://github.com/spring-projects/spring-graphql/issues/69

PR #2526, opened 2025-09-25
2507: Migrate CardQueryService --> CardQueryController

seluianova:
### Proposed Changes
- fix "graphql-export" command
- migrate GraphQLHandler to GraphQLConfiguration
- migrate CustomDataFetcherExceptionHandler to GraphQLExceptionHandler
- migrate CardQueryService to CardQueryController
- migrate VerifyCardTest

**_The last commit is temporary, it should be removed before merging this PR.
It removes all files depending on javalin and allows to run SpringBoot app and test the migrated functionality._**

### Side Effects
The "export-graphql" command takes longer, because it requires the app to be started... Not sure if this can be improved
image

### Testing
- "Generate GraphQL specs" command can be tested --> check specs/backend-api.graphql
- verifyCardInProjectV2 query can be tested via Postman/curl (also covered by autotests)

### Resolved Issues
Part of: #2507


f1sh1918:
Hm i played around a little bit with graphql schema generation but couldn't find a good approach to avoid running spring.
I think its okay for now and maybe we find an improvement later

Issue #1707, opened 2024-10-23
Replace download app page


f1sh1918:
**Is your feature request related to a problem? Please describe.**
Currently we use f.e. [Download](https://download.bayern.ehrenamtskarte.app/) to show the app links for iOS and Android (origins from the activation PDF). This looks more like a phishing attempt than an official site.

![Image](https://github.com/user-attachments/assets/e712e317-1bda-4b58-870c-6d28fb1e2c1b)

**Describe the solution you'd like**
  ![Image](https://github.com/user-attachments/assets/cae3e83b-da09-4020-b8ac-73746fd04346)
- create a separate download route in our administration frontend and show the store links and store icons
- the links, texts and icons have to be used from the particular `projectConfig` if data is missing it has to be added there

- create redirects from download.app to this route in salt to redirect the existing links in the pdfs (this can be done in a separate salt issue)
- the redirects should work for all three projects (this can be done in a separate salt issue)

hauf-toni:
Design can be found 📌 [here](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=2736-3203&t=mzp7xTHZsyKeZfzX-1). Happy to hear your thoughts on this @f1sh1918 @andrew8er

andrew8er:
Looks good, I would use "Datenschutzerklärung" instead of "privacy policy" to have a consistent language though.

connium:
@connium Check which imprint and data privacy page should be linked.

connium:
We already have public data privacy pages for each app, which are linked from both the app itself and the app stores.
A general imprint doesn't exist yet and needs to be added to resolve this task.

connium:
blocked by #2403

f1sh1918:
LGTM

toni1991:
Download page has been added. @f1sh1918 said we have to change the nginx config accordingly to redirect e.g. _http://download.bayern.ehrenamtskarte.app/_ to the new react route. @andrew8er can you tell me where those nginx configs are?

PR #2530, opened 2025-09-29
1707: Added download app page

toni1991:
### Proposed Changes

- Added a download app page to the administration app

### Side Effects

- No side effect to expect

### Testing

Go to `/download` URL route for the different entitlement card issuers and check that the download page is matching the issuer

### Resolved Issues

Fixes: #1707


administration/src/mui-modules/download/DownloadPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const AppIcon = styled('img')({
+  height: 100,
+  width: 100,
+  border: '1px solid rgba(0, 0, 0, 0.2)',
+  borderRadius: '20%',
+  marginBottom: '1rem',
+})
+
+const linkStyle = {
```
f1sh1918, line 19: 🔧 we are currently migrating to MUI.
I think it would be a better approach to use MUI link component instead of setting here a custom styling for anchor or react-router `Link`
https://mui.com/material-ui/react-link/
There are also props for underlining, etc.
In the `theme.ts` file (if needed) general styling adjustments of components can be made.

I see that this would require for internal (react routing) to wrap that MUILink with a `Link` of `react-router` but this can be fixed later.
https://mui.com/material-ui/integrations/routing/
This can also be done in a separate task to keep the pr size small.

administration/src/mui-modules/download/DownloadPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Link } from 'react-router'
+
+import AppStoreLinks from '../../bp-modules/components/AppStoreLinks'
+import { ProjectConfigContext } from '../../project-configs/ProjectConfigContext'
+import { getBuildConfig } from '../../util/getBuildConfig'
+
+const AppIcon = styled('img')({
+  height: 100,
+  width: 100,
+  border: '1px solid rgba(0, 0, 0, 0.2)',
```
f1sh1918, line 14: According to the design the border should be `transparent`
But i think its a good improvement here for the koblenz icon

administration/src/mui-modules/download/DownloadPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { ios, android } = getBuildConfig(window.location.hostname)
+
+  return (
+    <Box display='flex' flexDirection='column' alignItems='center' mx='auto' mt={10}>
+      <AppIcon alt={`${config.name} app icon`} src={`/icons/${config.projectId}.png`} aria-hidden />
+      <Typography variant='h5' component='h1'>
+        {config.name}
+      </Typography>
+      <StyledAnchorLink href='https://tuerantuer.de/digitalfabrik'>Tür an Tür - Digitalfabrik gGmbH</StyledAnchorLink>
+      <Typography variant='h6' component='h2' mt={6}>
```
f1sh1918, line 48: 🔧 actually both headlines have `24 px` font-size and should be h5. 
https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=2990-12&t=hCUBzJ3LHHxxM7jZ-0
You can just override the `fontWeigth` here to `500`

administration/src/mui-modules/download/DownloadPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const config = useContext(ProjectConfigContext)
+  const { ios, android } = getBuildConfig(window.location.hostname)
+
+  return (
+    <Box display='flex' flexDirection='column' alignItems='center' mx='auto' mt={10}>
+      <AppIcon alt={`${config.name} app icon`} src={`/icons/${config.projectId}.png`} aria-hidden />
+      <Typography variant='h5' component='h1'>
+        {config.name}
+      </Typography>
+      <StyledAnchorLink href='https://tuerantuer.de/digitalfabrik'>Tür an Tür - Digitalfabrik gGmbH</StyledAnchorLink>
```
f1sh1918, line 47: ❓ it was not actually mentioned in the issue description but it probably makes sense to open this link with target blank, since its external?

Please mind security issues with target blank links, as described here
https://mui.com/material-ui/react-link/#security

administration/src/mui-modules/download/DownloadPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { styled } from '@mui/system'
+import React, { ReactElement, useContext } from 'react'
+import { useTranslation } from 'react-i18next'
+import { Link } from 'react-router'
+
+import AppStoreLinks from '../../bp-modules/components/AppStoreLinks'
+import { ProjectConfigContext } from '../../project-configs/ProjectConfigContext'
+import { getBuildConfig } from '../../util/getBuildConfig'
+
+const AppIcon = styled('img')({
```
f1sh1918, line 11: 🙃 If you don't want to set custom borders, it's also possible to use MUI `Box` with `component='img`  that would there you could use the predefined borders from mui and adjust the colors
But that's really optional
https://mui.com/system/borders/

Feel free to keep your solution

f1sh1918 (commented):
Very good work 👍 
Everything works fine and is implemented really clean so far.

I just added some minor improvements to be a bit more convenient with the mui component library. Also one links should be target blank, since external.

Rest is super fine. thx for you first contribution

administration/src/mui-modules/download/DownloadPage.tsx
toni1991: Done: MUILinks with inline styling

administration/src/mui-modules/download/DownloadPage.tsx
toni1991: Done: Box with MUI border

administration/src/mui-modules/download/DownloadPage.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { ios, android } = getBuildConfig(window.location.hostname)
+
+  return (
+    <Box display='flex' flexDirection='column' alignItems='center' mx='auto' mt={10}>
+      <Box
+        component='img'
+        sx={{
+          height: 100,
+          aspectRatio: 1,
+          marginBottom: '1rem',
```
f1sh1918, line 25: 🔧 we try to use spacing from mui which is defined here
https://mui.com/material-ui/customization/spacing/
One step is `8px` currently as default.
We may think about adjusting that using rem. 
```suggestion
          marginBottom: 2,
```

f1sh1918 (approved):
One small thing rest looks really good

Docmale:
The icons for apps are missing:
image

Issue #2506, opened 2025-09-16
Setup configuration loading


seluianova:
### Describe the Problem
Spring has its own configuration mechanism:
https://docs.spring.io/spring-boot/how-to/properties-and-configuration.html
I think it would be nice to adapt it.
BackendConfiguration refactoring will be needed.

f1sh1918:
@seluianova not sure if this pr could make conflicts, since i added active profiles
https://github.com/digitalfabrik/entitlementcard/pull/2599

PR #2534, opened 2025-09-29
2507: Migrate card mutation controller to SpringBoot

seluianova:
### Short Description
Next part of the migration

### Proposed Changes
- migrate CardMutationService --> CardMutationController
- add AuthInterceptor to obtain auth context from the request and put it to the GraphQL context
- graphql path is now set to `/` instead of `/graphql`
- server port is set to `8000` instead of `8080`
- add DevCorsConfig for dev env to allow local testing
- migrate tests: ActivateCardTest, CreateCardFromSelfServiceTest

**The last commit is temporary, it should be removed before merging this PR.
It removes all files depending on javalin and allows to run SpringBoot app and test the migrated functionality.**

### Side Effects
n/a

### Testing
**We can already use the UI for testing!**

Before starting administration, remove "Generate Administration GraphQL" step from the run configuration:
`<option name="RunConfigurationTask" enabled="true" run_configuration_name="Generate Administration GraphQL" run_configuration_type="js.build_tools.npm" />`

Before starting native apps, remove "Set build config" step from the run configuration:
`<option name="RunConfigurationTask" enabled="true" run_configuration_name="Set build config bayern" run_configuration_type="ShConfigurationType" />`

Otherwise, they will fail to start because the GraphQL schema in the backend has not been fully migrated yet.

Currently we can test:
- card creation in self-service `/erstellen`
- card activation and verification in the app

### Resolved Issues
Part of: #2507


backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
```diff
@@ -41,18 +43,26 @@ import extensionStartDayOrNull
...
 import java.time.LocalDate
 import java.util.Base64
 import javax.crypto.KeyGenerator
 
-@Suppress("unused")
-class CardMutationService {
-    private val logger: Logger = LoggerFactory.getLogger(CardMutationService::class.java)
+@Controller
+class CardMutationController(
+    @Suppress("SpringJavaInjectionPointsAutowiringInspection")
```
f1sh1918, line 61: ❓ where can i find some docs where its explain what these autowirings do?

f1sh1918 (approved):
Great work! 👍 
Tested all use cases you mentioned
Test also ran successfully!

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
seluianova: well there is the large official documentation on the entire concept.
I think it has to be read from the very beginning: https://docs.spring.io/spring-framework/reference/core/beans.html

in a nutshell:
- a "bean" is an object managed by Spring;
- beans are annotated with specific annotations, e.g. `@Controller`, `@Configuration`, `@Service` and other;
- when some class is annotated as a bean, Spring will create an object of this class and put it to the Spring Context;
- beans from the Spring Context can be injected into other beans as constructor params without manually creating the objects; for example here I don't create backendConfiguration object manually, but it is injected by Spring from the context;
- "autowire" means injecting a bean from the Spring Context.

regarding this particular warning:
it will be removed when we do https://github.com/digitalfabrik/entitlementcard/issues/2506
this warning means that our current BackendConfiguration is not a normal "bean". It should have the @Configuration annotation, but this requires some refactoring to be done first.

f1sh1918:
@seluianova i have one general question. I think one of the main purposes of migrating to spring was that it support kotlin coroutines? Will we migrate all the mutations and queries then in a separate step?

seluianova:
> I think one of the main purposes of migrating to spring was that it supports kotlin coroutines. Will we migrate all the mutations and queries then in a separate step?

I was against it actually 😀
but overall, yes, this could be another round of migration.
this would involve reworking the entire content of queries/mutations; currently, I don't touch them and only apply general Spring concepts.

PR #2542, opened 2025-10-01
2474: Migrate menu bar

f1sh1918:
### Proposed Changes

- create NavigationItems.tsx component, add mui-types package
- migrate UserMenu.tsx component to MUI
- use NavigationItems in HomeController
- migrate NavigationBar.tsx to mui, add entitlement_icon_outline.svg icon
- fix NavigationBar.test.tsx

### Side Effects

- should be none

### Testing

1. Login and test all items in the navigation bar
2. Open user menu and test menu items
3. Test logout

### Resolved Issues

Fixes: #2474

### DIFF
--- a/administration/src/bp-modules/NavigationBar.tsx
+++ b/administration/src/bp-modules/NavigationBar.tsx
@@ -16,36 +16,34 @@ const NavigationBar = (): ReactElement => {
 
   return (
     <AppBar position='sticky' color='transparent'>
-      <Stack>
-        <Toolbar
-          sx={theme => ({
-            flexWrap: 'wrap',
-            [theme.breakpoints.down('lg')]: {
-              padding: 1.5,
-            },
-          })}>
-          <SvgIcon fontSize='large'>
-            <EntitlementIcon />
-          </SvgIcon>
-          <NavLink to='/' style={{ color: 'inherit', textDecoration: 'none', margin: '0 12px' }}>
-            <Stack>
+      <Toolbar
+        sx={theme => ({
+          flexWrap: 'wrap',
+          [theme.breakpoints.down('lg')]: {
+            padding: 1.5,
+          },
+        })}>
+        <SvgIcon fontSize='large'>
+          <EntitlementIcon />
+        </SvgIcon>
+        <NavLink to='/' style={{ color: 'inherit', textDecoration: 'none', margin: '0 12px' }}>
+          <Stack>
+            <Typography>
+              {config.name} {t('administration')}
+            </Typography>
+            {!region ? null : (
               <Typography>
-                {config.name} {t('administration')}
+                {region.prefix} {region.name} {`(${process.env.REACT_APP_VERSION})`}
               </Typography>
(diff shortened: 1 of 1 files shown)


f1sh1918:
Here is a first draft @hauf-toni and @connium 
image
Maybe you can give me a short opinion if thats okay, especially the user menu part

f1sh1918:
> > Here is a first draft @hauf-toni and @connium
> > Maybe you can give me a short opinion if thats okay, especially the user menu part
> 
> **App bar:** I recognized that some users don't recognize the menu bar. So I'm wondering if we should set it to dark blue background color and white text color. @hauf-toni I'm also fine with this solution, as it matches the current (non-MUI) design and the elevation maybe already tackles the issue.
> 
> **Menu:** I wouldn't set the email address in bold but use regular font weight instead. IMO, it catches too much focus.

Well switching to blue needs a lot of refactoring, since we cannot use our Text Buttons (wrapped by react-router nav-link) since they don't offer a white text variant, but have to create `MenuItems` with Icons and text to fit in the theme. To use this menu item as "internal links" we have to first migrate `MuiLinks` and configure them to use internal routing or apply custom theming to `NavLinks` which is just a temporary solution

I would recommend to adjust this in the separate issue when we migrated and configured the other components

connium:
If changing the background color is a lot of effort, then just skip that and keep the style you showed.

Regarding the font weight: At the moment there are two things, that "trigger" me:
1. The bold font of the user's email address
2. The different font size. The (important) menu items have a smaller font size than the informative texts.

@hauf-toni, can you suggest a solution here from, please? Feel free to vote against my opinion — focus on the user and not on me.

f1sh1918:
> If changing the background color is a lot of effort, then just skip that and keep the style you showed.
> 
> Regarding the font weight: At the moment there are two things, that "trigger" me:
> 
> 1. The bold font of the user's email address
> 2. The different font size. The (important) menu items have a smaller font size than the informative texts.
> 
> @hauf-toni, can you suggest a solution here from, please? Feel free to vote against my opinion — focus on the user and not on me.

Well typos will in any case change with the other pr, so i wouldn't too much focus on that. We have button font-sizes of 14px and could use `body2` (also 14px) for mail address

seluianova:
🙃 I think this separator looked nice
image

seluianova:
🙃 I think this looks like a tiny extra space
image

seluianova:
🙃 I think the tooltip text is incorrect, as this menu contains not only settings.
probably the tooltip is not really needed here, for me the icon is clear enough
image

seluianova:
> This "side effect" is a focus and has a11y reasons

@f1sh1918 ah, I think then it should be a rectangle around the entire element. just 2 lines don't look like a focus

f1sh1918:
> here

Are you sure you referenced(link) correctly. I even didn't find anything about currently selected item.
I found sth in `MenuItem` component which we currently don't use here.
I would tend to leave that issue open at the moment and handle it separate if needed

Issue #2555, opened 2025-10-07
Logout redirect does not work for /cards if form is filled


f1sh1918:
### Describe the Bug

Normally the url changed to default ('/') if we logout.
But if you start creating a new card (fill out a form field)
This redirect does not work (maybe blocked by `useBlockNavigation` ?) , because there should be also a prompt when leaving the page with a card form filled

### Steps to Reproduce

1. Login with region admin/manager
2. Karte erstellen
3. Type something in the form fields
4. Click on user menu and `logout`
5. Check that the url stays the same (/cards)

### Expected Behavior
After logout the user will be redirect to `/`
So after another login you will start on the dashboard

PR #2557, opened 2025-10-07
2555: Logout during card creation

f1sh1918:
### Short Description

Apparently redirect and signOut do not work when `useBlockNavigation`
Therefore i created a custom hook which first ensures that the navigation has to be finished that, before doing the signOut in a callback. That triggers the prompt correctly.

### Proposed Changes

- add custom hook
- use customHook to ensure the prompt will be triggered onSignout

### Side Effects

- none

### Testing

Check the bug description

### Resolved Issues

Fixes: #2555


f1sh1918:
If you feel there is a better or easier solution maybe by adjusting the `useBlockNavigation` let me know.
The only other working way i found was using a `setTimeout` which is way more hacky and unpredictable

Issue #2479, opened 2025-09-04
Replace by `notistack` components


connium:
We already have notistack (partly) setup, use it throughout the whole app. We should move the `SnackbarProvider` from `src/mui-modules/application-verification/ApplicationApplicantController.tsx` to `src/App.tsx`.
- `src/bp-modules/AppToaster.tsx` (remove)
- `src/bp-modules/cards/ImportCardsInput.test.tsx`
- `src/bp-modules/cards/hooks/useCardGenerator.test.tsx`
- `src/bp-modules/util/cardGenerationError.tsx`

For details, ask @andrew8er

PR #2562, opened 2025-10-08
2479: Replace toast by notistack

toni1991:
### Short Description

Replace Blueprint toast component with notistack.

### Proposed Changes

- AppToaster has been replaced by AppSnackbar
- All usages of notistack have been consolidated
- AppSnackbar has convenient methods to enqueue messages with different variants (error, info, success, etc.) which declutters the code a little bit
- Snackbar colors have been adjusted to use the ones from MUI

### Side Effects

No side effects to expect

### Testing

There should be no difference beside the slight UI differences of the actual toast/snackbar. Every toast that was shown before the changes has to be there with the new snackbar.

- Bavaria: Check that approve, reject and delete application results in a successful (green) snackbar 
- Koblenz card self service: Check that creating invalid card results in a error (red) snackbar
- Check that changing any region related setting shows a successful (green) snackbar
- Check that importing stores shows information on successful import in a default (white) snackbar
- Check that creating, editing and deleting a user results in successful (green) snackbars

### Resolved Issues

Fixes: #2479

### DIFF
--- a/frontend/assets/l10n/app_de.json
+++ b/frontend/assets/l10n/app_de.json
@@ -103,6 +103,7 @@
     "checkRequired": "Prüfung nötig",
     "checkingCode": "Der QR-Code wird durch eine Server-Anfrage geprüft.",
     "codeExpired": "Der eingescannte Code ist bereits am {{expirationDate}} abgelaufen.",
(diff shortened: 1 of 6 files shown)


f1sh1918:
@toni1991 i think the first testing bulletpoint belongs to Bavaria not koblenz?

Issue #2523, opened 2025-09-24
Migrate project settings to MUI


f1sh1918:
### Describe the Problem

Migrate the project settings to MUI

Issue #2484, opened 2025-09-04
Migrate personal settings to MUI


connium:
Replace the blueprint.js parts of the personal settings (e.g. activity log, change password) with MUI controls.
💡 Please raise your hands if you recognize that this will create a massive amount of work.

Issue #2482, opened 2025-09-04
Migrate region settings to MUI


connium:
Replace the blueprint.js parts of the region settings with MUI controls.
💡 Please raise your hands if you recognize that this will create a massive amount of work.

PR #2569, opened 2025-10-10
2482, 2484 & 2523: project settings, user settings and region settings use MUI components

toni1991:
### Short Description

Make user settings and region settings (and used components) use MUI components

### Proposed Changes

All dependencies to blueprint components have been removed and replaced by MUI components. Styling has been consolidated to some extend.

Be aware: a single dependency to blueprint is still there in `ChangePasswordForm.tsx` and `PepperSettings.tsx` that will be resolved with #2559

### Side Effects

- no side effects to expect

### Testing

- Compare project settings, region settings and user settings pages before and after changes. Everything should be functional like it was (also the dialogs for changes of the notes and data privacy texts)
- Slight UI changes are to expect, but should be minimal (e.g. font size/weight changes, borders of the UI Card components, input components)

### Resolved Issues

As the code of those tickets uses shared components and the UI idea is the same the following have been resolved at once:

- Fixes: #2482 
- Fixes: #2484 
- Fixes: #2523

### DIFF
--- a/administration/src/bp-modules/project-settings/ApiTokenSettings.tsx
+++ b/administration/src/bp-modules/project-settings/ApiTokenSettings.tsx
@@ -169,7 +169,7 @@ const ApiTokenSettings = ({ showPepperSection }: ApiTokenSettingsProps): ReactEl
                     <TableCell>{formatDate(item.expirationDate)}</TableCell>
                     <TableCell>
                       <Delete
-                        sx={{ pointer: 'cursor', display: 'block' }}
+                        sx={{ cursor: 'pointer', display: 'block' }}
                         color='error'
                         onClick={() => setTokenToDelete(item.id)}
                       />


administration/src/bp-modules/project-settings/ApiTokenSettings.tsx
```diff
@@ -116,58 +115,67 @@ const ApiTokenSettings = ({ showPepperSection }: ApiTokenSettingsProps): ReactEl
...
+                value={expiresIn}
+                label={t('validPeriod')}
+                onChange={e => setExpiresIn(e.target.value)}>
+                <MenuItem value={1}>1 {t('month')}</MenuItem>
+                <MenuItem value={3}>3 {t('months')}</MenuItem>
+                <MenuItem value={12}>1 {t('year')}</MenuItem>
+                <MenuItem value={36}>3 {t('years')}</MenuItem>
+              </Select>
+            </FormControl>
+            <Button sx={{ minWidth: 'auto' }} onClick={() => createToken({ variables: { expiresIn } })}>
```
andrew8er: ❓  Does the `sx={{ minWidth: 'auto' }}` have any effect? I couldn't spot any.

administration/src/bp-modules/project-settings/ApiTokenSettings.tsx
```diff
@@ -116,58 +115,67 @@ const ApiTokenSettings = ({ showPepperSection }: ApiTokenSettingsProps): ReactEl
...
+            </Button>
+          </Box>
           {createdToken !== null && (
             <>
-              <p>{t('newToken')}:</p>
-              <NewTokenText> {createdToken}</NewTokenText>
+              <Typography variant='body2'>{t('newToken')}:</Typography>
+              <NewTokenText>
+                <Typography variant='body1'>{createdToken}</Typography>
+              </NewTokenText>
```
andrew8er: 🔧 While you are tackling this component anyway, I suggest we also get rid of the `styled-components` usage. It is on maintenance mode now, and MUI provides an [alternative](https://mui.com/system/styled/) with a similar API. Here however, I would just use a custom `<Box sx={{...}}>`.

administration/src/bp-modules/regions/data-privacy-policy/DataPrivacyOverview.tsx
```diff
@@ -68,14 +64,24 @@ const DataPrivacyOverview = ({ dataPrivacyPolicy, regionId }: RegionOverviewProp
...
+        </Typography>
+        <TextField
+          sx={{
+            mb: 0,
+          }}
+          slotProps={{
+            input: {
+              sx: { fontSize: theme.typography.body2.fontSize },
+            },
+          }}
```
andrew8er: ❓ I'm not sure, why we are using `body2` everywhere instead of the default `body`, but I think it would be possible to define this is in the `mui-modules/theme.ts` file.

Maybe we can later sort this out in the theme without going through every component later.

administration/src/bp-modules/project-settings/ApiTokenSettings.tsx
toni1991: Unfortunately yes. In the Stack the Button gets squished by the select which expands. See example with removed minwidth and same button below, which has intended width:

image

administration/src/bp-modules/project-settings/ApiTokenSettings.tsx
toni1991: :+1: I removed all `styled-components` in the folders `regions`, `project-settings` and `user-settings`

administration/src/bp-modules/project-settings/ApiTokenSettings.tsx
```diff
@@ -118,62 +147,39 @@ const ApiTokenSettings = ({ showPepperSection }: ApiTokenSettingsProps): ReactEl
...
+                </TableRow>
+              </TableHead>
+              <TableBody>
+                {tokenMetaData.map(item => (
+                  <TableRow key={item.id}>
+                    <TableCell>{item.creatorEmail}</TableCell>
+                    <TableCell>{formatDate(item.expirationDate)}</TableCell>
+                    <TableCell>
+                      <Delete
+                        sx={{ pointer: 'cursor', display: 'block' }}
```
seluianova:
```suggestion
                        sx={{ cursor: 'pointer', display: 'block' }}
```

PR #2581, opened 2025-10-14
2507: Migrate application mutations to SpringBoot

seluianova:
### Short Description
Migrate application mutations

### Proposed Changes
- migrate EakApplicationMutationService --> EakApplicationMutationController
- implement MultipartGraphQLHandler to handle multipart requests (used for attachments upload)
- migrate tests
- refactor GraphQLRequestInterceptor --> now we can inject params from graphql context like this:
image

**The last commit is temporary, it should be removed before merging this PR.
It removes all files depending on javalin and allows to run SpringBoot app and test the migrated functionality.**

### Side Effects
Apparently, we have a rather complicated flow for handling application attachments.
To upload attachments, we use multipart requests, which are not supported by Spring Boot out of the box.

_As far as I understand, using multipart requests in GraphQL is generally not recommended:_
(see https://graphql.org/learn/file-uploads/ and https://www.apollographql.com/blog/file-upload-best-practices#approach-3-multipart-upload-requests-not-recommended)

I had to add a custom handler to process multipart requests; See **MultipartGraphQLHandler** for details - I tried to add a detailed KDoc.

We could probably refactor this logic in the future; for example, by splitting the process into two steps:
1. REST request to upload the files and receive file ids bzw. references in the response
2. GraphQL request that includes these file references in JSON to create the application

### Testing

**All actions with applications can be tested (creation & management)**

Before starting administration, remove "Generate Administration GraphQL" step from the run configuration:
`<option name="RunConfigurationTask" enabled="true" run_configuration_name="Generate Administration GraphQL" run_configuration_type="js.build_tools.npm" />`

### Resolved Issues
Part of: #2507

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/multipart/MultipartGraphQLHandler.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/multipart/MultipartGraphQLHandler.kt
@@ -45,21 +45,17 @@ class MultipartGraphQLHandler(
         // Collect all parts
         val partsMap = request.parts.associateBy { it.name }
 
-        // Parse operations
-        val operationsJson = partsMap["operations"]?.inputStream
-            ?: throw GraphQLMultipartParseException("Missing 'operations' part")
-        val operationsNode = mapper.readTree(operationsJson)
+        val operationsNode = partsMap["operations"]?.inputStream?.use { mapper.readTree(it) }
+            ?: throw GraphQLMultipartParseException("Missing 'operations' node")
 
-        // Parse file mapping
-        val mapJson = partsMap["map"]?.inputStream
-            ?: throw GraphQLMultipartParseException("Missing 'map' part")
-        val substitutions = mapper.readValue<Map<String, List<String>>>(mapJson)
+        val mapNode = partsMap["map"]?.inputStream?.use { mapper.readValue<Map<String, List<String>>>(it) }
+            ?: throw GraphQLMultipartParseException("Missing 'map' node")
 
         // Collect uploaded files
-        val files = request.parts.filter { substitutions.containsKey(it.name) }
+        val files = request.parts.filter { mapNode.containsKey(it.name) }
 
         // Substitute file references into the GraphQL variables
-        substitutions.forEach { (key, paths) ->
+        mapNode.forEach { (key, paths) ->
             if (key == "operations" || key == "map") {
                 throw GraphQLMultipartParseException("Invalid file key: '$key'")
             }
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/GraphQLSchemaConfiguration.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/GraphQLSchemaConfiguration.kt
@@ -92,8 +92,5 @@ class GraphQLSchemaConfiguration(
      * This prevents Spring from autoconfiguring its own, avoiding a conflict.
      */
     @Bean
-    fun runtimeWiringConfigurer(): RuntimeWiringConfigurer =
-        RuntimeWiringConfigurer { wiringBuilder ->
-            wiringBuilder.scalar(this.uploadScalar)
-        }
+    fun runtimeWiringConfigurer(): RuntimeWiringConfigurer = RuntimeWiringConfigurer { it.scalar(this.uploadScalar) }
 }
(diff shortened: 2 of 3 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/auth/SignInMutationController.kt
```diff
@@ -6,32 +6,32 @@ import app.ehrenamtskarte.backend.graphql.auth.types.AuthData
...
+import org.springframework.graphql.data.method.annotation.ContextValue
 import org.springframework.graphql.data.method.annotation.MutationMapping
 import org.springframework.stereotype.Controller
 
 @Controller
-class SignInMutationController(
-    private val request: HttpServletRequest,
-) {
+class SignInMutationController {
+    private val logger = LoggerFactory.getLogger(SignInMutationController::class.java)
```
andrew8er: 🔧 This could be declared at the file level, since it is not specific to the instance of the containing class. Even better, it could be declared with `by lazy {}`, to speed up the program startup.

backend/src/main/resources/application.properties
```diff
@@ -4,3 +4,6 @@ spring.graphql.schema.locations=classpath:/specs/
 spring.graphql.schema.file-extensions=graphql
 # TODO-2506: Server port is currently duplicated in config.yml but actually read from here; the config source should be unified
 server.port=8000
+spring.servlet.multipart.enabled=true
+spring.servlet.multipart.max-file-size=5MB
+spring.servlet.multipart.max-request-size=10MB
```
andrew8er: ❓ Depending on how exactly this is calculated, this might result in only one part being allowed, since a request always contains some extra data.

andrew8er (approved):
A few suggestions.

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/multipart/MultipartGraphQLConfiguration.kt
```diff
@@ -0,0 +1,26 @@
...
+@Configuration
+class MultipartGraphQLConfiguration {
+    /**
+     * Creates a router function to handle multipart/form-data requests for the GraphQL endpoint
+     */
+    @Bean
+    @Order(-10)
+    fun graphQlMultipartRouter(handler: WebGraphQlHandler): RouterFunction<ServerResponse> =
+        router {
+            val multipartHandler = MultipartGraphQLHandler(handler)
```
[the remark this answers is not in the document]
seluianova: I believe the current configuration complies with the single responsibility principle.
The router configures the route; the handler encapsulates multipart parsing / processing logic.

andrew8er:
> there is no quasi standard. I checked various libs and docs to create this handler to preserve the approach we used with javalin and keep the current API.

With "quasi spec" I meant, there is none in the GraphQL spec directly, but there is a spec outside of it, that seems at least to have a certain support: https://github.com/jaydenseric/graphql-multipart-request-spec

There is an implementation for Spring Boot, but it does not look like it's being widely used: https://github.com/nkonev/multipart-spring-graphql

We can keep it like this for now, but we should look for alternatives.

seluianova:
> With "quasi spec" I meant, there is none in the GraphQL spec directly, but there is a spec outside of it, that seems at least to have a certain support: https://github.com/jaydenseric/graphql-multipart-request-spec

it won't work for us as is, because we have to parse the request, extract files, modify the json.
I saw all these libs.

seluianova:
> We can keep it like this for now, but we should look for alternatives.

@andrew8er I added my suggestion to the PR description. feel free to add yours.

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/multipart/MultipartGraphQLHandler.kt
```diff
@@ -0,0 +1,130 @@
...
+        val future = graphQlHandler.handleRequest(graphQlRequest)
+            .map { response ->
+                ServerResponse.ok()
+                    .headers { it.putAll(response.responseHeaders) }
+                    .contentType(selectResponseMediaType(serverRequest))
+                    .body(response.toMap())
+            }
+            .toFuture()
+
+        return ServerResponse.async(future)
```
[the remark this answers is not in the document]
seluianova: I hoped that separating the statements would make this hell at least a little more readable.

andrew8er:
All good!

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/auth/SignInMutationController.kt
seluianova: okay, let's gradually update this everywhere so that we use the same approach in all classes

Issue #2591, opened 2025-10-21
Typography set body2 as default


f1sh1918:
### Describe the Problem

Since body2 is our most used typography variant, it should be used as default

### Describe Your Preferred Solution

- set `body2` as default variant
- remove `variant: 'body2'` from all `<Typography>` components
- set `body1` for typograhphy components without variant

PR #2592, opened 2025-10-21
2591: Typo body2 default

f1sh1918:
### Short Description

Since body2 is our most used typography variant, it should be used as default to avoid unneeded declarations

### Proposed Changes

- set `body2` as default variant
- remove `variant: 'body2'` from all `<Typography>` components
- set `body1` for typograhphy components without variant
- fix mentioned issues
- set `CSSBaseline` to have apply in the `body` html tag
- add `#root` styles in `theme.ts` and remove `index.css`

### Side Effects

- none

### Testing

Smoke test texts within the application

### Resolved Issues

Fixes: #2591
Fixes: #2605

### DIFF
--- a/administration/src/bp-modules/statistics/components/StatisticsLegend.tsx
+++ b/administration/src/bp-modules/statistics/components/StatisticsLegend.tsx
@@ -1,4 +1,4 @@
-import { Stack } from '@mui/material'
+import { Stack, Typography } from '@mui/material'
 import { styled } from '@mui/system'
 import React, { ReactElement } from 'react'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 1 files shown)


seluianova:
Different font size
image

seluianova:
Different font size for "Erweiterung der Datenschutzerklärung" configured in `/region`
image

seluianova:
"Mehr anzeigen" was smaller before (not sure which size we want here)
image

f1sh1918:
> 🙃 Different space between letters. _And probably off topic, but for the distance between the field and the error message is also different here._ image

Fixing this is super annoying because we have to remove the stack and the gap and create customs spaces for each item. I really don't wanna do this

Issue #2451, opened 2025-08-30
Adapt exception handling


connium:
### Describe the Problem

Migrate CustomDataFetcherExceptionHandler

PR #2594, opened 2025-10-23
2451: Migrate exceptions handling to SpringBoot

seluianova:
### Short Description

Migrate remaining exceptions handling to SpringBoot

### Proposed Changes

Exceptions handling was already migrated, I just added some missing details (see inline comments) and updated tests

### Side Effects
**Unauthorized** and **Forbidden** errors are now returned with HTTP status code **200** (previously **401** and **403**).
I added corresponding GraphQlExceptionCodes so that the correct errors are displayed in the Administration UI.

Spring Boot’s GraphQL implementation always returns HTTP 200 for valid GraphQL requests, even when execution errors occur.
This follows the GraphQL specification: a single request can contain multiple operations, some of which may fail while others succeed.
In such cases, the expected response includes both data and errors in the body.
So the idea is that GraphQL errors are passed in the json body and it is like a different layer of errors than HTTP errors.

**Perhaps it is technically possible to find a workaround and return 401 and 403 somehow but I could not find the way to do it.** Let me know if you think it's a problem for us.

### Testing
Different error cases can be checked in the administration UI and in the app.
Full list of our errors can be found here: https://github.com/digitalfabrik/entitlementcard/blob/9144ae13635795dc915cb774f0438e2095afe1e0/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/shared/types/GraphQLExceptionCode.kt

### Resolved Issues
Fixes: #2451

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/DevCorsConfigurer.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/DevCorsConfigurer.kt
@@ -2,7 +2,7 @@ package app.ehrenamtskarte.backend
 
 import app.ehrenamtskarte.backend.config.BackendConfiguration
 import app.ehrenamtskarte.backend.config.Environment
-import org.springframework.context.annotation.Configuration
+import org.springframework.stereotype.Component
 import org.springframework.web.servlet.config.annotation.CorsRegistry
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
 
@@ -11,8 +11,8 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
  * Without these headers, browsers block requests from the frontend to the backend API.
  * This setup allows all origins during development but is disabled in other environments.
  */
-@Configuration
-class DevCorsConfig(
+@Component
+class DevCorsConfigurer(
     @Suppress("SpringJavaInjectionPointsAutowiringInspection")
     private val config: BackendConfiguration,
 ) : WebMvcConfigurer {
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/EntryPoint.kt
@@ -25,6 +25,7 @@ import graphql.schema.GraphQLSchema
 import org.jetbrains.exposed.sql.exists
 import org.jetbrains.exposed.sql.transactions.transaction
 import org.slf4j.LoggerFactory
+import org.springframework.boot.WebApplicationType
 import org.springframework.boot.runApplication
 import org.springframework.context.support.beans
 import java.io.File
@@ -112,6 +113,7 @@ class GraphQLExport : CliktCommand("graphql-export") {
 
     override fun run() {
         val springContext = runApplication<BackendApplication> {
+            webApplicationType = WebApplicationType.NONE
             addInitializers(
                 beans {
                     bean { config }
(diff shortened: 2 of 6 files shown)


administration/src/errors/GraphQlErrorMap.tsx
```diff
@@ -161,9 +161,20 @@ const graphQlErrorMap = (extensions?: ErrorExtensions): GraphQLErrorMessage => {
...
-    default:
+    case GraphQlExceptionCode.NotImplemented: {
       return {
-        title: defaultError.title,
+        title: i18next.t('errors:functionNotAvailable'),
+      }
+    }
+    case GraphQlExceptionCode.ProjectNotFound: {
+      return {
+        title: i18next.t('errors:projectNotFound'),
```
seluianova: we didn't have an error message for "project not found" before and displayed "seite nicht gefunden"

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/EakApplicationMutationController.kt
```diff
@@ -235,7 +234,7 @@ class EakApplicationMutationController(
     ): Boolean =
         transaction {
             val application =
-                ApplicationEntity.findById(applicationId) ?: throw NotFoundException("Application not found")
```
seluianova: removed NotFoundException from queries/mutations because it is mapped to "seite nicht gefunded" error in administration

andrew8er:
In the [GraphQL best practices guide](https://graphql.org/learn/serving-over-http/#status-codes), they say that modern GraphQL server implementations should respond with an error code:

> However, for responses with the `application/graphql-response+json` media type, the server will reply with a 4xx or 5xx status code if a valid GraphQL request fails to execute.

This is probably a little bit misleading. What they want to say is (I guess) that a query which cannot produce any data should return 4xx/5xx (as you stated). Queries that can [return partial data can return that plus an error](https://nearform.com/open-source/urql/docs/basics/errors/).

So the question is, do we count unauthenticated requests as completely failed?

andrew8er:
> I also found this doc: [GraphQL over HTTP](https://graphql.github.io/graphql-over-http/draft/#sec-application-json), and it says:
>
> The server SHOULD use the 200 status code for every response to a well-formed GraphQL-over-HTTP, independent of any GraphQL request error or GraphQL field error raised.

This is for servers that use `application/json` which is considered legacy. We should probably use `application/graphql-response+json` instead.

I find the (quasi) standard pretty spotty and confusing in this regard. But maybe we should look at this from the client's perspective: how should a client handle errors? Currently we have network related errors (timeouts, DNS errors, etc), HTTP-level errors with some semantic meaning (unauthorized, etc) and application level errors. 

I would have no problem with ditching the HTTP-level errors in between and solely have the network errors (which we cannot avoid) and application level errors (which we can totally design to our needs). I always found a small but well designed list of application error codes (in addition to human readable error messages) to be easier to deal with as a client developer. The gRPC standard has a pretty good list of errors that we could take inspiration from: https://grpc.io/docs/guides/status-codes/

administration/src/util/translations/de.json
```diff
@@ -628,7 +628,8 @@
     "applicationToCsvError": "Es ist ein Fehler beim Export des Antrags aufgetreten.",
     "freinetPersonDataInvalid": "Erforderliche Personendaten von Freinet sind nicht verfügbar. Die Kartendaten konnten deshalb nicht übertragen werden.",
     "freinetCardDataInvalid": "Die Kartendaten können nicht an Freinet übertragen werden, weil diese ungültig sind.",
-    "freinetDataTransferNotPermitted": "Für diese Region ist keine Datenübermittlung erlaubt."
+    "freinetDataTransferNotPermitted": "Für diese Region ist keine Datenübermittlung erlaubt.",
```
andrew8er: ❓ The wording here is maybe a bit misleading and harsh. It should probably not be "erlaubt" but "eingerichtet" or "vorgesehen".

Issue #2583, opened 2025-10-15
Setup sentry for SpringBoot


seluianova:
### Describe the Problem

Sentry setup should be migrated / adapted to SpringBoot

It was introduced here: https://github.com/digitalfabrik/entitlementcard/pull/2209

PR #2599, opened 2025-10-27
2583: Setup sentry spring boot

f1sh1918:
### Short Description

Currently sentry was configured for flutter to collect errors but there is a particular implementation for spring boot that should be used and replace old sentry configuration

### Proposed Changes

- remove old sentry config (javalin)
- set active spring profile on backend start to be able to use different `application.properties` files (production, development)
- add `application-production.properties` including dns,release, sample rate for sentry for production
- add `application-development.properties` including disable sentry for development

### Side Effects

- We get a warning on startup, since Sentry tries to get options on first initialization before the application properties were read, but they get initialized afterwards. I tried a lot with the logback but couldn't solve the issue.

### Testing
- set environment to `production` in your `config.local.yml` and DON'T forget to revert this later :)
- add this in your `SignInMutationController`
```
        try {
            throw Exception("This is a test.")
        } catch (e: Exception) {
            Sentry.captureException(e)
        }
```
- rerun backend, you should see: 
```
c.g.ajalt.clikt.core.CoreCliktCommandKt  : The following 1 profile is active: "production"
```
- login with user
- you should find a new issue here: https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/?project=7&statsPeriod=90d
- to ensure that reports will not be send for testing add the same snippet to f.e. `approveApplicationStatus` and run `Test backend`
- change the backend environment to `development` and rerun backend service, you should see:
```
c.g.ajalt.clikt.core.CoreCliktCommandKt  : The following 1 profile is active: "development"
```
- if you now run tests or trigger error, these shouldn't be send to sentry

### Resolved Issues

Fixes: #2583

### DIFF
--- a/backend/src/main/resources/application-development.properties
+++ b/backend/src/main/resources/application-development.properties
@@ -1,4 +1,2 @@
-# Sentry
-sentry.enabled=false
-sentry.dsn=
-sentry.release=unknown
+# Add additional properties for development profile here
+
--- a/backend/src/main/resources/application.properties
+++ b/backend/src/main/resources/application.properties
@@ -10,3 +10,5 @@ spring.servlet.multipart.max-request-size=11MB
 spring.graphql.graphiql.enabled=true
 # Sentry
 sentry.enabled=false
+sentry.dsn=
+sentry.release=unknown


andrew8er (approved):
One small suggestion, otherwise LGTM.

I could live with the error for now. It should be fixed, when we update the upstream dependency.

Issue #2575, opened 2025-10-13
Migrate unstable_usePrompt to useBlocker


f1sh1918:
### Describe the Problem

We are still using an unstable function within our `useBlockNavigation`
There is a stable function now from `react-router`
We also use a workaround that was introduced here: https://github.com/digitalfabrik/entitlementcard/pull/2557 that fix the issue that was discussed here:https://github.com/digitalfabrik/entitlementcard/pull/2557#issuecomment-3399236910

### Describe Your Preferred Solution

- migrate to https://reactrouter.com/api/hooks/useBlocker
- remove custom hook `useNavigateWithCallback` if possible
- test if the issue in #2555 is fixed without that custom hook

### Related Issues

[](https://github.com/digitalfabrik/entitlementcard/pull/2557)

PR #2606, opened 2025-10-27
2575: Replace `unstable_usePrompt`

andrew8er:
### Short Description

Replaces our usage of `unstable_usePrompt` from `react-router`. This was more complicated than initially thought, since the `AuthContext.signOut()` caused a navigation even if it should be blocked by the blocker. It turned out that react-router's `navigate()` function does not return a rejected promise in this case, as I would have expected. The return type of the function is not [well defined anyway](https://github.com/remix-run/react-router/issues/12348).

Instead I introduced a new route, which does all logout actions and then redirects to the login screen.

### Proposed Changes

- Use `useBlocker()` for consistent cross-browser behavior.
- Add a `/logout` route that clears the login token and then redirects to home.
- Rename the `<KeepAliveToken>` component to `<AutomticLogoutDialog>` to better reflect what it does and is.
- Un-nest layout in `<AutomticLogoutDialog>`.

### Side Effects

None (?)

### Testing

- Add some cards manually in http://localhost:3000/cards/add
- Try to navigate away, observe the blocker dialog.
- Logout, confirm you are logged out by pasting an authenticated route like http://localhost:3000/applications
- To test the automatic logout dialog, temporarily set the JWT expiry time to something like 185 seconds in `JwtService.kt`: `.withExpiresAt(Date.from(Instant.now().plus(185, ChronoUnit.SECONDS)))` (line 24).

### Resolved Issues

Fixes: 
- #2575 
- #2555

### DIFF
--- a/administration/src/mui-modules/theme.ts
+++ b/administration/src/mui-modules/theme.ts
@@ -255,5 +255,12 @@ export const theme = (config: ProjectConfig): Theme =>
           },
         },
       },
+      MuiDialogContentText: {
+        styleOverrides: {
+          root: {
+            color: 'inherit',
+          },
+        },
+      },
     },
   })


administration/src/Logout.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { ReactElement, useContext } from 'react'
+import { Navigate } from 'react-router'
+
+import { AuthContext } from './AuthProvider'
+
+export const Logout = (): ReactElement => {
+  const { signOut } = useContext(AuthContext)
+
+  signOut()
```
f1sh1918: 🔧 i think you should use a `useEffect` here to avoid side effects die to rerendering

```
  useEffect(() => {
    signOut()
  }, [signOut])

```

administration/src/shared/components/BlockerDialog.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <Dialog
+      open={p.blocker.state === 'blocked'}
+      onClose={p.blocker.reset}
+      aria-labelledby='alert-dialog-title'
+      aria-describedby='alert-dialog-description'>
+      <DialogTitle id='alert-dialog-title'>{p.title}</DialogTitle>
+      <DialogContent>
+        <DialogContentText id='alert-dialog-description'>{p.message}</DialogContentText>
```
f1sh1918: 🔧 Here `<Typography>` component should be used to ensure that we use `body2` here like in other dialogs as body text

andrew8er: I set it in the theme to `inherit`.

andrew8er:
@seluianova 

>    open /cards/add
>    enter name
>    click the close (x) button on the form --> warning appears; but it doesn't matter whether I click "Abbrechnen" or "Bestätigen", the form closes anyway and the changes are lost
>
>Not sure if we need a warning in this case at all (there wasn't one before).

I fixed this by comparing the current/new nav path.

>    open /cards/add
>    enter data in any fields except name (e.g. Geburtsdatum or Nürnberg-Pass ID if it is the Nürnberg project)
>    go to user menu - Benutzereinstellungen (or logout or any other action) --> no warning appears, the changes are lost
>
>I think it might be fine to fix that separately as minor issues, but let me know what you think

I expanded the check to also test the birth date, but checking the extensions would be a little bit more involved. We would need a method for every extension that checks, if it deviates from its default state.

Issue #2573, opened 2025-10-13
Setup graphiql for SpringBoot


seluianova:
### Describe the Problem
Setup graphiql (or similar tool)
It should be available under `/graphiql`

_(we had graphiql for javalin, but currently it's not available and should be configured/migrated)_

PR #2613, opened 2025-10-28
2573: Setup graphiql

f1sh1918:
### Short Description

For documentation and testing of our graphql mutations and queries, we want to setup graphiql

### Proposed Changes

- remove old graphiql resource files
- enable graphiql

### Side Effects

- none

### Testing
1. Go to http://localhost:8000/graphiql
2. Check that graphiql is available and open the docs
image

Schema should be available

### Resolved Issues

Fixes: #2573

### DIFF
--- a/backend/src/main/resources/application.properties
+++ b/backend/src/main/resources/application.properties
@@ -7,3 +7,4 @@ server.port=8000
 spring.servlet.multipart.enabled=true
 spring.servlet.multipart.max-file-size=5MB
 spring.servlet.multipart.max-request-size=11MB
+spring.graphql.graphiql.enabled=true


f1sh1918:
I've tried couple of time to get it work with auto-installation, as far as i see it should only be:
`spring.graphql.graphiql.enabled=true`
I couldn't find any additional recent documentation
But i got no resource found `renderExample.js` when i open the endpoint
So i created the `RestController` manually

You can test if you can find a solution for that. Just comment out /delete the `GraphiQLController` and add this to (application.properties)
```
`spring.graphql.graphiql.enabled=true`
```

seluianova (approved):
perfect, works like a charm now :)

PR #2616, opened 2025-10-29
2506: Backend configuration improvements

seluianova:
### Short Description
"Lazy" implementation of the task #2506.
Initially I wanted to merge config.yml and application.properties to create a single configuration managed by SpringBoot.
But it seems it would be better to postpone that because:
- this would require many changes on our side AND in the salt repo;
- anyway we plan to refactor our configuration later (`we were thinking about creating one configuration package in this monorepo and store there all the configuration and check how we can pass the needed parts to web, native and backend (c)`)

### Proposed Changes
- get rid of the warning "SpringJavaInjectionPointsAutowiringInspection"
- make the app use server.port and server.host from the config.yml, because we override these values in the salt repo

### Side Effects
no

### Testing
Just a basic check that the application starts and works

### Resolved Issues
Fixes: #2506

### DIFF
--- a/backend/src/test/kotlin/app/ehrenamtskarte/backend/IntegrationTestConfig.kt
+++ b/backend/src/test/kotlin/app/ehrenamtskarte/backend/IntegrationTestConfig.kt
@@ -1,6 +1,7 @@
 package app.ehrenamtskarte.backend
 
 import app.ehrenamtskarte.backend.config.BackendConfiguration
+import kotlinx.io.files.FileNotFoundException
 import org.springframework.boot.test.context.TestConfiguration
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Primary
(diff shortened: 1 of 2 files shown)


backend/src/test/kotlin/app/ehrenamtskarte/backend/IntegrationTestConfig.kt
```diff
@@ -3,10 +3,12 @@ package app.ehrenamtskarte.backend
 import app.ehrenamtskarte.backend.config.BackendConfiguration
 import org.springframework.boot.test.context.TestConfiguration
 import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Primary
 
 @TestConfiguration
 class IntegrationTestConfig {
     @Bean
+    @Primary
     fun backendConfiguration(): BackendConfiguration {
```
f1sh1918: 🙃  This is optional, since you didn't touch it
But shouldn't this throw a `FileNotFoundException` ?
This can also be inlined

```suggestion
    fun backendConfiguration(): BackendConfiguration = BackendConfiguration.load(ClassLoader.getSystemResource("config.test.yml")
        ?: throw FileNotFoundException("src/test/resources/config.test.yml")
    )
```
And it could be inlined also

Issue #2580, opened 2025-10-14
Improve keyboard navigation for user menu


f1sh1918:
### Describe the Problem

Currently the content of the UserMenu can't be accessed via keyboard.
This seem to rely that inside a MUI `Menu` component you have to deal with the focus by yourself.
Another idea was to switch to `Popover` component instead of menu, then the items inside can be accessed but then you are trapped inside the user menu

### Describe Your Preferred Solution

- find a solution either with `Menu` or `Popover` that you can select the Menu content (buttons) but also be able to leave the Menu by keyboard

f1sh1918:
Yes it looks like other elements block the accessibility like Stack and Typography

PR #2621, opened 2025-10-30
2580: User menu keyboard navigation

f1sh1918:
### Short Description

We cannot access the user menu items by keyboard

### Proposed Changes

- remove stack and adjust styling

### Side Effects

- none

### Testing

1. Login, navigate via keyboard
2. Open the menu, then tab to access the menu items and use "Up" and "Down"

### Resolved Issues

Fixes: #2580


seluianova:
@kalairubachandrakumaran-oss it's about this menu. when you open it, press the "tab" key, and you should be able to switch between menu items using the "up" and "down" keys
image

kalairubachandrakumaran-oss:
> @kalairubachandrakumaran-oss it's about this menu. when you open it, press the "tab" key, and you should be able to switch between menu items using the "up" and "down" keys image

Okay then it works fine :)

Issue #2612, opened 2025-10-28
Verification request can be re-sent for approved application


seluianova:
### Describe the Bug

The administrator can resend the verification request **if the application has already been approved** but verification action is no longer available to the contact person.

### Steps to Reproduce
1. Go to '/beantragen', submit an application that requires verification
2. Login as region admin, approve the created application --> **the status of the application is now "approved" but the button to re-send a verification request is still available**

<img width="609" height="526" alt="Image" src="https://github.com/user-attachments/assets/6c6198cd-bb36-4e5c-afb7-7841499b91c6" />

When admin clicks "Anfrage erneut senden" - the contact person receives the verification request BUT approval or rejection is no longer available.

<img width="760" height="69" alt="Image" src="https://github.com/user-attachments/assets/6e94c819-abf5-45bc-925e-f6ab01aaa263" />

### Expected Behavior

When application is approved, it should not be possible to re-send the verification request

PR #2625, opened 2025-11-03
2612: Verification request pending only

f1sh1918:
### Short Description

Currently it's possible to send a verification re-request mail even if the application is already approved

### Proposed Changes

- enable sending verification request mail only for pending applications
- fix alignment of status in `VerificationListItem`

### Side Effects

- none

### Testing

- Approve an application, ensure that a verification request mail cannot be sent anymore

### Resolved Issues

Fixes: #2612


bahaaTuffaha (approved):
Tested on Chrome ... works as expected ✅️

Issue #2511, opened 2025-09-20
Statistics ~ Tooltip disappears when hovering near bar edges & too much spacing between tooltip text and number


kalairubachandrakumaran-oss:
### Describe the Problem
Problem 1: When hovering over the bar, the tooltip is not always visible -> if the cursor is near the edge of the bar, the tooltip is only partially shown
Hint: when hovering in the middle of the bar, the tooltip works correctly

Problem 2: Tooltip text and number (for example "Erstelle Karten" & "10" ) have too much spacing

### Describe Your Preferred Solution

Solution to Problem 1: Adjust tooltip positioning so it stays visible even when hovering at the bar's edge.
Solution to Problem 2: Reduce the spacing between the label and the value in the tooltip so that the text and number appear closer together

### Related Issues

#2347 

### Additional Information

Problem 1:

<img width="300" height="100" alt="Image" src="https://github.com/user-attachments/assets/31ad55ab-02bd-4e4f-9b1d-a09e3cc057ba" />

https://github.com/user-attachments/assets/c241f7b2-ed1a-40dc-801a-b2056826d6a2

Problem 2:

<img width="300" height="70" alt="Image" src="https://github.com/user-attachments/assets/5a1b378f-be78-4147-968e-70c5e8158482" />

PR #2642, opened 2025-11-04
2511: Fix statistic tooltip

f1sh1918:
### Short Description

Check the two issue in the issue description

### Proposed Changes

- fix styling of the tooltip (square), move the appearance to the right of the mouse cursor

### Side Effects

- none

### Testing

- go to statistics (Region Admin or Project Admin) and hover over the bar item

### Resolved Issues

Fixes: #2511


andrew8er (approved):
Works 👍

f1sh1918:
As we cannot know how long the label will be, we show it in multiple lines. (I have to set a fixed width for an absolute positioned element.) 
And that the number is vertically centered is fine for me.
From a design perspective it will look odd if I put the number to the bottom

We can ask @hauf-toni for a better design/alignment here if you think it's really worth to spend more time in it. I think it's good enough

Issue #2645, opened 2025-11-05
Improve test render wrapper


f1sh1918:
### Describe the Problem

TBD

PR #2646, opened 2025-11-05
2645: Improve test render wrapper

f1sh1918:
### Short Description

I created this little draft to improve our structure for mocking HOCs like Providers etc.
I found it a bit annoying to have multiple functions like `renderWithRouter` or `renderWithTranslation` and in the end `renderWithTheme`. Here we can now granularly say what we want to have by setting a boolean. And we could also enable provider by default.

### Proposed Changes

- create one single function that includes all our providers (easily extendable)
- also a custom provider can be passed if needed (wrapper prop)

### Side Effects

- none

### Testing

Running tests should be enough

### Resolved Issues

Fixes: #2645

### DIFF
--- a/administration/src/bp-modules/cards/CreateCardsButtonBar.test.tsx
+++ b/administration/src/bp-modules/cards/CreateCardsButtonBar.test.tsx
@@ -3,11 +3,13 @@ import React from 'react'
 
 import { initializeCard } from '../../cards/Card'
 import bayernConfig from '../../project-configs/bayern/config'
-import { renderWithOptions } from '../../testing/render'
+import { CustomRenderOptions, renderWithOptions } from '../../testing/render'
 import { getTestRegion } from '../user-settings/__mocks__/Region'
 import CreateCardsButtonBar from './CreateCardsButtonBar'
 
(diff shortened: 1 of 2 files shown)


administration/src/testing/render.tsx
```diff
@@ -43,3 +30,78 @@ export const renderWithTranslation = (
...
+    localization = false,
+    snackbar = false,
+    wrapper,
+  } = options
+  const wrappers: ((props: { children: ReactNode }) => ReactNode)[] = []
+  if (wrapper) {
+    const CustomWrapper = wrapper
+    const customWrapper = (props: { children: ReactNode }) => <CustomWrapper {...props} />
+    wrappers.push(customWrapper)
+  }
```
andrew8er: 🔧  I think this is a case where React's `createElement()` should be used:

```suggestion
  const wrappers: ((props: { children: ReactNode }) => ReactNode)[] = []

  if (wrapper) {
    wrappers.push((props: { children: ReactNode }) => createElement(wrapper, props, null))
  }
```

andrew8er (approved):
Tested 👍

Issue #2444, opened 2025-08-30
Upgrade backend to Gradle 9.2


connium:
### Describe the Problem

For our backend service, we should upgrade to the current Gradle version (at the time of writing ~~8.14.2~~ 9.2.0) for both the backend and the Android frontend build. This would theoretically allow us to reuse Gradle daemons and safe a little bit a memory (if the IDE is set up correctly), although this is optional.

Gradle could also use the newish [toolchain](https://docs.gradle.org/current/userguide/toolchains.html) support to state desired toolchain used to build the project. This would simplify the initial project setup slightly.

### Describe Your Preferred Solution

- Upgrade to the latest Gradle release for `/backend` as well as `/fontend/android`

andrew8er:
Heads up, current Gradle version as of 2025-08-31 is [Gradle 9.0](https://gradle.org/whats-new/gradle-9/) with interesting features.

PR #2654, opened 2025-11-10
2444: Upgrade `/backend` to Gradle 9.2.0

andrew8er:
### Short Description

Upgrade the backend build system to Gradle 9.2.0 and update plugins and dependencies.

### Proposed Changes

- Upgrade Gradle 9.2.0
- Upgrade all build plugins
- Upgrade most dependencies (except testcontainers, due to missing `org.testcontainers:postgresql` dependency)
- Fix some version references
- Remove redundant JUnit dependencies, since `spring-boot-starter-test` [already defines those](https://docs.spring.io/spring-boot/reference/testing/spring-applications.html)

### Side Effects

None

### Testing

- Start backend
- Run backend tests

### Resolved Issues

Fixes: #2444

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -91,8 +91,6 @@ dependencies {
     runtimeOnly(libs.postgresql.postgresql)
 
     testImplementation(libs.expediagroup.graphql.kotlin.client)
-    testImplementation(libs.junit.jupiter)
-    testImplementation(libs.junit.jupiter.params)
     testImplementation(libs.jetbrains.kotlin.test)
     testImplementation(libs.jetbrains.kotlin.test.junit5)
     testImplementation(libs.jetbrains.kotlinx.coroutines.test)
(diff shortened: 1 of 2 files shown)


andrew8er:
> 💭 just wondering why this triggered Coverage report for administration

Maybe the PR is to be merged into a base PR which is then to be merged into main ultimately? The [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-files-changed-in-a-pull-request) aren't entirely clear about this, but the `branch` filter alone would not otherwise allow a run here.

> 🙃 there is a deprecation warning that could be addressed, but currently not critical I guess
Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.

Yes, I have those practically always in all Gradle projects. They originate from plugins, but the should be resolved, when those get updated. The worst that can happen is, that we cannot immediately upgrade to newer Gradle versions, should they arrive.

f1sh1918:
@andrew8er i would suggest to finally merge the spring migration pr and not merge more prs into it

PR #2661, opened 2025-11-12
2660: Upgrade `/frontend/android` to Gradle 9.2

### LINKED ISSUE #2205: Setup Android Gradle project
### Describe the Problem
We have a gradle project in the android frontend module, but it's not properly configured.
Image

When I try to set it up, I get the error:
```
The project is using an incompatible version (AGP 8.10.0) of the Android Gradle plugin. Latest supported version is AGP 8.8.0-alpha05
```

I assume this means that the android plugin is not working properly, although we haven't encountered any side effects so far.

Also there is a warning:

```
WARNING: Unsupported Kotlin plugin version.
The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `2.0.21` that might work differently than in the requested version `2.1.21`
```

I think it would be nice to add gradle project setup and get rid of errors/warnings (dependencies upgrade/downgrade seems to be required) because they look confusing.

Leave as is

### LINKED ISSUE #2445: Use new Gradle features to improve developer experience
### Describe the Problem

After upgrading Gradle, we can make use of new fetures:
- reuse Gradle daemons and safe a little bit a memory (if the IDE is set up correctly), although this is optional.
- use the newish [toolchain](https://docs.gradle.org/current/userguide/toolchains.html) support to state desired toolchain used to build the project. This would simplify the initial project setup slightly.

### Related Issues
#2444

### LINKED ISSUE #2660: Upgrade frontend/android to Gradle 9.2
### Describe the Problem

Update the Gradle version used in the frontend/android build to 9.2.0 to enable usage of JDK 25. This also requires updating Flutter to the latest version.

- Upgrade Flutter to 3.35.7
- Upgrade Gradle to 9.2.0

### Related Issues

- #2281

andrew8er:
### Short Description

To make this project fully compatible with the new JDK 25 (LTS) release, we need to upgrade the `/frontend/android` Gradle version to 9.2.0, which also necessitates upgrading Flutter to the newest version.

This PR also cleans up and removes a few IntelliJ project files/directories, which I think are unnecessary, confusing and generate a lot of churn, since they are constantly changed by IntelliJ and show up as modified in git.

If you have any objections or use cases for these, please raise your hand.

### Proposed Changes

- Upgrade Flutter to 3.35.7
- Upgrade Gradle in `/frontend/android` to 9.2.0
- Remove the nested IntelliJ project files in `/frontend`: this directory marks an IntelliJ project root. I do not think we have a nested project here, since the frontend is configured as an IntelliJ module and it is quiet confusing to have three projects nested inside of each other. Instead, the IntelliJ module definition file is moved from `/frontend/.idea` to `/frontend`.
   We already have a nested IntelliJ project root in `/frontend/android`, which currently cannot get rid of, since IntelliJ does apparently not support nested Gradle projects. 
- Remove all files in the nested IntelliJ project directory `/frontend/android/.idea` and ignore its content by default. All of the module information will be inferred from the Gradle file, so we do not need them. We can selectively whitelist files individually, should the need arise.

### Testing

Build and test frontend for Android.

### Resolved Issues

- Fixes #2660 
- Fixes #2205
- Fixes #2445

### DIFF
--- a/frontend/android/app/build.gradle.kts
+++ b/frontend/android/app/build.gradle.kts
@@ -31,7 +31,7 @@ android {
     ndkVersion = "28.1.13356709"
 
     defaultConfig {
-        minSdk = 23
+        minSdk = flutter.minSdkVersion
         targetSdk = 36
         multiDexEnabled = true
         versionCode = flutter.versionCode
--- a/frontend/ios/Podfile.lock
+++ b/frontend/ios/Podfile.lock
@@ -146,7 +146,7 @@ EXTERNAL SOURCES:
 SPEC CHECKSUMS:
   connectivity_plus: cb623214f4e1f6ef8fe7403d580fdad517d2f7dd
   device_info_plus: 21fcca2080fbcd348be798aa36c3e5ed849eefbe
-  Flutter: e0871f40cf51350855a761d2e70bf5af5b9b5de7
+  Flutter: cabc95a1d2626b1b06e7179b784ebcf0c0cde467
   flutter_secure_storage: 1ed9476fba7e7a782b22888f956cce43e2c62f13
   geolocator_apple: 1560c3c875af2a412242c7a923e15d0d401966ff
   GoogleDataTransport: aae35b7ea0c09004c3797d53c8c41f66f219d6a7


f1sh1918:
Are you sure that modules.xml can be deleted safely?
Didn't find too much proper documentation

Why do we stick to java 17 in compile options (`app/build/gradle`)?
```
  compileOptions {
        sourceCompatibility = JavaVersion.VERSION_17
        targetCompatibility = JavaVersion.VERSION_17
    }
```

For me everything works fine on ios and android, so i guess it should be fine

Some files have to be pushed (Podfile.lock, AppFrameworkInfo.plist (since the minimun ios version changed from 12.0 to 13.0, same for project.pbxproj)

I added a commit with the meta data changes

f1sh1918:
> @f1sh1918
> 
> > Are you sure that modules.xml can be deleted safely?
> > Didn't find too much proper documentation
> 
> The file references the `frontend.iml` file in the same directory, which is _also_ referenced from the `/.idea/modules.xml` of the outer project. This is a pretty confusing setup: the same module was part of two nested projects.
> 
> The `/frontend/.idea/modules.xml` file is only relevant if you open the `/frontend` directory in IntelliJ. I don't think, there is a need for that, please correct me if I'm wrong here. Instead, I moved the `frontend.iml` to `/frontend/frontend.iml` and adjusted the path of the outer `/.idea/modules.xml`.
> 
> > Why do we stick to java 17 in compile options (app/build/gradle)?
> 
> I did not touch that, this is not relevant to the toolchain in use, it just configures what language level constructs are permitted and which bytecode is emitted by the compiler.
> 
> > Also build.gradle.kts changed with the update of flutter
> 
> This was automatically changed by either `fvm use 3.38.0` or `fvm flutter pub get`, I'm not 100% sure.

Well i followed the update steps and found these files as changed that were not commited, so i added them

I'm a bit confused about your mentioned version, since we use `3.35.7` here afaics.

f1sh1918:
is there any blocker for this pr @andrew8er ? i smoke tested it looks fine

seluianova (approved):
lgtm 👍
smoke tested android & ios

Issue #2295, opened 2025-06-24
Fix multiple tag creation during release


f1sh1918:
### Describe the Bug

It appears to be that if we do a release for a particular platform, f.e. trigger `delivery_backend_administration` also the `-all` tag will be created and not only `-web` tag.
In general we only need one tag depending if its a all platform release or a specific platform.
Additionally basing the last release notes (2025.6.4-web on 2025.5.0-all) ended it wrong release notes (changelog)
Lots of changelog items were missing.
This was manually fixed by basing on `2025.5.0-web` , since we did some hotfix release web only.

### Steps to Reproduce
Not really sure how to reproduce that
tba.

PR #2662, opened 2025-11-13
2295: Update app toolbelt

f1sh1918:
### Short Description

We don't have to set platform for git tags, since we only create one tag per release without tag now.
There are issues in finding previous tag for github when creating release notes, so we should have the same tag structure for all releases.
We only name the release that is create on this tag including platform (which was hardcoded before)

So if we create a release for web/backend it will be called f.e.

[web] 2025.11.0 (225)
[native] 2025.11.0 (225)

or just 
2025.11.0 (225)
for not platform specific release (all)

There are also some breaking changes in the app-toolbelt for releases, that i addressed here

### Proposed Changes

- add platform parameter to notify release to be able to set a particular release name as described in the description
- remove platform parameter for bump-up (create tag) because we don't use it there
- update to recent app-toolbelt (not released yet)
- add formatted release notes to mattermost
- add promote message for mattermost

### Side Effects

- should be none

### Testing

n/A

### Resolved Issues

Fixes: #2295

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1339,7 +1339,7 @@ workflows:
             - promote_github_release:
                 context:
                     - deliverino
-                platform: web
+                platform: native
                 requires:
                     - promote_android
                     - promote_ios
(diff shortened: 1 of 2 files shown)


steffenkleinle (approved):
LGTM

Issue #2533, opened 2025-09-29
Adjust documentation


seluianova:
### Describe the Problem

- [ ] Working with the GraphQL API `./docs/working-with-graphql-api`

Check if there are more places where adjustments are needed.

PR #2669, opened 2025-11-17
2533: Actualize docs for SpringBoot

seluianova:
### Short Description
Actualize docs for SpringBoot

### Proposed Changes
- adjusted items 1-5 & fixed numeration

### Side Effects
no

### Testing
n/a

### Resolved Issues
Fixes: #2533

### DIFF
--- a/docs/working-with-graphql-api.md
+++ b/docs/working-with-graphql-api.md
@@ -1,30 +1,55 @@
 # Working with the GraphQL API
 
-> This article is about how to create your first query in the backend and use it in the web frontend.
-> In this case you want to create a simple graphql query that fetches applications for a region.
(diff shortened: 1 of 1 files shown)


andrew8er (commented):
One thing that is still missing, is a short introduction to the overall workflow. It is hidden in this text way down, but I think a paragraph at the top would help to set the reader's mental model. 

It is important to know that we have a code-first approach here, where the spec is generated from the server code, not the other way.

It is also worth mentioning that we have one unified API for both the administration and public queries for whatever reason.

I'm thinking of the process:
- Server code defines server API spec
- Spec is generated from server code into file xy
- File xy is used to validate queries for administration and frontend

seluianova:
@andrew8er I reworked the doc and added the points you mentioned, pls have a look

andrew8er (approved):
Very good 👍

Issue #2626, opened 2025-11-03
Replace styled components with MUI


connium:
### Describe the Problem

Once the MUI migration is complete, the remaining styled components should be replaced by MUI components.

PR #2671, opened 2025-11-18
2626: Replace styled components

f1sh1918:
### Short Description

Replace the last styled component usages and use mui instead

### Proposed Changes

- Replace styled components with mui styled
- remove dependency

### Side Effects

- none

### Testing

- check the components where styled was changed, like login screen centered
- application status bar

### Resolved Issues

Fixes: #2626


bahaaTuffaha (approved):
Tested on chrome by creating an application ✅

Issue #2418, opened 2025-08-19
Incorrect error message when user attempts to withdraw an approved application


seluianova:
### Describe the Bug
When user attempts to withdraw an **approved** application, incorrect error message is shown:

<img width="1003" height="724" alt="Image" src="https://github.com/user-attachments/assets/9bb1a85c-3fc3-404f-948e-83b80020c84b" />

### Steps to Reproduce
We don't show the "Antrag zurückziehen" button for approved application.
However, if the user opens the link before the application is approved (or rejected), the button will be available:

1. Create an application
2. Open /antrag-einsehen/{accessKey} page
3. At the same time, log in as a regional admin **in a new tab** and approve the created application
4. Return to /antrag-einsehen/{accessKey} tab, do not refresh the page
5. Click "Antrag zurückziehen" button, see misleading error

### Expected Behavior
The message should not be misleading. 
I think it would be nice show the current status in the message, like "Der Antrag wurde bereits akzeptiert."

PR #2674, opened 2025-11-18
2418: Withdraw approved application

f1sh1918:
### Short Description

We have here an issue that only happens if the applicant has an open applicant view (mail link) and want to withdraw the application while in the same moment the application was rejected or withdrawn.
I implemented a more general error, since this issue is very rare and will likely not happen. I think its not worth in implementing separate errors for all possible`Status`

### Proposed Changes

- add a check if the application is still in a `Status.Pending`, and throw an `OutofSync` error if that's the case
- adjusted onCompleted, since false will never happen. Instead i throw an error if application will not be found. I think the error message before was completely pointless

### Side Effects

- no

### Testing

1. Create an application with an organization to verify
2. Open the link that the applicant received and keep it in your browser
3. Open a new tab and login and approve the application
4. Switch to the other tab (Applicant view) and withdraw the application
5. You should receive a message that the current state is out of sync and you should refresh

### Resolved Issues

Fixes: #2418

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/EakApplicationMutationController.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/EakApplicationMutationController.kt
@@ -123,9 +123,12 @@ class EakApplicationMutationController(
         transaction {
             val application = ApplicationEntity.find { Applications.accessKey eq accessKey }
                 .singleOrNull() ?: throw InvalidLinkException()
-            if (application.status != Status.Pending) throw InvalidApplicationStatusException()
-            application.status = Status.Withdrawn
-            true
+            if (application.status != Status.Pending) {
+                throw InvalidApplicationStatusException()
+            } else {
+                application.status = Status.Withdrawn
+                true
+            }
         }
 
     @GraphQLDescription("Verifies or rejects an application verification")


backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/EakApplicationMutationController.kt
```diff
@@ -120,12 +121,11 @@ class EakApplicationMutationController(
...
-                ApplicationEntity.find { Applications.accessKey eq accessKey }.single().status = Status.Withdrawn
-                true
-            } catch (e: IllegalArgumentException) {
-                false
-            }
+            val application = ApplicationEntity.find { Applications.accessKey eq accessKey }
+                .singleOrNull() ?: throw InvalidLinkException()
+            if (application.status != Status.Pending) throw InvalidApplicationStatusException()
+            application.status = Status.Withdrawn
+            true
```
andrew8er: 🔧 
- I strongly suggest to always use curly braces (`{}`) and an indented block for `if` constructs, unless it is used as an expression.
- Also, when in a tail position, it is easier to see when (or when not) the code following the if clause is executed:

```suggestion
            val application = ApplicationEntity.find { Applications.accessKey eq accessKey }
                .singleOrNull() ?: throw InvalidLinkException()
         
            if (application.status != Status.Pending) {
                throw InvalidApplicationStatusException()
            } else {
                application.status = Status.Withdrawn
                true
            }
```

❓ The return type here does not carry any useful information?

andrew8er (commented):
A more general question:

If the application was already approved/rejected and the applicant withdraws their application, would it not be less irritating to just silently drop the application? That would likely mean less puzzled faces and probably less support requests, even if this should be a pretty rare case. Showing such an unspecific error message does not help anyone IMHO.

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/application/EakApplicationMutationController.kt
andrew8er: You are right, mutations need to return some data.

PR #2682, opened 2025-11-19
Fix pack_backend job to pick up the correct artifact

seluianova:
### Short Description

`./gradlew assemble` produces 2 artifacts by default: backend.tar and backend-boot.tar
but the pack_backend job expected one tar only.

### Proposed Changes
- disable creation of backend-boot.tar

### Side Effects

-

### Resolved Issues

Fixes: #

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,1 +1,1 @@
-{"versionName":"2025.11.2","versionCode":214}
+{"versionName":"2025.11.3","versionCode":215}


seluianova:
> i added a suggestion :)

ok, I don't remember what I tried and why it didn't work, but yes, your suggestion works :)
thanks!

andrew8er (approved):
Looks good.

I think, in the long term, we should seriously consider transitioning to containers. That would make a whole bunch of custom scripts obsolete and also completely eliminate issues like https://tasks.tuerantuer.org/projects/infra/work_packages/4793/

Issue #2603, opened 2025-10-27
Distinguish components and functions by using case sensitive naming


f1sh1918:
### Describe the Problem

Currently we also use Pascal case naming for file that are no react components.

### Describe Your Preferred Solution

- use Pascal case for react components and typescript classes
- use lowercase for functions, hooks, etc

### Related Issues

[](https://github.com/digitalfabrik/entitlementcard/pull/2590#discussion_r2462740266)

connium:
Maybe you can add linting rules for this, or at least document it in coding guidelines.

connium:
I rarely used snake_case because I found it unusual to write. To me, kebap-case was the most fluent way to write. But I'm also a dinosaur 🦖

PR #2736, opened 2025-12-10
2603: Improve file naming and code style

andrew8er:
### Short Description

Updates `prettier` to the latest version, which has some formatting changes. Also slightly adjusts the prettier config.

### Proposed Changes

- Update `prettier` to version 3.7.4
- Adjust prettier [config](https://prettier.io/docs/options):
  - Set `printWidth` to `100`.
  - Set `bracketSameLine` to `false`.
  - Set `endOfLine` to `lf` (Unix standard).
  - Set `trailingComma` to `all`.
- Colocate test files with the files they test. Rationale:
  - If you need to rename/move a file, you need to rename the test file as well.
  - If you make significant changes to a file, you probably need to adjust tests as well.
  - Test file imports become simpler.
  - File directory nesting decreases by one.
  - We already have test colocated for files such as `/administration/src/components`.
- Use snakeCase convention for non-component files.

### Open question

I'm still not 100% sure about the camelCase file naming. There are some edge cases:

- Extensions under `/administration/src/cards/extensions` sometimes contain React components, sometimes not. Yet, it would look weird if their naming convention differs.
- Directories under `/administration/src/routes` follow URL path conventions, since they reflect URL path segments.

### Testing

Not testable

### Resolved Issues

Fixes: #2603


seluianova (approved):
just briefly checked (since it's a lot), smoke tested, looks good

PR #2746, opened 2025-12-16
2743: Note is not included in confirmation of receipt

andrew8er:
### Short Description

The regions `applicationConfirmationMailNote` is not included in the initial applicant mail. This is because our mailer DSL is very error prone. When not including the initial unary `+` method in the `Paragraph` builder block, the DSL will silently drop a text fragment. Additionally, there was an error in the `Paragraph.plain()` builder method.

### Proposed Changes

- Set sequence start values for tables with test data.
- Add `maildev` service to the docker-compose file, for easier mail debugging.
- Smaller code improvements.
- Inline some methods from `RegionRepository` used in `RegionMutationController`, which where pointless. Also pull out code from transactions that does not need to execute in a transaction.
- Add a `t()` method to the `Paragraph` builder class and use it instead of an unary `+` method.
- Fix the error in `Paragraph.plain()`.

### Open questions

The text for the notification is mixed into the standard text we generate by default. The way regions seem to use this feature, creates really weird text flows with multiple salutations, a kind of mail in mail situation.

### Testing

- Log in as a region admin in the admin UI: http://localhost:3000/ and set a note in "/Region verwalten/Eingangsbestätigung Antrag". Also check the "Notiz aktivieren" button.
- Go to http://localhost:3000/beantragen and create an application for that region.
- Verify the mail that was created: http://localhost:5026/

### Resolved Issues

Fixes: #2743

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -73,8 +73,8 @@ services:
     # Dev mail server, open at http://localhost:5026
     image: maildev/maildev
     environment:
-      - MAILDEV_SMTP_PORT=25
-      - MAILDEV_WEB_PORT=80
+      - MAILDEV_SMTP_PORT=1025
+      - MAILDEV_WEB_PORT=1080
       - MAILDEV_MAIL_DIRECTORY=/var/maildev
       - MAILDEV_INCOMING_USER=maildev@localhost.local
       - MAILDEV_INCOMING_PASS=maildev
@@ -82,8 +82,8 @@ services:
       - type: tmpfs
         target: /var/maildev
     ports:
-      - 5025:25
-      - 5026:80
+      - 5025:1025
+      - 5026:1080
     networks:
       - network
 networks:


docker-compose.yml
```diff
@@ -69,6 +69,23 @@ services:
...
       - network
+  maildev:
+    # Dev mail server, open at http://localhost:5026
+    image: maildev/maildev
+    environment:
+      - MAILDEV_SMTP_PORT=25
+      - MAILDEV_WEB_PORT=80
+      - MAILDEV_MAIL_DIRECTORY=/var/maildev
+      - MAILDEV_INCOMING_USER=maildev@localhost.local
+      - MAILDEV_INCOMING_PASS=maildev
```
f1sh1918: ❓ Wouldn't it make sense to also configure maildev for sending mails? afaics we only use it to receive mails

f1sh1918:
Just a general question to mailDev.
The service is running and i checked the configuration and afaics i have to use `maildev@localhost.local` in the application form to receive mails there.
Unfortunately i get an error that a mail can not be send to this mail address when submitting the application form.
Did you test this?
These two errors i get in the backend
```
Caused by: jakarta.mail.SendFailedException: Invalid Addresses
Caused by: org.eclipse.angus.mail.smtp.SMTPAddressFailedException: 550-Requested action not taken: mailbox unavailable
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/regions/RegionMutationController.kt
```diff
@@ -46,18 +48,17 @@ class RegionMutationController {
...
+            throw ForbiddenException()
+        }
+
         transaction {
-            if (!admin.mayUpdateSettingsInRegion(regionId)) {
-                throw ForbiddenException()
+            RegionsRepository.findRegionById(regionId).apply {
+                this.activatedForApplication = activatedForApplication
+                activatedForCardConfirmationMail = activatedForConfirmationMail
             }
```
seluianova:
```suggestion
            Regions.update({ Regions.id eq regionId }) {
                it[this.activatedForApplication] = activatedForApplication
                it[activatedForCardConfirmationMail] = activatedForConfirmationMail
            }
```
🙃 can be changed to 1 query instead of 2

f1sh1918:
> > Unfortunately i get an error that a mail can not be send to this mail address when submitting the application form.
> 
> @f1sh1918 I think the `maildev@localhost.local` should be used in the project config:
> ```
>     smtp:
>       host: localhost
>       port: 5025
>       username: maildev@localhost.local
>       password: maildev
> ```

Okay then I don't understand the docker config since there is the incoming user mapped. 
For me it looks like the current config only provides a local mailbox

PR #2753, opened 2026-01-08
fix-fvm-installation

f1sh1918:
### Short Description

A new fvm version requires to store the fvm path variable in the circle ci environment

### Proposed Changes
- store the fvm path in $BASH_ENV

### Side Effects

- none

### Testing

Check `commit` workflow
### Resolved Issues

Fixes: #


seluianova (approved):
thx!

Issue #2690, opened 2025-11-21
Use searchAcceptingStoresV2 in administration


f1sh1918:
### Describe the Problem

in [#2689](https://github.com/digitalfabrik/entitlementcard/issues/2689) we created a new endpoint. Use this endpoint to ListAcceptingStores

### Describe Your Preferred Solution

- adjust the query in web app (StoresListTable)

Issue #2689, opened 2025-11-20
Create searchAcceptingStoresInProjectV2


f1sh1918:
### Describe the Problem

Since we want to deliver descriptions in different languages (currently only for nuernberg), we have to create a new query.

### Describe Your Preferred Solution

- create a new query that delivers the same data like `searchAcceptingStoresInProject` but remove `descriptions` string and return descriptions array with `locale`<key>:`description`<value>
- mark `searchAcceptingStoresInProject` `deprecated`

### Related Issues

Follow-up of #2648

PR #2755, opened 2026-01-08
2689 Search accepting stores v2

f1sh1918:
### Short Description

Since we adjusted the database structure to have localized descriptions, we should deliver them.
Mind that we have to keep the old query for app backend compatibility, therefore a new v2 query will be created.
I decided to also  implement the administration part to be able to test it properly

### Proposed Changes

- add deprecation for function and data class
- add new function and data class
- adjust graphql query for administration
- initialize new description values

### Side Effects

- none

### Testing

1. Import acceptance stores for nuernberg including english description
2. Login as ProjectStoreManager
3. Check the StoreOverview Table and click "Edit" (... Menu)
4. Check you can find german english descriptions in the particular form fields

### Resolved Issues

Fixes: #2689 
Fixes: #2690

### DIFF
--- a/specs/backend-api.graphql
+++ b/specs/backend-api.graphql
@@ -57,7 +57,7 @@ type AcceptingStoreV2 {
   categoryId: Int!
   contact: Contact!
   contactId: Int!
-  description: [LocalizedDescription!]
+  descriptions: [LocalizedDescription!]
   id: Int!
   name: String
   physicalStore: PhysicalStore


backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/stores/AcceptingStoreQueryController.kt
```diff
@@ -93,4 +97,37 @@ class AcceptingStoreQueryController(
...
+    ): List<AcceptingStoreV2> {
+        val projectConfig = backendConfig.getProjectConfig(project)
+        val filteredStores = transaction {
+            AcceptingStoresRepository.findBySearch(
+                project,
+                params.searchText,
+                params.categoryIds,
+                params.coordinates,
+                params.limit ?: Int.MAX_VALUE,
+                params.offset ?: 0,
```
andrew8er: 💭 This amount of arguments warrants named args.

Issue #436, opened 2021-12-03
Disable implicit-dynamic option


steffenkleinle:
**Is your feature request related to a problem? Please describe.**
The option [implicit-dynamic](https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks) could be set to false.

**Describe the solution you'd like**
To be discussed:
Enable the rule, set its severity to `error` and fix all occurrences.

PR #2756, opened 2026-01-08
436: Enable strict-inference type check and fix issues

seluianova:
### Short Description
Enable strict-inference rule (`implicit-dynamic` from the issue description is legacy for now)

### Proposed Changes
- enable strict-inference rule and fix issues

UPD. I also enabled strict-raw-types rule so that we have all strict type checks
https://dart.dev/tools/analysis#enabling-additional-type-checks

### Side Effects
no

### Testing
**Dev only**:
Run `fvm flutter analyze` command

### Resolved Issues
Fixes: #436

### DIFF
--- a/frontend/analysis_options.yaml
+++ b/frontend/analysis_options.yaml
@@ -5,6 +5,7 @@ analyzer:
     - build/**
     - lib/graphql_gen/**
     - lib/build_config/build_config.dart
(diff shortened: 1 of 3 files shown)


f1sh1918 (approved):
LGTM

andrew8er (approved):
LGTM

PR #2759, opened 2026-01-12
Fix fvm path

f1sh1918:
### Short Description

The previous fix had the issue that it used a fixed directory for the path. But depending on environment (ios/android) a different home directory is used

### Proposed Changes

- use $HOME in path

### Side Effects

- none

### Testing

https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/10244/workflows/3734ee24-fe9e-4c7a-ad41-4a75630173c1

### Resolved Issues

Fixes: #


andrew8er (approved):
Maybe we also need `source "$BASH_ENV"`, as shown here: https://circleci.com/docs/guides/security/set-environment-variable/#set-an-environment-variable-in-a-shell-command

But let's try this first.

seluianova (approved):

Issue #2770, opened 2026-01-14
Reduce width of the modal for creating/changing users.


Docmale:
### Describe the Problem

The width of the modal for creating/changing users is now oriented at the browser window. These excessively wide input fields seem strange to me and should be adjusted to the necessary size.

### Describe Your Preferred Solution

Reduce width of the modal for creating/changing users to a smaller fixed size

PR #2775, opened 2026-01-15
2770: Set dialog breakpoints

f1sh1918:
### Short Description

I guess it happened due to setting fullwidth property to `<Dialog>` , so if no `maxWidth` is set it will be fullwidth.
I propose to set the `maxWidth` explicitly and fallback to the most used `maxWidth` breakpoint.
I think it's clearer for a developer what width will be used.

### Proposed Changes

- set default `maxWidth` to 'sm' since its the most used
- add maxWidth to all other usages of the <ConfirmDialog>

### Side Effects

- none

### Testing

- Test all dialogs, f.e. UserEdit/UserCreation, Koblenz SelfService form, Adding a note or rejection, deleting an application

### Resolved Issues

Fixes: #2770

### DIFF
--- a/administration/src/components/TextAreaDialog.tsx
+++ b/administration/src/components/TextAreaDialog.tsx
@@ -47,6 +47,7 @@ const TextAreaDialog = ({
       open={isOpen}
       title={title}
       id={id}
+      maxWidth='md'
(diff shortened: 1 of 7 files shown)


seluianova (approved):
lgtm!

Issue #2522, opened 2025-09-24
Application is rejected ~ a notification that the email has been sent is missing


kalairubachandrakumaran-oss:
### Describe the Bug

The user (Regionsadministrator or Regionsverwalter) does not receive any notification that an email has been sent.
At the moment, only a toast message appears saying that the request was rejected.

### Steps to Reproduce

1. Log in in to https://staging.bayern.ehrenamtskarte.app/ as role Regionsadministrator or Regionsverwalter
2. Go to the page with all entries/Anträge and switch to the tab for approved applications
3. Find an application with status open
4. click on "Antrag ablehnen" , select a reason and save
5. Check the toast messages

### Expected Behavior
User should receive a notification that an email has been sent
-> For example, a toast message confirming that an email was sent (see comment from @connium in #2341).

### Additional Information

[](https://github.com/user-attachments/assets/42a5458c-5135-48b4-b3a1-cb303584b555)

seluianova:
@connium should we show 2 toasts ? currently it looks like this:

<img width="543" height="371" alt="Image" src="https://github.com/user-attachments/assets/17a9ae76-1a4c-4ba1-b474-a37d914b53db" />

PR #2776, opened 2026-01-16
2522: Adjust application rejections note

f1sh1918:
### Short Description

Currently its not clear for the service user that the applicant will be informed about the rejection.

### Proposed Changes

- add a text to the toast

### Side Effects

- none

### Testing

- go to "Eingehenden Aufträge" and reject an application

### Resolved Issues

Fixes: #2522


seluianova:
> okay i think we should create a separate ticket for that

ok. I think we need 3 messages here:
- application rejected & email sent
- application rejected & email not sent
- failed to reject application

PR #2791, opened 2026-01-20
2788 allow plus sign + as character in names hotfix

andrew8er:
Hotfix PR of #2790


f1sh1918 (approved):
LGTM
I think we have to solve the conflicts to merge this. Because if you run the pipeline it will increase the version number. And that has to be merged

Issue #2472, opened 2025-09-03
Add an acceptance store


connium:
### Describe the Problem

As an acceptance store manager,
I want to add a new acceptance store
so that I can enrich the list of available acceptance stores.

### Describe Your Preferred Solution

- An action is displayed above the list that allows me to add a new acceptance store.
- The requirements for adding an acceptance store equals those for editing an existing one.

### Related Issues

#2470, #2471, #2497, #2498

PR #2804, opened 2026-01-26
2472: Add accepting store

f1sh1918:
### Short Description

Here we add the possibility to add an acceptance store

### Proposed Changes

- add `addAcceptingStore` mutation
- split houseNumber and street + unit test
- add houseNumber field and validation to the form
- add validation unit  tests
- refetch store data after adding a new store

### Side Effects

- none

### Testing

1. Login as a ProjectStore Manager
2. Go to Akzeptanzstellen verwalten
3. Click "Neu hinzufügen"
4. Type in required form data (especially check street and houseNumber field)
5. Add the store
6. Check if the new store can be found in table
7. Also check that streetAndHouseNumber (which is currently one field in the database) was splitted correctly into the form fields if you edit a store

### Resolved Issues

Fixes: #2472

### DIFF
--- a/backend/src/test/kotlin/app/ehrenamtskarte/backend/helper/CSVAcceptanceStoreBuilder.kt
+++ b/backend/src/test/kotlin/app/ehrenamtskarte/backend/helper/CSVAcceptanceStoreBuilder.kt
@@ -1,6 +1,6 @@
 package app.ehrenamtskarte.backend.helper
 
-import app.ehrenamtskarte.backend.generated.inputs.CSVAcceptingStoreInput
+import app.ehrenamtskarte.backend.generated.inputs.AcceptingStoreInput
 
 object CSVAcceptanceStoreBuilder {
     fun build(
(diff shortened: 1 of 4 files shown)


connium:
> As I understand, the "Hauptstr. B305" is the name of the street❓ and there is no house number. But this is not accepted as well @connium Even if I put "Hauptstr" as a street and "B305" as a house number, the coordinates are not resolved.
> 
> 💭 Just assuming that we may receive such questions from users later... (doesn't have to be addressed in this PR)

"B305" is the designation of the road (Bundesstraße). Actually it's just a place next to the road and the full address would be "Hauptstraße, 83487 Marktschellenberg", without a house number.

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/stores/AcceptingStoresMutationController.kt
```diff
@@ -46,6 +47,36 @@ class AcceptingStoresMutationService {
...
+            val acceptingStore = mapCsvToStore(store)
+            val existingStoreId = AcceptingStoresRepository.getIdIfExists(
+                acceptingStore,
+                authContext.admin.projectId,
+                regionEntity.id,
+            )
+            if (existingStoreId != null) {
+                throw StoreAlreadyExistsException()
+            }
+            AcceptingStoresRepository.createStore(mapCsvToStore(store), authContext.admin.projectId, regionEntity.id)
```
seluianova:
```suggestion
            AcceptingStoresRepository.createStore(acceptingStore, authContext.admin.projectId, regionEntity.id)
```

PR #2812, opened 2026-01-30
Improve the visual appearance of the category filter

seluianova:
### Short Description
Just tried to improve the visual appearance of the category filter a bit 

### Proposed Changes
Before & after:
image image

### Side Effects
no

### Testing
Check filter by category in the apps

### Resolved Issues
n/a

### DIFF
--- a/frontend/lib/search/filter_bar_button.dart
+++ b/frontend/lib/search/filter_bar_button.dart
@@ -3,8 +3,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter_svg/svg.dart';
 
 class FilterBarButton extends StatefulWidget {
-  static const double width = 80;
-  static const double height = 80;
+  static const double width = 74;
+  static const double height = 74;
 
   final CategoryAsset asset;
   final void Function(CategoryAsset, bool) onCategoryPress;
@@ -32,7 +32,7 @@ class _FilterBarButtonState extends State<FilterBarButton> {
       height: FilterBarButton.height,
       child: Material(
         color: _selected ? chipTheme.selectedColor : chipTheme.backgroundColor,
-        shape: shape.copyWith(side: _selected ? BorderSide.none : chipTheme.side),
+        shape: shape.copyWith(side: _selected ? BorderSide(color: theme.colorScheme.outline) : chipTheme.side),
         child: InkWell(
           customBorder: shape,
           onTap: () {
--- a/frontend/lib/themes.dart
+++ b/frontend/lib/themes.dart
@@ -84,9 +84,9 @@ ThemeData get lightTheme {
       fillColor: WidgetStatePropertyAll(primaryColor),
     ),
     chipTheme: ChipThemeData(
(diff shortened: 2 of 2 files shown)


hauf-toni:
Thanks @seluianova for taking this on! Your improvements are much needed. 

> i think the contrast between selected and not selected is not good enough. I would tend to use a white color type
> 
> What do you think @hauf-toni

**Here are some numbers when checking with accessibility guidelines in mind:** 
The contrast for the selected item and the background is not high enough in the new implementation:
Bildschirmfoto 2026-02-16 um 14 16 30

unfortunately, even in the new suggestions, the contrast is not sufficient:
Bildschirmfoto 2026-02-16 um 14 21 06

💡 Maybe we could add an outline in a dark grey to enhance the appearance and accessibility?

This is a bit out of scope but I did experiment a bit with other ideas for the categories. together with the figma ai I came up [with this idea](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=4149-1855&t=0Jn2fQ5076RfOG0d-1): 

Bildschirmfoto 2026-02-16 um 15 22 29

Just wanted to include it here if we have a chance to rework the design in the future.

hauf-toni:
> > Thanks @seluianova for taking this on! Your improvements are much needed.
> > > i think the contrast between selected and not selected is not good enough. I would tend to use a white color type
> > > What do you think @hauf-toni
> > 
> > 
> > **Here are some numbers when checking with accessibility guidelines in mind:** The contrast for the selected item and the background is not high enough in the new implementation: Bildschirmfoto 2026-02-16 um 14 16 30
> > unfortunately, even in the new suggestions, the contrast is not sufficient: Bildschirmfoto 2026-02-16 um 14 21 06
> > 💡 Maybe we could add an outline in a dark grey to enhance the appearance and accessibility?
> > This is a bit out of scope but I did experiment a bit with other ideas for the categories. together with the figma ai I came up [with this idea](https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=4149-1855&t=0Jn2fQ5076RfOG0d-1):
> > Bildschirmfoto 2026-02-16 um 15 22 29
> > Just wanted to include it here if we have a chance to rework the design in the future.
> 
> @connium i think thats out of scope now but we could think about that in the future
> 
> For now i would tend to go with the one with be highest contrast

Would still recommend an outline since the contrast in the examples above is not high enough for Accessibility Standards. A darker outline could look like this: https://www.figma.com/design/SDoGnXIjCCXpE1tAAJkRVc/entitlementcard?node-id=4149-1339&t=dbNNfTcbxw12HtXg-1

f1sh1918:
> I would really want to keep this. It would not take that much effort to get it into a mergeable state and besides the very noticeable visual improvement of the buttons, it also cleans up the code a little.

Sure just changing the selected border color. I also would do this like that.
I just wanted to point out that we shouldn't remove the main visual improvement of this pr

andrew8er (approved):

PR #2820, opened 2026-02-03
2795: Fix Sentry reports for backend

### LINKED ISSUE #2611: Fix failed to init Sentry DSN required
### Describe the Bug

Currently we are facing this issue, when starting backend after #2583 

```
Failed to init Sentry during appender initialization: DSN is required. Use empty string or set enabled to false in SentryOptions to disable SDK.
```

### Steps to Reproduce

1. Start backend
2. See error

### Expected Behavior

No error occurs

### Related Issues

This is a upstream issue from `sentry-spring-boot-starter-jakarta`
https://github.com/getsentry/sentry-java/issues/4531

### Additional Information

Probably we can only fix this issue by updating and `io.sentry.jvm.gradle` when this issue is fixed in `sentry.java`.
If we remove the `sentry-jvm-gradle` the issue is gone

### LINKED ISSUE #2795: Fix sentry reports backend
### Describe the Problem

Somehow we don't get any reports for the backend.
I can't really believe that there aren't any issue at all...
So we should check the server backend logs

- check why no issue be sent

- if there are really no issues in the server log, close the issue here

### LINKED ISSUE #2853: [DEV] Sentry release version not set for backend
### Describe the Bug

For our backend error reports we don't proper version information
https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/9073/?project=7

### Steps to Reproduce

1. Go to sentry backend report
2. Check release is unknown

### Expected Behavior

Current release number is shown

andrew8er:
### Short Description

I tested Sentry reporting with an Exception in a GraphQL handler path and could confirm that it actually works:
https://sentry.tuerantuer.org/organizations/digitalfabrik/issues/?project=7&statsPeriod=14d

Maybe we really don't have any unhandled errors. A portion of those might also be "swallowed" by the `GlobalExceptionHandler.handleGeneralException()` method. We probable should re-throw the exception here, so Sentry can pick it up for reporting. This doesn't affect exceptions thrown in GraphQL controller methods, though, which make up most of our code base.

### Proposed Changes

- Update the Sentry dependencies
- Add doc links to the config files
- Slightly improve the build config: instead of having a custom property defined for CI builds, rely on the quasi-default `CI` environment variable to detect, if we are running in a CI environment.

### Side Effects

None

### Testing

We could deploy a special version with an intentional unhandled error on staging, to have a final test. Alternatively, we could create an external error condition such as temporarily shutting down the DB and then see of there are any Sentry reports.

### Resolved Issues

Fixes: 
- Fixes #2795
- Fixes #2853
- Fixes #2611

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -108,7 +108,6 @@ dependencies {
     implementation(libs.piwik.matomo.java.tracker)
     implementation(libs.postgis.jdbc)
     implementation(libs.projectreactor.reactor.kotlinextensions)
-    implementation(libs.sentry.springbootstarter)
     implementation(libs.simplejavamail)
     implementation(libs.springdoc.openapi.starter)
     implementation(libs.springframework.boot.starter.mail)
--- a/backend/gradle/libs.versions.toml
+++ b/backend/gradle/libs.versions.toml
@@ -4,7 +4,7 @@
 
 expediagroup-graphql = { id = "com.expediagroup.graphql", version.ref = "graphql" }
 gitlab-arturbosch-detekt = { id = "io.gitlab.arturbosch.detekt", version = "1.23.8" }
-gmazzo-buildconfig = { id = "com.github.gmazzo.buildconfig", version = "5.7.1"}
+gmazzo-buildconfig = { id = "com.github.gmazzo.buildconfig", version = "5.7.1" }
 google-protobuf = { id = "com.google.protobuf", version = "0.9.6" }
 jetbrains-kotlin-spring = { id = "org.jetbrains.kotlin.plugin.spring", version.ref = "kotlin" }
 jetbrains-kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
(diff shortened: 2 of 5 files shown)


andrew8er:
@f1sh1918 

Should we re-throw exceptions in `GlobalExceptionHandler.handleGeneralException()`?

andrew8er:
> I mean what about adding Sentry.captureException to the else block of handleGraphQlException

I didn't even know of this… something I dislike about Sping Boot: this is configured implicitly with a decorator rather than explicitly, so without knowing what to look for, this is really hard to find.

I added `Sentry.captureException()` statements to both `GraphQLExceptionHandler` and `GlobalExceptionHandler`. Let's see, if we get some errors now.

andrew8er:
The release version name is set by a property in file `/backend/src/main/resources/application-production.properties` via
```
sentry.release=${NEW_VERSION_NAME:unknown}
```
but this hasn't changed actually. I'm also a bit stumped, because this should never have worked. The `NEW_VERSION_NAME` is set in the CI build, and is not present when running on the server.

Options are:
- Ditching `sentry-spring-boot-starter-jakarta` autoconfiguration of Sentry and doing it manually, as explained [here](https://medium.com/javarevisited/%EF%B8%8F-a-senior-architects-guide-to-production-ready-sentry-integration-in-spring-boot-jakarta-ee-e92f12234f16)
- Configure a filter step in Gradle that replaces the `sentry.release` value in the property file, as explained [here](https://www.perplexity.ai/search/in-sentry-for-spring-boot-how-pUEyCT4JTVeCNNe_tlaWwA)

Both options are not ideal, but I tend to go with the former here.

f1sh1918 (approved):
Looks good. Sent a testing error and worked.

When i implemented it i tried to avoid manual init but for me now it feels cleaner and should also solve:

#2853
#2611

Please add them to the description :)

PR #2821, opened 2026-02-03
2696: Do not show loading widget when we get store summary from cache

seluianova:
### Short Description
When a user clicks on a store on the map, we use `FetchPolicy.cacheAndNetwork` for this request.
This means it returns cached data first (if available), while also triggering a background network request to ensure the data is up to date.

The problem is that we always show the loading widget `AcceptingStorePreviewCard(isLoading: true)`, even when the data is immediately available from the cache.

### Proposed Changes
In this PR, we show the loading widget only when the data is actually loaded from the backend.
If the data is loaded from the cache, it is displayed immediately.

### Side Effects
no

### Testing
1. Open the Map tab in the app
2. Click on any store → the **loading** widget is shown for a moment, then replaced by the widget with summary data.
3. Click another store and then click the first store again → no loading widget should be shown; the summary widget should be displayed immediately.

### Resolved Issues
Part of #2696

### DIFF
--- a/frontend/lib/map/preview/accepting_store_preview.dart
+++ b/frontend/lib/map/preview/accepting_store_preview.dart
@@ -23,33 +23,26 @@ class AcceptingStorePreview extends StatelessWidget {
         ),
       ),
       builder: (result, {refetch, fetchMore}) {
-        try {
-          final exception = result.exception;
+        final data = result.parsedData;
+        if (result.isLoading && data == null) {
+          return const AcceptingStorePreviewCard(isLoading: true);
+        }
 
-          if (result.hasException && exception != null) {
-            throw exception;
+        if (result.hasException) {
+          debugPrint('AcceptingStores query failed: ${result.exception}');
+          // If we have cached data, show it even if the network request fails
+          if (data == null) {
+            return AcceptingStorePreviewCard(isLoading: false, refetch: refetch);
           }
+        }
 
-          final fetchedData = result.parsedData;
-
-          if (result.isLoading && fetchedData == null) {
-            return const AcceptingStorePreviewCard(isLoading: true);
-          }
-
-          if (fetchedData == null) {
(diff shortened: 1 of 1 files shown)


frontend/lib/map/preview/accepting_store_preview.dart
```diff
@@ -30,18 +32,19 @@ class AcceptingStorePreview extends StatelessWidget {
 
           final fetchedData = result.parsedData;
 
-          if (result.isLoading || fetchedData == null) {
+          if (result.isLoading && fetchedData == null) {
```
seluianova: if result is loading but fetched data is not null - it means we have data in cache.
so we should only show `isLoading: true` if result is loading AND fetched data is null

frontend/lib/map/preview/accepting_store_preview.dart
```diff
@@ -15,38 +16,33 @@ class AcceptingStorePreview extends StatelessWidget {
...
+          return const AcceptingStorePreviewCard(isLoading: true);
+        }
 
-          if (result.isLoading || fetchedData == null) {
-            return const AcceptingStorePreviewCard(isLoading: true);
+        if (result.hasException) {
+          debugPrint('AcceptingStores query failed: ${result.exception}');
+          // If we have cached data, show it even if the network request fails
+          if (data == null) {
+            return AcceptingStorePreviewCard(isLoading: false, refetch: refetch);
```
andrew8er: ❓ Can `result.hasException` be true and `data` be not null? What would happen then? I could not find anything on this in the docs: https://pub.dev/documentation/graphql_flutter/latest/graphql_flutter/QueryResult/data.html

f1sh1918: i think the latest cache will be used or try to refetch

andrew8er (approved):
LGTM 👍

Issue #2282, opened 2025-06-18
Enable virtual threads for local development


andrew8er:
### Describe the Problem

Since we want to get some insights into virtual threads, it should be made easy to enable it locally and do some experiments.

### Related Issues

Depends on #2281

PR #2822, opened 2026-02-03
2282: Enable virtual threads for development

andrew8er:
### Short Description

Enables Java Virtual Threads for development.

### Side Effects

Hopefully none.

### Testing

None

### Resolved Issues

Fixes: #[2282](https://github.com/digitalfabrik/entitlementcard/issues/2282)

### DIFF
--- a/backend/src/main/resources/application-staging.properties
+++ b/backend/src/main/resources/application-staging.properties
@@ -1,0 +1,7 @@
+spring.threads.virtual.enabled=true
+
+# Sentry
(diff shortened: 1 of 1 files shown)


backend/src/main/resources/application-staging.properties
```diff
@@ -0,0 +1,7 @@
+spring.threads.virtual.enabled=true
```
f1sh1918: ❓ how will this profile be used? How does spring know that its staging environment? we probably have to add some start parameter to choose the profile, no?

f1sh1918 (approved):

Issue #2786, opened 2026-01-20
Reload dosn't work


Docmale:
### Describe the Bug
During update of acceptance points with deletings, it can occur: If the app   ‘Sozialpass Nürnberg’ (or  'Kblenzpass') is loaded and then a store manager imports a new file in which one or more acceptance points have been deleted, then the user can performs one of two actions: 
1. Opening the tab ‘Karte’, the information is not found, and a reload always results in the same status, so a reload doesn’t make sense. Error; “Fehler beim Laden der Daten”
2. Switching to the 'Search' tab, the already deleted locations are still displayed in the list. Clicking on a deleted location leads to an error with a reload option. If this option is used, no action is triggered, and the only option is to exit the app and remove it from the background list. Only then the app be restarted, at which point the current data is loaded. Error: “Akzeptanzstelle nicht gefunden”

### Steps to Reproduce
See discription

### Expected Behavior

In both scenarios all data should be reloaded

### Additional Information

<img width="378" height="55" alt="Image" src="https://github.com/user-attachments/assets/9a9980fe-155c-4fd6-b1cc-6effa9f36bdf" />

<img width="379" height="91" alt="Image" src="https://github.com/user-attachments/assets/767a0e16-9361-4dfb-b5e3-b92f2dc80f89" />

f1sh1918:
okay it looks like the searchList is cached somehow and the map not  (there was only 1 store in the backend)
i reuploaded the stores and it seem to work.
Nevertheless this caching issue has to be solved somehow.
We have to check how to reproduce that

seluianova:
> i think this is a duplicate? @seluianova

maybe "related issue" ?

PR #2833, opened 2026-02-10
2786: Adjust reload

f1sh1918:
### Short Description

Currently we get a misleading error message if a user searches for an accepting store and wants to see it on the map while in this moment a store manager deleted this store

### Proposed Changes

- show proper error message that the store is not available
- clear the store cache when this issue occurs to ensure that the next search won't show this store
- report to sentry if error occurs

### Side Effects

- none

### Testing

1. Search for an acceptance store in the app
2. Delete this acceptance store in the administration gui
3. Go back to the app and go to store details and show on map
4. There should be a valid error that the store is not available anymore
5. Go back to search and search for the same acceptance store. It should not be found anymore

### Resolved Issues

Fixes: #2786


f1sh1918:
> Looks good, but I'm unable to provoke the error condition, the app reloads when it gets focused. Is this intentional?



> Looks good, but I'm unable to provoke the error condition, the app reloads when it gets focused. Is this intentional?

hm the hot reload should only happen for code changes or triggered manually not by focus. no idea tbh.

Issue #2834, opened 2026-02-10
Refresh store data if app comes in Foreground


f1sh1918:
### Describe the Problem

Currently if the app stays in background and will be opened, the stores will not be refetched.
So if in the mean time a store will be deleted it will not be reflected in the app until the user triggers a new search but not for the current search list

### Describe Your Preferred Solution

- implement an app lifecycle observer
- trigger a refetch of the store data if app comes on foreground

PR #2835, opened 2026-02-10
2834: Refresh store on foreground

f1sh1918:
### Short Description

This pr reloads the store data when the app comes to foreground to avoid data inconsistencies.
For example the app lifes in the background and in the mean time a store was deleted, it still may be shown in the store search result list

### Proposed Changes

- [add app lifecycle observer and AppResumeNotifier](https://github.com/digitalfabrik/entitlementcard/pull/2835/changes/c1dc5bcb7a2f8426fd5174f710552394594d85fd)
- [refetch store data if app comes to foreground](https://github.com/digitalfabrik/entitlementcard/pull/2835/changes/865537cb686515979db069fd118e9d9941294389)
### Side Effects

- cache will be deleted once to ensure an actual reload

### Testing

1. Open the app. Check for a store in the first 20 results maybe the first one
2. Go to administration. Login as Project Store Manager and delete this store
3. Bring the app in background and foreground again. 
4. Check that the store search result should be refreshed and the deleted store is not available anymore

### Resolved Issues

Fixes: #2834

### DIFF
--- a/frontend/lib/search/results_loader.dart
+++ b/frontend/lib/search/results_loader.dart
@@ -10,6 +10,7 @@ import 'package:ehrenamtskarte/graphql_gen/graphql_queries/stores/accepting_stor
 import 'package:ehrenamtskarte/l10n/translations.g.dart';
 
 import 'package:ehrenamtskarte/home/home_page.dart';
+import 'package:provider/provider.dart';
 
 class SliverResultsLoader extends StatefulWidget {
   final Input$CoordinatesInput? coordinates;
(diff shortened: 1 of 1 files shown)


andrew8er (commented):
When I test this on iOS with a debugger, I see that `AppResumeNotifier.emitResumed()` gets executed, but shouldn't `UserCodeModel.initialize()` also get executed with the callback configured in `AppLifecycleObserver` from `app.dart`? Or am I misunderstanding something here?

f1sh1918:
> Generally works.
> 
> One thing I find not so great is that when getting the app back to the foreground the reload triggers the display of the loader animation and rebuilds the result list afterwards. I would prefer a more subtle invisible "background" update that just adds/removes entries as the result arrives. Things like these tend to make an app "feel" sluggish.

Well i see the point but thats only visible if you are on the search page and reopen the app from background and not from any other tabs.
To adjust that we probably would have to encapsulate all the data fetches and make a general loading screen which would require lots of refactoring i don't see this here in the ticket now

I'm also fine with not implementing this refresh

PR #2864, opened 2026-02-17
Docs import self service data

f1sh1918:
### Short Description

Add a short documentation for importing data for the card self service

### Proposed Changes

- add some instructions
- extend end dates

### Side Effects

- none

### Resolved Issues

Fixes: #


f1sh1918:
> How about including a API design client software (like [Bruno](https://www.usebruno.com/)) config in the repository? I frequently use this for tests in other projects.
> 
> Also useful for GraphQL.

Feel free to open a ticket. We have graphiql for graphql already and the amount if rest endpoints is quite low

Issue #2673, opened 2025-11-18
Reject application without option should give feedback


f1sh1918:
### Describe the Bug

Currently a user does not get any feedback if a rejection was executed without selecting a message option

### Steps to Reproduce

1. Go to /applications
2. Click on Antrag ablehnen
3. Click Ablehnung bestätigen without choosing a message
4. Modal is closed without any feedback

### Expected Behavior

If the user clicks "Ablehnung bestätigen" without selecting a reason there should be a form error message or some info that a reason is required

f1sh1918:
i guess @hauf-toni would prefer to show a hint from ux perspective instead of disabling a button

PR #2866, opened 2026-02-18
2673: No rejection reason feedback

f1sh1918:
### Short Description

Currently a user gets no feedback if no rejection reason was selected and the rejection was confirmed.
Even the application was not rejected there should be a proper feedback instead of just closing the modal

### Proposed Changes

- close modal only if reason provided onConfirm
- show proper error message if the autoComplete field was left or the confirmButton was clicked without a selected reason
- update node version in GH actions

### Side Effects

- none

### Testing

1. Reject an application and don't select a reason. Check the feedback. Then choose a reason and reject the application

### Resolved Issues

Fixes: #2673

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -25,7 +25,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v6
         with:
-          node-version: '20.18.0'
+          node-version: '24.13.0'
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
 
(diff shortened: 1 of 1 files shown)


administration/src/routes/applications/components/ApplicationCard.tsx
```diff
@@ -90,9 +90,12 @@ const RejectionDialog = (props: {
   const { t } = useTranslation('applicationsOverview')
   const rejectionMessages = t('rejectionReasons', { returnObjects: true }) as string[]
   const [reason, setReason] = useState<string | null>(null)
+  const [touched, setTouched] = useState(false)
```
andrew8er: 💭 The word 'touched' has an odd primary meaning, a better UX related term would be 'interacted'.

PR #2869, opened 2026-02-20
2824: Render box not laid

f1sh1918:
### Short Description

The root of this issue seems to be that the MapLibreContainer is rendered in a stack but has to Layout container until its fully rendered. This can cause issue on ios (KI answer)

Since i can not really reproduce the issue, i'm not sure that it fixes the problem but it sound reasonable and this Container does not harm. We can either keep the bug ticket open or wait until it escalates again after we deployed a new version with this "fix"

### Proposed Changes

- enforce map to be filled even before it was rendered by adding `Position.filled` 

### Side Effects

- none

### Resolved Issues

Fixes: #2824

### DIFF
--- a/frontend/lib/map/map/map.dart
+++ b/frontend/lib/map/map/map.dart
@@ -74,6 +74,7 @@ class _MapContainerState extends State<MapContainer> implements MapController {
 
     return Stack(
       children: [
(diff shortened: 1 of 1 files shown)


andrew8er (approved):
LGTM, I would add a note referencing the issue above the extra container though.

Issue #2678, opened 2025-11-19
Include links to imprint and data protection on status/confirmation page


connium:
### Describe the Problem

Links to the imprint and to the data privacy are required on public facing pages.
The status page (for the applicant) and the confirmation page (for the contact person) currently don't refer to the imprint or the data privacy.

### Describe Your Preferred Solution

For EAK Bayern, add links to the customer specific imprint and data privacy at the bottom of the respective pages.

PR #2879, opened 2026-02-24
2678: Links application public pages

f1sh1918:
### Short Description

Our public faces applicant and verifier view doesn't have imprint and data privacy yet

### Proposed Changes

- add links to applicant view
- add links to verifier view

### Side Effects

- none

### Testing

- Create an application for bayern and check the applicant page that you get view mail links
- Also check the verifier page that you get via mail link

### Resolved Issues

Fixes: #2678

### DIFF
--- a/administration/src/components/PageLayout.tsx
+++ b/administration/src/components/PageLayout.tsx
@@ -1,27 +1,30 @@
-import { Box, BoxProps } from '@mui/material'
+import { Stack, StackProps } from '@mui/material'
 import React, { ReactElement, ReactNode } from 'react'
 
 import Footer from './Footer'
 
-type PageLayoutProps = BoxProps & {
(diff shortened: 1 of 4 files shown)


administration/src/components/PageLayout.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import Footer from './Footer'
+
+type PageLayoutProps = BoxProps & {
+  children: ReactNode
+  showDataPrivacy?: boolean
+}
+
+const PageLayout = ({ children, showDataPrivacy, ...boxProps }: PageLayoutProps): ReactElement => (
+  <Box
```
andrew8er: 🔧 `<Stack>` is the same as `<Box>`, but with `display: flex` already set.

Issue #2832, opened 2026-02-10
Use consistent date format


connium:
### Describe the Problem

I would prefer all dates with leading zeros:
<img width="385" height="144" alt="image" src="https://github.com/user-attachments/assets/c0b2a70d-d1ae-4990-9d69-4f46db06dcdc" />

_Originally posted by @Docmale in https://github.com/digitalfabrik/entitlementcard/issues/2659#issuecomment-3584987876_

This will make the UI more consistent.

### Describe Your Preferred Solution

Write all dates in the format `dd.mm.yyyy`, when using German language.

### Describe Possible Alternatives

Keep it as it is.

andrew8er:
Since we are using a wild mix of `date-fns` and a half baked `temporal` polyfill (in `PlainDate.ts`) of unclear provenance, I'd like to replace both with a proper modern polyfill like https://github.com/js-temporal/temporal-polyfill.

I use it in another project, and so far it works great. Also it is easy to replace with the native implementation, once it becomes generally available.

PR #2880, opened 2026-02-24
2832: Use consistent date format

andrew8er:
### Short Description

Completely revises the date/time/timestamp handling.

### Proposed Changes

- Add a [`Temporal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) polyfill: https://github.com/fullcalendar/temporal-polyfill.
- Replace usage of our homegrown `Temporal` quasi-polyfill `PlainDate`.
- Replace usage of `date-fns` functions with `Temporal` and `Intl` functionality. We still cannot remove the `date-fns` dependency though, since MUI's date pickers need it for date calculations. See also https://github.com/mui/mui-x/issues/4399. An alternative would be to use https://github.com/DASPRiD/mui-temporal-pickers.
- Replace usage of legacy `Date` where possible. MUI `<DatePicker>` only supports `Date` but our `<CustomDatePicker>` is now refactored to use `Temporal.PlainDate`.
- Move almost all user-facing date string formatting to `i18next` translation strings (exception see below).
- Teach `i18next` to handle `Temporal` types directly.
- Remove a little bit of excessive logging in `exportApplicationToCsv()`.
- Add timezone data to test mock data  in `applicationData.ts`.
- Fix tooltip in `<StatisticsFilterBar>`. The tooltip would never trigger when the button was disabled.
- Move a misplaced test file.

### Open questions

- `<JsonFieldElemental>` still uses manual string formatting for displaying (date) values. Could be changed, but all translation strings need to be adjusted.
- There are quiet a few places, where `formatDateDefaultGerman()` is used, e. g. in `BirthdayExtension`. I think, in general, we should avoid having a hard-coded locale aware string formatting. The strings produced here are also used for the CSV export. This _could_ potentially be replaced with ISO formatted dates, but this should be tested in common office suites, as it might be a breaking change.
- The `ProjectConfig.timezone` value is now unused and its purpose is not really clear. We actually should not have the need for a hard-coded timezone value IMHO. Should we remove it?

#### Bayern EAK

- Test date form elements in http://localhost:3000/beantragen
- Test applicant view in http://localhost:3000/antrag-einsehen/access_key_06
- Test verifier view in http://localhost:3000/antrag-verifizieren/access_key_06
- Test application views in http://localhost:3000/applications (login as `region-admin@bayern.ehrenamtskarte.app`)

#### Koblenz

- Test forms in http://localhost:3000/erstellen

### Nürnberg

- Test date displays in admin UI:  http://localhost:3000/ (login as `region-admin@nuernberg.sozialpass.app`)

### Resolved Issues

Fixes: #2832

### DIFF
--- a/administration/src/routes/statistics/components/StatisticsOverview.tsx
+++ b/administration/src/routes/statistics/components/StatisticsOverview.tsx
@@ -8,7 +8,7 @@ import RenderGuard from '../../../components/RenderGuard'
 import { CardStatisticsResultModel, Region, Role } from '../../../generated/graphql'
 import { ProjectConfigContext } from '../../../project-configs/ProjectConfigContext'
 import downloadDataUri from '../../../util/downloadDataUri'
-import { generateCsv, getCsvFileName } from '../utils/csvStatistics'
+import { csvFileName, generateCsv } from '../utils/csvStatistics'
 import StatisticsBarChart from './StatisticsBarChart'
 import StatisticsFilterBar from './StatisticsFilterBar'
 import StatisticsLegend from './StatisticsLegend'
@@ -60,7 +60,7 @@ const StatisticsOverview = ({
           try {
             downloadDataUri(
               generateCsv(statistics, cardStatistics),
-              getCsvFileName(`${dateStart}_${dateEnd}`, region),
+              csvFileName(dateStart, dateEnd, region),
             )
           } catch {
             enqueueSnackbar(t('exportCsvNotPossible'), { variant: 'error' })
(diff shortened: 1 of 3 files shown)


administration/src/cards/extensions/BirthdayExtension.tsx
```diff
@@ -76,26 +87,34 @@ const BirthdayExtension: Extension<BirthdayExtensionState> = {
...
+      Temporal.PlainDate.compare(birthday, today) <= 0
+    )
   },
   fromString: (value: string) => {
-    const birthday = PlainDate.safeFromCustomFormat(value)
-    return birthday === null ? null : { birthday }
+    const date = safeParseGermanPlainDateString(value)
+    return date !== null
+      ? {
+          [BIRTHDAY_EXTENSION_NAME]: safeParseGermanPlainDateString(value),
```
seluianova:
```suggestion
          [BIRTHDAY_EXTENSION_NAME]: date,
```

PR #2890, opened 2026-03-02
Init Freinet HTTP client as a Spring bean

seluianova:
### Short Description

Currently we create a new http client instance for every sendApplicationAndCardDataToFreinet request, which is an expensive operation.

### Proposed Changes

- create freinetHttpClient as a Spring bean (single instance will be re-used for all requests)

### Side Effects

no, just better performance

### Testing

**Dev only:**

Just check that the Freinet transfer still works

### Resolved Issues

n/a

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/freinet/util/FreinetApi.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/freinet/util/FreinetApi.kt
@@ -16,14 +16,12 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import io.ktor.client.HttpClient
 import io.ktor.client.request.request
 import io.ktor.client.request.setBody
-import io.ktor.client.statement.bodyAsChannel
 import io.ktor.client.statement.bodyAsText
 import io.ktor.http.ContentType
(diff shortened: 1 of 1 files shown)


andrew8er:
Too me it looks like the `accessKey` and `agencyId` should be an argument of every call to this API, since it is different for every call. We have this info already there in every place we call it anyway.

> the bean is named as freinetHttpClient, so it was intended for freinet only. or what do you mean?

I don't exactly know how this whole bean stuff works, as far as I know, it's some sort of a dependency injection system… Do the beans get resolved by their name or by their type? If it's by name, then no problem, if it's by type however, we might accidentally inject this if we ever need a different HTTP client in another place.

seluianova:
> Too me it looks like the accessKey and agencyId should be an argument of every call to this API, since it is different for every call. We have this info already there in every place we call it anyway.

I think having `accessKey` and `agencyId` as `FreinetApi` instance fields is not so bad; it helps to keep the method signatures of `FreinetApi` cleaner. 
creating a new `FreinetApi` object is lightweight, so it wouldn't bring any visible performance implication.

in the case of `HttpClient` the reuse is more critical because it manages heavy resources like thread pools and connection selectors. reusing it prevents resource exhaustion and allows for connection pooling.

Issue #2778, opened 2026-01-19
Add email delivery and Freinet integration for triggered card creation


connium:
### Describe the Problem

As a region manager or region administrator,
I want to trigger a card creation from Freinet by clicking on a link,
so that the applicant receives an activation email, if automated mail sending is enabled.

### Describe Your Preferred Solution

At the moment, there is already a link in Freinet, which directs to the parameterized URL endpoint of the digital print shop. They pass the name, the card type and the expiration date.

- Define the names of the parameters for the volunteer's email address and their Freinet-ID.
- Accept the additional parameters.
- Any of these two parameters must be optional.
- If no email address is passed, no email should be sent.
- If a valid email address is passed and
   -  automatic automated mail sending is enabled for this region, an activation email should be sent to the person.
   -  automatic automated mail sending is disabled for this region, no email should be sent to the person.
- If an invalid email address is passed, an error should be logged and a notification should be shown to the user.
- If no Freinet-ID is passed, no communication should take place with Freinet.
- If a Freinet-ID is passed and
   -  Freinet data transfer is enabled for this region, the issued card should be stored in Freinet.
   -  Freinet data transfer is disabled for this region, no communication should take place with Freinet.

### Related Issues

#2779 

### Technical information

- cardConfig erweitern in ProjectConfig
- queryParam userID initialisieren
- freinetUserId übergeben an cardGenerator hook
- neue Mutation sendCardDataToFreinet erstellen
- ~~Freinet Endpoint um Gültigkeit der userId zu prüfen: (https://www.freinet-online.de/query/api/PersonServiceEndpoint.php?agencyId=52&accessKey=gcBc66&personId=40334)~~
- dann sendCardInformation nutzen

connium:
> > cardConfig erweitern in ProjectConfig
> 
> when I add a `FreinetUserIdExtension`, it enables the use of this parameter in the csv import. do we want to keep this as is? or we need different configurations for csv columns and for query params… currently all extensions automatically work as both.
[...]

PR #2897, opened 2026-03-09
2778: Add email delivery and freinet integration for triggered card creation

seluianova:
### Short Description
At the moment, there is already a link in Freinet that directs to a parameterized URL endpoint of the digital print shop. It passes the name, the card type, and the expiration date.

We need to add one more query parameter: `UserId`.
If `UserId` is not null, we need to transfer the card data to Freinet.

### Proposed Changes
- add `FreinetUserIdExtension` to the card config
- add `sendCardDataToFreinet` mutation in the backend
- call the `sendCardDataToFreinet` mutation on the card creation form when the `UserId` query parameter is not null

### Side Effects
no

### Testing

**⚠️ For dev and staging environment you can only test with region "Freinet Demo"**

1. Log in to the administration as a region admin
2. Go to region settings `/region` and make sure the following options are enabled:
image

image

4. Use the following link to open the prefilled card creation form:

`/cards/add?Name=Freinet%20User&Ablaufdatum=09.03.2029&UserId=1396940&MailNotification=test@test.com`

`UserId` must be a real user ID from the Freinet portal: https://www.freinet-online.de/
_You can find the credentials for the portal in Passbolt._

`MailNotification` must be your email address so you can verify the incoming email.

**When the data has been successfully transferred to Freinet and email has been successfully sent, the following toasts should appear:**

image

5. After creating the card using the link with the query parameters, check the following:
- A new card has been created in Freinet (check in https://www.freinet-online.de, "Freiwillige" module)
- A confirmation email has been sent to the user

### Resolved Issues
Fixes: #2778

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/freinet/FreinetApplicationMutationController.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/freinet/FreinetApplicationMutationController.kt
@@ -34,7 +34,12 @@ class FreinetApplicationMutationController(
 ) {
     private val logger = LoggerFactory.getLogger(FreinetApplicationMutationController::class.java)
 
-    @GraphQLDescription("Send application and card information to Freinet")
+    @GraphQLDescription(
+        """Send application and card information to Freinet.
+    Returns:
+    - true: Data was successfully sent to Freinet.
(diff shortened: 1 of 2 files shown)


andrew8er (approved):
LGTM, some suggestions.

PR #2909, opened 2026-03-17
Improve Api Token settings

f1sh1918:
### Short Description

During review of #2906 i found an unneeded useEffect that i removed here

### Proposed Changes

- remove unneeded use effect
- use primary contained button for "Erstellen"

image
  

### Side Effects

- none

### Testing

1. Login for bavaria as External API user
2. Go to "Project verwalten"
3. Create and delete api tokens. Check that the table with tokens updates correctly if you add and remove tokens

### Resolved Issues

Fixes: #

### DIFF
--- a/administration/src/routes/project/components/ApiTokenSettings.tsx
+++ b/administration/src/routes/project/components/ApiTokenSettings.tsx
@@ -80,7 +80,12 @@ const ApiTokenGeneration = (): ReactElement => {
             <MenuItem value={36}>3 {t('years')}</MenuItem>
           </Select>
         </FormControl>
-        <Button sx={{ minWidth: 'auto' }} onClick={() => createToken({ variables: { expiresIn } })}>
+        <Button
+          color='primary'
+          sx={{ minWidth: 'auto' }}
+          variant='contained'
+          onClick={() => createToken({ variables: { expiresIn } })}
+        >
           {t('create')}
         </Button>
       </Stack>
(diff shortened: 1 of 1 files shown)


andrew8er (commented):
While you are at it, there is another problem with this component:

console log
>`Warning: validateDOMNesting(...): <table> cannot appear as a descendant of <p>.
table`

The solution is to simply remove the `<Typography>` component in line 150.

andrew8er (approved):

seluianova (approved):
tested 👍

kalairubachandrakumaran-oss:
**Test result:**
A token can be created and deleted successfully. ✅ 
-> The expiration date is also shown correctly depending on the selected duration
@f1sh1918 @connium 
**Question / Observation:**
The token shown in the field always seems to be the next token that has not been created yet.
As soon as I click “Create”, a new token is immediately displayed in the field.

According to the information text, it sounds like the token should be saved after creation, because it says the newly created token is only shown once: "Ein **neu erzeugtes Token** wird nur einmalig angezeigt und kann danach nicht wieder abgerufen werden. Bitte speichern Sie dieses Token an einem sicheren Ort."

Can you please confirm if the current behavior is correct?
https://github.com/user-attachments/assets/e2c8d3d2-0986-4882-b8d8-9f52944a24d4

connium:
You are right. The behavior isn't intuitive.

In most software products I know, the token is displayed in a popup, so you have to actively dismiss it to continue working with the UI. From your screencast it looks, like a token would be displayed, even if it was deleted. That's definitely a bug.

Considering the current behavior, I'd expect, that the token shown is always the latest generated token.

I think we can do better with UIX in this case. I'll create two issues for this.
1. Don't show the token, if another action (e.g., delete) was executed.
2. Improve the flow from an UIX perspective.

PR #2911, opened 2026-03-17
2702: Misaligned close button

### LINKED ISSUE #2702: Misaligned close button on the toaster
### Describe the Bug
If the message is long (more than 1 line), the button (x) is slightly displaced.
Image
I think it should be centered.

### Steps to Reproduce
An example of a ‘long’ message can be reproduced by following the steps described in the PR: https://github.com/digitalfabrik/entitlementcard/pull/2674

### Expected Behavior
Smthg like
Image

f1sh1918:
### Short Description

Currently the close button is aligned to the bottom right if the snackbar contains multiline text

### Proposed Changes

- reduce the `maxWidth` of the snackbar text container to align the close icon centered

### Side Effects

- none

### Testing

1. Create a card for koblenz /erstellen
2. Type in not existing data to force an error

Bildschirmfoto 2026-03-17 um 17 18 43

### Resolved Issues

Fixes: #2702

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,110 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Repository Overview
+
+**Digitale Berechtigungskarte** — a whitelabel platform for digital entitlement/benefit cards in Germany. Three main components:
+
+- **`/frontend`** — Flutter mobile app (Android & iOS)
+- **`/backend`** — Kotlin/Spring Boot GraphQL + REST API
+- **`/administration`** — React/TypeScript web portal
+- **`/build-configs`** — Shared whitelabel project constants (Bayern, Nuernberg, Koblenz)
+- **`/specs`** — Shared GraphQL schema and Protobuf specs
+
+Root `package.json` has npm workspaces for `administration` and `build-configs`. The backend uses Gradle independently.
+
+## Commands
+
+### Administration (`/administration`)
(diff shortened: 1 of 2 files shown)


andrew8er (approved):
LGTM. 

Instead of ignoring `Claude.md`, why not include it?

andrew8er (approved):

seluianova (approved):

kalairubachandrakumaran-oss:
@f1sh1918 @andrew8er @seluianova 
Test result:
Two questions regarding the current snackbar result:
1. The info icon on the left side is missing now in the koblenz error message. Is that okay?
2. There is still empty space between the text and the close icon. Is this okay?
image
image

andrew8er:
In `AppSnackbarProvider.tsx` we hide all notification icons, so there should not be any icons at all, regardless of the project. Also, it is yellow?

kalairubachandrakumaran-oss:
> In `AppSnackbarProvider.tsx` we hide all notification icons, so there should not be any icons at all, regardless of the project. Also, it is yellow?

> It is not yellow, it is my marker :D

kalairubachandrakumaran-oss:
> > In `AppSnackbarProvider.tsx` we hide all notification icons, so there should not be any icons at all, regardless of the project. Also, it is yellow?
> 
> > It is not yellow, it is my marker :D

> And space is okay?

andrew8er:
> And space is okay? 

I personally would avoid it, but I think there is no real need to change it now.

PR #2932, opened 2026-03-25
2816: Replace apollo-client with urql

### LINKED ISSUE #2816: Upgrade/replace `@apollo/client`
### Describe the Problem

While investigating what it needs to upgrade to React 19, I notices that `@apollo/client` and `apollo-upload-client` are dragging in React 18 as dependencies, which creates non-obvious runtime errors.

Upgrading to the current `@apollo/client` version 4.1.3 and `apollo-upload-client` 19.0.0 would fix this but needs a bit more refactoring: https://www.apollographql.com/docs/react/migrating/apollo-client-4-migration. This particularly affects our error handling.

Another major change is that latest `@apollo/client` is not compatible with they way we generate our query/mutation hooks, and in fact, it is also not recommended anymore: https://www.apollographql.com/docs/react/development-testing/graphql-codegen#a-note-about-the-client-preset. An AI code agent could be of help here.

Since we need to upgrade `@apollo/client`/`apollo-upload-client` with some larger changes, this would also open the possibility to replace those packages with a more compact alternative, like [urql](https://github.com/urql-graphql/urql):

- urql natively implements the [GraphQL Multipart Request spec](https://github.com/jaydenseric/graphql-multipart-request-spec): https://nearform.com/open-source/urql/docs/advanced/persistence-and-uploads/#file-uploads
- Is considerably smaller: https://pkg-size.dev/urql vs https://pkg-size.dev/@apollo%2Fclient (rxjs contributes significantly to apollos size here).
- Has good integration with the GraphQL codegen package: https://the-guild.dev/graphql/codegen/docs/getting-started#from-the-frontend

I use urql in a personal project and can attest that it is pretty easy to set up and works without any problems so far.

We should also upgrade our current `graphql-codegen.yml` to [`codegen.ts`](https://the-guild.dev/graphql/codegen/docs/config-reference/codegen-config), which is now the recommended config format. We still need a minimal `graphql-codegen.yml`, since this is the file most IDEs (at least VS Code and IntelliJ) use to configure their GraphQL support.

- Provide `codegen.ts`, adjust `package.json` scripts.
- Provide minimal `graphql-codegen.yml` for IDEs.
- Replace `@apollo/client`/`apollo-upload-client` with `urql`.
- Replace generated React hooks with urql hook call and generated request objects.

- Provide `codegen.ts`, adjust `package.json` scripts.
- Provide minimal `graphql-codegen.yml` for IDEs.
- Upgrade `@apollo/client`/`apollo-upload-client` to latest version
- Replace generated React hooks with apollo-client hook call and generated request objects.

### Related Issues

- #2693
- #2696

andrew8er:
### Short Description

Replaces apollo-client with [urql](https://nearform.com/open-source/urql/).

### Proposed Changes

- Add urql dependencies, remove apollo dependencies.
- Add `graphql-codegen.ts`. This is the [recommended](https://the-guild.dev/graphql/codegen/docs/config-reference/codegen-config) way to configure the graphql code generator now. A separate `graphql-codegen.yml` still exists for IDEs to configure their code completion.
- Move all generated source to directory `src-gen` (from `src/generated`) for better separation.
- Implement `<UrqlClientProvider>`.
- Refactor code to use urql's `useQuery()`/`useMutation()` (with help from Claude Code 🤖).
- Refactor tests to use urql mock exchange for mocking requests.
- Narrow some component property types. I'm not exactly sure why, but a lot of components needed narrowing of their component types. I mostly used `Pick<Type, 'prop1' | 'prop2'>` to only specify the properties that are actually used by the component. This should also make them more generally useful, since the properties returned differ from query to query.
- Smaller code improvements, mainly convert if/else to ternary expressions and removal of redundant intermediate variables.
- Merge the error handling objects/functions into one file, since they are closely related and not very long.
- Fix/suppress newly emerged eslint errors.
- Fix type error in `WorkAtOrganizationForm.tsx` (see "Open Issues" below).

### Open issues

- ~~In `WorkAtOrganizationForm.tsx`, TS complains about a missing `isAlreadyVerified` property when assigning to type `WorkAtOrganizationInput`. Aside from the fact that the type error happens now (which could stem from the rather complicated type mapping going on in `formType.ts` involving `null` and `undefined`), I'm a but stumped as to what this property actually means and if it should be in an "input" type, rather than computed as an "output" type from the application data. This smells a bit of a data misuse opportunity to me.~~ (Solved)

### Side Effects

None

### Testing

Since this PR touches every place that issues GraphQL queries, everything has to be checked unfortunately.

### Resolved Issues

Fixes: #2816


f1sh1918 (commented):
Acutally good first draft 👍 

I added some comments. (even i know this is WIP)
A general problem is that in lots of cases after a mutation update the refetch does not work. F.e. for notes or for userEdit (i edit the mail address of a user and UI does not change f.e. in the table after saving dialog) and an entire page reload is needed

Statistics is also broken
Bildschirmfoto 2026-03-27 um 11 51 16


A second issue that i also commented is what i talked about in our discussion that the missing possibility of callbacks usage (in queries and mutations, which only called once) may result in some rerendering issues if queries are not explicitly called in async functions f.e. onButton click.
These cases might have to be wrapped in `useEffects` to avoid unneeded rerenderings (also added a comment).

For the type problem i didn't find a better solution tbh. i guess removing `immutableTypes: true` ( to avoid generating types `readonly`) is also not a good practise and may cause other issues

administration/src/provider/UrqlClientProvider.tsx
```diff
@@ -0,0 +1,32 @@
...
+import getApiBaseUrl from '../util/getApiBaseUrl'
+import { AuthContext } from './AuthProvider'
+
+const UrqlClientProvider = ({ children }: { children: ReactElement }): ReactElement => {
+  const authContext = useContext(AuthContext)
+
+  return (
+    <Provider
+      value={
+        new Client({
```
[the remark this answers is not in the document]
andrew8er: Fixed with `useMemo()`.

andrew8er:
@f1sh1918 

> A second issue that i also commented is what i talked about in our discussion that the missing possibility of callbacks usage (in queries and mutations, which only called once) may result in some rerendering issues if queries are not explicitly called in async functions f.e. onButton click.
These cases might have to be wrapped in useEffects to avoid unneeded rerenderings (also added a comment).

I actually prefer the urql API, since the error handling is a bit more complicated in real world scenarios than the Apollo callback based API assumes: application level errors show up in the `onSuccess` callback as well. The rerendering issues are more likely due to our sloppy state handling, judging from the issues I fixed up until now for this PR.

> 1.UserEdit is still broken: Edit user name and save nothing happens in the UI. Update: This is also an issue on main. Will create a ticket for that: Here is PR https://github.com/digitalfabrik/entitlementcard/pull/2944

I actually came across this as well while rebasing and came to the same solution.

> 2. "Benutzereinstellungen" were not saved correctly. If there are no user notifications and i enable them and save. Then go to other page and back to this page they are still not set. Probably also an UI issue with missing refetch.

I improved this component. On errors, it will now reset to it's original state. 

> 3. i faced the issue when i create an application and then go back to the administration area i get a broken pipe error and can not reach the backend anymore

Isn't this more likely a backend error? I need to test this…

administration/src/routes/applications/components/ApplicationCard.tsx
```diff
@@ -249,66 +251,63 @@ const ApplicationCard = ({
...
-        const { title } = getMessageFromApolloError(error)
-        enqueueSnackbar(title, { variant: 'error' })
-      })
+    if (result.error) {
+      const { title } = messageFromGraphQlError(result.error)
+      enqueueSnackbar(title, { variant: 'error' })
+    } else if (result.data) {
+      setRejectionDialogOpen(false)
+      onChange({ ...application, ...result.data.application })
+      enqueueSnackbar(t('applicationRejectedToastMessage'), { variant: 'success' })
```
seluianova: there is a merge issue: we don't have this translation anymore.
we have `applicationRejectedEmailNotSent` and `applicationRejectedEmailSent`

f1sh1918 (approved):
Great work 👍 
Tested all projects and main functions

administration/src/routes/applications/components/ApplicationCard.tsx
andrew8er: Thanks, that was an annoying rebase…

I tweaked the conditions a little, namely the last branch that catches all errors with no user data, can you recheck:

```ts
    const errors = result.error?.graphQLErrors ?? []

    if (result.data?.application) {
      setRejectionDialogOpen(false)
      onChange({ ...application, ...result.data.application })

      if (errors.length > 0) {
        enqueueSnackbar(t('applicationRejectedEmailNotSent'), { variant: 'warning' })
      } else {
        enqueueSnackbar(t('applicationRejectedEmailSent'), { variant: 'success' })
      }
    } else if (result.error) { // <-- was (errors && errors.length > 0)
      const { title } = messageFromGraphQlError(result.error)
      enqueueSnackbar(title, { variant: 'error' })
    }
```

seluianova (approved):
I couldn't review every single change, but based on my testing so far, I haven't found any issues 👍

PR #2944, opened 2026-03-31
2943: Current user edit fix

f1sh1918:
### Short Description

Currently the user table will not be updated after the current user was edited.

### Proposed Changes

- also execute onSuccess for the current user to update the user table

### Side Effects

- none

### Testing

See issue instructions

### Resolved Issues

Fixes: #2943


seluianova (approved):
works like a charm 👍

PR #2946, opened 2026-03-31
2792: Improve application rejection notification handling

### LINKED ISSUE #2355: Add tests for reject application mutation
### Describe the Problem
Add integration tests for rejectApplicationStatus() mutation

### LINKED ISSUE #2792: Improve Application rejection notification handling
### Describe the Problem

Follow up of #2522 

Currently we do not catch all possible errors and give feedback to the user

These scenarios should be covered:
- application rejected & email sent
- application rejected & email not sent
- failed to reject application

### Related Issues

#2522

seluianova:
### Short Description
When an admin rejects an application and the email to the applicant fails to send, the exception is ignored. As a result, no error is shown to the admin, and the message "Der Antragsteller oder die Antragstellerin wurde per E-Mail informiert" is displayed instead.

### Proposed Changes
- if an application has been successfully rejected but the email fails to send, return a "partial success" response: both data and an error
- add a new toast message for this case

### Side Effects
no?

### Testing

**Dev only:**
Adjust the `sendApplicationRejectedMail()` function to throw a `MailNotSentException`, then try to reject an application in the administration UI.

**Staging environment:**
Only the positive flow can be tested:

1. Create an application via `/beantragen`.
2. Log in as a region admin and reject the application:
   - The application status should change to "rejected"
   - An email notification should be sent to the applicant
   - The toast `Der Antrag wurde abgelehnt. Der Antragsteller oder die Antragstellerin wurde per E-Mail informiert.` should be displayed to the admin

### Resolved Issues
Fixes: #2792 
Fixes: #2355


f1sh1918 (approved):
Tested. Looks good 👍 
Couldn't find any issues

PR #2948, opened 2026-04-01
Log store import result

f1sh1918:
### Short Description

Currently we have no information how many stores will be created or deleted each import run. Due to our favorites function (not in usage) it would be good to know how stable stores are in the database.

I just tested this to get an impression how many store will be recreated. Maybe this is not 100% needed

Here is the salt pr: https://git.tuerantuer.org/DF/salt/pulls/631

### Proposed Changes

<!-- Describe this PR in more detail. -->

 - update gitignore                                                                                                                                                                                                                                                                                                   
  - update docs
  - add import result logging                                                                                                                                                                                                                                                                                          
  - add output path  

### Side Effects

<!-- List all related components that have not been explicitly changed but may be affected by this PR -->

- none

### Testing

- add `importResultOutput: src/main/resources/import/import_result.csv` to your `config.local.yml`
- run "DB import online data`
- check `import_result.csv`

### Resolved Issues

<!-- List all issues which should be closed when this PR is merged. -->

Fixes: #

### DIFF
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -10,7 +10,6 @@ local.properties
 src/main/resources/config/config.local.yml
 
 src/main/resources/freinet_url.txt
-src/main/resources/import/filtered_stores.csv
-src/main/resources/import/import_result.csv
+logs/
 
 rejectedStores.json
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -10,5 +10,7 @@ local.properties
 src/main/resources/config/config.local.yml
 
 src/main/resources/freinet_url.txt
+src/main/resources/import/filtered_stores.csv
+src/main/resources/import/import_result.csv
 
 rejectedStores.json
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
@@ -35,7 +35,7 @@ data class ProjectConfig(
     val matomo: MatomoConfig?,
     val freinet: FreinetConfig?,
     val filteredStoresOutput: String?,
-    val importResultOutput: String?,
+    val importLogFilePath: String?,
 )
 
 data class ServerConfig(val dataDirectory: String, val host: String, val port: String)
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
@@ -35,6 +35,7 @@ data class ProjectConfig(
     val matomo: MatomoConfig?,
     val freinet: FreinetConfig?,
     val filteredStoresOutput: String?,
+    val importResultOutput: String?,
 )
 
 data class ServerConfig(val dataDirectory: String, val host: String, val port: String)
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
@@ -11,8 +11,10 @@ import java.time.LocalDate
 
 class StoreImportResultCsvOutput(config: ImportConfig) :
     PipelineStep<StoreImportReturnResultModel, StoreImportReturnResultModel>(config) {
+    private val logger = LoggerFactory.getLogger(StoreImportResultCsvOutput::class.java)
+
     override fun execute(input: StoreImportReturnResultModel): StoreImportReturnResultModel {
-        val outputPath = config.project.importResultOutput
+        val outputPath = config.project.importLogFilePath
         if (!outputPath.isNullOrBlank()) {
             writeCsvWithImportResult(outputPath, input)
         }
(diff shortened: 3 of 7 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
```diff
@@ -0,0 +1,58 @@
...
+    override fun execute(input: StoreImportReturnResultModel): StoreImportReturnResultModel {
+        val outputPath = config.project.importResultOutput
+        if (!outputPath.isNullOrBlank()) {
+            writeCsvWithImportResult(outputPath, input)
+        }
+        return input
+    }
+
+    private fun writeCsvWithImportResult(outputPath: String, result: StoreImportReturnResultModel) {
+        val logger = LoggerFactory.getLogger(StoreImportResultCsvOutput::class.java)
```
seluianova: 🙃 loggers are conventionally defined on the class level

seluianova (approved):
tested, lgtm!

backend/src/main/kotlin/app/ehrenamtskarte/backend/config/BackendConfiguration.kt
```diff
@@ -35,6 +35,7 @@ data class ProjectConfig(
     val matomo: MatomoConfig?,
     val freinet: FreinetConfig?,
     val filteredStoresOutput: String?,
+    val importResultOutput: String?,
```
andrew8er: 🔧 I'd name this `importLogFilePath` to better reflect what it is, "output" is way to generic.

backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
andrew8er: In Kotlin, I usually have them on the file level, since often I want to call them from (extension) functions on the file level as well. From a resource usage/initialization standpoint there's no difference with class-level vs file level, both are globally initialized at startup and hang around the JVM forever.

backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
```diff
@@ -0,0 +1,58 @@
...
+                format.print(writer).use { printer ->
+                    printer.printRecord(
+                        LocalDate.now(config.project.timezone),
+                        result.storesCreated + result.storesUntouched,
+                        result.storesCreated,
+                        result.storesDeleted,
+                        result.storesUntouched,
+                    )
+                }
+            }
```
andrew8er: 🙃 
```suggestion
            Files.newBufferedWriter(
                Paths.get(outputPath),
                java.nio.file.StandardOpenOption.CREATE,
                java.nio.file.StandardOpenOption.APPEND,
            ).use { writer ->
                CSVFormat.RFC4180.builder()
                    .apply {
                        if (!Files.exists(Paths.get(outputPath))) {
                            setHeader(
                                "Date",
                                "StoresTotal",
                                "StoresCreated",
                                "StoresDeleted",
                                "StoresUntouched",
                            )
                        }
                    }
                    .get().print(writer).use { printer ->
                        printer.printRecord(
                            LocalDate.now(config.project.timezone),
                            result.storesCreated + result.storesUntouched,
                            result.storesCreated,
                            result.storesDeleted,
                            result.storesUntouched,
                        )
                    }
            }
```

backend/.gitignore
```diff
@@ -10,5 +10,7 @@ local.properties
 src/main/resources/config/config.local.yml
 
 src/main/resources/freinet_url.txt
+src/main/resources/import/filtered_stores.csv
+src/main/resources/import/import_result.csv
```
andrew8er: ❌ Files from the `resource` paths will be _compiled into_ the resulting JAR on the next build. We should create and exclude a `logs` directory instead.

andrew8er (changes requested):
Works well, a few things:

- The naming of the config option could be better, on the long run ,we should rename similar options as well.
- The dev path for the log outputs will cause a bloated jar on the next build.

backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
f1sh1918: as we have it mostly on the class level i moved it to the class level

backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
f1sh1918: this leads to no header will be written in the csv for first run.
the file existence check happens after the writer has already created the file
I adjusted the implementation

backend/.gitignore
f1sh1918: i create a log folder and log them there. Good point

backend/src/main/kotlin/app/ehrenamtskarte/backend/import/stores/common/steps/StoreImportResultCsvOutput.kt
andrew8er: Ah yes, your're right.

andrew8er (approved):

PR #2969, opened 2026-04-13
2968: Optimize user csv import

### LINKED ISSUE #2968: Optimize `UserImportController.handleUserImport()`
### Describe the Bug

The current implementation is inefficient and produces timeout errors on larger datasets.

### Steps to Reproduce

The error only happens behind a proxy request timeouts set, but these instructions aid to test the affected code path:

1. Go to http://localhost:3000/
2. Switch project to Koblenz, login as `project-admin@koblenz.sozialpass.app`:`Administrator!`
3. Go to "Projekt verwalten", generate API key
4. Call `POST /users/import`, with generated bearer token and multipart test file from `administration/resources/self-service/import_userentitlements.csv`
5. See error

andrew8er:
### Proposed Changes

- Uses DataSource API for connecting to DB.
- Enables batching of user entitlements inserts. 
- Add a test case for the detection of duplicate user hashes in inputs.
- Add a [Bruno](https://www.usebruno.com/) workspace with rest calls

### Side Effects

~~The call will now abort with a 400 (bad request) if the imported list contains duplicate user hashes. Previously, this would be accepted and later rows would silently overwrite previous rows. I believe duplicate user hashes are an error, but if this is a problem with real world data, we would need to re-implement the previous behavior.~~

### Testing

- Go to http://localhost:3000/
- Switch project to Koblenz, login as project-admin@koblenz.sozialpass.app:Administrator!
- Go to "Projekt verwalten", generate API key
- Run the SQL file `/bruno-workspace/sample-data/userentitlements/insert_userentitled_cards.sql` against the DB.
- Call `POST /users/import`, with generated bearer token and included test files (could use Bruno).
- Before the import took around 20s and it should be now around ~5s, maybe slightly more for the second import 

### Resolved Issues

Fixes: #2968


f1sh1918:
i created a new token and tried to import but i get
```
Authorization token expired, invalid or missing
```

seluianova:
🪲 Cards are not revoked, when user entitlements have been revoked

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/TokenAuthenticator.kt
```diff
@@ -3,34 +3,33 @@ package app.ehrenamtskarte.backend.shared
...
-        val tokenHash = PasswordCrypto.hashWithSHA256(token.toByteArray())
+    val apiTokenEntity = ApiTokensRepository.findByTokenHash(tokenHash)
+        ?: throw UnauthorizedException()
 
-        return transaction {
-            val apiToken = ApiTokensRepository.findByTokenHash(tokenHash)
-                ?.takeIf { it.expirationDate > LocalDate.now() }
-                ?: throw UnauthorizedException()
+    return when {
+        // TODO Why do these cases have a different exception?
```
seluianova: you mean, the unauthorized exception for the expired token and the forbidden exception for a wrong token type?
I think this is standard usage of http status codes 🤔

https://www.rfc-editor.org/rfc/rfc7235#section-3.1

> The 401 (Unauthorized) status code indicates that the request has not
   been applied because it lacks valid authentication credentials

expired = not valid

f1sh1918:
Some improvements i see:
I tested a batchUpsert in my pr (https://github.com/digitalfabrik/entitlementcard/pull/2970) and its way faster (6s vs 2s) with a 20k dataset.
Additionally i would set an index on `Cards.EntitlementId` or it will scan the entire sequence.
I think the property that you added to `Database.kt` only optimizes plain `Inserts` but not on `Upserts`afaics but i'm not 100% sure

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/TokenAuthenticator.kt
andrew8er: Wouldn't both assume an invalid token?

andrew8er:
@f1sh1918 
> I tested a batchUpsert in my pr (https://github.com/digitalfabrik/entitlementcard/pull/2970) and its way faster (6s vs 2s) with a 20k dataset.

Did you really get batch upserts? Because I could not, and this seems to be a deficit of Exposed (see the bug report in the description)

f1sh1918:
> * Support of batch update JetBrains/Exposed#943

Well `batchUpdates` seem not to work but `batchUpsert` works
You can check my pr. I can also provide some better description

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/TokenAuthenticator.kt
seluianova: I don't think so.
wrong token type  = valid token, but missing permissions

andrew8er:
Batch upserts do not work:
```
…
SQL: INSERT INTO userentitlements ("userHash", "startDate", "endDate", revoked, "regionId", "lastUpdated") VALUES ($argon2id$v=19$m=19456,t=2,p=1$nt732mZtSnRnBuODoDUzAvtnI/dm+dKs8FMOJcQVnow, '2026-10-01', '2026-12-31', FALSE, 96, '2026-04-13T13:51:16.819675') ON CONFLICT ("userHash") DO UPDATE SET "startDate"=EXCLUDED."startDate", "endDate"=EXCLUDED."endDate", revoked=EXCLUDED.revoked, "regionId"=EXCLUDED."regionId", "lastUpdated"=EXCLUDED."lastUpdated"
SQL: INSERT INTO userentitlements ("userHash", "startDate", "endDate", revoked, "regionId", "lastUpdated") VALUES ($argon2id$v=19$m=19456,t=2,p=1$MNMo2a4Mysn//WXHkHO0bZ50VO0EpVB1dQQegzwC2G0, '2026-07-01', '2026-12-31', FALSE, 96, '2026-04-13T13:51:16.819675') ON CONFLICT ("userHash") DO UPDATE SET "startDate"=EXCLUDED."startDate", "endDate"=EXCLUDED."endDate", revoked=EXCLUDED.revoked, "regionId"=EXCLUDED."regionId", "lastUpdated"=EXCLUDED."lastUpdated"
SQL: INSERT INTO userentitlements ("userHash", "startDate", "endDate", revoked, "regionId", "lastUpdated") VALUES ($argon2id$v=19$m=19456,t=2,p=1$Hw0oQiDO5YLTFefn8E0vXL91a81Fp5P4+mfGq8J7oSg, '2026-10-01', '2028-06-30', FALSE, 96, '2026-04-13T13:51:16.819675') ON CONFLICT ("userHash") DO UPDATE SET "startDate"=EXCLUDED."startDate", "endDate"=EXCLUDED."endDate", revoked=EXCLUDED.revoked, "regionId"=EXCLUDED."regionId", "lastUpdated"=EXCLUDED."lastUpdated"
…
```

Just a bunch of single inserts…

andrew8er:
@f1sh1918 
Your version is pretty much the same as mine. Difference is:

- I return the revoked updates/insert entitlement IDs and update the cards with this after all updates are done.
- I use `upsertReturning` instead of `batchUpsert` because batching does not work anyway and there is currently no `batchUpsertReturning`.

But I stole your migration though…

f1sh1918:
> @f1sh1918 Your version is pretty much the same as mine. Difference is:
> 
> I return the revoked updates/insert entitlement IDs and update the cards with this after all updates are done.
> 
> But I stole your migration though…

Hm i will retest. I think the way of the bulk upsert was different. At least last time i tested.
Now i get
```
Caused by: org.apache.tomcat.util.http.fileupload.FileUploadException: the request was rejected because no multipart boundary was found
```

andrew8er:
Your body is probably not formatted as MIME multipart

f1sh1918:
> Your body is probably not formatted as MIME multipart

why did this change? i just want to ensure nothing breaks for koblenz
I was able to upload before
we used `multipart/form-data` before
Pretty sure this change will break their import

andrew8er:
I din't change anything for the server side in this regard. I meant your client is not formatting the request accordingly. Multipart/form-data should also work, although it is the absolutely worst format for that.

andrew8er:
This is what I use with Bruno:
Screenshot 2026-04-13 at 16 15 28

f1sh1918:
> I din't change anything for the server side in this regard. I meant your client is not formatting the request accordingly. Multipart/form-data should also work, although it is the absolutely worst format for that.

🙈 sorry my bad. i didn't have the testfile on that branch

~~Well i also realized that "batch" is not really working apparently...~~
But the import on my branch is 5s faster. 💭 
I'm about to investigate why. I think first creating the lists in-memory might be faster

Update: Claude says that the batch is working but exposed logs each row in the sql log (even its batched)

andrew8er:
@f1sh1918 
I re-instated the batching and optimized a few other smaller things. I now get import speeds comparable with your implementation – indeed, they are identical on the DB level.

I also added a check for duplicate user hashes, which your Claude generated test files contain plenty of. This would error at the batch insert call with a `batch upsert ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time`. The included test file in the `/bruno-workspace` contains a cleaned version.

f1sh1918:
> @f1sh1918 I re-instated the batching and optimized a few other smaller things. I now get import speeds comparable with your implementation – indeed, they are identical on the DB level.
> 
> I also added a check for duplicate user hashes, which your Claude generated test files contain plenty of. This would error at the batch insert call with a `batch upsert ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time`. The included test file in the `/bruno-workspace` contains a cleaned version.

Alright i will retest all tommorow. I will cleanup the test files though

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/TokenAuthenticator.kt
```diff
@@ -3,34 +3,32 @@ package app.ehrenamtskarte.backend.shared
...
 import java.time.LocalDate
 
-// todo: this will be removed or refactored after #2452
-object TokenAuthenticator {
-    private fun authenticateToken(header: String?, neededType: ApiTokenType): ApiTokenEntity {
-        val token = header
-            ?.takeIf { it.startsWith("Bearer ") }
-            ?.removePrefix("Bearer ")
-            ?: throw UnauthorizedException()
+private val messageDigestSha256 by lazy { MessageDigest.getInstance("SHA-256") }
```
f1sh1918: 🔧 MessageDigest is stateful and not thread-safe. Concurrent calls to .digest() will corrupt each  
  other's results, causing valid tokens to be rejected or wrong hashes to match. The old PasswordCrypto.hashWithSHA256() correctly instantiated a fresh MessageDigest per call.

I'm not sure if we may have concurrent request and this will be a real issue

backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
```diff
@@ -70,13 +71,15 @@ object Database {
         log: Boolean = true,
     ): org.jetbrains.exposed.v1.jdbc.Database =
         connect(
-            config.postgres.url,
-            driver = "org.postgresql.Driver",
-            user = config.postgres.user,
-            password = config.postgres.password,
+            datasource = PGSimpleDataSource().apply {
```
f1sh1918, line 74: ❓ I just read that we could still improve this a bit by using connection pooling f.e. with `HikariCP` (also used in https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-jdbc
But i think we are talking about 20-50ms. So this is very optional

f1sh1918 (commented):
Codewise looks good and the import speed is amazing. Good work 👍 
Less than 2s 🚀 

One "maybe" issue and one minor optional improvement i found

I added some test data.
The pr description should be updated, because we should also test bulk card revokes
Import file a, then run sql script and import file b

backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
andrew8er: I have this in one of my other projects. Is pretty easy to set up, but it has some config gotchas one needs to be aware of.

andrew8er: The speedup is really dependent on the amount of concurrent requests, but a connection pool is really standard for backend services.

f1sh1918: okay if you have the mood to implement it, feel free. We could also do it in a separate pr but i think its not that much effort

andrew8er: I'd do that in a separate PR…

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/TokenAuthenticator.kt
andrew8er: `MessageDigest` is now local and recreated for every use.

f1sh1918 (approved):
LGTM

backend/src/main/kotlin/app/ehrenamtskarte/backend/routes/UserImportController.kt
```diff
@@ -52,127 +59,182 @@ class UserImportController(
...
+                        } else {
+                            throw UserImportException(it.recordNumber, "Missing data")
+                        }
+                    }
+
+                    // Check for duplicate user hashes, first do a coarse check with contentHashCode()
+                    entries.groupBy { it.userHash.contentHashCode() }.forEach { (_, entries) ->
+                        // Do an exact check with toList()
+                        if (entries.size != entries.distinctBy { it.userHash.toList() }.size) {
+                            throw UserImportException("Duplicate user hash found with ${entries.size} entries")
```
seluianova: ❓maybe add line numbers?
the file provided by Christoph also have duplicates but without line numbers it is not possible to locate them and fix the file

f1sh1918: @connium thats important to mention, since current import file would break

f1sh1918: i will provide a list with duplicates to you. 
It looks like they don't remove old permissions (start and endDate) but just create new entries with new start/end date.

f1sh1918: The big question is, if we do a revoke for the first entry (dates expired), will the second entry revoke the revokation?

f1sh1918: i think we have to include that a card can be re-revoked!

seluianova: > i think we have to include that a card can be re-revoked!

but they didn't request this feature. they only requested "revoke"

f1sh1918: yes i know but i think this make sense in general and would lead to issues

andrew8er: The problem here is that this call was never formally specified in detail. This should be an absolute requirement for every public call/query.

> ❓maybe add line numbers?

We accept pretty borked input data now, so there should not be an immediate problem here.

For the future however, I would like to specify a detailed reporting format that should contain these issues to debug, but the format should be well thought out. When we provide this data, our users will rely on it and changing will become a hassle. So I would purposely _not_ include this info until we have the format sorted out.

The PR provides some handwavy stats in the `message` property, but this is meant to be strictly informational and not to be processed further.

Besides that: does anybody even see the response of this call and would act on it?

f1sh1918:
Here is some query and duplicates from the csv for testing. It has to be ensured the last entry state will be reflected in the database for `userentitlement.revoked` and `card.revoked`
```
$argon2id$v=19$m=19456,t=2,p=1$xr+oE7ZdRMOBnKLdo+hdQfZJslmym40CjpZQMJniG+c (120: revoked=false, 8298: revoked=true)                                                                                                                                                                                                  
  $argon2id$v=19$m=19456,t=2,p=1$Hfw6p3lE+t3GRDxywlNyiclDG5c2/QRsdsb6aSpCUK4 (1855: revoked=false, 8672: revoked=true)                                                                                                                                                                                                 
  $argon2id$v=19$m=19456,t=2,p=1$ssft8cqiHWiGSxKMnGIyxF93bK4frAvnLq86hCx/39o (1913: revoked=true, 9178: revoked=false)                                                                                                                                                                                                 
  $argon2id$v=19$m=19456,t=2,p=1$G7XGPe+XPxyQ08dypTmSnPdfuttoImyoVXNjJB6XWCM (1943: revoked=true, 4763: revoked=false)                                                                                                                                                                                                 
  $argon2id$v=19$m=19456,t=2,p=1$aBg80b9RdNrGuq1J9ZRdBGqaDjYIBE/0ZdakdWM19mw (4117: revoked=false, 4118: revoked=true)                                                                                                                                                                                                 
  $argon2id$v=19$m=19456,t=2,p=1$nD2mM8yByBGFuH/A4q2b8PK835vYm3wYKaG4kX9nsgQ (4625: revoked=false, 4626: revoked=true)                                                                                                                                                                                                 
  $argon2id$v=19$m=19456,t=2,p=1$wL2jWfAuAL6I7HQJwz2TQ1F098h6T+yNCIBB2eh88jA (4852: revoked=true, 4853: revoked=false) 
```

sql
```
SELECT
    encode(ue."userHash", 'escape')  AS "userHash",
    ue."startDate",
    ue."endDate",
    ue."revoked"                     AS "entitlement_revoked",
    ue."lastUpdated",
    c.id                             AS "card_id",
    c."codeType",
    c."revoked"                      AS "card_revoked",
    c."firstActivationDate",
    c."issueDate"
FROM userentitlements ue
    LEFT JOIN cards c ON c."entitlementId" = ue.id
WHERE encode(ue."userHash", 'escape') = ANY(ARRAY[
    '$argon2id$v=19$m=19456,t=2,p=1$xr+oE7ZdRMOBnKLdo+hdQfZJslmym40CjpZQMJniG+c',
    '$argon2id$v=19$m=19456,t=2,p=1$Hfw6p3lE+t3GRDxywlNyiclDG5c2/QRsdsb6aSpCUK4',
    '$argon2id$v=19$m=19456,t=2,p=1$ssft8cqiHWiGSxKMnGIyxF93bK4frAvnLq86hCx/39o',
    '$argon2id$v=19$m=19456,t=2,p=1$G7XGPe+XPxyQ08dypTmSnPdfuttoImyoVXNjJB6XWCM',
    '$argon2id$v=19$m=19456,t=2,p=1$aBg80b9RdNrGuq1J9ZRdBGqaDjYIBE/0ZdakdWM19mw',
    '$argon2id$v=19$m=19456,t=2,p=1$nD2mM8yByBGFuH/A4q2b8PK835vYm3wYKaG4kX9nsgQ',
    '$argon2id$v=19$m=19456,t=2,p=1$wL2jWfAuAL6I7HQJwz2TQ1F098h6T+yNCIBB2eh88jA'
    ])
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/routes/UserImportController.kt
f1sh1918: I think thats a question for @connium I guess they may only check at some point if the data was not update or hopefully they will be informed if the import fails.
I mean we have swagger for api documentation. i think we should define it there and inform the customer to properly check this documentation
I can add something

backend/src/main/kotlin/app/ehrenamtskarte/backend/routes/UserImportController.kt
```diff
@@ -37,142 +43,195 @@ class UserImportController(
...
+                    this[UserEntitlements.regionId] = it.regionId
+                    this[UserEntitlements.lastUpdated] = now
+                }
             }
-        }
 
-        return ResponseEntity.ok(mapOf("message" to "Import successfully completed"))
-    }
+            val activeCount = deduplicatedEntries.filter { !it.revoked }
+                .chunked(4096).map { updateCardRevocationStatus(it, false) }.sum()
```
seluianova, line 139:
```suggestion
            val activeCount = deduplicatedEntries.filter { !it.revoked }
                .chunked(4096).sumOf { updateCardRevocationStatus(it, false) }
```

🙃 same for revokedCount

seluianova: 🪲 but actually this value is not accurate. if the card’s status hasn’t changed, it is still counted. 
e.g. it says “50 newly revoked cards,” even if those cards were revoked before.

backend/src/main/kotlin/app/ehrenamtskarte/backend/routes/UserImportController.kt
```diff
@@ -37,142 +43,195 @@ class UserImportController(
...
-        startDate: LocalDate,
-        endDate: LocalDate,
-        revoked: Boolean,
-        regionId: Int,
+private fun updateCardRevocationStatus(entries: List<CsvEntry>, revoked: Boolean) =
+    Cards.update(
+        where = {
+            Cards.entitlementId inSubQuery UserEntitlements
+                .select(UserEntitlements.id)
+                .where { UserEntitlements.userHash inList entries.map { it.userHash.toByteArray() } }
```
seluianova, line 233: 🙃 we include every single entry from csv here, even when the import contains no changes (no entitlements were actually updated).
I think, ideally, we could collect only updated `UserEntitlements.id` and update the cards by `Cards.entitlementId`.
but considering that the current implementation already works very quickly, it's just a thought.

**UPD**. 
> we could collect only updated `UserEntitlements.id`

but I guess this could have own performance implications so I'm not really sure what is better

backend/src/main/kotlin/app/ehrenamtskarte/backend/routes/UserImportController.kt
```diff
@@ -37,142 +43,195 @@ class UserImportController(
...
+            | regionKey  | string          | Region identifier (Amtlicher Gemeindeschlüssel), e.g. `07111` |
+            | userHash   | string          | Argon2id hash of the user identifier |
+            | startDate  | dd.MM.yyyy      | Start date of the entitlement period |
+            | endDate    | dd.MM.yyyy      | End date of the entitlement period, must be after startDate |
+            | revoked    | true/false      | Whether the entitlement is revoked |
+
+            **Example CSV:**
+            ```
+            regionKey,userHash,startDate,endDate,revoked
+            07111,${'$'}argon2id${'$'}v=19${'$'}m=19456,t=2,p=1${'$'}AAAA...,01.01.2026,31.12.2026,false
```
seluianova, line 62: 🙃 userHash should be enclosed in quotation marks because it contains commas

seluianova (approved):
works good as far as I tested, couldn't find any issues (just a few minor/optional notes).
the performance improvement looks great ⭐

backend/src/test/kotlin/app/ehrenamtskarte/backend/routes/UserImportTest.kt
```diff
@@ -298,7 +298,6 @@ internal class UserImportTest : IntegrationTest() {
         val response = importUsers(csvFile)
 
         assertEquals(HttpStatus.OK, response.statusCode)
-        assertEquals("Import successfully completed", response.body?.message)
```
seluianova, line 301: 🙃 I think ideally we could update the expected message here, to test it as well

andrew8er: I see this message a purely informational. It is just a short, human readable string. It should not be relied upon, not be parsed and expected to change at any time. This is why I would not test this at all actually.

We should ad some properties for better reporting at some point, when we know what our users need.

PR #2970, opened 2026-04-13
User import improvements

f1sh1918:
### Proposed Changes

- create UserEntitlementImportEntry data class to carry import row data
- validate and collect all CSV rows into a `List<UserEntitlementImportEntry>` in memory, failing fast on any validation error before touching the DB. 
- batchUpsert() using Exposed's batchUpsert with ON CONFLICT (userHash) DO UPDATE, chunked at 1000 rows to stay under PostgreSQL's 65535 bind-parameter limit
- `shouldReturnGeneratedValues = false` avoids fetching generated keys after each batch
- bulk revocation of cards
- add index ` cards_entitlementid_idx` to speed up bulk revocation
- provide batch test data

### Side Effects

-

### Testing

1. Run migration
2. Create api token (koblenz projekt manager)
3. Use postman or curl and use some csv data from `/resources`
4. Run the provided sql file to created cards that are related with entitlements
5. Update with the next csv import file (next month)

### Resolved Issues

Fixes: #


f1sh1918:
here is some draft with batch card revoke, updated index and some test data @andrew8er 
Maybe you can take some data or ideas out of it

Reduced importing times from ~15s to ~2s with 20k imports

f1sh1918:
in favor of: #2969

PR #2972, opened 2026-04-14
2961: PDF form fields not filled

### LINKED ISSUE #2961: PDF form fields not filled
### Describe the Bug

For nuernberg we have a letter header with optional address data that can be filled within the bulk card creation or query params. Currently its not visible on the pdf.
Adobe acrobat doesn't show it at all. Some like chrome pdf integration shows it if you click on the form

### Steps to Reproduce

1. Login with nuernberg
2. Click on "Mehrere Karten erstellen"
3. use csv with address data (f.e. from resources) https://github.com/digitalfabrik/entitlementcard/blob/b7775507bc683d98688047723209cc9cb1e1c9bf/administration/resources/cards/sample-bulk-import-nuernberg.csv
4. Create the pdf
5. Open the pdf and check that the form fields contain the particular content from the csv

### Expected Behavior

Address data should be printed in the form fields

andrew8er:
### Short Description

Updates the PDF templates to fix form transparency problems.

### Proposed Changes

- Update template files with PDF-X/3 export settings.
- Slightly improve PDF config and drawing code.
- Update test file
- Use short year formatting for Nürnberg Sozialpass to avoid text wraps.
- Add the template design files and export instructions.

### Side Effects

None

### Testing

Create PDFs, see https://github.com/digitalfabrik/entitlementcard/pull/2836 for detailed instructions.

Fixes: #2961


administration/src/cards/pdf/elements.ts
```diff
@@ -22,26 +16,27 @@ const DEFAULT_QUIET_ZONE_SIZE = 4 // pt
...
   font: PDFFont,
+  formFonfigs: FormConfig[],
   formElementProps: PdfFormElementProps,
 ): void => {
+  const form = page.doc.getForm()
   const lineHeight = font.heightAtSize(formElementProps.fontSize) + 6
 
-  formFields.forEach((formField, index) => {
-    formField.addToPage(page, {
+  formFonfigs.forEach((config, index) => {
```
f1sh1918: ❓ Typo?

f1sh1918:
My personal preference is that i would keep the birthday in dd.mm.yyyy on pdf. since i think there is sufficient space and it's more relevant than for valid period

andrew8er:
@f1sh1918 

I tried that and had a line wrap on some occasions.
[berechtigungskarten-1.pdf](https://github.com/user-attachments/files/26741771/berechtigungskarten-1.pdf)

f1sh1918:
> @f1sh1918
> 
> I tried that and had a line wrap on some occasions. [berechtigungskarten-1.pdf](https://github.com/user-attachments/files/26741771/berechtigungskarten-1.pdf)

i'm only talking about the birthday not the validity period. Birthday shouldn't be a problem

administration/src/cards/pdf/pdfFactory.ts
```diff
@@ -166,11 +161,8 @@ export const generatePdf = async (
       type: 'application/pdf',
     })
   } catch (error) {
-    // TODO What is the point of this?
-    if (error instanceof Error) {
-      throw new PdfError(error.message)
-    }
-    throw error
+    console.log('PDF generation error:', error)
```
f1sh1918: 🙃 please remove

f1sh1918 (approved):
Looks good so far. 👍 

One change that should be mentioned, is that the location is fixed on the forth line.
But immo this is the better/correct way @connium ?
Left (new) right (old)

Image

administration/src/project-configs/nuernberg/config.ts
```diff
@@ -107,3 +112,26 @@ export const config: ProjectConfig = {
   showBirthdayExtensionHint: false,
   locales: buildConfigNuernberg.common.appLocales,
 }
+
+// eslint-disable-next-line prefer-arrow/prefer-arrow-functions
+function createAddressFormFields(pageIndex: number, info: CardInfo, card: Card): FormConfig[] {
```
f1sh1918, line 117: ❓ why we can't use arrow function here?

andrew8er: Because a function declaration (`function foo()`) is not completely equivalent to a function expression (`const function = () => {}`). The former is callable even when it appears below the call site. 

This is why I think this eslint rule is actually utter BS. It effectively enforces a source layout that is bottom up instead of top-down. It actively makes the code worse.

f1sh1918:
@andrew8er 
Test instructions for koblenz should also be added. Since the misplace of the pdf pass (content printed over label) is also an important issue to fix

administration/src/project-configs/nuernberg/config.ts
f1sh1918: Feel free to disable it

seluianova (approved):
tested 👍

andrew8er:
@f1sh1918 
> Test instructions for koblenz should also be added. Since the misplace of the pdf pass (content printed over label) is also an important issue to fix

Do you mean the issue reported by @connium ? I could not reproduce this and this PR has no specific fix for this. It might have been a cache issue.

administration/src/project-configs/nuernberg/config.ts
andrew8er: I wanted to go through all the eslint rules at some point. I have the feeling that we have too much of them, with quiet some overlap, and the AirBnb rules in particular cause dependency blocks:  https://github.com/airbnb/javascript/issues/2961

f1sh1918:
> Looks good so far. 👍
> 
> One change that should be mentioned, is that the location is fixed on the forth line. But immo this is the better/correct way @connium ? Left (new) right (old)
> 
> Image

@connium mentioned that empty lines are not allowed.
Please revert this change @andrew8er

connium:
> Looks good so far. 👍
> 
> One change that should be mentioned, is that the location is fixed on the forth line. But immo this is the better/correct way @connium ? Left (new) right (old)
> Image


The spec says something like:
_Beachten Sie, dass im gesamten Anschriftfeld keine Leerzeilen erlaubt sind – die Anschriftzone wird von der gedachten Linie abwärts beschriftet, die Zone mit der Rücksendeangabe und Zusatzangaben von unten nach oben_

andrew8er:
> Beachten Sie, dass im gesamten Anschriftfeld keine Leerzeilen erlaubt sind – die Anschriftzone wird von der gedachten Linie abwärts beschriftet, die Zone mit der Rücksendeangabe und Zusatzangaben von unten nach oben

This is just peak German bureaucracy…

f1sh1918:
> @f1sh1918
> 
> > Test instructions for koblenz should also be added. Since the misplace of the pdf pass (content printed over label) is also an important issue to fix
> 
> Do you mean the issue reported by @connium ? I could not reproduce this and this PR has no specific fix for this. It might have been a cache issue.

Yes i also realized that. then i have no clue why this happens on staging. But it also means that we don't need this in the hotfix release

connium:
> > Beachten Sie, dass im gesamten Anschriftfeld keine Leerzeilen erlaubt sind – die Anschriftzone wird von der gedachten Linie abwärts beschriftet, die Zone mit der Rücksendeangabe und Zusatzangaben von unten nach oben
> 
> This is just peak German bureaucracy…

In a former job I configured scanners and their recognition software and I learned, that algorithms hate blank lines if the content is meant to be a single object. :nerd_face: I guess this rule origins from postal service providers that are keen on optimizing the sorting lines to be as fast and robust as possible.

f1sh1918:
retested look good! 👍

PR #2980, opened 2026-04-16
2979: Statistics filter fix

### LINKED ISSUE #2979: Statistics can't be filtered
### Describe the Bug

If a user choose a startDate and an endDate statistics should reflect that

### Steps to Reproduce

1. Login with region admin or project admin (bayern)
2. Go to statistics
3. Use the filterbar
4. See error that nothing changes

### Expected Behavior

The results will be filtered

### Related Issues

Not sure but could be the urql issue

f1sh1918:
### Short Description

Changes of the DateInput fields don't apply to the statistics query, so the data will never change

### Proposed Changes

- use the dateRange state for the query
- simplify on apply filter

### Side Effects

- none

### Resolved Issues

Fixes: #2979


seluianova (approved):
great finding, thanks for the quick fix!

andrew8er (approved):

PR #2986, opened 2026-04-20
2981: Rename mail notification

### LINKED ISSUE #2981: Rename "MailNotification" parameter
### Describe the Problem

At the moment at EAK Bayern we accept a parameter named `MailNotification`. This seems to be less intuitive. We already had a discussion whether we can rename it.
The parameter hasn't been used so far, and we can safely rename it.

Rename the property to a more yet generic but meaningful name.

### Related Issues

#2778, #2779

### Additional Information

This parameter is used for CSV import as well as for the parameterized URL.
Also check if card creation from application uses this parameter somehow (not sure)

f1sh1918:
### Short Description

We want to change our MailNotification query parameter and extension to something more general

### Proposed Changes

- Rename mail notification to email                                                                                                                                               
- adjust tests                                                                                                                                                                    
- add documentation                                                                                                                                                               
- adjust sample data

### Side Effects

- none

### Testing

1. Login with bavaria region admin/manager
2. Ensure that sending confirmation mail is enabled (Region verwalten)
3. use these query params for card creation. Use a proper mail address that you have access to 
```
/cards/add?Name=Freinet+User&Ablaufdatum=09.03.2029&EMail=test@test.com
```
4. Check if card creation works and you received a confirmation mail
5. Go to bulk import and use this csv data (use also proper mail addresses)
```
Name,Ablaufdatum,Kartentyp,EMail
Bärta Bäumeister,03.3.2027,Standard,test@test.de
Rudi Dutschke,,gold,test@test.de
Erika Muster,,Goldkarte,test@test.de
```
### Resolved Issues

Fixes: #2981


administration/src/cards/extensions/EMailExtension.ts
```diff
@@ -0,0 +1,25 @@
+import type { Extension } from './extensions'
+
+export const EMAIL_EXTENSION_NAME = 'email'
```
andrew8er, line 3: ❓ The original name was `emailNotification`. Won't we have problems with that in the future? Where is this extension data stored? Is it only a link parameter?

If yes, wouldn't it make existing links unusable?

andrew8er (approved):
LGTM.

I wonder, however, if we create a (small) problem with outdated links that were created before this PR.

administration/src/cards/extensions/EMailExtension.ts
connium: The `MailNotification` property hasn't been used in the past for EAK. Neither for CSV import nor for links. So we shouldn't encounter any issues.

If I understand the extension mechanism correctly, the extensions are bound to projects and this extension is only used for EAK Bayern, right?

andrew8er: I see, in this case I don't expect any problems 👍 .

administration/src/project-configs/bayern/config.ts
```diff
@@ -54,7 +54,7 @@ const applicationJsonToCardQuery = (json: JsonField<'Array'>): string | null =>
     query.set(config.card.extensionColumnNames[cardTypeExtensionIdx] ?? '', value)
     if (personalData.emailAddress) {
       const applicantMailNotificationExtensionIdx = config.card.extensions.findIndex(
```
bahaaTuffaha: ❓ Do we need to change the name of this variable too?

administration/src/project-configs/common/applicationFeatures.ts
```diff
@@ -41,7 +41,7 @@ export const applicationJsonToCardQuery = (json: JsonField<'Array'>): string | n
 
     if (personalData.emailAddress) {
       const applicantMailNotificationExtensionIdx = config.card.extensions.findIndex(
```
bahaaTuffaha: ❓ Do we need to change the name of this variable too?

docs/card-query-params.md
```diff
@@ -0,0 +1,67 @@
...
+/cards/add?Name=Thea+Test&Ablaufdatum=03.03.2026&Geburtsdatum=01.01.2000&Passnummer=12345678&Pass-ID=123
+/cards/add?Name=Thea+Test&Ablaufdatum=03.03.2026&Geburtsdatum=01.01.2000&Passnummer=12345678&Pass-ID=123&Adresszeile+1=Teststraße+3
+/cards/add?Name=Thea+Test&Ablaufdatum=fehler&Geburtsdatum=fehler&Passnummer=abc&Pass-ID=abc
+```
+
+| Parameter      | Required | Format / Values                    | Description                                              |
+|----------------|----------|------------------------------------|----------------------------------------------------------|
+| `Name`         | yes      | String                             | Full name of the card holder                             |
+| `Ablaufdatum`  | yes      | `dd.MM.yyyy`                       | Expiry date                                              |
+| `Startdatum`   | yes      | `dd.MM.yyyy`                       | Activation start date. Defaults to today if omitted      |
```
bahaaTuffaha, line 44: ❓ can be omitted? so this not required?

bahaaTuffaha (commented):
Looks good ✨ 
Tested card creation/bulk at bayerns all generated an email with correct email address ✅

bahaaTuffaha (approved):

administration/src/project-configs/bayern/config.ts
f1sh1918: make sense good find

docs/card-query-params.md
f1sh1918: yes then today will be used

Docmale:
@f1sh1918 should card creation happens automatically or do I have to manually create the card?

f1sh1918:
In both cases the card creation has to be triggered manually. 
First case single card creation the values should be prefilled in the form (mail address is not visible) 
Second case normal bulk creation for card. Select csv and create bulk of cards

PR #2989, opened 2026-04-21
2988: Add workflow e2e tests

### LINKED ISSUE #2988: Add workflow for e2e tests
### Describe the Problem

Currently we are facing quite a lot production issues.
Therefore we want to create e2e-tests to automatically test recurring user flows.
This workflow should at least run successfully before creating a new beta release

- the workflow can be triggered manually
- the workflow will be automatically triggered when creating a release/hotfix branch
- the workflow runs e2e tests against local backend

f1sh1918:
### Short Description

Since we had quite some hotfix release, we want to add one more quality gate and add e2e tests in our pipeline that can access a local backend.

### Proposed Changes

add e2e administration job:
- spin up local backend, db migration, import dev test data
- install and run playwright
- fix e2e tests (due to mui migration) 

My idea would to schedule this workflow every monday morning. So even if we don't release very frequently we get reporting quite frequently

### Testing

- https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/11010/workflows/cf69a5b2-82c2-4816-bb7d-0e5729e855bf/jobs/64552

### Resolved Issues

Fixes: #2988


.circleci/src/jobs/e2e_administration.yml
```diff
@@ -0,0 +1,79 @@
...
+      keys:
+        - v1-gradle-{{ checksum "backend/build.gradle.kts" }}
+  - run:
+      name: Setup submodules
+      command: |
+        git submodule sync
+        git submodule update --init
+  - run:
+      name: Build backend
+      command: ./gradlew classes processResources && touch /tmp/backend-build.done || touch /tmp/backend-build.failed
```
f1sh1918, line 25: not sure if there is better way. But this is for the wait-step to check if backend build is ready. If we would just run them after each other it would spend more running time in total

andrew8er: Aren't tasks within a job always in sequence?

andrew8er (commented):
A few remarks/questions:
- The steps within a job are always in sequence, so why do we need to wait for the backend build?
- I feel that we have too many workflows, and understanding the difference among them becomes quiet tedious. Why don't we have one `commit` workflow and one `commit_main` (maybe we find a better name) that does the same with more intensive testing. The latter one can also be triggered by periodically or via any other event.

f1sh1918:
> A few remarks/questions:
> - The steps within a job are always in sequence, so why do we need to wait for the backend build?
> - I feel that we have too many workflows, and understanding the difference among them becomes quiet tedious. Why don't we have one `commit` workflow and one `commit_main` (maybe we find a better name) that does the same with more intensive testing. The latter one can also be triggered by periodically or via any other event.

Because the build backend command runs in background (won't block) . If we run all strict in sequence this job would take at least 2m more

Well if we include this job in the commit main workflow we trigger it very frequently and I expect this workflow to take at least 12-15m with a lot of future tests

But if we run in commit_main Android builds that take 12m we could think about that. 
Good point
We could also use the build backend artifact there to save time.
I will make a draft in a separate branch to check how much longer the commit_main would run

andrew8er:
- I would run e2e tests only on `commit_main`.
- If I'm not mistaken, the only thing that runs parallel is the installation of Playwright and the building of the backend. Is that really that much of a speedup? If not, I would suggest running everything sequentially, moving parallelism to jobs. For example, we could have a `frontend_build` and `backend_build` job running in parallel and then have a `frontend_e2e` job (that depends on both) starting and running both the backend and the frontend with [shared workspaces](https://circleci.com/docs/guides/orchestrate/workflows/#using-workspaces-to-share-data-between-jobs)

f1sh1918:
closed in favor of #2993

PR #3000, opened 2026-04-29
2999: Implement db connection pooling

### LINKED ISSUE #2999: Implement database connection pooling
### Describe the Problem

Our current implementation with `PGSimpleDataSource` opens a new physical TCP connection to PostgreSQL for every request and closes it afterwards. Under concurrent load — especially with virtual threads, which make high concurrency trivially cheap — this can quickly exhaust PostgreSQL's `max_connections` limit, causing the server to hang. `HikariCP` maintains a fixed pool of reusable connections, so requests beyond the pool size queue instead of opening new database connections. This bounds the load on PostgreSQL regardless of how many concurrent requests are in flight.

- implement connect pooling

f1sh1918:
### Short Description

Our current implementation with PGSimpleDataSource opens a new physical TCP connection to PostgreSQL for every request and closes it afterwards. Under concurrent load — especially with virtual threads, which make high concurrency trivially cheap — this can quickly exhaust PostgreSQL's max_connections limit, causing the server to hang. HikariCP maintains a fixed pool of reusable connections, so requests beyond the pool size queue instead of opening new database connections. This bounds the load on PostgreSQL regardless of how many concurrent requests are in flight.

### Proposed Changes

- add HicariCP plugin
- adjust datasource

### Side Effects

- should be none

### Testing

N/A

### Resolved Issues

Fixes: #2999

### DIFF
--- a/backend/gradle/libs.versions.toml
+++ b/backend/gradle/libs.versions.toml
@@ -80,3 +80,4 @@ springframework-boot-testcontainers = { module = "org.springframework.boot:sprin
 testcontainers = { module = "org.testcontainers:testcontainers", version.ref = "testcontainers" }
 testcontainers-junit-jupiter = { module = "org.testcontainers:junit-jupiter", version.ref = "testcontainers" }
 testcontainers-postgresql = { module = "org.testcontainers:postgresql", version.ref = "testcontainers" }
+zaxxer-hickaricp = { module = "com.zaxxer:HikariCP", version = "7.0.2" }
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
@@ -11,16 +11,18 @@ import app.ehrenamtskarte.backend.db.setup.insertOrUpdateProjects
 import app.ehrenamtskarte.backend.db.setup.insertOrUpdateRegions
 import app.ehrenamtskarte.backend.graphql.auth.types.Role
 import app.ehrenamtskarte.backend.graphql.freinet.util.FreinetAgenciesLoader
+import com.zaxxer.hikari.HikariConfig
+import com.zaxxer.hikari.HikariDataSource
 import org.jetbrains.exposed.v1.core.DatabaseConfig
 import org.jetbrains.exposed.v1.core.StdOutSqlLogger
 import org.jetbrains.exposed.v1.core.statements.StatementType
 import org.jetbrains.exposed.v1.jdbc.Database.Companion.connect
 import org.jetbrains.exposed.v1.jdbc.transactions.TransactionManager
 import org.jetbrains.exposed.v1.jdbc.transactions.transaction
-import org.postgresql.ds.PGSimpleDataSource
 import java.io.BufferedReader
 import java.io.File
 import java.io.InputStreamReader
+import java.util.Properties
 
 object Database {
     fun executeSqlResource(path: String) {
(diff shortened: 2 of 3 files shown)


backend/gradle/libs.versions.toml
```diff
@@ -46,6 +46,7 @@ favre-lib-bcrypt = { module = "at.favre.lib:bcrypt", version = "0.10.2" }
 google-protobuf-kotlin = { module = "com.google.protobuf:protobuf-kotlin", version.ref = "protobuf" }
 google-zxing-core = { module = "com.google.zxing:core", version = "3.5.4" }
 grundig-geojson-jackson = { module = "de.grundid.opendatalab:geojson-jackson", version = "1.14" }
+hikaricp = { module = "com.zaxxer:HikariCP" }
```
andrew8er: 🔧 We should pin all dependency versions (I also always namespace dependencies based on their origin/organisation):
```suggestion
zaxxer-hickaricp = { module = "com.zaxxer:HikariCP", version = "7.0.2" }
````

**Review by andrew8er — APPROVED** [review]
LGTM.

File: backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt [hunk 2]
```diff
@@ -71,17 +73,24 @@ object Database {
...
+                    // Lazily initialize the connection pool
+                    initializationFailTimeout = -1
+                    isAutoCommit = false
+                    maxLifetime = 1000 * 60 * 8 // 8 minutes
+                    // Set session time zone to UTC, to make timestamps work properly in every configuration.
+                    connectionInitSql = "SET TIME ZONE 'UTC'"
+                    dataSourceProperties = Properties().apply {
+                        setProperty("url", config.postgres.url)
+                        setProperty("reWriteBatchedInserts", "true")
+                    }
```
seluianova, line 91:
```suggestion
                HikariConfig().apply {
                    username = config.postgres.user
                    password = config.postgres.password
                    jdbcUrl = config.postgres.url
                    maximumPoolSize = 8
                    // Lazily initialize the connection pool
                    initializationFailTimeout = -1
                    isAutoCommit = false
                    maxLifetime = 1000 * 60 * 8 // 8 minutes
                    // Set session time zone to UTC, to make timestamps work properly in every configuration.
                    connectionInitSql = "SET TIME ZONE 'UTC'"
                    dataSourceProperties = Properties().apply {
                        setProperty("reWriteBatchedInserts", "true")
                    }
```
I believe if we set jdbcUrl we don't need dataSourceClassName

andrew8er: It is not recommended by the [documentation](https://github.com/brettwooldridge/hikaricp):

> This property directs HikariCP to use "DriverManager-based" configuration. We feel that DataSource-based configuration (above) is superior for a variety of reasons (see below), but for many deployments there is little significant difference. When using this property with "old" drivers, you may also need to set the driverClassName property, but try it first without. Note that if this property is used, you may still use DataSource properties to configure your driver and is in fact recommended over driver parameters specified in the URL itself.

backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
```diff
@@ -71,17 +73,24 @@ object Database {
...
+            datasource = HikariDataSource(
+                HikariConfig().apply {
+                    dataSourceClassName = "org.postgresql.ds.PGSimpleDataSource"
+                    username = config.postgres.user
+                    password = config.postgres.password
+                    maximumPoolSize = 8
+                    // Lazily initialize the connection pool
+                    initializationFailTimeout = -1
+                    isAutoCommit = false
+                    maxLifetime = 1000 * 60 * 8 // 8 minutes
```
seluianova: ❓🙃 the default maxLifetime is 30 minutes. are there any specific reasons why we need 8?
I just tend to think default values are usually optimal (in this case - from db load perspective) unless there’s a clear reason to override them.

backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
seluianova: in the current setup the `dataSourceClassName` is just excessive.
we set `dataSourceClassName = "org.postgresql.ds.PGSimpleDataSource"` AND `setProperty("url", config.postgres.url)` - it's a mix of DriverManager-based and DataSource-based configuration.
if we want to switch to the DataSource-based configuration, we need to set serverName, port, database instead of url.

seluianova: > for a variety of reasons (see below)

btw I couldn't find the mentioned reasons in the doc...

> We recommended using dataSourceClassName instead of jdbcUrl, but either is acceptable. We'll say that again, either is acceptable.

andrew8er: > btw I couldn't find the mentioned reasons in the doc...

It's explained directly below in "username" and "password". 

The initial motivation to change to a DataSource based configuration was the need to set driver properties, namely `reWriteBatchedInserts`.

backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
f1sh1918: postgres sql has a default idle timeout of 10minutes and the value should be shorter with a small buffer.
I can add a comment

backend/src/main/kotlin/app/ehrenamtskarte/backend/db/Database.kt
f1sh1918: hm if we switch to dataSource based we also have to adjust the configs in salt quite a bit.

I think the mixture is a pragmatic choice for our current config. If we don't do any driver updates i don't see things breaking tbh.

But i see that it might be confusing

seluianova: > I think the mixture is a pragmatic choice for our current config

jdbcUrl approach - works for us.
dataSourceClassName approach - kind of works, but it is not used in a consistent way; you still have to pass jdbcUrl via setProperty to make it work.
there is no need to mix them (setting jdbcUrl is enough). not sure why you mean it's pragmatic.

> i don't see things breaking tbh.

anyway that's true so.. I'm fine to keep this as is

seluianova (approved):
nice idea to improve this 👍

PR #3011, opened 2026-05-05
Release 2026.5.0

f1sh1918:
### Short Description

We had minor issues with the hotfix workflow

### Proposed Changes

- add a safe exit 0 when there is no hotfix

### Side Effects

- none

### Resolved Issues

Fixes: #

### DIFF
--- a/.circleci/src/jobs/bump_version.yml
+++ b/.circleci/src/jobs/bump_version.yml
@@ -22,7 +22,7 @@ steps:
         - run:
             name: Bump git version
             command: |
-              HOTFIX_FLAG=$([[ "${CIRCLE_BRANCH}" == hotfix* ]] && echo "--hotfix")
+              HOTFIX_FLAG=$([[ "${CIRCLE_BRANCH}" == hotfix* ]] && echo "--hotfix" || true)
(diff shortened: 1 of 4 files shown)


.circleci/src/jobs/bump_version.yml
```diff
@@ -22,7 +22,7 @@ steps:
         - run:
             name: Bump git version
             command: |
-              HOTFIX_FLAG=$([[ "${CIRCLE_BRANCH}" == hotfix* ]] && echo "--hotfix")
+              HOTFIX_FLAG=$([[ "${CIRCLE_BRANCH}" == hotfix* ]] && echo "--hotfix" || true)
```
seluianova, line 25: 💭 perhaps this would be more consistent / straightforward ?

```
if [[ "${CIRCLE_BRANCH}" == hotfix* ]]; then
  HOTFIX_FLAG="--hotfix"
else
  HOTFIX_FLAG=""
fi
```

`|| true` just forces exit code 0 when the condition fails?

seluianova (approved):

.circleci/src/jobs/bump_version.yml
f1sh1918: well the echo in practise cannot fail?! But if it would it would be swallowed
So i think the solution should be good enough IMO and we can save some lines of code

andrew8er: As a longer term option, `app-toolbelt` could actually figure this out by itself, by looking at the branch name. Is this something we want?

andrew8er (approved):

Issue #3013, opened 2026-05-06
Improve notification settings


f1sh1918:
### Describe the Problem

Currently we are updating the state in a useEffect which is bad practise.
Due to a eslint and typescript update in #3003 our pipeline fails

### Describe Your Preferred Solution

- avoid useEffect with state update

### Related Issues

#3003

PR #3014, opened 2026-05-06
3013: Improve notification settings

f1sh1918:
### Short Description

I refactored the `NotificationSettings` to avoid `useEffect` that updates the state to avoid linting error

### Proposed Changes

- introduce null in state to reset on error, remove useEffect hook

### Side Effects

- none

### Testing

1. Login as a region admin/region manager for bayern
2. Click on the avatar icon in the righ upper corner
3. Go to Benutzereinstellungen
4. Select and Deselect the checkbox in "Benachrichtigungen" and Save
5. Reload the page and ensure that the your selections are saved

### Resolved Issues

Fixes: #3013

### DIFF
--- a/administration/src/routes/user-settings/components/NotificationSettings.tsx
+++ b/administration/src/routes/user-settings/components/NotificationSettings.tsx
@@ -13,11 +13,13 @@ import {
 } from '../../../graphql'
 import getQueryResult from '../../../util/getQueryResult'
 
+type NotificationState = { notificationOnApplication: boolean; notificationOnVerification: boolean }
+
 const NotificationSettings = (): ReactElement => {
   const { t } = useTranslation('userSettings')
   const { enqueueSnackbar } = useSnackbar()
-  const [pendingActivation, setPendingActivation] = useState<boolean | null>(null)
-  const [pendingVerification, setPendingVerification] = useState<boolean | null>(null)
+  const [pendingSettings, setPendingSettings] = useState<NotificationState | null>(null)
+  const [lastSaved, setLastSaved] = useState<NotificationState | null>(null)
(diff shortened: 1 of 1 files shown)


andrew8er (commented):
The component would previously reset its state to the last loaded state if it could not save the new state (in the `reset()` function). 

I think this behavior is desirable.

f1sh1918:
> The component would previously reset its state to the last loaded state if it could not save the new state (in the `reset()` function).
> 
> I think this behavior is desirable.

But this behavior still persists @andrew8er 
Set the pendings to `null` on error will fallback to the `savedSettings`
Thats exactly what happened before i think

andrew8er:
I'll recheck...

andrew8er:
@f1sh1918 
> But this behavior still persists 

The behavior is not 100% regardless of this PR. On error, the state switches back to the one at "page load" time, not the last saved state:

- Start backend and frontend
- Login, go to user settings
- Change a notification setting and save
- Stop backend
- Change another notification setting and save 
- The settings will switch to the ones at "page load", not the last saved setting.

f1sh1918:
> @f1sh1918
> 
> > But this behavior still persists
> 
> The behavior is not 100% regardless of this PR. On error, the state switches back to the one at "page load" time, not the last saved state:
> 
> * Start backend and frontend
> * Login, go to user settings
> * Change a notification setting and save
> * Stop backend
> * Change another notification setting and save
> * The settings will switch to the ones at "page load", not the last saved setting.

okay thx for the explanation. now i got it.
updated the pr @andrew8er

andrew8er (approved):
Works, thanks!

PR #3023, opened 2026-05-18
3022 install precompiled ruby

### LINKED ISSUE #3022: Update Ruby version and use precompiled binaries for CI
### Describe the Problem

Our current build fails with a timeout since the configured Ruby version cannot be installed from precompiled binaries and will fall back to compiling from source.

https://app.circleci.com/pipelines/github/digitalfabrik/entitlementcard/11099/workflows/9d8a0ddb-1702-4e6e-a33d-7722a50929f3/jobs/64881

Use precompiled binaries.

### Related Issues

- #2996

andrew8er:
### Short Description

Manage development tools with [mise](https://mise.en.dev/), to be able to install precompiled Ruby versions for different platforms.

Since we need Ruby on both macos-arm64 and linux-x86_64, we can neither rely on Ubuntu packages nor on Docker base images for providing Ruby. Turns out that having precompiled Ruby versions for macos-arm64 are quiet a rarity.

Some options I considered:
- https://rubies.travis-ci.org/ (no ARM builds)
- https://rvm.io/binaries/ (almost no macos builds, no ARM builds)
- https://github.com/ruby/ruby-builder/releases/tag/toolcache
- https://github.com/jdx/ruby/releases

The last two options are viable, so I added the mise utility to automatically download and install Ruby, among other dependencies.

### Proposed Changes

- Add mise to the CI setup for CircleCI and Github Actions.
- Define and lock the tools used for every sub-project (`/administration`, `/backend`, `/frontend`).
- Update Playwright, since it would otherwise hang on installing Chromium.
- Use mise to download and run tools for CI.
- Created a [fine grained PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) for the Github user [`DigitalfabrikCI`](https://github.com/DigitalfabrikCI) and added it to the CircleCI [environment](https://app.circleci.com/settings/project/github/digitalfabrik/entitlementcard/environment-variables). This token has no privileges and no expiry date, its only purpose is to overcome rate limits imposed by Github. A similar token with the same environment variable name is created by Github automatically for every Github Action run.
- Removed `install_fastlane` since Cocoapods and Fastlane are installed via mise.
- Rename `install_flutter` to `install_flutter_packages`. It only installs dependency packages.
- mise replaces the following CI tasks/actions:
  - `ruby/install`
  - Installation of ninja (for Android builds)
  - Installation of npm
  - Installation of fvm. Flutter itself is now installed via mise. We only need to download dependency packages.
- For certain scripts (Gradle build files, Fastlane files), a wrapper is provided that either calls the mise tool or fvm. This can be removed if the project commits to using mise for developers also.

### Further considerations

Mise is currently only configured to provide tools for CI, but it could equally well support project setup and general development. If this is accepted, we could:
  - Remove all `Gemfile` and `Gemfile.lock`, since mise can install gems itself. We only need it to install Fastlane and Cocoapods.
  - Remove fvm, since mise can provide project specific flutter versions.
  - Have a stable directory linked into the project, where the Flutter SDK can be found. Currently, IntelliJ needs its SDK path reconfigured if the Flutter version changes.
  - Have exactly the same tools (JRE, Ruby, Node.js, etc…) in CI as on the dev setup.
  - Have a vastly simplified project setup (Install mise globally once, do `mise install`).

Fixes: #3022


.circleci/src/jobs/deliver_ios.yml
```diff
@@ -1,33 +1,31 @@
...
-  xcode: 26.2.0
-resource_class: m4pro.medium
 environment:
   FASTLANE_SKIP_UPDATE_CHECK: true
 working_directory: ~/project/frontend/ios
 steps:
   - prepare_workspace
+  - install_mise
   - install_app_toolbelt
-  - ruby/install:
```
f1sh1918, line 18: 🔧 if you remove the usage of `ruby/install` orb, please also remove the `orbs.yml` since it shouldn't be required

f1sh1918 (commented):
This pr looks very promising, thx for all the effort 👍 

Just one minor finding, rest looks good.

The general discussion as you already raised in the pr desciption is that it would make sense to also use mise for local dev.
I'm not 100% sure if everyone wants to use that, because some people already use nix and i'm not sure how well they work together. Additionally for new developers the initial setup could be more time consuming (actually i didn't use mise so far,so i dunno)

Keeping the local dev as it is might bring in some confusing and some overhead which i would not like not maintain, so i'm actually fine with using mise for local dev but this might be discussed in a bigger round

Actually using `ruby/setup-ruby` in GH  actions and a small script pully from jdx/ruby (as mise does) could also do the trick IMO

andrew8er:
I would also be interested in using Nix, haven't had any experience with if so far.

As there are no Nix-related config files in the repo, I was not aware of people using it. We should settle on one - not-optional - solution, and configure this to solve everyone's needs.

I chose mise because it solves a lot of pain points with a very simple config file and a pretty easily understandable mechanism without having the overhead of the devcontainers route, but if anyone wants to provide a well commented Nix setup, we can use this as well.

andrew8er:
This is something we could discuss on the next TaT conference!

f1sh1918:
> This is something we could discuss on the next TaT conference!

Yes definitely. I think mise is less complicated to setup and i would tend to choose this. actually i think there are just 1-2 people using nix and its way more complicated to setup. I really would not like to make using nix mandatory. I think mise solves our needs here.

f1sh1918 (approved):
Thx for this great improvement!
LGTM

bahaaTuffaha (commented):
- There no steps how to test this PR...
- I think we need to update the docs at `development-setup.md`.
- when I run `mise install` :

```bash
 mise install
mise config files in ~/Repos/entitlementcard are not trusted. Trust them? Yes
[WARN] migrate: error parsing config file: ~/Repos/entitlementcard/mise.toml
mise config files in ~/Repos/entitlementcard are not trusted. Trust them? Yes
mise ERROR error parsing config file: ~/Repos/entitlementcard/mise.toml
mise ERROR failed create_dir_all: ~/.local/state/mise/trusted-configs
mise ERROR Permission denied (os error 13)
mise ERROR Version: 2026.7.11 macos-arm64 (2026-07-20)
mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
```
I think it shouldn't need sudo right?

bahaaTuffaha:
> * I think we need to update the docs at `development-setup.md`.

oh I just realised that this is only for CI 😅

andrew8er:
>    There no steps how to test this PR...
>    I think we need to update the docs at development-setup.md.
>    when I run mise install :
>
>  oh I just realised that this is only for CI 😅

Yeah this is only for CI right now. I would however add another PR which converts this project fully to mise, if we are satisfied with it, along with updates to the docs.

Unfortunately, there is nothing to test here, other than letting the CI workflows run.

> mise install
> mise config files in ~/Repos/entitlementcard are not trusted. Trust them? Yes

You need to trust the config once with `mise trust mise.toml`.

> [WARN] migrate: error parsing config file: ~/Repos/entitlementcard/mise.toml
> mise config files in ~/Repos/entitlementcard are not trusted. Trust them? Yes
> mise ERROR error parsing config file: ~/Repos/entitlementcard/mise.toml
> mise ERROR failed create_dir_all: ~/.local/state/mise/trusted-configs
> mise ERROR Permission denied (os error 13)
> mise ERROR Version: 2026.7.11 macos-arm64 (2026-07-20)
> mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
>
> I think it shouldn't need sudo right?

No, `sudo` is not necessary. I suspect your file/dir permissions are somehow borked. What are your permissions like? Mine is:

```sh
❯ ls -l ~/.local/share/mise
total 0
drwxr-xr-x@ 21 andrewachter  staff   672 Jul 21 21:25 downloads/
drwxr-xr-x@ 22 andrewachter  staff   704 Jul 21 21:25 installs/
drwxr-xr-x@  3 andrewachter  staff    96 Jul 14 15:58 migrations/
drwxr-xr-x@ 84 andrewachter  staff  2688 Jul 21 21:25 shims/

❯ ls -l ~/.local/state/mise/
total 0
drwxr-xr-x@  4 andrewachter  staff  128 May 18 18:40 deps/
drwxr-xr-x@  6 andrewachter  staff  192 Jun 16 22:56 hints/
drwxr-xr-x@  3 andrewachter  staff   96 Jun  5 10:29 ignored-configs/
drwxr-xr-x@ 15 andrewachter  staff  480 Jul  6 22:50 tracked-configs/
drwxr-xr-x@ 13 andrewachter  staff  416 Jul  6 22:45 trusted-configs/
```

bahaaTuffaha:
> No, `sudo` is not necessary. I suspect your file/dir permissions are somehow borked. What are your permissions like? Mine is:

most of them root:
```
drwxr-xr-x@  6 root           staff  192 Jul 21 22:04 downloads
drwxr-xr-x@  7 root           staff  224 Jul 21 22:04 installs
drwxr-xr-x@  3 digitalfabrik  staff   96 Jul 21 22:03 migrations
drwxr-xr-x@ 12 root           staff  384 Jul 21 22:04 shims
```

bahaaTuffaha (approved):
Thanks for the huge effort. ✨

andrew8er:
> most of them root:

Ah that is the culprit. It should be owned by your user:

```sh
sudo chown -R $(id -u):$(id -g) ~/.local/state/mise ~/.local/share/mise
```

should fix it…

PR #3070, opened 2026-07-20
3069: Refactor `Matomo` class

### LINKED ISSUE #3069: Refactor `Matomo`
### Describe the Bug

The `Matomo.kt` object has some major design flaws:

- It uses a singleton `tracker: MatomoTracker` instance, but initializes it in a non thread-safe way.
- All methods take a configuration object, of which the only the first call will initialize the tracker with the config.
- The request are run in a `CoroutineScope(Dispatchers.IO)`, but then use their own async mechanism (via `…Async()` methods.

The functionality should be refactored:
- Have one class with one instance injected into every place where necessary.
- Initialize once with the appropriate config.
- Commit to one async style. I would favor using the builtin async methods and remove `CoroutineScope` usage. As far as I can tell, this is the only site we were using coroutines in this project, and we might remove the whole dependency after refactoring here.

### Environment

- Device/Browser: N/A
- OS version: N/A
- App version: N/A

### Related Issues
 
- #3062

andrew8er:
### Short Description

Refactors the `Matomo` class and a few adjacent code places.

### Proposed Changes

- Refactor the `Matomo` class to only have one level of method nesting.
- Only use `sendBulkRequestAsync()`, since it can handle single events as well.
- Only use the client built-in concurrency method, remove Kotlin Coroutine usage.
- Commit some IntelliJ files that keep getting changed automatically.
- Remove the `libs.jetbrains.kotlin.stdlib` library from the build file, since this gets added automatically by the Kotlin plugin.
- Remove transactions from the `Matomo` and `CardVerifier` class. These classes had buried transactions that queried for the same entity. They now take the entities as arguments, removing redundant DB queries. I suspect we have a lot more of deeply buried transactions that we should get rid of. Transactions should only occur at the very top level of our code, i.e. the controller code. Sprinkling those willy nilly into utility code only creates hard to follow DB interaction patterns and badly performing code.
- Remove the `@Synchronized` annotation from `generateTotpSecret()`. The function uses one generator object per call and should therefore be fully thread-safe already. Please object, if I'm wrong here.

### Side Effects

None

### Testing

Backend tests should trigger some tracking calls.

### Resolved Issues

Fixes: #3069

### DIFF
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
@@ -158,8 +158,9 @@ class CardMutationController(
         @GraphQLIgnore @ContextValue request: HttpServletRequest,
         dfe: DataFetchingEnvironment,
     ): CardCreationResultModel {
-        val config = backendConfiguration.getProjectConfig(project)
-        if (!config.selfServiceEnabled) {
+        val projectConfig = backendConfiguration.getProjectConfig(project)
+
+        if (!projectConfig.selfServiceEnabled) {
             throw NotImplementedException("Self-Service is not enabled in the project")
         }
 
@@ -216,8 +217,7 @@ class CardMutationController(
         }
 
         matomo.trackCreateCards(
-            backendConfiguration,
-            config,
+            projectConfig = projectConfig,
             request,
             dfe.field.name,
             userEntitlements.regionId.value,
@@ -303,8 +303,7 @@ class CardMutationController(
 
         if (regionId != null) {
             matomo.trackCreateCards(
-                backendConfiguration,
-                projectConfig,
+                projectConfig = projectConfig,
                 request,
                 dfe.field.name,
                 regionId,
@@ -384,8 +383,7 @@ class CardMutationController(
             return@t CardActivationResultModel(ActivationState.success, encodedTotpSecret)
         }
         matomo.trackActivation(
-            backendConfiguration,
-            projectConfig,
+            projectConfig = projectConfig,
             request,
             dfe.field.name,
             cardHash,
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
@@ -394,7 +394,7 @@ class CardMutationController(
             request = request,
             query = dfe.field.name,
             cardEntity = cardEntity,
-            successful = activationResult.activationState == ActivationState.success,
+            cardActivationState = activationResult.activationState,
         )
 
         return activationResult
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
@@ -62,6 +62,7 @@ import kotlin.io.encoding.Base64
 class CardMutationController(
     private val backendConfiguration: BackendConfiguration,
     private val mailerService: Mailer,
+    private val matomo: Matomo,
 ) {
     private val logger: Logger = LoggerFactory.getLogger(CardMutationController::class.java)
 
@@ -214,7 +215,7 @@ class CardMutationController(
             )
         }
 
-        Matomo.trackCreateCards(
+        matomo.trackCreateCards(
             backendConfiguration,
             config,
             request,
@@ -301,7 +302,7 @@ class CardMutationController(
         val regionId = authContext.admin.regionId?.value
 
         if (regionId != null) {
-            Matomo.trackCreateCards(
+            matomo.trackCreateCards(
                 backendConfiguration,
                 projectConfig,
                 request,
@@ -382,7 +383,7 @@ class CardMutationController(
             )
             return@t CardActivationResultModel(ActivationState.success, encodedTotpSecret)
         }
-        Matomo.trackActivation(
+        matomo.trackActivation(
             backendConfiguration,
             projectConfig,
             request,
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
@@ -467,19 +467,16 @@ private const val cost = 10
 fun hashActivationSecret(rawActivationSecret: ByteArray): ByteArray =
     BCrypt.withDefaults().hash(cost, rawActivationSecret)
 
-@Synchronized
-private fun generateTotpSecret(): ByteArray {
-    // https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HTOP)
-    // https://tools.ietf.org/html/rfc4226#section-4 - R6 (How long should a shared secret be?
-    // -> 160bit)
-    // https://tools.ietf.org/html/rfc4226#section-7.5 - Random Generation (How to generate a
-    // secret? -> Random))))
-    val algorithm = TimeBasedOneTimePasswordGenerator.TOTP_ALGORITHM_HMAC_SHA256
-    val keyGenerator = KeyGenerator.getInstance(algorithm)
-    keyGenerator.init(TOTP_SECRET_LENGTH * 8)
-
-    return keyGenerator.generateKey().encoded
-}
+/**
+ * https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HOTP)
+ * https://tools.ietf.org/html/rfc4226#section-4 - R6 (How long should a shared secret be? -> 160bit)
+ * https://tools.ietf.org/html/rfc4226#section-7.5 - Random Generation
+ */
+private fun generateTotpSecret(): ByteArray =
+    KeyGenerator.getInstance(TimeBasedOneTimePasswordGenerator.TOTP_ALGORITHM_HMAC_SHA256).run {
+        init(TOTP_SECRET_LENGTH * 8)
+        generateKey().encoded
+    }
 
 private fun verifyActivationSecret(rawActivationSecret: ByteArray, activationSecretHash: ByteArray): Boolean =
     BCrypt.verifyer().verify(rawActivationSecret, activationSecretHash).verified
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
@@ -331,64 +331,72 @@ class CardMutationController(
         val rawActivationSecret = Base64.decode(activationSecretBase64)
 
         // Avoid race conditions when activating a card.
-        val activationResult = transaction(transactionIsolation = TRANSACTION_REPEATABLE_READ) t@{
+        val (cardEntity, activationResult) = transaction(transactionIsolation = TRANSACTION_REPEATABLE_READ) {
             this.maxAttempts = 1
 
-            val card = CardRepository.findByHash(project, cardHash)
-            val activationSecretHash = card?.activationSecretHash
+            val cardEntity = CardRepository.findByHash(project, cardHash)
+            val activationSecretHash = cardEntity?.activationSecretHash
 
-            if (card == null || activationSecretHash == null) {
-                logger.info(
-                    "$remoteIp failed to activate card, card not found with cardHash:$cardInfoHashBase64",
-                )
-                return@t CardActivationResultModel(ActivationState.not_found)
+            when {
+                cardEntity == null || activationSecretHash == null -> {
+                    logger.info(
+                        "$remoteIp failed to activate card, card not found with cardHash:$cardInfoHashBase64",
+                    )
+                    Pair(cardEntity, CardActivationResultModel(ActivationState.not_found))
+                }
+
+                !verifyActivationSecret(rawActivationSecret, activationSecretHash) -> {
+                    logger.info(
+                        "$remoteIp failed to activate card with id:${cardEntity.id} and overwrite: $overwrite",
+                    )
+                    Pair(cardEntity, CardActivationResultModel(ActivationState.wrong_secret))
+                }
+
+                CardVerifier.isExpired(cardEntity.expirationDay, projectConfig.timezone) -> {
+                    logger.info(
+                        "$remoteIp failed to activate card with id:${cardEntity.id} and overwrite: " +
+                            "$overwrite because card is expired",
+                    )
+                    Pair(cardEntity, CardActivationResultModel(ActivationState.expired))
+                }
+
+                cardEntity.revoked -> {
+                    logger.info(
+                        "$remoteIp failed to activate card with id:${cardEntity.id} and overwrite: " +
+                            "$overwrite because card is revoked",
+                    )
+                    Pair(cardEntity, CardActivationResultModel(ActivationState.revoked))
+                }
+
+                !overwrite && cardEntity.totpSecret != null -> {
+                    logger.info(
+                        "Card with id:${cardEntity.id} did not overwrite card from $remoteIp",
+                    )
+                    Pair(cardEntity, CardActivationResultModel(ActivationState.did_not_overwrite_existing))
+                }
+
+                else -> {
+                    val totpSecret = generateTotpSecret()
+                    val encodedTotpSecret = Base64.encode(totpSecret)
+
+                    CardRepository.activate(cardEntity, totpSecret)
+
+                    logger.info(
+                        "Card with id:${cardEntity.id} and overwrite: $overwrite was activated from $remoteIp",
+                    )
+                    Pair(cardEntity, CardActivationResultModel(ActivationState.success, encodedTotpSecret))
+                }
             }
-
-            if (!verifyActivationSecret(rawActivationSecret, activationSecretHash)) {
-                logger.info(
-                    "$remoteIp failed to activate card with id:${card.id} and overwrite: $overwrite",
-                )
-                return@t CardActivationResultModel(ActivationState.wrong_secret)
-            }
-
-            if (CardVerifier.isExpired(card.expirationDay, projectConfig.timezone)) {
-                logger.info(
-                    "$remoteIp failed to activate card with id:${card.id} and overwrite: " +
-                        "$overwrite because card is expired",
-                )
-                return@t CardActivationResultModel(ActivationState.expired)
-            }
-
-            if (card.revoked) {
-                logger.info(
-                    "$remoteIp failed to activate card with id:${card.id} and overwrite: " +
-                        "$overwrite because card is revoked",
-                )
-                return@t CardActivationResultModel(ActivationState.revoked)
-            }
-
-            if (!overwrite && card.totpSecret != null) {
-                logger.info(
-                    "Card with id:${card.id} did not overwrite card from $remoteIp",
-                )
-                return@t CardActivationResultModel(ActivationState.did_not_overwrite_existing)
-            }
-
-            val totpSecret = generateTotpSecret()
-            val encodedTotpSecret = Base64.encode(totpSecret)
-            CardRepository.activate(card, totpSecret)
-            logger.info(
-                "Card with id:${card.id} and overwrite: $overwrite was activated from $remoteIp",
-            )
-            return@t CardActivationResultModel(ActivationState.success, encodedTotpSecret)
         }
+
         matomo.trackActivation(
             projectConfig = projectConfig,
-            request,
-            dfe.field.name,
-            cardHash,
-            activationResult.activationState == ActivationState.success,
+            request = request,
+            query = dfe.field.name,
+            cardEntity = cardEntity,
+            successful = activationResult.activationState == ActivationState.success,
         )
+
         return activationResult
     }
 
--- a/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/utils/CardVerifier.kt
+++ b/backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/utils/CardVerifier.kt
@@ -1,5 +1,6 @@
 package app.ehrenamtskarte.backend.graphql.cards.utils
 
+import app.ehrenamtskarte.backend.db.entities.CardEntity
 import app.ehrenamtskarte.backend.db.entities.UserEntitlementsEntity
 import app.ehrenamtskarte.backend.db.repositories.CardRepository
 import com.eatthepath.otp.TimeBasedOneTimePasswordGenerator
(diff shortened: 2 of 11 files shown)


backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
```diff
@@ -34,227 +30,155 @@ enum class MatomoEventName(val value: String) {
...
-            config,
-            projectConfig.matomo,
-            MatomoRequest.request()
-                .actionName(query)
-                .searchCategory(params.categoryIds?.joinToString(","))
-                .searchQuery(params.searchText ?: "")
-                .searchResultsCount(numResults.toLong())
-                .also { attachRequestInformation(it, request) },
-        )
+        if (params.searchText != null && params.categoryIds != null) {
```
f1sh1918: ❌ this inversion of the if clause is not the same!
Now we do not track anything if `searchText` exists but `categoryId` not.
In the previous early return we only return if both are null but if f.e. searchText is not null but categoryId we do track

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/utils/CardVerifier.kt
```diff
@@ -16,20 +17,16 @@ val TIME_STEP: Duration = Duration.ofSeconds(30)
...
-    }
-
-    fun verifyDynamicCard(project: String, cardHash: ByteArray, totp: Int, timezone: ZoneId): Boolean {
-        val card = transaction { CardRepository.findByHash(project, cardHash) } ?: return false
-        return !isExpired(card.expirationDay, timezone) &&
-            isYetValid(card.startDay, timezone) &&
-            !card.revoked &&
-            isTotpValid(totp, card.totpSecret)
-    }
+    fun verifyStaticCard(project: String, cardEntity: CardEntity, timezone: ZoneId): Boolean =
```
f1sh1918: 🙃 `project` parameter is now unused. please remove

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/utils/CardVerifier.kt
```diff
@@ -16,20 +17,16 @@ val TIME_STEP: Duration = Duration.ofSeconds(30)
...
+        !isExpired(cardEntity.expirationDay, timezone) &&
+            isYetValid(cardEntity.startDay, timezone) &&
+            !cardEntity.revoked
+
+    fun verifyDynamicCard(project: String, cardEntity: CardEntity, totp: Int, timezone: ZoneId): Boolean =
```
f1sh1918: 🙃 `project` parameter is now unused. please remove

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
```diff
@@ -460,19 +467,16 @@ private const val cost = 10
 fun hashActivationSecret(rawActivationSecret: ByteArray): ByteArray =
     BCrypt.withDefaults().hash(cost, rawActivationSecret)
 
-@Synchronized
```
f1sh1918, line 463: 💭 i'm even not pretty sure if we need the `@Synchronized` but i also see no harm in keeping it, since this totp thing is quite security-sensitive tbh.

f1sh1918 (changes requested):
Very good refactoring
But i realized that the tracking behavior "silently" changed

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
andrew8er: I missed to turn the `&&` into `||`, fixed now.

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
andrew8er: The [docs](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/javax/crypto/KeyGenerator.html#getInstance(java.lang.String)) say, that a new generator object is returned. We use it once and then it will be GCed. 

It was moved from `/backend/src/main/kotlin/app/ehrenamtskarte/backend/cards/service/CardActivator.kt` in 4906cc97521d75534216eda829e5df0e69ad642d and was ultimately committed in 968d9b358f94dd14d2db54e32aab2363d935f402.

#### Analysis by Claude:

❯ @src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationControl 
  ler.kt Eximanine the usage of the @Synchronized annotation in line 475. Is it necessary?                                                                    
  ⎿  Read src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt (484 lines)

⏺ No, it's not necessary. Here's the analysis.

  What @Synchronized actually locks here

  `generateTotpSecret()` is a top-level (file-level) function, so Kotlin compiles it to a static method on the file-facade class `CardMutationControllerKt`.
  `@Synchronized` emits the `ACC_SYNCHRONIZED` flag, and for a static method that acquires the monitor on the Class object (`CardMutationControllerKt::class`). The effect is that every concurrent call is serialized globally.

  Is there anything to protect?

  No. Look at what the body touches:

```kt
  private fun generateTotpSecret(): ByteArray =
      KeyGenerator.getInstance(TimeBasedOneTimePasswordGenerator.TOTP_ALGORITHM_
  HMAC_SHA256).run {
          init(TOTP_SECRET_LENGTH * 8)
          generateKey().encoded
      }
```

  - `KeyGenerator.getInstance(...)` returns a fresh instance on every call — it's a local, never stored in a field, never shared across threads.
  - `TOTP_ALGORITHM_HMAC_SHA256` and `TOTP_SECRET_LENGTH` are compile-time constants (read-only).
  - The `init()` / `generateKey()` mutations happen only on that thread-local `KeyGenerator`.
  - The JCA `getInstance` factory itself is thread-safe, and the underlying `SecureRandom` each generator uses is not shared between invocations.

  There is no shared mutable state, so there's no data race for the annotation to prevent.

  The downside of keeping it

  It's not just a no-op — it's a needless bottleneck. `generateTotpSecret()` is called inside the activateCard transaction (line 379). With `@Synchronized`, all concurrent card activations across the whole JVM would serialize on the class monitor while generating keys, for zero correctness benefit.

  Git context worth flagging

  The committed history already settled this — commit 4c6d36845 ("Improve generateTotpSecret() and remove @Synchronized") deliberately removed it. Your uncommitted working-tree change re-adds it:

```
  +@Synchronized
   private fun generateTotpSecret(): ByteArray =
```

  So this isn't a new decision — it re-introduces something the branch already dropped on purpose. I'd revert that working-tree hunk (`git restore -p / git checkout` -- on that file) to keep the annotation removed.

  Verdict: unnecessary, and slightly harmful. Drop it.

f1sh1918 (approved):
Nice refactoring 👍

backend/src/main/kotlin/app/ehrenamtskarte/backend/DI.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import app.ehrenamtskarte.backend.config.BackendConfiguration
+import app.ehrenamtskarte.backend.shared.Matomo
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+class DI {
+    @Bean
+    fun matomo(backendConfig: BackendConfiguration) = Matomo(backendConfig)
```
seluianova: you can add the `@Service` annotation to the Matomo class to have Spring create the bean automatically.

`@Configuration` is typically needed when we want create multiple beans of the same type with different configurations, or register a class that we cannot annotate (for example, a class from an external library), or when bean creation requires some custom logic.
in simple cases like this you can just add `@Service` or `@Component` to the class itself.

backend/src/main/kotlin/app/ehrenamtskarte/backend/graphql/cards/CardMutationController.kt
```diff
@@ -460,19 +467,16 @@ private const val cost = 10
...
-    // https://tools.ietf.org/html/rfc4226#section-7.5 - Random Generation (How to generate a
-    // secret? -> Random))))
-    val algorithm = TimeBasedOneTimePasswordGenerator.TOTP_ALGORITHM_HMAC_SHA256
-    val keyGenerator = KeyGenerator.getInstance(algorithm)
-    keyGenerator.init(TOTP_SECRET_LENGTH * 8)
-
-    return keyGenerator.generateKey().encoded
-}
+/**
+ * https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HTOP)
```
seluianova:
```suggestion
 * https://tools.ietf.org/html/rfc6238#section-3 - R3 (TOTP uses HOTP)
```

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
```diff
@@ -34,227 +30,155 @@ enum class MatomoEventName(val value: String) {
...
+                        it.authToken(matomoConfig.accessToken)
+                        it.headerAcceptLanguage(AcceptLanguage.fromHeader(httpRequest.getHeader("Accept-Language")))
+                        it.headerUserAgent(httpRequest.getHeader("User-Agent"))
+                        it.visitorIp(httpRequest.remoteAddr)
+                        it.build()
+                    },
+                ).exceptionally {
+                    when (it) {
+                        is IOException -> {
+                            logger.debug("Could not send request to Matomo")
```
seluianova: is it an intentional change? it was `logger.error` before

backend/src/main/kotlin/app/ehrenamtskarte/backend/DI.kt
andrew8er: `DI.kt` was removed and `@Service` added on class `Matomo`.

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
andrew8er: `sendBulkTrackingRequest()` used `debug()`, while `sendTrackingRequest()` used `error()`. Since I reused the former, I accidentally kept `debug()`. Fixed now.

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
```diff
@@ -34,227 +30,155 @@ enum class MatomoEventName(val value: String) {
...
-        )
+        if (params.searchText != null || params.categoryIds != null) {
+            sendTrackingRequests(
+                projectConfig = projectConfig,
+                httpRequest = request,
+                requestBuilders = listOf(
+                    MatomoRequest.request()
+                        .actionName(query)
+                        .searchCategory(params.categoryIds?.joinToString(","))
+                        .searchQuery(params.searchText)
```
seluianova:
```suggestion
                        .searchQuery(params.searchText ?: "")
```
otherwise MatomoException occurs when searchText is null

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
```diff
@@ -34,227 +31,156 @@ enum class MatomoEventName(val value: String) {
...
+                tracker.sendBulkRequestAsync(
+                    requestBuilders.map {
+                        it.siteId(matomoConfig.siteId)
+                        it.authToken(matomoConfig.accessToken)
+                        it.headerAcceptLanguage(AcceptLanguage.fromHeader(httpRequest.getHeader("Accept-Language")))
+                        it.headerUserAgent(httpRequest.getHeader("User-Agent"))
+                        it.visitorIp(httpRequest.remoteAddr)
+                        it.build()
+                    },
+                ).exceptionally {
```
seluianova: `exceptionally` block only catches exceptions thrown by CompletableFuture.
but synchronous exceptions can still be thrown before the future is returned.

e.g. try to set an invalid api token and perform an action that sends a matomo request. 
this results in an unhandled IllegalArgumentException that propagates to the client:

```
2026-07-22T11:17:19.008Z ERROR 7840 --- [backend] [         task-1] a.e.b.graphql.GraphQLExceptionHandler    : Unhandled exception in GraphQL controller, field='createCardsByCardInfos'

java.lang.IllegalArgumentException: Auth token must be exactly 32 characters long
	at org.matomo.java.tracking.RequestValidator.validate(RequestValidator.java:47) ~[matomo-java-tracker-core-3.4.0.jar:na]
	at org.matomo.java.tracking.Java8Sender.sendBulkAsync(Java8Sender.java:334) ~[matomo-java-tracker-3.4.0.jar:na]
	at org.matomo.java.tracking.MatomoTracker.sendBulkRequestAsync(MatomoTracker.java:327) ~[matomo-java-tracker-core-3.4.0.jar:na]
	at org.matomo.java.tracking.MatomoTracker.sendBulkRequestAsync(MatomoTracker.java:299) ~[matomo-java-tracker-core-3.4.0.jar:na]
	at app.ehrenamtskarte.backend.shared.Matomo.sendTrackingRequests(Matomo.kt:162) ~[main/:na]
	at app.ehrenamtskarte.backend.shared.Matomo.trackCreateCards(Matomo.kt:51) ~[main/:na]
	at app.ehrenamtskarte.backend.graphql.cards.CardMutationController.createCardsByCardInfos(CardMutationController.kt:305) ~[main/:na]
```

I think we need both a try/catch block AND an exceptionally handler here to cover both synchronous and asynchronous failures.

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
```diff
@@ -34,227 +31,156 @@ enum class MatomoEventName(val value: String) {
...
+                    when (it) {
+                        is IOException -> {
+                            logger.error("Could not send request to Matomo")
+                        }
+
+                        is ExecutionException, is InterruptedException -> {
+                            logger.error("Error while getting response")
+                        }
+                    }
+                    null
```
seluianova:
```suggestion
                ).exceptionally { ex ->
                    val cause = ex.cause ?: ex
                    when (cause) {
                        is MatomoException -> {
                            logger.error("Could not send request to Matomo", cause)
                        }
                        else -> {
                            logger.error("Unexpected error while sending request to Matomo", cause)
                        }
                    }
                    null
```

async exceptions are currently not logged properly and are silently ignored.

quick test: set an invalid Matomo URL and trigger a request. the request will fail, but no error will be logged.

**UPD**. the reason is that the exception in the `exceptionally` block needs to be unwrapped. In this case, the exception is a `MatomoException` (not sure if it ever expose an `IOException` here).

also there was no fallback handling for unexpected exception types, so some failures can be ignored silently (I added an else-block in my suggestion to cover that).

backend/src/main/kotlin/app/ehrenamtskarte/backend/DI.kt
seluianova: perfect, thanks

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
andrew8er: Added a `try catch` block and improved error logging. The exceptions type and message is now logged as well. I did not add the exception as an extra argument, since that would print the whole stack trace, which I think is not necessary here.

andrew8er:
@seluianova Thanks for your review so far. Did you by chance use the "comment" function instead of the "review" function? Because I get notifications for every comment of yours, which is draining my attention capacity quiet a bit. Also I never know, when you are actually finished this way.

seluianova:
> Did you by chance use the "comment" function instead of the "review" function? Because I get notifications for every comment of yours, which is draining my attention capacity quiet a bit.

yes, I did. I never thought this could cause any inconvenience, to be honest 😀

I personally don't check every incoming notification unless I am ready to see what was written or switch back to working on the PR.

at the same time, when a review takes a long time (in my case, I often check everything thoroughly, so it can take a while), it is visible that the review is in progress, and some comments can already be fixed in parallel.

but sure, I can try to do it differently if that works better for you.

andrew8er:
No problem, I usually have a notification utility running, that alerts me of github notifications, so I can react quickly if necessary. I would prefer review comments in a "batch" with a short overall comment and verdict (comment, approve, request changes), even if that takes a little bit longer to receive. I can do other stuff in between.

In backend terms, I would think of this as optimizing for throughput instead of latency 🤓

seluianova (commented):
So, now I am done :)
The improvements look good.

There is one unresolved comment: async exception handling is currently not working.

However, there is one caveat. When I applied the fix I suggested, I noticed the following issue: when we send a `trackSearch` request, Matomo correctly processes the request, but returns HTTP 500.

```
2026-07-22T13:05:36.022Z ERROR 29808 --- [backend] [moJavaTracker-1] a.ehrenamtskarte.backend.shared.Matomo   : Could not send request to Matomo

org.matomo.java.tracking.MatomoException: Tracking endpoint responded with code 500
	at org.matomo.java.tracking.Java8Sender.checkResponse(Java8Sender.java:197) ~[matomo-java-tracker-4.0.0.jar:na]
	at org.matomo.java.tracking.Java8Sender.sendBulk(Java8Sender.java:302) ~[matomo-java-tracker-4.0.0.jar:na]
	at org.matomo.java.tracking.Java8Sender.sendBulkAsync(Java8Sender.java:358) ~[matomo-java-tracker-4.0.0.jar:na]
	at org.matomo.java.tracking.Java8Sender.lambda$sendBulkAsync$1(Java8Sender.java:346) ~[matomo-java-tracker-4.0.0.jar:na]
```

I checked the container logs and found the following cause:

`[Wed Jul 22 12:05:20.334783 2026] [php:notice] [pid 68:tid 68] [client 172.20.0.1:45608] - Error in Matomo (tracker): Error query: Error query: SQLSTATE[HY000]: General error: 1020 Record has changed since last read in table 'matomo_user_token_auth' In query: UPDATE matomo_user_token_auth SET last_used = ? WHERE idusertokenauth = ? Parameters: array ( 0 => '2026-07-22 12:05:20', 1 => 3, ) In query: UPDATE matomo_user_token_auth SET last_used = ? WHERE idusertokenauth = ? Parameters: array ( 0 => '2026-07-22 12:05:20', 1 => 3, )`

I am not sure how to fix this yet, or whether this is only an issue with my local setup.
I think this should either be investigated further or moved to a separate issue.

What do you think? Can you reproduce that?

**UPD:** Ah, wait, I found better reproduction steps.

If you type just one character in the search field, it sends a single request and there is no error.

However, if you quickly type multiple letters, for example "abc", it sends three requests: "a", "ab", and "abc", and the error occurs.

So it seems to be related to a race condition or something similar.

**UPD-2:** I was wrong here:

> Matomo correctly processes the request, but returns HTTP 500

in the case above Matomo correctly processes the requests "a" and "ab", but "abc" really fails and does not get tracked. so it's a real regression issue.

**UPD-3:** I think this is related to the change from `sendRequestAsync` to `sendBulkRequestAsync` for all requests.
during search, we can now end up sending multiple bulk requests at the same time, which seems to trigger the error.
I guess this should be reverted:

> Only use sendBulkRequestAsync(), since it can handle single events as well.

andrew8er:
@seluianova 
> UPD-3: I think this is related to the change from sendRequestAsync to sendBulkRequestAsync for all requests.
during search, we can now end up sending multiple bulk requests at the same time, which seems to trigger the error.
I guess this should be reverted:

Interesting. I will investigate this. As far as I understand the [docs](https://javadoc.io/doc/org.piwik.java.tracking/matomo-java-tracker-core/latest/org/matomo/java/tracking/MatomoTracker.html), [sendBulkRequestAsync](https://javadoc.io/static/org.piwik.java.tracking/matomo-java-tracker-core/4.0.0/org/matomo/java/tracking/MatomoTracker.html#sendBulkRequestAsync(java.util.Collection)) should be usable just as [sendRequestAsync](https://javadoc.io/static/org.piwik.java.tracking/matomo-java-tracker-core/4.0.0/org/matomo/java/tracking/MatomoTracker.html#sendRequestAsync(org.matomo.java.tracking.MatomoRequest)), and maybe even preferable, since it uses POST instead of GET, which is semantically correct.

andrew8er:
My preliminary understanding is that this is a transaction rollback error on the Matomo side. For me this happens when I type characters relatively fast. When entering single characters slowly, it works.

See also https://mariadb.com/docs/server/reference/error-codes/mariadb-error-codes-1000-to-1099/e1020

Query: 
```sql
UPDATE matomo_log_visit 
SET 
  idvisitor = ?, 
  profilable = ?, 
  visit_last_action_time = ?, 
  visitor_seconds_since_order = ?, 
  visit_exit_idaction_name = ?, 
  visit_exit_idaction_url = ?,  
  visit_total_actions = visit_total_actions + 1 ,  
  visit_total_interactions = visit_total_interactions + 1 ,  
  visit_total_searches = visit_total_searches + 1 , 
  config_resolution = ?, 
  visit_total_time = ? 
WHERE 
  idsite = ? AND 
  idvisit = ? 
```

Could it be that this is run in a transaction (probably with some queries preceeding it, to get the initial data) and failing, because other track events arrive?

I cannot find anything in the Matomo forum or the Github project, but I'm pretty sure that this is actually a Matomo problem. The question is, do we need to deal with this now?

https://matomo.org/faq/on-premise/how-to-configure-matomo-for-speed/

seluianova:
> The question is, do we need to deal with this now?

afaics the potential fix is to revert the change that uses sendBulkRequestAsync for all requests.
when I switched back to sendRequestAsync instead of sendBulkRequestAsync, I could no longer reproduce the issue.
(would be nice to double check though)

I mean, the change looks nicer code-wise, but if it introduces a regression, even a rare one (only for fast-typing users), I'm not sure the trade-off is worth it 🤷🏻‍♀️

_UPD: I also couldn't find in the docs why sendBulkRequestAsync should be preferred for single events_

_If we want to add some optimization here_, I think a better approach would be to collect events internally into batches and then send them via sendBulkRequestAsync.

andrew8er:
Using `sendRequestAsync()` instead of `sendBulkRequestAsync()` reports the following error:

```
[Thu Jul 23 12:31:55.441854 2026] [php:notice] [pid 48:tid 48] [client 192.168.65.1:40883] - Error in Matomo (tracker): Tracker API 'cip' was used, requires valid token_auth

192.168.65.1 - - [23/Jul/2026:12:31:55 +0000] "GET /matomo.php?token_auth=9cb10e07c67a327be1513c775fa73099&rec=1&idsite=1&action_name=searchAcceptingStoresInProjectV2&apiv=1&_id=609eb141eda9f651&ua=de.nrw.it.giz.ehrensache.bayern.android.development%2F1.0+%281%29+Android%2F37+google+sdk_gphone16k_arm64&search=dsdd&search_count=0&token_auth=9cb10e07c67a327be1513c775fa73099&cip=127.0.0.1&send_image=0&rand=11d7726441013fdcac34 HTTP/1.1" 400 639 "-" "Java/25.0.3"
```

This also happens on `main`, so this is independent from this PR. Note that the `token_auth` appears twice in the URL, but is correct. Tbh, I would leave it as is for now, and we need to check on staging, if we have correct traces.

I have refactored the class a little more: the `MatomoRequest` objects are now created in one function instead of spread around several, so it is easier to see, what gets set.

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
```diff
@@ -33,223 +31,206 @@ enum class MatomoEventName(val value: String) {
...
+                    ),
+                )
+            }
+        }
+    }
+
+    private fun sendTrackingRequest(matomoRequest: MatomoRequest) {
+        try {
+            tracker.sendRequestAsync(matomoRequest).exceptionally(::logMatomoErrors)
+        } catch (e: Throwable) {
```
seluianova, line 167: 🙃 I tend to catch `Exception` because `Throwable` also includes errors that aren't supposed to be caught (StackOverflowError, OutOfMemoryError, etc.), not critical though

backend/src/main/kotlin/app/ehrenamtskarte/backend/shared/Matomo.kt
```diff
@@ -33,223 +31,206 @@ enum class MatomoEventName(val value: String) {
...
+            try {
+                tracker.sendBulkRequestAsync(matomoRequests).exceptionally(::logMatomoErrors)
+            } catch (e: Throwable) {
+                logger.error("Error while sending tracking request: $e")
+            }
+        }
+    }
+
+    private fun logMatomoErrors(exception: Throwable): Nothing? {
+        when (exception) {
```
seluianova:
```suggestion
        val cause = exception.cause ?: exception
        when (cause) {
```
🙃 I believe this callback receives a `CompletionException` from `CompletableFuture`, not the original `MatomoException`.
To handle `MatomoException` specifically, it should be unwrapped first; otherwise, it will likely always end up in the `else` branch.

seluianova (approved):
Re-tested, looks good to me!
Just minor optional suggestions.

**UPD**. Btw, why? Matomo is not configured in the test config, so there should not be any calls. Or what do you mean?

> Backend tests should trigger some tracking calls.

andrew8er:
@seluianova Merged with your latest suggestions.

PR #3095, opened 2026-08-10
3091:  Same application created multiple times

### LINKED ISSUE #3091: Same application is created multiple times
### Describe the Bug

In some cases, an application appears multiple times.
The identical application was created 3 times. The only differences are:
- `id`: subsequent IDs
- `createdDate`: dates differ by milliseconds (`.278078Z`, `.278113Z`, `.322777Z`)
- `verificationId`: subsequent IDs but not in ascending order (`..179`, `..178`, `..180`)

I can verify that this is not a single occurrence but seems to happen from time to time. Sometimes there are 2 identical applications and sometimes there are 3.

### Steps to Reproduce

I didn't manage to reproduce it yet.

### Expected Behavior

Only a single application is created when an applicant submits the application form.

### Environment

v2026.7.0

### Related Issues

N/A

### Additional Information

A redacted example of such an application.

[eak-applications.edit.json](https://github.com/user-attachments/files/30824912/eak-applications.edit.json)

f1sh1918:
### Short Description

Unfortunately we removed the disabling of all input with a `useEffect` that was triggered when `loading` prop changed here:
https://github.com/digitalfabrik/entitlementcard/commit/b899c726f8c88910aa0952e3c2aca6d7b583ff4a#diff-1d00c38aee25bc9ec894ff59dad4a4e8edbd1a8abd2c7a5a675f57adb1187f14L51-R76

This resulted in the user could push the "Abschicken" button x-times until the backend sends a result

Note: That we might also add some backend checks but i think this should be good enough as it also worked before we mistakenly removed the `useEffect`

### Proposed Changes

- `disableAllInput` via formContext memo, so loading also triggers disableAllInput that avoids sending same application multiple times by disabling the button directly after first click

### Side Effects

- N/A

### Testing

1. Create an application (bavaria) and click on "Absenden"
2. It should not be possible to click the button multiple times
3. Check the "Eingehende Anträge" to verify that only one application was sent with the same data

### Resolved Issues

Fixes: #3091

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,10 +24,11 @@ commands:
         description: This command builds the circle config from the files in src and validates that it is up-to-date and valid.
         steps:
             - run:
-                command: curl -fLSs https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh | sudo bash
+                command: mise install "github:CircleCI-Public/circleci-cli"
                 name: Install CircleCI CLI
+                working_directory: ~/project
             - run:
-                command: scripts/circleci-update-config
+                command: mise x -- scripts/circleci-update-config
                 name: Build circle config
                 working_directory: ~/project
             - run:
@@ -42,12 +43,12 @@ commands:
                       echo "The CircleCI config is up to date."
                       exit 0;
                     else
-                      echo "The CircleCI config is not up to date. You can update it by running the `./scripts/circleci-update-config` script."
+                      echo "The CircleCI config is not up to date. You can update it by running the \`./scripts/circleci-update-config\` script."
                       exit 1;
                     fi
                 name: CircleCI config up to date
             - run:
-                command: circleci config validate
+                command: mise x -- circleci config validate
                 name: Validate circle config
                 working_directory: ~/project
     install_app_toolbelt:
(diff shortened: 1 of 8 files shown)


seluianova (approved):
works 👍
thanks for the investigation and quick fix!

bahaaTuffaha (approved):
Tested on Chrome (created 2 applications and there was no duplicates) ✅

I think it would be better to check duplicate ids from the server side and accept the first one only.

connium:
> Tested on Chrome (created 2 applications and there was no duplicates) ✅
> 
> I think it would be better to check duplicate ids from the server side and accept the first one only.

I didn't find any duplicated IDs ([see original issue](https://github.com/digitalfabrik/entitlementcard/issues/3091)).
If I understand @f1sh1918 correctly, the UI doesn't get locked immediately and thus people might push the submit button multiple times (I observe young people pushing any button (traffic light, automatic doors, apps, ...) continuously until the desired action happens).

f1sh1918:
Yes we don't have duplicated ID, we create multiple applications with different ids. so thats a valid fix here
We could check for identical jsonValues additionally on backend side but as i said i just fix the original issue.

PR #3096, opened 2026-08-10
Fix circle ci cli

f1sh1918:
### Short Description

With a new circle cli cli version the indentation changed.
We commited a config.yml with old indentation in the repo and the CI runs in with a new cli version and detects changes and then the check of the config file fails

### Proposed Changes

- use a fixed circle ci cli version from mise

### Side Effects

- N/A

### Testing

N/A

### Resolved Issues

Fixes: #


seluianova (approved):
thanks!

PR #3108, opened 2026-08-25
3076: Improve region assignment

### LINKED ISSUE #3076: Automatic region assignment seems not to be reliable
### Describe the Bug

To me, it looks like the automatic region assignment in the application form is not reliable.

### Steps to Reproduce

In the application form of the Ehrenamtskarte Bayern, I tested with some arbitrary (valid) zip codes.
It has happened, that it said that the region could not be determined. Then I entered another zip code, which was found successfully. After that I entered the previous zip code again, and now it was found.

I cannot reproduce this behavior reliably.

### Expected Behavior

After the user has entered the zip code of their home location, the form should reliably select the corresponding region.

### Environment

- Device/Browser: Laptop, Firefox
- OS version: Fedora Linux
- App version: production

### Related Issues

N/A

### Additional Information

Context: A region reached out to me, that some applications are directed to their neighbor region.

f1sh1918:
### Short Description

We were facing some race conditions for the region fetch via postal code.
I added some debouncing to reduce requests if the the user makes a rapid input change
Furthermore i checked if the postal code value that was typed, fits with the current result

### Proposed Changes

- add debouce hook
- remove RegionNotFound exception, since it's a valid result to find no region
- use debouncing and add race condition guard, show region not found for empty result list

### Side Effects

- N/A

### Testing

- Go to `/beantragen` and type in some postal codes that exist and don't exists and check that the resulting region that was selected is correct

### Resolved Issues

Fixes: #3076

### DIFF
--- a/administration/src/hooks/useDebouncedValue.ts
+++ b/administration/src/hooks/useDebouncedValue.ts
@@ -1,0 +1,14 @@
+import { useEffect, useState } from 'react'
+
+const useDebouncedValue = <T>(value: T, delayMs: number): T => {
+  const [debouncedValue, setDebouncedValue] = useState(value)
+
+  useEffect(() => {
+    const timeout = setTimeout(() => setDebouncedValue(value), delayMs)
+    return () => clearTimeout(timeout)
+  }, [value, delayMs])
+
+  return debouncedValue
+}
+
+export default useDebouncedValue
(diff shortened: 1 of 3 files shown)


bahaaTuffaha (approved):
It works as expected. I tried `86152` ✅
At first I thought the debounce for the input itself but here used for the pause logic which I think it's better.

andrew8er (approved):
Nice work.

For the future, we could look into solutions like https://tanstack.com/form/latest to simplify the form handling.

PR #3133, opened 2026-09-02
2744: Collection minor issues

### LINKED ISSUE #2744: Collection of minor issues in the administration area
### Describe the Bug

**🐛 No. 1 - User management:** When editing the own user, a warning is shown. This has a low contrast.

Image

**🐛 No. 2 - Project management (as external API user):** The button for creating a new token IMO should be a primary button.

Image

**🐛 No. 3 - Application management:** When deleting an application, a confirmation dialog is displayed. After confirmation, there is no notification shown (e.g. "Successfully deleted"). For other actions, we show such notifications.

**🐛 No. 4 - Card creation:** When importing cards from a CSV file, it is possible, to define a gold card with an expiry date. This looks strange. IMO this should be a validation error.

Image

**🐛 No. 5 - Acceptance store management:** When importing acceptance stores, the messages states that longitude and latitude are required. While the columns must be present, they don't need to contain any data. So they shouldn't be mentioned as required fields.

Image

**🐛 No. 6 - Acceptance store management:** When trying to import an empty file or a file with just the column titles, an error notification is shown. It feels like this notification vanishes much faster than others (approx. 2 seconds).

Image

**🐛 No. 7 - Acceptance store management:** If the file contains invalid data, the corresponding cells are highlighted, but there is no error notification. Especially if the error(s) are in a column far right, I don't get it. If the geolocation cannot be resolved, we show an error notification.

Image

**🐛 No. 8 - Acceptance store management:** When executing a dry run or actually importing stores, a notification is shown. Shouldn't this notification be a green one (success)? @hauf-toni 

Image

f1sh1918:
### Short Description
We had some minor issues and design glitches, i've fixed here.

### Proposed Changes

- fix the issues shown here #2744 

### Side Effects

- N/A

### Testing

See issue description #2744 
- Test single card creation (gold and standard) and also bulk import cards. Best for all regions 

### Resolved Issues

Fixes: #2744

### DIFF
--- a/administration/src/cards/card.test.ts
+++ b/administration/src/cards/card.test.ts
@@ -103,13 +103,13 @@ describe('Card', () => {
     it('should correctly set and get value', () => {
       const dateString = '03.04.2022'
       const date = parseGermanPlainDateString(dateString)
-      const line = ['Thea Test', dateString, 'Goldkarte']
+      const line = ['Thea Test', dateString, 'Standard']
       const headers = ['Name', 'Ablaufdatum', 'Kartentyp']
       const card = initializeCardFromCSV(cardConfig, line, headers, region)
 
       expect(card.fullName).toBe('Thea Test')
       expect(card.expirationDate).toEqual(date)
-      expect(card.extensions[BAVARIA_CARD_TYPE_EXTENSION_NAME]).toBe('Goldkarte')
+      expect(card.extensions[BAVARIA_CARD_TYPE_EXTENSION_NAME]).toBe('Standard')
 
       expect(isValueValid(card, cardConfig, 'Name')).toBeTruthy()
       expect(isValueValid(card, cardConfig, 'Ablaufdatum')).toBeTruthy()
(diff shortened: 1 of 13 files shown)


seluianova:
❓not sure if I correctly understand what was desired here:

> 🐛 No. 5 - Acceptance store management: When importing acceptance stores, the messages states that longitude and latitude are required. While the columns must be present, they don't need to contain any data. So they shouldn't be mentioned as required fields.

longitude and latitude were removed from the list:

image

but I would assume that we just need to remove the `*` from them?
because it’s still possible to load these fields; they’re just not mandatory.
or was the idea that we don’t want to allow users to upload pre-filled longitude and latitude at all?

seluianova (approved):
Looks nice to me 👍 
One questionable point (in the comment above) and one unrelated finding:

I just noticed that we lost the usage of the `dryRunImportResult` translation at some point. We now show the same success message for both the dry run and the normal import. _This might be adjusted separately, though, up to you._

Tested issues:

No. 2 - Project management (as external API user): The button for creating a new token IMO should be a primary button ✅
No. 3 - Application management: When deleting an application, a confirmation dialog is displayed. After confirmation, there is no notification shown (e.g. "Successfully deleted") ✅
No. 4 - Card creation: When importing cards from a CSV file, it is possible, to define a gold card with an expiry date ✅
No. 5 - Acceptance store management: When importing acceptance stores, the messages states that longitude and latitude are required ❓
No. 6 - Acceptance store management: When trying to import an empty file or a file with just the column titles, an error notification is shown. It feels like this notification vanishes much faster than others (approx. 2 seconds). ✅

andrew8er (approved):
Loks good, nice cleanups.

f1sh1918:
> ❓not sure if I correctly understand what was desired here:
> 
> > 🐛 No. 5 - Acceptance store management: When importing acceptance stores, the messages states that longitude and latitude are required. While the columns must be present, they don't need to contain any data. So they shouldn't be mentioned as required fields.
> 
> longitude and latitude were removed from the list:
> 
> image
> but I would assume that we just need to remove the `*` from them? because it’s still possible to load these fields; they’re just not mandatory. or was the idea that we don’t want to allow users to upload pre-filled longitude and latitude at all?

Yes IMO the goal was to not show them as "required" fields but it should still be possible to upload csvs with that fields as optional fields @connium ?

connium:
> > ❓not sure if I correctly understand what was desired here:
> > > 🐛 No. 5 - Acceptance store management: When importing acceptance stores, the messages states that longitude and latitude are required. While the columns must be present, they don't need to contain any data. So they shouldn't be mentioned as required fields.
> > 
> > 
> > but I would assume that we just need to remove the `*` from them? because it’s still possible to load these fields; they’re just not mandatory. or was the idea that we don’t want to allow users to upload pre-filled longitude and latitude at all?
> 
> Yes IMO the goal was to not show them as "required" fields but it should still be possible to upload csvs with that fields as optional fields @connium ?

My intention was, to remove the required-indicator from the _longitude_ and _latitude_, as it is with the other optional fields like email, homepage, … . The fields should still be mentioned in the list, as this is the specification of the import file.

I just recognized, that the label says "Erforderliche Spalten". :roll_eyes: 

My goal is this:
Bildschirmfoto am 2026-09-24 um 11 25 05

Not quite sure about the "Maximalanzahl an Einträgen". We mention this for the card import. If there is no limit for acceptance stores, remove it.

